PR #264, opened 2012-07-11
Initial sphinx documentation

kushaldas:
This include inital sphinx documentation. I used details from README.md and autodocs from classes.

You can see the current status at http://kushal.fedorapeople.org/redis-py/html/
If you want I can use use a different theme like http://retask.readthedocs.org/en/latest/

My next goal is update the docstrings for various commands with proper return and argument details for documentation and adding up docstrings for the exception classes.

After adding the initial docstring updates, I would prefer to add more examples for each command (in docstrings).

Tell me what you think


RoeyPrat:
Hi @kushaldas, 
Thank you, and apologies that it's taken so long to get back to you on this.
The repo has sphinx documentation (probably added after you made this PR).
I'm closing the PR, but you are very welcome to make documentation improvements over the master branch - we will be much more responsive.

PR #294, opened 2012-10-30
support unix_socket_path in from_url

graingert:
unfortunately no way to encode the DB in this implementation

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -240,6 +240,7 @@ class StrictRedis(object):
         For example::
 
             redis://username:password@localhost:6379/0
+            redis://:password@/var/run/redis/redis.sock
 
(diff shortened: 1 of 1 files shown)


gyermolenko:
hi @graingert 
The code for `from_url` method has moved and evolved since the time of PR creation.
Could you check if this PR is still relevant? 

As for your question "Is there an 'official' redis URI definition?" some links were added to the [method's docstring](https://github.com/andymccurdy/redis-py/blob/master/redis/connection.py#L831).

Thank you.

andymccurdy:
Yes, this was resolved long ago. Use `unix://` as the scheme to connect via a unix socket connection.

graingert:
Yeah I've lost all context in this, sorry

PR #443, opened 2014-03-13
Fresh version of #93: Add logging of commands, connects and disconnects

alanjds:
This branch [is argued](https://github.com/andymccurdy/redis-py/pull/93#issuecomment-668917) to add logging capabilities costing only 0.2% slowdown.

Please consider merging it and then updating the docs on the usage of the new feature, as I am unsure on how to enable, if different from standard.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -58,6 +58,7 @@ class Connection(object):
         "Connects to the Redis server if not already connected"
         if self._sock:
             return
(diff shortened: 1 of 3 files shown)


tuukkamustonen:
Kinda surprised there is no logging at all in the current package... so +1.

jvanasco:
it pains me that there is no logging yet.

PR #472, opened 2014-05-07
Fixes #471 - raise ValueError instead of implicitly casting values to strings

### LINKED ISSUE #471: Don't implicitly cast values to strings
When saving something that can't automatically be deserialized properly the client should raise an exception instead of coercing the value into a string.

Current behavior:

``` python
>>> sample_dict = {'nested': {'dict': 'here'}}
>>> r.hmset('dict', sample_dict)
True
>>> r.hgetall('dict') == sample_dict
False
>>> r.hgetall('dict')
{'nested': "{'dict': 'here'}"}
```

dbravender:
It's not always possible to deserialize an object from its str response. This makes developers aware that redis can only handle strings and they are responsible for serializing and deserializing before storing objects in redis.


andymccurdy:
Fixed in redis-py 3.0

PR #492, opened 2014-06-06
restore default Lock token storage, add toggle to make it thread-local

chillipino:
after thinking about this a bit, always putting the token in thread local storage makes it impossible to intentionally acquire a lock that's shared throughout the process.  it's likely that existing code will depend on the [previous behavior](http://xkcd.com/1172/), so i think we should control it with a toggle.

this change adds a `thread_local=False` parameter to `Lock.__init__()` and `StrictRedis.lock()`, and only puts the token in thread-local storage if the parameter is `True`.  i'm open to suggestions if you don't like the way this was done (i'm a bit iffy on the `dummy` class, myself).

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -478,7 +478,7 @@ class StrictRedis(object):
                     continue
 
     def lock(self, name, timeout=None, sleep=0.1, blocking_timeout=None,
-             lock_class=None):
+             lock_class=None, thread_local=False):
         """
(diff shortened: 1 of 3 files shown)


zhaopangithub:
How do I use this parameter thread-local. 
Can you give me a slightly more detailed demo.
Because I don't find any difference, when it's true or false

PR #556, opened 2014-11-17
Check if Script.sha is None. 

endophage:
It's nonsensical to make a network call that's guaranteed to fail.

When a script object is instantiated, unless the dev manually sets the script_instance.sha attribute, it will always call evalsha the first time with an empty string. This will always fail, fall through to the except, and set the sha.

Especially in a large scale applications, there is a high probability the redis server already has the script.

p.s. I talked through this change with my colleagues and I initially wanted to just set the sha attribute up front in the init function. My colleagues suggested that the current implementation is probably to provide additional laziness for the dev that instantiates all their scripts up front. This commit therefore attempts to maintain the same level of laziness as before, only calculating the sha once we know we're actually going to use the script.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2642,7 +2642,7 @@ class Script(object):
     def __init__(self, registered_client, script):
         self.registered_client = registered_client
         self.script = script
-        self.sha = ''
+        self.sha = None
 
     def __call__(self, keys=[], args=[], client=None):
         "Execute the script, passing any required ``args``"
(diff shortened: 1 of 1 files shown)


Suor:
+1 for solving this, it makes no sense to make `EVALSHA` requests with empty hash.

You  approach, however, won't work as filled `.sha` in client doesn't mean that current client is connected to a server, which knows script. This is because script objects a reusable with different connections, also script object lifetime is long enough to see redis server restart.

I believe that the right way is to just calculate hash on script creation and then act optimistically as done now.

endophage:
@Suor glad somebody took a look at last. I hear what you're saying and I'd suggest the right way to solve it might be to call client.script_load both before the first eval and inside the except.

I've switched jobs and languages since submitting this and haven't written any Python in 6 months now. I don't have any ongoing motivation to fix this up. Feel free to close/modify/etc...

andymccurdy:
This has been resolved by PR #867

Issue #555, opened 2014-11-16
Connection leak when no explicit cleanup of pubsub object


yoav-steinberg:
Since the registation of "on_connect" callbacks were added to `PubSub`, pubsub objects aren't freed when they loose all app references (because the callback is still referened from within the connection pool's connection object).
Try this:

``` python
r = redis.Redis()
r.ping()
for i in range(100):
    s = r.pubsub()
    s.subscribe('chn')
```

I'd expect the connection used for `s` to be reused, but instead 100 subscribe connections are created and never freed (even if we move to a scope where `s` isn't defined anymore).

I undestand that for "resubscribe" feature we need this on_connect callback. So maybe there isn't a perfect solution for this. But in such a case we should:
- Document that s needs to be explicitly cleaned up (via a call to .close())
- Have the `PubSub` class implement the python context manager protocol so we can do something like:

``` python
for i in range(100):
    with r.pubsub() as s:
        s.subscribe('chn')
```

MattFisher:
I don't know if it's the same issue, but Celery leaks connections when using redis as its broker: https://github.com/celery/celery/issues/4465

Issue #531, opened 2014-09-03
ConnectionPool repr fails when db is not specified


rancerbeta:
This gives a key error because the repr for connectionpool uses the underlying connection object's repr string template. In this case, the connection object repr expects a db variable to be set because connection's init has a default db value of 0. There is no such thing in the connectionpool kwargs, so a key error is thrown when the string template is used.

```
import redis
r = redis.ConnectionPool(host='localhost', port=6379)
repr(r)
```

```
>>> x = redis.ConnectionPool(host='localhost', port=6379, max_connections=1)
>>> x.connection_kwargs
{'host': 'localhost', 'port': 6379}
>>> x.connection_class.description_format
'Connection<host=%(host)s,port=%(port)s,db=%(db)s>'
>>> x.connection_class.description_format % x.connection_kwargs
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'db'
>>> 
```

gyermolenko:
as I checked for #768 :
 no longer an issue - all format arguments are passed with defaults.

andymccurdy:
This was resolved in redis-py 3.0.

PR #558, opened 2014-11-19
simple fix for issue #531

mihirvj:
this commit uses the fact that connection has db set to 0 by default. If
a db value is not present in kwargs, it is defaulted to 0.

targets issues #555 and #531

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2062,10 +2062,10 @@ class PubSub(object):
             pass
 
     def __enter__(self):
-      return self
+        return self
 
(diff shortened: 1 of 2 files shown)


vaultah:
Any updates?

sharepusher:
I experienced this issue when using pipeline, the version is 2.10.5, may I ask whether the issue fixed or not ? thanks

gyermolenko:
In general PR should target only one issue. Plus #531 is obsolete.

andymccurdy:
The description format issue was resolved in redis-py 3.0.

PR #569, opened 2014-12-10
Atomic redis.set when acquiring a lock

23doors:
Both LUA and standard versions updated.

As referenced in: https://github.com/andymccurdy/redis-py/issues/554


c-nichols:
@andymccurdy what's the status on getting this merged?  We traced some performance issues back to the lua implementation and think this would have helped us avoid them.

andymccurdy:
This makes sense to me. Thanks!

PR #570, opened 2014-12-10
Implement connection retries in connect().

nfvs:
This PR adds the ability to attempt reconnecting when there is a connection failure, with a configurable number of retries and wait duration.

`reconnect_on_timeout` enables a reconnection attempt when the connection is broken, but only once, and if `connect()` fails an exception is thrown. This is maintained by default with this PR.

But by passing a `connection_retries=X` argument to `Connection()`, a reconnection will be retried X number of times. A `connection_retry_wait=Y` makes it possible to wait Y seconds between connection retries.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -6,6 +6,7 @@ import os
 import socket
 import sys
 import threading
+import time
 import warnings
 
 try:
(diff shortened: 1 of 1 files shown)


vanadium23:
Can we merge this? It'll be useful for some mean time infrastructure problems.

shashankmehra:
Is there a drawback to this approach? Seems like something I was looking for.

dejlek:
I would add exponential backoff strategy here mixed with number of retries...

PR #575, opened 2015-01-01
Allow pubsub WorkerThread to be shut down while in a callback 

efwahl:
Allow pubsub WorkerThread to be shut down while in a callback by ignoring the RuntimeError thrown when join is called on the running thread.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2310,7 +2310,11 @@ class PubSub(object):
 
             def stop(self):
                 self._running = False
-                self.join()
+                try:
+                    self.join()
+                except RuntimeError:
+                    # safe to ignore this, called stop from WorkerThread
+                    pass
 
         thread = WorkerThread()
         thread.start()


andymccurdy:
Hi Edward,

I agree we should able to shutdown the worker thread from a callback. However looking at the current code and your proposed fix I realize the `stop` method should be doing a bit more work. We're not currently unsubscribing from any patterns or channels and we're not cleaning up the underlying pubsub instance.

I've fixed that here: 69ce3269dfcb87e972ef29a84772d781d1613c9c

Can you please test that commit in your setup and make sure it works for you? Thanks!

efwahl:
Hi Andy,

I tested, and yes your solution works cleanly for my code that fails 
with the current 2.10.3 version. And that looks like a much cleaner 
solution to the problem than what I suggested.

Thanks for the quick response.

regards,
-- Ed

andymccurdy:
Wonderful. I'm going to get a few other bugs cleaned up and will push a new build to PyPI in the next couple days.

PR #577, opened 2015-01-05
Daemonize PubSubWorkerThread

jkaye2012:
Prevents Python process from hanging when using run_in_thread without
explicitly stopping.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2305,6 +2305,7 @@ class PubSubWorkerThread(threading.Thread):
         self.pubsub = pubsub
         self.sleep_time = sleep_time
         self._running = False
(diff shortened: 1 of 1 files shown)


AndrewGuenther:
:+1: 

I would also like the ability to do this, but a better implementation may be to simple pass along the keyword arguments from run_in_thread so that you don't break backwards compatibility.

AndrewGuenther:
Hey @jkaye2012, I created a new PR which takes this value as a parameter rather than assuming `True`, check it out: #640

jkaye2012:
@AndrewGuenther seems to have a better solution.

PR #578, opened 2015-01-08
add 'decrby' method

derek-dchu:
Add 'decrby' method following the style of 'incrby' method which is an alias for 'decr' method.


c-bata:
+1

Issue #579, opened 2015-01-09
Url encoding passwords with invalid url characters causes connection failure


lBowlin:
Cloud services that required passwords such as "iAITMl7lD127c/VZEl0OZl8qviaXHIoKekaJjP1J1HI=" can contain a variety of characters including forward slashes which are invalid in a url string.

When entered in the correct format "rediss://:password@hostname:port" the slash cause in improper breakage of the netloc  during urlparse. Encoding the forward slash into a valid url safe character (%2F) fixes the urlparse issue, but is never decoded after parsing prior to assignment as the password.

This is particularly an issue with Azure Redis Caches which have keys that may contain such values. I think a simple url decode would fix the issue, but have not tested this.

yozik04:
Also using Azure Redis Cache. Have same problems.

andymccurdy:
Hey guys, this issue was fixed a while back here: https://github.com/andymccurdy/redis-py/pull/586

ddennis:
Thank you for documenting this. Have the same problem using node

PR #586, opened 2015-02-10
Handle percent-encoded URLs in parsing code

paulkeene:
Hi,

Clients of redis-py that attempt to instantiate a `ConnectionPool` object by using the `from_url` class method can run into problems if components of the passed-in URL contain illegal characters (e.g. `rediss://:my/password@localhost:6379`). Understandably the forward slash in this example URL causes it to be parsed incorrectly, but even if a client attempts to percent-encode their URL (e.g. `rediss://:my%2Fpassword@localhost:6379`) this will not fix the problem since `from_url` will not know to decode the password after the URL is parsed.

This commit attempts to fix the problem by introducing a new argument, `decode_components`, to `from_url` that allows URL components to be decoded after the URL has been parsed. I used a flag here because existing clients might be relying on the current behavior.

I added some unit tests and verified that they pass with python2 and python3.

This fixes issue #579 

Thanks,
Paul

### DIFF
--- a/redis/_compat.py
+++ b/redis/_compat.py
@@ -3,6 +3,7 @@ import sys
 
 
 if sys.version_info[0] < 3:
+    from urllib import unquote
     from urlparse import parse_qs, urlparse
     from itertools import imap, izip
     from string import letters as ascii_letters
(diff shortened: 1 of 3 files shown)


yozik04:
Great. But sad it was not released to pypi yet.

PR #587, opened 2015-02-10
Fix syntax highlighting in README

frewsxcv:


Grokzen:
@andymccurdy This PR is no longer valid/needed. All instances it tries to fix has been resolved long time ago. This can be closed.

PR #588, opened 2015-02-10
Fix pep8 errors

paulkeene:
This fixes the E731 errors that are showing up in the Travis CI pep8 builds. All unit tests pass locally with python2 and python3.


andymccurdy:
Great, thanks. For some reason I couldn't get the errors to show up locally, even with using the latest pep8 (1.6.1).

PR #597, opened 2015-03-14
Improve performence of transactions / pipeline requests which involve large chunks of data.

tzickel:
Similar to the performance fix of pack_command from a year ago, this adds the same logic into the pack_commands, which is used in transactions and pipeline commands.

The speedup can be huge depending on the size of chunks being send.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -461,9 +461,11 @@ class StrictRedis(object):
         string_keys_to_dict('BGREWRITEAOF BGSAVE', lambda r: True),
         {
             'CLIENT GETNAME': lambda r: r and nativestr(r),
+            'CLIENT ID': int,
             'CLIENT KILL': bool_ok,
             'CLIENT LIST': parse_client_list,
             'CLIENT SETNAME': bool_ok,
+            'CLIENT UNBLOCK': lambda r: r and int(r) == 1 or False,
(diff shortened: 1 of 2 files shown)


tzickel:
Ok, I see we need to provide some speedup context... For the following example, the current redis-py code (2.10.5) runs at about **1.4 seconds per iteration** (with a very high variance of up to an extra second), yet with my modification it runs at about **0.6 seconds per iteration** with a comparatively small variance of 0.1 second.

The reason is because the current code can still call "SYM_EMPTY.join(pieces)" where pieces might hold a giant blob of data (in the pack_command for example it cannot happen, because large arg is appended directly to output).

```python
import redis
import time

r = redis.Redis()
_ = r.info()

data = 'a' * 100000

for j in range(3):
  with r.pipeline() as p:
    for i in range(10000):
      p.set('a', data)
    s = time.clock()
    _ = p.execute()
    e = time.clock()
  print e - s
```

joshowen:
This would be very useful for us too.  @andymccurdy thoughts?

tzickel:
@andymccurdy any chances on getting this merged ?

tzickel:
This seems like a bug in github (I did not close this issue, but simply updated the branch), as I cannot seem to reopen it, i'll issue a new PR...

PR #614, opened 2015-05-04
Connection.encode complains when value is an object having unicode characters in its printable representation

duxiaoyao:


andymccurdy:
Thanks!

PR #616, opened 2015-05-07
Added new BatchedPipeline class and tests

johnistan:
### What is the problem / feature ?

This is utility that I have found useful in my own work. Please take a look a see if it belongs in the library. 

It is a BatchedPipeline object that takes a batch_size argument and then executes all buffered commands after the user has added that many commands. I find it useful when executing large number of commands that I want to gain the speed of pipeline without swamping the server with one huge pipelined set of commands at the end. Or without have to keep track.

I found myself writing lots of code like this:

``` py
pipe = r.pipline()

for i, item enumerate(huge_iter):
    if i % 10000 == 0:
        pipe.execute()
    pipe.set(i, item)
pipe.execute()
```

With the BatchedPipeline you can use a context manager to achieve the same thing:

``` py
with r.batched_pipeline(10000) as pipe:
    for i, item enumerate(huge_iter):
        pipe.set(i, item)
```
### How did it get fixed / implemented ?

I subclassed `BasePipeline` overriding `__init__`, `pipeline_execute_command` and `__exit__`
### How can someone test / see it ?

I have duplicated all existing pipeline tests with the new class and added two new ones to test the context manager and normal use. I'm sure I could get away with not duplicating all the test by using some py.test parameterize fu. But wanted see if this was of interest.

I wanted to see if this is of interest before I add the docs and clean up the doc strings. So this is not ready to go as is.

Specifically what needs to be done is docs, de-dupe test and doc strings.

Here is a cute animal picture for your troubles...
pandas

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -470,6 +470,23 @@ class StrictRedis(object):
             transaction,
             shard_hint)
 
+    def batched_pipeline(self, batch_size, transaction=True, shard_hint=None):
+        """
+        Return a new batched pipeline object that can queue multiple commands
+        for later execution. ``batch_size`` represent the maximum number of
+        commands to buffer before executing them against the server
+        ``transaction`` indicates whether all commands should be executed
+        atomically. Apart from making a group of operations atomic, pipelines
+        are useful for reducing the back-and-forth overhead between the client
+        and server.
+        """
+        return BatchedPipeline(
+            batch_size,
+            self.connection_pool,
+            self.response_callbacks,
+            transaction,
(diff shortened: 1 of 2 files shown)


jofusa:
Any thoughts on this?

joshowen:
@andymccurdy Any thoughts on this?

jofusa:
still willing to add necessary docs. and any changes needed.

szwed:
+1

saturnboy:
@andymccurdy i need this too

PR #620, opened 2015-05-22
Support multiple keys for pfcount call

cwilkes:
http://redis.io/commands/pfcount supports multiple keys in the request, it internally merges them together in a temporary key, returns the result, and deletes the temp key.

I find it handy when wanting to sum up a lot of keys and don't care to keep the pfmerge around.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1813,12 +1813,12 @@ class StrictRedis(object):
         "Adds the specified elements to the specified HyperLogLog."
         return self.execute_command('PFADD', name, *values)
 
-    def pfcount(self, name):
+    def pfcount(self, *sources):
         """
         Return the approximated cardinality of
-        the set observed by the HyperLogLog at key.
+        the set observed by the HyperLogLog at key(s).
         """
-        return self.execute_command('PFCOUNT', name)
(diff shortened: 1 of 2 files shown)


andymccurdy:
Thanks!

cwilkes:
My original use case was to support ipython usage of
  pfcount('a', 'b')
but this causes problems when I pass in pfcount(['a', 'b']) as the ending call is execute('pfcount', ['a', 'b']) which doesn't work -- it needs to be execute('pfcount', 'a', 'b').

Now I don't think that pfcount('a', 'b') should be supported, kind of like numpy's hstack requires a tuple to be passed in even though it seems natural to call it like hstack(ary0, ary1), forcing people to call hstack((ary0, ary1))

Figuring out how to do this in code is somewhat tricky, this is what I've come up with:

def pfcount(sources):
    if type(sources) in (list, tuple):
        return execute_command('PFCOUNT', *sources)
    else:
        return execute_command('PFCOUNT', sources)

which looks ugly but gets the job done.

What do you think?

andymccurdy:
I believe it should be identical to other commands in redis-py that support multiple keys. All the other commands use *args, so this one should to.

PR #627, opened 2015-06-11
Tiny addition: `.pubsub()` now works for Sentinels

rozbb:
Sentinels publish plenty of status messages, and there was no way to see them using redis-py. Here is a list of those message: http://redis.io/topics/sentinel/#pubsub-messages

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -2,7 +2,7 @@ import os
 import random
 import weakref
 
-from redis.client import StrictRedis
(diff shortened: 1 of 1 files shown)


mnisjk:
:+1:

amuttsch:
For my current project I want to subscribe to the sentinel status messages as well. I'm using a workaround currently, but having this would cleanup the code. Thanks!

JiaqiLiu:
Really need it!

Issue #613, opened 2015-05-03
Add ability to close connections/clients


harlowja:
It'd be nice to have a way to close a client (stopping it from further use) for situations/libraries that want to take custom clients (or provide there own) but want to themselves provide close methods (to stop further object usage).

I have something that does this at the following and can submit a PR if you are fine with the concept...

See: https://review.openstack.org/#/c/176951/38/taskflow/utils/redis_utils.py (search for ``class RedisClient`` in that).

Thoughts?

dmk23:
+1

The functionality is needed for cases when too many connections get opened and they need to be manually closed to avoid exceeding the limits

jackyfkc:
Hi harlowja, I have a doubt if it is required to implement an explicit close() method since the underlying connections would be automatically closed when the **client object** ran out of scope.

The redis.connection.Connection has implemented __del__ method.

harlowja:
Sometimes explicit is better than implicit (one of the guiding principles of python); I feel a explicit ``close`` is nicer (and matches usage of https://docs.python.org/2/library/contextlib.html#contextlib.closing), of course a `__del__`` method is ok, and nice, but both existing seems to be good imho.

andymccurdy:
`ConnectionPool`s already have a `disconnect` method that closes all open connections.

I don't really see the point of marking a client instance as "closed" to prevent it from executing future commands. If you absolutely want to prevent your application from using the client instance, just remove your application's reference to the instance after you call the `ConnectionPool.disconnect` method.

woky:
The use case is e.g. a Flask application. I don't fully understand the details of Flask and _uWSGI_ but long story short, each request can happen in a different process, so it's very important to have assurance that all IO has been done and nothing is leaking behind when a request (and potentially its process) ends.

It'd be great if someone knowledgeable explained the life-cycle in the documentation. I've discovered this _answer_ after I started writing issue and GitHub showed this issue to me when I was writing title. This was preceded by usual Googling and code reading of course.

PR #628, opened 2015-06-20
Allow the client (and its connection_pool) to be closed

harlowja:
For situations where it is not acceptable to keep a client
and its pool open when not in use it is very nice to have the
ability to close a client and its pool to release the resources
associated with it.

Fixes issue #613


andymccurdy:
See comments in #613.

Issue #576, opened 2015-01-05
Python process hangs without joining following run_in_thread


jkaye2012:
Currently, if one uses PubSub.run_in_thread and neglects to stop the thread before the program exits, the Python process will hang on the processing thread. In my testing, this issue is resolved easily by making the worker thread a daemon before starting it; in this way, the processing thread is terminated when its parent dies.

Are there any issues with daemonizing the worker thread in this way?

Commit: 48ae31aabb1497779cadd525a9b6d865d5b8b90a

jkaye2012:
Any word on this?

jkaye2012:
I would really love to not have to fork this project and build my own version of this package.. Can any maintainer comment on this?

jackyfkc:
You could attempt to create a new pull request

andymccurdy:
This was resolved long ago. Sorry for not closing this sooner.

PR #640, opened 2015-07-29
Added the option to call run_in_thread as daemon

AndrewGuenther:
Since PubSub.run_in_thread() starts the thread before returning it, it
is impossible for the caller to mark it as a daemon, which can be
desirable for many reasons.

This change adds an additional optional parameter which will cause the
thread to be run as a daemon. It is a keyword argument and defaults to
false so the current behavior will not change and it will not break any
current consumers on upgrade.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2295,7 +2295,7 @@ class PubSub(object):
 
         return message
 
-    def run_in_thread(self, sleep_time=0):
+    def run_in_thread(self, sleep_time=0, daemon=False):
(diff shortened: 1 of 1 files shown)


jkaye2012:
+1. Closing my PR in support of this one.

jkaye2012:
#576

soluwalana:
:+1:

andymccurdy:
This functionality was added in redis-py 3.0. Sorry that this wasn't closed when that version was released.

PR #642, opened 2015-07-30
Python 3 support in Connection.send_packed_command

kosonya:
In Python 3 the argument of a socket.sendall must be a list of byte sequences, not strings. I changed the type check in send_packed_command as to preserve the old behavior in Python 2, expect a <class 'bytes'> or a list of those as an argument, and raise TypeError if a string is passed instead.

(Maxim Kovalev, Zugata)

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -542,11 +542,11 @@ class Connection(object):
                 if python_version == 2:
                     command = [command]
                 elif python_version == 3:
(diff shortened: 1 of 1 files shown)


andymccurdy:
I don't believe this is an issue at this point. Everything that calls `send_packed_command` already ensures that all values are encoded to bytes. If I'm missing something, feel free to reopen.

PR #646, opened 2015-08-18
Fix tests under Redis 3.x; we can be of an "embedded string" format now.

lamby:
Signed-off-by: Chris Lamb <chris@chris-lamb.co.uk>


andymccurdy:
thanks!

PR #652, opened 2015-09-02
Try the next sentinel host on socket timeout

### LINKED ISSUE #651: Next sentinel host is not contacted after socket timeout
If a there is a socket timeout with a sentinel host, `redis.exceptions.TimeoutError` is returned and none of the other sentinel hosts are contacted.

If there is a connection timeout or connection refused, the next host is tried. It would great if there was way to try the next sentinel host in the same way for socket timeout errors.

(You can `retry_on_timeout=True` to retry the same sentinel host once, but if you get a socket timeout a second time, `redis.exceptions.TimeoutError` is returned.)

svend:
This fixes #651.

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -4,7 +4,8 @@ import weakref
 
 from redis.client import StrictRedis
 from redis.connection import ConnectionPool, Connection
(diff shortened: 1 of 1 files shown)


blopker:
Any way we can get this merged? I think it's causing issues for us too.

andymccurdy:
This was merged from #684. Thanks!

PR #655, opened 2015-09-18
 convert read_response()'s results to int or float.

sunminghong:
add convertToNumber() to convert "HMGET,HGETALL" 's response to int or float.

{'mode': '0', 'lasttime': '1442546304', 'name': 'a name',
    'type': '0', 'regtime': '1442546304'}

to:
{'mode': 0, 'lasttime': 1442546304, 'name': 'a name',
    'type': 0, 'regtime': 1442546304}

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -73,13 +73,8 @@ def convertToNumber(a):
                 t=1
 
     if t == 1:
-        a = float(a)
-    else:
-        print("a:", type(a), a)
-        a = int(a)
-        print("a2:", type(a), a)
-
-    return a
+        return float(a)
+    return int(a)
 
 
 class Token(object):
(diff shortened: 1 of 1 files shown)


andymccurdy:
We never want to implicitly convert values that look to be numeric to numbers. That's not nice for people that need strings that just happen to contain all numbers.

If you want this behavior I'd suggest registering your own response callback. See this section of the docs: https://github.com/andymccurdy/redis-py#response-callbacks

```python
r = redis.StrictRedis(...)
r.set_response_callback('HGETALL', convert_to_number)
```

With the above code, anytime you call `r.hgetall` the normal return value will be passed to your `convert_to_number` function where you can convert it to any representation you like.

PR #665, opened 2015-10-13
adding max_connections option to StrictRedis Class

ohadpartuck:
adding max_connections option to StrictRedis Class. This is totally needed to control this option.


ohadpartuck:
@andymccurdy Please Review

andymccurdy:
Thanks. Most people who need this option create connection pools manually. But allowing it to be specified at the client level could save a line of code in some situations.

The default value of `max_connections` should be `None` however. I'll update it after the merge.

ohadpartuck:
cool

PR #669, opened 2015-10-26
parsing 'max number of clients reached' as ConnectionError

th13f:
Fix for https://github.com/andymccurdy/redis-py/issues/668 issue.


andymccurdy:
thanks!

PR #674, opened 2015-11-06
Fix: retry_on_timeout test condition was not correct

gst:


gst:
oups my bad, problem was not because of the not but because of the exception not being a TimeoutError instance. while what I have got at this point : 

Error connecting to localhost:6379. **timed out**.

Gonna make another PR..

PR #675, opened 2015-11-06
Fix: must raise socket.timeout as TimeoutError

gst:
otherwise retry_on_timeout feature can't detect the timeout.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -491,6 +491,8 @@ class Connection(object):
                     sock.close()
 
         if err is not None:
+            if isinstance(err, socket.timeout):
+                raise TimeoutError(err)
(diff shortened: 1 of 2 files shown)


gst:
> otherwise retry_on_timeout feature can't detect the timeout.

in client.execute_command() function.

andymccurdy:
@gst thanks for raising this error. I fixed in a slightly different way that's identical to how other TimeoutErrors are raised.

gst:
you're welcome.

effectively the place where your raise the TimeoutError is better.

What about the 3 tests I've just added ? without the fix/change they'd fail normally.. so maybe it'd be useful to keep them ?

PR #679, opened 2015-11-25
fix getitem syntax for redis get when value is the empty string

candeira:
When using `__getitem__` syntax to retrieve an empty string, an exception is produced, just as if the key were unset:

```
>>> r = Redis(**connection_data)
>>> print(r.get('bah'))
None
>>>r['bah']
KeyError: 'bah'
>>> r['bah'] = ""
>>> r.get('bah')
b''
>>> r['bah']
KeyError: 'bah'
```

The behaviour should be the same as for any other set value, and the exception only produced for unset keys:

```
>>> r['bah'] = ""
>>> r.get('bah')
b''
>>> r['bah']
b''
```

This patch provides a test and a fix.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -885,7 +885,7 @@ class StrictRedis(object):
         doesn't exist.
         """
         value = self.get(name)
-        if value:
+        if value is not None:
             return value
         raise KeyError(name)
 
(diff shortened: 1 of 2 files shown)


andymccurdy:
Thanks!

Issue #682, opened 2015-11-26
I found this unused variable in redis-py library


PramodBisht:
Correct me if I am wrong but can't see the any usage of `command_name` argument in `get_connection` method of `BlockingConnectionPool` and `ConnectionPool` class. It is passed unneccessarily from everywhere it is called.

andymccurdy:
Copied from #683:

The arguments are there on purpose. They allow user-authored subclasses to vary the connection returned based on the command name and/or keys involved.

A great example of this would be a ConnectionPool that automatically routes read-only queries to a set of slave servers and write queries to a master server. In this scenario get_connection needs to know the command name. Key names might be used to implement a ConnectionPool that supports client-side clustering via a hash ring.

PR #683, opened 2015-11-27
Fixed Issue #682, Deleted unwanted code

PramodBisht:
Hey, I checked code and found that,  `get_connection` method had unnecessary argument which it was not using in its implementation, So, I deleted those arguments in method definition and call.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -567,7 +567,7 @@ class StrictRedis(object):
         "Execute a command and return a parsed response"
         pool = self.connection_pool
         command_name = args[0]
-        connection = pool.get_connection(command_name, **options)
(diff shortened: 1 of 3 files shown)


andymccurdy:
The arguments are there on purpose. They allow user-authored subclasses to vary the connection returned based on the command name and/or keys involved.

A great example of this would be a ConnectionPool that automatically routes read-only queries to a set of slave servers and write queries to a master server. In this scenario `get_connection` needs to know the command name. Key names might be used to implement a ConnectionPool that supports client-side clustering via a hash ring.

andymccurdy:
By the way, I should also mention that I appreciate the help.

PR #684, opened 2015-11-27
FIX #651 try next Sentinel node on TimeoutError

### LINKED ISSUE #651: Next sentinel host is not contacted after socket timeout

blopker:
A failure we are seeing in production is that Sentinel nodes are timing out instead of just throwing ConnectionErrors. This keeps the clients from getting the new master and brings down the app.

Same PR as #652, but with tests.

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -4,7 +4,8 @@ import weakref
 
 from redis.client import StrictRedis
 from redis.connection import ConnectionPool, Connection
-from redis.exceptions import ConnectionError, ResponseError, ReadOnlyError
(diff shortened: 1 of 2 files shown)


andymccurdy:
Thanks!

PR #685, opened 2015-12-01
Redis cluster commands

iandyh:
In some cases, you still need `redis-py` to perform cluster related commands, e.g. `slaves`, `info`, `replicated`. This PR implements all the cluster commands (`cluster sub-command args`).

Because of the special reply from Redis Cluster, the tests mock out all the responses from Redis server and test it against mock response. 

Cheers.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -279,6 +279,35 @@ def parse_slowlog_get(response, **options):
     } for item in response]
 
 
+def parse_cluster_info(response, **options):
+    return dict([line.split(':') for line in response.splitlines() if line])
+
+
+def _parse_node_line(line):
+    line_items = line.split(' ')
+    node_id, addr, flags, master_id, ping, pong, epoch, \
+        connected = line.split(' ')[:8]
+    slots = [sl.split('-') for sl in line_items[8:]]
+    node_dict = {
+        'node_id': node_id,
+        'flags': flags,
+        'master_id': master_id,
+        'last_ping_sent': ping,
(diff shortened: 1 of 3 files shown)


iandyh:
Currently the CI fails because the lengthy mocked Redis raw response breaks the pep8 check. I prefer to keep the raw response as it is with an ignore comment. I could also break down the long string into tuples and then join them later. It's up to maintainer's decision.

iandyh:
@andymccurdy Hi, is there a timetable for review the PR? It is quite essential for anyone who is running Redis Cluster.

iandyh:
@andymccurdy Ping. I've pep8ed the code and please review the PR.

andymccurdy:
Hey, sorry for the delay. This seems mostly fine. It might be nice to create separate methods for each cluster command (similar to how the sentinel commands are implemented.

I'll go ahead and merge this anyway. Thanks for doing this.

iandyh:
@andymccurdy That was quick. Thanks for the merge! I'll consider the advice. Current design aligns with the cluster command in ruby client.

iandyh:
@drabaioli Hi! @Grokzen has explained most of the things. I'll add a bit more. Basically, even though redis-py-cluster provides most of the features you need to talk to Redis Cluster, you still need the ability to talk to one specific node inside the cluster. e.g. `cluster replicate node-id` will configure a node as a slave of another node and clearly you don't want to send this kind of command to every node in the cluster.

PR #691, opened 2015-12-10
Added tests for transaction() and transaction(value_from_callable).

nfvs:
Both of these were mostly inspired by the awesome [fakeredis](https://github.com/jamesls/fakeredis).


andymccurdy:
Thanks! I added your test in 9cbb48a. Sorry this took so long.

PR #692, opened 2015-12-10
Add differentiation for lock not owned errors

harlowja:
It is quite useful to be able to distinguish between
between general lock errors and lock errors that are releated
to ownership lose (and handle the cases differently).

To be able to do this add a new lock error subclass and
use it in the locations where lock error was raised previously
when ownership lose was detected.

### DIFF
--- a/redis/exceptions.py
+++ b/redis/exceptions.py
@@ -69,3 +69,7 @@ class LockError(RedisError, ValueError):
     # NOTE: For backwards compatability, this class derives from ValueError.
     # This was originally chosen to behave like threading.Lock.
     pass
+
+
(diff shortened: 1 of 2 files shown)


harlowja:
Travis failued due to the following, probably just needs to be refreshed (told to run again):

```
Cloning into 'andymccurdy/redis-py'...

fatal: unable to access 'https://github.com/andymccurdy/redis-py.git/': Couldn't resolve host 'github.com'
```

andymccurdy:
This was resolved in redis-py 3.0. Sorry that this wasn't closed when that release happened.

PR #696, opened 2015-12-26
added after-script hook for investigating redis version.

mumumu-test:


mumumu-test:
opend by mistake, sorry.

PR #695, opened 2015-12-26
Added Redis-Geo functionality + tests

Darkless012:
Added geospatial functionality for beta version of Redis (3.2.0)
http://redis.io/commands/geoadd, geodist, geohash, geopos, georadius, georadiusbymember

PR #697, opened 2015-12-27
Update test environment to redis 3 and python 3.5

mumumu:
Redis 3.2 is currently testing status (RC1) and we already get 3.x specific feature PR like Redis-Geo ( #695 ), but redis-py test environment is old. (Redis 2.8.9, and Python 2.6 to 3.3)

So, it's about time to update redis-py test environment to the latest. 
How do you think?


Grokzen:
Would it be possible to have travis run the test suite with both redis 2.8.x, 3.0.x and 3.2.x for example? The matrix would grow alot but it might be worth it if possible.

andymccurdy:
This has been resolved long ago.

PR #698, opened 2015-12-27
implemented REPLACE modifier of restore command

mumumu:
http://redis.io/commands/restore

implemented REPLACE modifier for RESTORE command on Redis 3.0 or greater. Now we can safely replace value even if key already exists.

Note: This PR works on Redis 3.0 environment. ( #697 )


andymccurdy:
thanks! merged.

PR #718, opened 2016-03-11
HMGET variant returning a dict.

mastermatt:
When calling HMGET, I almost always need the results wrapped up with the keys in a dict. This would simply allow the caller to designate their desired return type.

Test included.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1954,6 +1954,11 @@ class StrictRedis(object):
         args = list_or_args(keys, args)
         return self.execute_command('HMGET', name, *args)
 
+    def hmget_dict(self, name, keys, *args):
+        "Return a dict with keys equal to ``keys`` from within hash ``name``"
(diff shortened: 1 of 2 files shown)


andymccurdy:
@mastermatt 

Thanks for the patch. However, this doesn't work in pipelines. Anytime you want to do casting/converting of the response type you need to do it in a response callback. For example, something like this would work:

```python

class StrictRedis(object):
    ...
    def hmget(self, name, keys, *args):
        # note that any kwargs will get passed to response callbacks
        return self.execute_command('HMGET', name, *args, keys=args)

def hmget_dict(response, **options):
    keys = options['keys']
    return dict(izip(keys, response))


r = StrictRedis(...)
r.set_response_callback('HMGET', hmget_dict)
my_dict = r.hmget('my-hash', 'key1', 'key2')
```

PR #720, opened 2016-03-16
Make it easier to use a custom ConnectionPool

friedcell:
Adding a property to make it easier to switch ConnectionPool and augment Connection down the line to customize decoding in Connection or Parser.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -441,8 +441,8 @@ class StrictRedis(object):
                  socket_connect_timeout=None,
                  socket_keepalive=None, socket_keepalive_options=None,
                  connection_pool=None, unix_socket_path=None,
-                 encoding='utf-8', encoding_errors='strict',
-                 charset=None, errors=None,
+                 pool_class=ConnectionPool, encoding='utf-8',
+                 encoding_errors='strict', charset=None, errors=None,
                  decode_responses=False, retry_on_timeout=False,
                  ssl=False, ssl_keyfile=None, ssl_certfile=None,
                  ssl_cert_reqs=None, ssl_ca_certs=None,
@@ -491,7 +491,7 @@ class StrictRedis(object):
                         'ssl_cert_reqs': ssl_cert_reqs,
                         'ssl_ca_certs': ssl_ca_certs,
                     })
-            connection_pool = ConnectionPool(**kwargs)
+            connection_pool = pool_class(**kwargs)
         self.connection_pool = connection_pool
         self._use_lua_lock = None
 


andymccurdy:
I don't understand why this is needed. Can you elaborate please?

friedcell:
It makes it easier to modify classes down the line (ConnectionPool and/or Connection) without losing any features of the simple redis_conn = StrictRedis(...).

The reason for using an augmented Connection class in my case is to change the way it decodes responses in py3.

andymccurdy:
The `StrictRedis.__init__` setup is really only for normal, out of the box setups. If you're modifying Connection classes etc., which is perfectly fine to be doing, you can get more control by creating a connection pool manually:

```python
r = redis.StrictRedis(connection_pool=YourCustomPool(host, port, other_options...))
```

friedcell:
I know, but that means you lose the whole convenience that __init__ provides. Since ConnectionPool already has a connection_class param in __init__ it makes sense to provide it in StrictRedis.

friedcell:
Your suggestion is good for when you're connecting via code but is much more complicated when you're trying to provide the connection info through a settings object. It also means that you can't just switch the Redis connection class (StrictRedis for CustomRedis), but you have to change the way the connection is created.

I tried to follow the same style redis-py already uses in ConnectionPool that has a `connection_class=Connection` and in Connection with `parser_class=DefaultParser`. It seems like a useful addition with an upside for some and no downsides.

friedcell:
Any chance this ever gets in?

I have code on py3 that replaces read_response in Connection but can't make it a reusable class that extends StrictRedis unless I duplicate the whole `__init__` method of StrictRedis (some 50 lines and a whole lot of ugliness more than with this fix)...

andymccurdy:
Again, I don't see the issue here using what already exists. Maybe a code sample of what you're trying to do would be helpful. Below is how I've personally customized this stuff before at previous companies. Using this pattern there's exactly only place to change any of the classes (Client, Connection, or ConnectionPool) in the entire code base.

```python

### my_custom_redis.py ############################
class MyCustomClient(redis.StrictRedis):
    pass

# optional
class MyCustomConnectionPool(redis.ConnectionPool):
    pass

# optional
class MyCustomConnection(redis.Connection):
    pass

### settings.py #############################
from my_custom_redis import MyCustomConnection

REDIS_SETTINGS = {
    'hostname': 'myredis.mycompany.com',
    'connection_class': MyCustomConnection
    ...
}


### redis_connection.py ################################
from my_custom_redis import *
from settings import REDIS_SETTINGS

pool = MyCustomConnectionPool(**REDIS_SETTINGS)
client = MyCustomClient(connection_pool=pool)



### app.py ##################################

from redis_connection import client

# execute commands on the client class
client.get('foo')
```

andymccurdy:
Also, what are you finding in Python 3 that requires modifying `read_response`? redis-py _should_ be completely compatible with py2/py3.

friedcell:
Currently redis has two options in python 3:
1) have every string returned as byte
2) have every string converted.
Problem is if you use bitops strings (that end up being invalid utf-8). Sample:
```
r.setbit("b", 0, 1)
r.get("b")
  File "/not-relevant/lib/python3.4/site-packages/redis/client.py", line 880, in get
    return self.execute_command('GET', name)
  File "/not-relevant/lib/python3.4/site-packages/redis/client.py", line 573, in execute_command
    return self.parse_response(connection, command_name, **options)
  File "/not-relevant/lib/python3.4/site-packages/redis/client.py", line 585, in parse_response
    response = connection.read_response()
  File "/not-relevant/lib/python3.4/site-packages/redis/connection.py", line 577, in read_response
    response = self._parser.read_response()
  File "/not-relevant/lib/python3.4/site-packages/redis/connection.py", line 280, in read_response
    response = response.decode(self.encoding)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```
One option is to return bytes when conversion fails. Second - that I use now - is to provide a custom ConnectionPool class that knows which values not to convert depending on the key.

The code you wrote is only interchangeable with StrictRedis(**REDIS_SETTINGS) when these settings don't include variables that the code in StrictRedis.__init__ affects are not present (charset, errors, unix_socket_path, connection_class) and what I'm targeting is full interoperability of code - only change is which class you import.

friedcell:
Any progress on this?

The proposed change does not affect anyone negatively, but gives a significant advantage in cases where custom implementations are needed. I see no good reason to not put this in ASAP.

friedcell:
@andymccurdy ping?

friedcell:
@andymccurdy ?

laixintao:
@friedcell 
I got a different result run your code:

```
Python 3.6.1 (default, Apr  4 2017, 09:40:21)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import redis

In [2]: r = redis.StrictRedis()

In [3]: r.setbit("b",0,1)
Out[3]: 0

In [4]: r.get("b")
Out[4]: b'\x80'
```

Still I don't understand your requirement exactly, keeping returned string as bytes isn't enough

friedcell:
@laixintao the issue is when you have conversion to strings enabled.

The point is that since quite a few classes are taken from defaults in __init__ functions (eg. connection_class=Connection in ConnectionPool, queue_class=LifoQueue in BlockingConnectionPool, parser_class=DefaultParser in Connection), by using ConnectionPool the same way, there is more extensibility and zero downsides for other users.

laixintao:
@friedcell I understand you now, and I agree that it won't influence others, but make code kind of harder to read, I think. Leave @andymccurdy to decide.

friedcell:
@laixintao that would be a perfectly good argument if the same approach wasn't already used in other places in the library.

Let's wait for @andymccurdy then...

andymccurdy:
My initial (and current) thinking is that it's confusing to accept a ConnectionPoolClass and an instance-of-ConnectionPool in the same callable. It's a minor complaint, but it's the absolute first callable someone using redis-py is exposed to.

All this seems to be stemming from you using an option (`decode_responses=True`) that applies to all connections when in fact you don't want _every_ response to be decoded. Special casing a ConnectionPool to avoid decoding certain responses seems like a bad idea overall. I appreciate its pragmatism, but it doesn't scale very well.

I'd rather see us build in the ability for finer-grain control of response decoding. The first thing that comes to mind is using a context manager to temporarily turn on (or off) response decoding. e.g., something ilke:

```python
r = redis.StrictRedis(..., decode_responses=False)
my_bytes = r.get('foo')
with r.response_strings:
    my_string = r.get('bar')

### or the inverse

r = redis.StrictRedis(..., decode_responses=True)
my_bytes = r.get('foo')
with r.response_bytes:
    my_string = r.get('bar')
```

I'm not committed to the client names (`response_strings` and `response_bytes`). But such a solution seems to be more general purpose and useful to others.

Also, making this works with pipelines and pubsub might be difficult, but I *think* doable.

Thoughts?

friedcell:
That sounds like a good solution - not one I would dare to tackle though...

PR #638, opened 2015-07-20
[WIP] Zadd options

chester89:

PR #721, opened 2016-03-25
ZADD options (Redis 3.0.2 or greater)

Vic020:
Hello,

First of all, this PR is not a complete patch for zadd options.

AND, Yes, i know this -> #638  ,really excellent.

but, it's look like not good for
1.old code because of this change
2.Strict Redis
3.Easy to use

My purpose is like this (and my temporary code already use this change):
`redis.zadd('incr', 1.1, 'key')`

SO, i just give a idea( if accepted, i could code more), forgive me.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1626,6 +1626,14 @@ class StrictRedis(object):
         """
         pieces = []
         if args:
+          
+            args = list(args)
+            for each in args:
+                if 'INCR' in str(each).upper():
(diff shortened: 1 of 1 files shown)


Vic020:
@andymccurdy

stdrickforce:
+1

andymccurdy:
Fixed in redis-py 3.0

Vic020:
Welcome back!

PR #723, opened 2016-03-29
Extend ConnectionPool.to_url to parse querystring arguments to correct type.

timsavage:
Previously if a value for socket_timeout was supplied as part fo the URL an error would be raised when a socket was created with an invalid type, this change fixes that by parsing `socket_timeout`, `socket_connect_timeout` to float values.
In addition the boolean values `socket_keepalive` and `retry_on_timeout` are parsed to bool types taking into account the usage of True/False, Yes/No strings.


andymccurdy:
Looks good! Thanks for the help.

PR #737, opened 2016-05-17
add daemon option to PubSubWorker so ctrl-c will work

kainswor:
There isn't a good way to create a pubsub thread which can be killed with ctrl-c. For previous projects I've had to start the thread from a subthread just to get the effect of thread.daemon=True.


andymccurdy:
Thanks!

PR #742, opened 2016-05-24
Use `if` instead `try` when get avaliable connection

MrKiven:


andymccurdy:
Your change makes the ConnectionPool not threadsafe.

MrKiven:
> Your change makes the ConnectionPool not threadsafe.

I'm newer, please tell why, thanks.

andymccurdy:
With your code, the `_available_connections` will be looked at twice. Once in the `if` check, and again in the `.pop()`. When running multiple threads, two different threads could be trying to acquire a connection at the same time. For example:

```
# assume there's 1 connection in _available_connections.

thread-1: tries to acquire a connection. "if self._available_connections" passes because there's one connection available.
thread-2: tries to acquire a connection. "if self._available_connections" passes because there's one connection available.
thread-1: executes "self._available_connections.pop()", which removes the connection from the list.
thread-2: executes "self._available_connections.pop()", but there's no connection available because thread-1 already took it. exception gets raised, program crashes.
```

MrKiven:
oh, I see. 
I'm using redis-py in my pro first time, so..I may have some misunderstand.
Thanks again

PR #744, opened 2016-05-29
Check also `int` not only `long` for PyPy.

youknowone:
This is a patch to enhance PyPy compatibility.
I don't know well about the code - but it seems it is related to binary unpacking.
In some case of that, PyPy can generate `int` instead of `long`.


yoloseem:
How about using `numbers.Integral`

andymccurdy:
thanks!

PR #745, opened 2016-05-30
Client can create blocking pool with `from_url`

MrKiven:


pfreixes:
Hi, why don't inject the instance `BlockingConnectionPool` using the `connection_pool` keyword of the `StrictRedis` class ?

```python
>>> from redis import client
>>> from redis.connection import BlockingConnectionPool
>>> c = client.StrictRedis(connection_pool=BlockingConnectionPool.from_url("redis://localhost:6379/0"))
>>> c.set("foo", 1)
True
>>> c.get("foo")
'1'
```

This param is ready to get no only pools implemented by redis-py, also other pools implemented  by the user - inherited from the `ConnecitonPool` - can be injected.

MrKiven:
@pfreixes yeah, obviously！ 
I mean `from_url` func can also use custom pools implemented by others, not just `c = Redis()` or `c = StrictRedis() `

andymccurdy:
I think what @pfreixes is asking is why do we need an extra argument to this method? Everything you need is already available by calling `BlockingConnectionPool.from_url(...)` and passing the return value to the `connection_pool` argument of a client class.

It's actually (slightly) more to type your way:

```python
# already possible with no code changes
r = StrictRedis(connection_pool=BlockingConnectionPool.from_url('redis://localhost:6379/0'))

# your way
r = StrictRedis.from_url('redis://localhost:6379/0', connection_pool_class=BlockingConnectionPool)
```

pfreixes:
Right now the redis client allows to the developer use the BlockingConnectionPool through the `connection_pool` keyword, therefore the developer is not restricted on that.

Accepting this MR  *redis-py* will get a bit of more technical debt, IMHO the current interfaces are a bit messy and they should be revisited to get a more friendly interface.

andymccurdy:
I don't think this makes sense to merge for the reasons above. You can already do everything you want to do with the existing API.

PR #747, opened 2016-06-05
Implemented support for the GEO commands for Redis 3.2.0

pfreixes:
The following commands have been implemented GEOADD, GEODIST,
GEOHASH, GEOPOS, GEORADIUS, GEORADIUSBYMEMBER.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -308,6 +308,37 @@ def parse_cluster_nodes(response, **options):
     return dict([_parse_node_line(line) for line in raw_lines])
 
 
+def parse_georadius_generic(response, **options):
+    if options['store'] or options['store_dist']:
+        # `store` and `store_diff` cant be combined
+        # with other command arguments.
+        return response
+
+    if type(response) != list:
+        response_list = [response]
+    else:
+        response_list = response
+
+    if not options['withdist'] and not options['withcoord']\
+            and not options['withhash']:
+        # just a bunch of places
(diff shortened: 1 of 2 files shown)


andymccurdy:
Thanks. I noticed you left all the return values as strings. It might be nice to convert some of them to Python types. Specifically:

GEOADD returns an int
GEODIST returns a double
GEORADIUS with the WITHDIST option returns a double

Thoughts?

pfreixes:
@andymccurdy fair enough, sorry for the delay I was busy with some work stuff. Before tomorrow it will be done.

Cheers

pfreixes:
A bit later but at last committed a new version where geo commands return Python native types, you can found these types in the UnitTests [1]

[1] https://github.com/andymccurdy/redis-py/pull/747/commits/f1abb79fb28ce58035e0a1d5c93688fb4d37f968

andymccurdy:
Looks great, thanks. I'm merging now.

Question: Should we be returning doubles rather than floats? The GEO command docs [1] indicate the redis server returns doubles. Do Python floats offer enough precision?

[1] http://redis.io/commands/geodist

benjolitz:
@andymccurdy Python floats are internally doubles.

You can see it here: https://github.com/python/cpython/blob/c30098c8c6014f3340a369a31df9c74bdbacc269/Include/floatobject.h#L17

PR #751, opened 2016-06-12
Improve performance for Python 3.x

cjsimpson:
This issue attempts to improve performance under Python 3.x while either maintaining or slightly improving 2.x performance. 

Some of the reasons Python 3.x is slow than 2.x:

- Strings need to be encoded into bytes before sending to redis
- Many of these strings are encoded over and over again (get, set, incr, etc..)

These redis commands are represented by the connection.Token object. To speed things up, I've implemented a cache of Token objects and modified the Token object to store its encoded value. This greatly reduces the number of times redis commands need to be encoded. It also reduces the number of times Token objects are created, although that has a smaller impact.

I also wrote a basic benchmark (basic_operations.py) modeled on redis-benchmark (although much more limited). It tests the performance of some of the basic commands.

```
usage: basic_operations.py [-h] [-n] [-P] [-s]

optional arguments:
  -h, --help  show this help message and exit
  -n         Total number of requests (default 100000)
  -P         Pipeline <numreq> requests. Default 1 (no pipeline).
  -s         Data size of SET/GET value in bytes (default 2)
```

Testing on my machine has shown a 10-20 % improvement for Python 3.5. It should make the biggest improvement to workloads where there are small and frequent requests to redis.

### DIFF
--- a/benchmarks/basic_operations.py
+++ b/benchmarks/basic_operations.py
@@ -1,0 +1,197 @@
+from __future__ import print_function
+import redis
+import time
+import sys
+from functools import wraps
+from argparse import ArgumentParser
+
+if sys.version_info[0] == 3:
+    long = int
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('-n',
+                        type=int,
+                        help='Total number of requests (default 100000)',
+                        default=100000)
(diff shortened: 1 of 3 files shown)


andymccurdy:
This looks great. Thanks!

PR #755, opened 2016-06-23
redis.connection.ConnectionPool.from_url for wrong or missing schema …

### LINKED ISSUE #728: redis.connection.ConnectionPool.from_url should raise an exception when passed an obviously bad URL value
I'm using version 2.10.5 of redis-py.

The documentation of [`from_url()`](https://github.com/andymccurdy/redis-py/blob/2.10.5/redis/connection.py#L739) is pretty clear:

> ```
>     Three URL schemes are supported:
>         redis:// creates a normal TCP socket connection
>         rediss:// creates a SSL wrapped TCP socket connection
>         unix:// creates a Unix Domain Socket connection
> ```

There are only 3 schemes supported, so anything falling outside of these schemes should result in an immediate error. However, as of 2.10.5, if passed schemes that it does not understand, it will silently misinterpret them. For instance, passing `"/dev/null"` will result in a client connecting to the redis instance running on localhost at port 6379. Same with passing `"uni:/dev/null"` (this is a case simulating a typo in the scheme). If I pass `"unix:/dev/null"`, then I will get a connection error because the URL will be interpreted correctly. Here's a session illustrating what I'm taking about:

```
>>> from redis.client import StrictRedis
>>> c = StrictRedis.from_url("/dev/null")
>>> c
StrictRedis<ConnectionPool<Connection<host=None,port=6379,db=0>>>
>>> c.echo("foo")
'foo'
>>> c = StrictRedis.from_url("uni:/dev/null")
>>> c.echo("foo")
'foo'
>>> c = StrictRedis.from_url("unix:/dev/null")
>>> c.echo("foo")
  File "/home/ldd/Documents/mangalam/btw/software/btw_env/local/lib/python2.7/site-packages/redis/client.py", line 644, in echo
    return self.execute_command('ECHO', value)
  File "/home/ldd/Documents/mangalam/btw/software/btw_env/local/lib/python2.7/site-packages/redis/client.py", line 578, in execute_command
    connection.send_command(*args)
  File "/home/ldd/Documents/mangalam/btw/software/btw_env/local/lib/python2.7/site-packages/redis/connection.py", line 563, in send_command
    self.send_packed_command(self.pack_command(*args))
  File "/home/ldd/Documents/mangalam/btw/software/btw_env/local/lib/python2.7/site-packages/redis/connection.py", line 538, in send_packed_command
    self.connect()
  File "/home/ldd/Documents/mangalam/btw/software/btw_env/local/lib/python2.7/site-packages/redis/connection.py", line 442, in connect
    raise ConnectionError(self._error_message(e))
redis.exceptions.ConnectionError: Error 111 connecting to unix socket: /dev/null. Connection refused.
```

Note that amending the URLs in the illustration above to add the double forward slash that _should_ appear under the scheme (so that they become `"uni:///dev/null"` and `"unix:///dev/null"` ) changes nothing except what socket path is reported in the exception in the last case.

depaolim:
…[fix #728]

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -859,7 +859,7 @@ class ConnectionPool(object):
                 'connection_class': UnixDomainSocketConnection,
             })
 
-        else:
+        elif url.scheme in ['redis', 'rediss']:
             url_options.update({
                 'host': hostname,
                 'port': int(url.port or 6379),
@@ -876,6 +876,8 @@ class ConnectionPool(object):
 
             if url.scheme == 'rediss':
                 url_options['connection_class'] = SSLConnection
+        else:
+            raise ValueError('invalid scheme "%s"' % url.scheme)
 
         # last shot at the db value
         url_options['db'] = int(url_options.get('db', db or 0))
--- a/tests/test_connection_pool.py
+++ b/tests/test_connection_pool.py
@@ -407,6 +407,18 @@ class TestSSLConnectionURLParsing(object):
         assert pool.get_connection('_').cert_reqs == ssl.CERT_REQUIRED
 
 
+class TestConnectionPoolWeirdURLParsing(object):
+    def test_unknown_scheme_raise_exception(self):
+        with pytest.raises(ValueError) as excinfo:
+            redis.ConnectionPool.from_url('uni:/dev/null')
+        assert 'invalid scheme "uni"' in str(excinfo.value)
+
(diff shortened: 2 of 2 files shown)


gyermolenko:
target issue #728 is already fixed . Unfortunately this PR is obsolete

andymccurdy:
This was resolved as part of redis-py 3.0.

PR #759, opened 2016-07-10
Fix transposition of GEORADIUS arguments

alexdebrie:
Latitude and longitude are transposed in the `georadius` [method signature](https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L2109) (compare to [official Redis docs](http://redis.io/commands/georadius)). No effect on functionality, assuming users were using the order specified by the official docs, but it makes the [redis-py docs](http://redis-py-doc.readthedocs.io/en/master/#redis.StrictRedis.georadius) more clear.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2106,7 +2106,7 @@ class StrictRedis(object):
         """
         return self.execute_command('GEOPOS', name, *values)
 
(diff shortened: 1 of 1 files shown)


andymccurdy:
thanks!

PR #763, opened 2016-07-14
Fix multiple script loading in load_scripts

ereOn:
I've been playing with LUA scripts and pipelines and noticed that if the same script is called several times in a given pipeline (with or without different parameters), it may be checked and loaded several times before the `MULTI`/`EXEC` block.

This PR fixes that.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2816,17 +2816,24 @@ class BasePipeline(object):
 
     def load_scripts(self):
         # make sure all scripts that are about to be run on this pipeline exist
-        scripts = list(self.scripts)
+        scripts = {}
+
+        for script in self.scripts:
+            scripts.setdefault(script.sha, []).append(script)
+
         immediate = self.immediate_execute_command
-        shas = [s.sha for s in scripts]
+        shas = [sha for sha in scripts if sha]
         # we can't use the normal script_* methods because they would just
         # get buffered in the pipeline.
         exists = immediate('SCRIPT', 'EXISTS', *shas, **{'parse': 'EXISTS'})
         if not all(exists):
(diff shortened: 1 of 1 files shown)


andymccurdy:
Are you using multiple instances of the same script object? Script objects are designed such that you create them once and use the same instance over and over again.

ereOn:
@andymccurdy Yes, I am indeed using several instances of the same script. My scripts are generated (by [redis-lua](https://github.com/ereOn/redis-lua)) and it can happend that two equivalent scripts, with identical SHA, are represented by different instances.

Let me know if my explanations are unclear.

andymccurdy:
`Script` instances remember the client they were created with so that users don't have to explicitly pass `client=my_client` for every invocation. Omitting the `client` will execute the script using the same client/connection pool that was used to create the script instance.

The same script instance can be used with many different clients and pipelines. It's safe to think of a script object as any other Python callable. Define it once in your code and use it over and over again, optionally with different clients/pipelines.

All that said, looking at the code I would think that any duplicate loading of the Lua code would be coming from the `script_load_for_pipeline` method (https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L2881). That is called from the script objects `__call__`.

ereOn:
@andymccurdy I had completely missed the `client` attribute when calling the script, hence my incorrect need for several instances.

Indeed, once I keep the same reference around, everything works as intended and I don't have duplicated loads anymore.

I'm closing this PR as it doesn't bring much.

Thanks for your explanations, I always suspected I was missing something. On my way to patch [redis-lua`](https://github.com/ereOn/redis-lua) now !

PR #765, opened 2016-07-29
'with' statement for PubSub

sstepashka:
Simple implementation of 'with' statement of PubSub.


sstepashka:
Hi @andymccurdy,
Could you merge it?

PR #767, opened 2016-08-09
Add thread to maintain avaliable connections.

MrKiven:


MrKiven:
A idea to keep connections in backend, @andymccurdy

PR #768, opened 2016-08-11
fix bug: KeyError for description_format

Brightcells:
```
redis.StrictRedis(connection_pool=redis.ConnectionPool(**connect_settings))
```

When `connect_settings` doesn't have key 'host', 'port', or 'db', will raise `KeyError:`

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -911,6 +911,10 @@ class ConnectionPool(object):
         if not isinstance(max_connections, (int, long)) or max_connections < 0:
             raise ValueError('"max_connections" must be a positive integer')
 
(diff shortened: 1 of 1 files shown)


gyermolenko:
it is hard to tell from description but if `description_format` from title is [this one](https://github.com/andymccurdy/redis-py/blob/master/redis/connection.py#L395)
than this is no longer an issue - all format arguments are passed to `__init__` with defaults.

andymccurdy:
I believe this had to do with the `repr()` of Connection/ConnectionPool/Redis instances which has been fixed since 3.0. Closing this for now. If there's still an issue with this, please feel free to reopen.

PR #770, opened 2016-08-15
set weight of sentinel nodes balancer, avoid the first sentinel node too busy

BingerCoder:
As the order of sentinel in queue,the first sentinel node always received more connections,so shuffle the sentinel to make balance as same as possible.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -830,12 +830,12 @@ to discover Redis nodes. You need to have at least one Sentinel daemon running
 in order to use redis-py's Sentinel support.
 
 Connecting redis-py to the Sentinel instance(s) is easy. You can use a
-Sentinel connection to discover the master and slaves network addresses:
+Sentinel connection to discover the master and slaves network addresses.Also you can set weight of each instance optionally:
 
 .. code-block:: pycon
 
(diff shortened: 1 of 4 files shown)


5977862:
In our project, use three sentinel nodes,and so many connections in first sentinel node,the other two sentinel nodes almost haven't no connections.

SO 

1.First node is very busy and left nodes are very free.High load make more wait tasks,more wait tasks make more load~ vicious cycle~
2.If first node crash down, all connections are unused,but if we balance them,just part of them are unused.

I think it is useful~

MrKiven:
there more other things to do, like:

1. set a weight for every nodes
2. record connections, then pick least connections node
3. just `random.choice`

just my point 😄

5977862:
I think the first point is good,set weight for nodes can choice best one we knew.     

If use random,connections of every node may as many as same.Or we can set weight for every nodes,it still be a predictable rate~

So I don't support the second point~   

😋

MrKiven:
> I think the first point is good,set weight for nodes can choice best one we knew.

yeah..weighted round robin..

5977862:
thank you for your advice,I will change this pr ~

PR #776, opened 2016-08-25
allow client to send remapped command

iandyh:
It's useful for redis-py to manage production Redis instances. 

usage:

```
renamed_commands = {
    'CONFIG GET': 'xxconfig get',
}
r = Redis(host='redis', port=6379, renamed_commands=renamed_commands)
```

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -443,7 +443,7 @@ class StrictRedis(object):
             'CLUSTER SETSLOT': bool_ok,
             'CLUSTER SLAVES': parse_cluster_nodes,
             'GEOPOS': lambda r: list(map(lambda ll: (float(ll[0]),
-                                         float(ll[1])), r)),
+                                                     float(ll[1])), r)),
             'GEOHASH': lambda r: list(map(nativestr, r)),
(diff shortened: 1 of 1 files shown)


andymccurdy:
I don't understand why you would want to do this. Can you elaborate?

iandyh:
Sure. For some companies like us, we renamed dangerous commands: `shutdown` -> `nevershutdown`. After renaming, server will not process any dangerous command sent from application side. However, we still need to perform certain commands for administration purposes. This is why I made the PR.

andymccurdy:
I think this is better left to a special case in your code.

For what it's worth, if I needed something like this, I'd do the check within the `Connection` class so commands sent within pipelines are sanitized as well. And I'd also just raise an exception that explains that the command isn't available rather than renaming the command, which is going to produce a generic `RedisError` with a much more cryptic error message.

iandyh:
@andymccurdy Adding renaming into `Connection` class would be needed. The exceptions are not needed since a `CommandError` will be thrown anyway and that's enough. 

IMO, the changes are still helpful otherwise it will be difficult for administrators to perform certain operations.

PR #779, opened 2016-08-31
added default optional parameter to get command (as in Python dicts)

szborows:


szborows:
my changes don't seem to be related to reported failures, are they?

redis/client.py
```diff
@@ -959,11 +959,12 @@ def expireat(self, name, when):
...
         return self.execute_command('EXPIREAT', name, when)
 
-    def get(self, name):
+    def get(self, name, default=None):
         """
         Return the value at key ``name``, or None if the key doesn't exist
         """
-        return self.execute_command('GET', name)
+        return_value = self.execute_command('GET', name)
+        return [return_value, default][return_value is None]
```
MrKiven: why not
```python
return return_value or default
```
?

andymccurdy:
You can't modify return values within command functions like this. They'll get lost in pipelines. This type of stuff needs to done in a response callback.

redis/client.py
szborows: because `[], {}, ''` etc. would also yield `True`, so for empty key it could return the default instead of empty string

szborows:
@andymccurdy oh, I see.

I was thinking how to redesign it and came up with an idea of stack of default values. That would work both for `.get`, `__getitem__` and pipelines. However:
* this stack must be kept somewhere. `StrictRedis` is natural candidate, but that would mean introducing state to this class. I don't think it's way to go.
* response callbacks aren't bound to the `StrictRedis` instance, hence all of them must be free functions. My change would have violated this, because of this introduced state.
* introducing state also would make class non-thread-safe. Some locking could possibly mitigate this issue but at what cost...

Do you have any other ideas how to solve this issue? I've also discovered `set_response_callback` which provides a way to get what I need without changing library. Maybe solving this inside the library is overengineering?

andymccurdy:
Looking at this again, I don't think that adding this behavior to redis-py is necessary. It's quite simple to do something like:

```python
my_value = r.get('foo') or 'default value'
```

Thanks for the suggestion though!

Issue #732, opened 2016-05-06
Race condition in connection disconnect


jpulec:
It seems like there is a race condition in the code for `Connection.disconnect()`. I run redis as a celery broker, so I seem to be running into a situation where I get a TypeError: NoneType has no attribute `_close()`. It would appear that in one process `disconnect()` is being called and setting the `self._sock` to None while another process has already executed the if statement that checks if `self._sock` is None.  

``` python
    def disconnect(self):
        "Disconnects from the Redis server"
        self._parser.on_disconnect()
        if self._sock is None:
            return
        try:
            self._sock.shutdown(socket.SHUT_RDWR)
            self._sock.close()
        except socket.error:
            pass
        self._sock = None
```

rolette:
Here's part of a backtrace showing basically the same problem, but with a SentinelManagedConnection:

```python  
File "/usr/local/lib/python3.5/site-packages/redis/client.py", line 880, in get
    return self.execute_command('GET', name)
  File "/usr/local/lib/python3.5/site-packages/redis/client.py", line 572, in execute_command
    connection.send_command(*args)
  File "/usr/local/lib/python3.5/site-packages/redis/connection.py", line 563, in send_command
    self.send_packed_command(self.pack_command(*args))
  File "/usr/local/lib/python3.5/site-packages/redis/connection.py", line 538, in send_packed_command
    self.connect()
  File "/usr/local/lib/python3.5/site-packages/redis/sentinel.py", line 44, in connect
    self.connect_to(self.connection_pool.get_master_address())
  File "/usr/local/lib/python3.5/site-packages/redis/sentinel.py", line 106, in get_master_address
    self.disconnect()
  File "/usr/local/lib/python3.5/site-packages/redis/connection.py", line 921, in disconnect
    connection.disconnect()
  File "/usr/local/lib/python3.5/site-packages/redis/connection.py", line 530, in disconnect
    self._sock.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

rolette:
I've pretty much convinced myself ConnectionPool doesn't have the locking it needs. The disconnect() method is ripping out connections that are in use and I suspect the call in Connection.disconnect() to self._sock.shutdown() ends up being a system call, so you lose the GIL and any protection you were counting on from that.

I did a quick experiment to catch and ignore the AttributeError on self._sock.close() and ended up with a RuntimeError in ConnectionPool.disconnect(). See the stack trace below:

```python
[...]

PR #784, opened 2016-09-10
Fix for disconnect issues in #732

rolette:
I haven't finished all of my testing yet, but initial signs are all good and it passes your tox tests. There were a couple of behavior changes required to fully fix it, so I wanted to get your feedback on my approach.

1) The `socket.shutdown()` call in Connection.disconnect() wasn't safe if the process was forked and the child process inherited the file descriptor. Since the `shutdown()` call is desirable for faster cleanup in the non-forked case, I added a new `fork_safe` option to `Connection.__init__()` that let's the code do the safe thing by default or take advantage of faster socket cleanup in a non-forked environment.

That also allowed us to skip doing anything in `_checkpid()` if we know the process isn't forked.

2) The other change fixes `ConnectionPool.disconnect()` so it doesn't rip connections that are in-use out from under the threads that own them. I use a generation counter to flush connections safely.

A side-effect of this is that `ConnectionPool.disconnect()` doesn't actually disconnect anything immediately. Connections are disconnected gradually. I added an option to force the old behavior of "damn the torpedoes!" killing all connections, but the default is the safe path.

Let me know what you think.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -1,4 +1,5 @@
 from __future__ import with_statement
+from copy import copy
 from distutils.version import StrictVersion
 from itertools import chain
 import os
@@ -409,7 +410,8 @@ class Connection(object):
                  socket_keepalive=False, socket_keepalive_options=None,
                  retry_on_timeout=False, encoding='utf-8',
                  encoding_errors='strict', decode_responses=False,
-                 parser_class=DefaultParser, socket_read_size=65536):
+                 parser_class=DefaultParser, socket_read_size=65536,
+                 **kwargs):
         self.pid = os.getpid()
         self.host = host
         self.port = int(port)
@@ -431,6 +433,14 @@ class Connection(object):
             'db': self.db,
         }
         self._connect_callbacks = []
+
+        # If the connection isn't used in a process that forks, there are
+        # certain optimizations we can make. Default assumes that we have to
+        # be safe for process forks.
+        self.fork_safe = kwargs.get('fork_safe', True)
+
+        # Connection pool generation id
+        self.pool_generation = 0
 
     def __repr__(self):
         return self.description_format % self._description_args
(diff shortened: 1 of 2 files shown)


rolette:
Extended testing still looks good on the fix.

andymccurdy:
@rolette I don't see any tests included with the PR. Maybe you forgot to push them?

rolette:
@andymccurdy I didn't create any new redis-py tests... I ran the existing tests to make sure I didn't break anything. The rest of the testing I was referring to was testing we do in my product where I was seeing the various issues originally.

rolette:
@andymccurdy To add a little more context, I'm not really sure how you'd go about adding a test that fits into the redis-py test infrastructure.

The errors that are generated due to the disconnects ripping connections out from under the owner are very timing dependent. It requires some runtime to generate the problem and there's really no fixed number of iterations you can do and say definitively that it is fixed.

I was previously able to repro some form of the issue in a reasonable amount of time within my product, but it's a complex environment (see my previous description in the bug thread). With this patch, I haven't been able to reproduce it yet.

Beyond gaining more confidence in the fix with accumulated runtime, IMO, code review to validate the fix is probably the way to go.

rolette:
I've been running a month with this fix and haven't seen the disconnect issues from #732 since applying it. I'm calling it done from my side. Can we get this merged into the main repo @andymccurdy?

jhgg:
Would love to see this merged. We run into a similar issue as well - especially when using the sentinel connection pool. When the sentinel elects a new master, we get a lot of exceptions until fully restarting the API servers

jhgg:
We've merged this into our fork - and found that this impl does not work w/ the BlockingConnectionPool. I implemented similar logic ontop of that onto our fork, https://github.com/hammerandchisel/redis-py/pull/1

@rolette - take a look and see what you think if you have a chance - and if you want to add that to this PR.

rolette:
@jhgg - Yeah, we don't use `BlockingConnectionPool` in my product, so I didn't touch it.

It's been 5 months since I submitted the the PR, so not sure @andymccurdy has any interest in merging it, but I'll try to take a look at your additions this weekend.

jhgg:
Thanks! I also implemented some tests for the blocking connection pool logic.

andymccurdy:
@rolette I've finally got around to resolving this. We already have a "unique-ish" id in the PID on both the Connection and the ConnectionPool so I'm using that as the guard against a child calling shutdown on a socket that wasn't created by them. I've also added some tests using the multiprocessing lib (that uses fork) to make sure we're doing the right thing.

Sorry again for the length of time but thanks for raising this issue and putting together this PR. It was definitely helpful.

andymccurdy:
3.2.0 has been released with these changes.

PR #788, opened 2016-09-28
WIP: add Lock.locked property

timfeirg:
Note that `Lock` lack a common method (or property) `lock.locked`, and I think it seems pretty obvious that `Lock.local.token` can be used to implement the `locked` method, but I haven't been reading the code very carefully.

I'll add docs later if this implementation is approved.


timfeirg:
it's obviously wrong, sorry

PR #792, opened 2016-10-12
Multi-platform management of TCP socket keepalive

ivendor:
I've changed the current interface for keepalive parameters which works only for Linux Platform using code snippet from: https://github.com/markokr/skytools/blob/master/python/skytools/sockutil.py

Now it works also for Windows and macOS platforms, but not all the parameters are considered (see the comment inside _set_tcp_keepalive method)

I've tested this commit on Windows and Linux platforms in my office.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -443,7 +443,7 @@ class StrictRedis(object):
             'CLUSTER SETSLOT': bool_ok,
             'CLUSTER SLAVES': parse_cluster_nodes,
             'GEOPOS': lambda r: list(map(lambda ll: (float(ll[0]),
-                                         float(ll[1])), r)),
+                                                     float(ll[1])), r)),
             'GEOHASH': lambda r: list(map(nativestr, r)),
             'GEORADIUS': parse_georadius_generic,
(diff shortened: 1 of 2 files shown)


andymccurdy:
Thanks for the PR. I'd like to merge it, but I'd like to see two things first:

1. Let's fix the PEP8 errors.
2. Can we keep the `socket_keepalive_options` argument to the client and connection classes (in addition to the new ones you're adding) to preserve backwards compatibility? People are already using keepalive (at least on linux hosts) and have code written to pass `socket_keepalive_options`. I'd prefer not to break those people.

ivendor:
Ok i'll rewrite the request but now the keepalive options can set any tcp socket options. I prefer to handle only the three options regarding the keepalive settings.

ivendor:
Update: i cannot mantain the previous interface as the socket module on Windows platform doesn''t have the same parameters of the linux one (TCP_KEEP* entries) so the interface needs to be changed

dejlek:
I do not think that anyone will look at this code after 3 years unless someone else takes over...

PR #798, opened 2016-11-07
Fix a syntax error on doc.

jiajunhuang:


jiajunhuang:
have a look at here, please: https://redis-py.readthedocs.io/en/latest/#redis.StrictRedis.georadius

PR #802, opened 2016-11-14
Allow setting clientname during connection construction

Habbie:
This makes the SELECT also log the client name, streamlining the
logging. Default clientnames have been provided which in some cases
will already make debugging easier even if no explicit clientname
has been provided.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -402,18 +402,21 @@ else:
 
 class Connection(object):
     "Manages TCP communication to and from a Redis server"
-    description_format = "Connection<host=%(host)s,port=%(port)s,db=%(db)s>"
+    description_format = \
+        "Connection<host=%(host)s,port=%(port)s,db=%(db)s,clientname=%(clientname)s>"
 
     def __init__(self, host='localhost', port=6379, db=0, password=None,
                  socket_timeout=None, socket_connect_timeout=None,
                  socket_keepalive=False, socket_keepalive_options=None,
                  retry_on_timeout=False, encoding='utf-8',
                  encoding_errors='strict', decode_responses=False,
-                 parser_class=DefaultParser, socket_read_size=65536):
+                 parser_class=DefaultParser, socket_read_size=65536,
+                 clientname="redis-py"):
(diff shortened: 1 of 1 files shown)


Habbie:
I see I'm failing some tests here; will try to look at that this week. Comments are welcome meanwhile!

AngusP:
@Habbie this would be a neat feature, though according to [redis.io/commands/client-setname](https://redis.io/commands/client-setname) you need to ensure the clientname contains no spaces, otherwise it'll get a  `(error) ERR Syntax error...` from Redis.

Also [here](https://github.com/andymccurdy/redis-py/pull/802/files#diff-b3be2d0fa4e0cf1f1c403f2de43ecdb4R539) you check if a client name is given, which it will always be as there's a default provided. If the default were `None`, and the check was instead `if type(self.clientname) is str` (or something similar) you'd be able to disable its use, which would be useful to people using redis-py in performant environments where many clients may be spawning, where the traffic from `CLIENT SETNAME`s would be unnecessary bloat

Habbie:
Yes, I got around to taking a short look and noticed whitespace issues. I'll try to get back to it soon.

daveisfera:
This would be a very helpful feature. What needs to be done to get this merged?

andymccurdy:
I agree, this is a nice feature. Happy to merge once the tests pass.

Habbie:
Hello! I'm glad this is getting some attention.

One, some context, I also filed https://github.com/antirez/redis/pull/3592 which would make this way more useful.

Two, I'll see what I can do about the tests sometime soon.

andymccurdy:
I've fixed the tests and cleaned this up a bit. The `client_name` argument can now be passed to `Redis` or `ConnectionPool` instances. `CLIENT SETNAME` is only executed if the `client_name` argument is supplied. `client_name` also only appears within the `repr()` strings if `client_name` is supplied.

These changes have been merged into master and will go out in the next release. Thanks!

Habbie:
Wonderful, thank you for compensating for my lack of action on this!

PR #813, opened 2017-01-03
InterruptedError does not exist in Python < 3.3.

lddubeau:
Upon a interrupted `select` call, the code prior to this commit would fail in 2.7 with a complaint that `InterruptedError` is not defined. The updated code should work fine on any existing version of Python from 2.7 onwards.

Please let me know what additional work needs to be done to get this out in a released version. Seeing a released version that incorporates this and #743 is becoming a pressing matter for me.  As soon as I upgraded to Celery 4.0.2 I've started running into the `EINTR` issue. My Django test suite cannot pass unless I use a version of `redis-py` that contains the `EINTR` fixes.


andymccurdy:
thanks!

PR #815, opened 2017-01-06
Only allow NX or XX

phuong:
`ResponseError: syntax error` If somehow both nx and xx are set to True


itamarhaber:
I agree with @laixintao - closing but feel free to reopen.

Issue #526, opened 2014-08-27
Add support for PUBSUB subcommands


dgouldin:
As documented here:

http://redis.io/commands/pubsub

For example:

```
>>> import redis
>>> client = redis.from_url('redis:///')
>>> client.execute_command('PUBSUB', 'NUMSUB', 'foo')
['foo', '1']
```

Grokzen:
@andymccurdy This one can be closed as resolved, based on the merged MR #819 that implemented the pubsub subcommands.

andymccurdy:
Thanks for catching this. Agreed, this have been resolved.

PR #819, opened 2017-01-24
Implement PUBSUB * commands

AngusP:
PUBSUB CHANNELS, PUBSUB NUMSUB and PUBSUB NUMPAT are implemented as per https://redis.io/commands/pubsub and suggested in Issue #526.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -519,6 +519,22 @@ cannot be delivered. When you're finished with a PubSub object, call its
     >>> ...
     >>> p.close()
 
+
+The PUBSUB set of subcommands CHANNELS, NUMSUB and NUMPAT are also
+supported:
+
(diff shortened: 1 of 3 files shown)


andymccurdy:
This looks great, thanks. One request: can you add a `@skip_if_server_version_lt` decorator to the tests so that they only run when testing against a compatible server? For reference: https://github.com/andymccurdy/redis-py/blob/master/tests/test_commands.py#L62

Happy to merge afterwards.

AngusP:
Hi @andymccurdy, sure, I'll do that and push soon

AngusP:
@andymccurdy added and pushed, builds passing

andymccurdy:
thanks. merging.

PR #830, opened 2017-01-30
Fix from_url on Python 3

maruno:
Added check in front of python 2 specific code.


andymccurdy:
fixed in redis-py 3.0

PR #834, opened 2017-02-17
Adding Sentinel Commands

otherpirate:
Sentinel API [commands](https://redis.io/topics/sentinel)

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -295,3 +295,64 @@ class Sentinel(object):
         connection_kwargs.update(kwargs)
         return redis_class(connection_pool=connection_pool_class(
             service_name, self, **connection_kwargs))
+
+    def _execute_command(self, command):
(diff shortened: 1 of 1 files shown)


otherpirate:
@andymccurdy Could you please review this PR?

lalinsky:
There are already sentinel commands in `redis.client`, I think these should be added there:

https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L792

otherpirate:
What do you think @andymccurdy ? Should I take @lalinsky suggestion?

chayim:
@otherpirate Thanks for the contribution. This is great! I'd like to get it in right after [this pr](https://github.com/andymccurdy/redis-py/pull/1534) if that's alright with you.

Alternatively if this is no longer active, I'll progress with it. Please let me know!

PR #839, opened 2017-03-02
Fix TypeError issue with socket_connect_timeout

markwingerd:
This fixes my issue with UnixDomainSocketConnection not handling socket_connect_timeout. https://github.com/andymccurdy/redis-py/issues/838

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -715,7 +715,7 @@ class UnixDomainSocketConnection(Connection):
     def __init__(self, path='', db=0, password=None,
                  socket_timeout=None, encoding='utf-8',
                  encoding_errors='strict', decode_responses=False,
-                 retry_on_timeout=False,
+                 retry_on_timeout=False, socket_connect_timeout=None,
                  parser_class=DefaultParser, socket_read_size=65536):
         self.pid = os.getpid()
         self.path = path


redis/connection.py
```diff
@@ -715,7 +715,7 @@ class UnixDomainSocketConnection(Connection):
     def __init__(self, path='', db=0, password=None,
                  socket_timeout=None, encoding='utf-8',
                  encoding_errors='strict', decode_responses=False,
-                 retry_on_timeout=False,
+                 retry_on_timeout=False, socket_connect_timeout=None,
```
lddubeau: This commit has been suggested as a fix for a [problem in Celery](https://github.com/celery/celery/issues/2903). 

I disagree with this change. `socket_connect_timeout` is a setting that has no effect for Unix domain sockets. Adding it to the constructor gives the wrong impression that it is a setting developers could use meaningfully, when it isn't. Moreover, `socket_connect_timeout` is not the *only* setting that is meaningful for a TCP connection but meaningless for a Unix domain socket. What about `socket_keepalive` and `socket_keepalive_options`? It is rather arbitrary to add just one setting but not the others.

The problem is with Celery and the solution has been outlined [in this comment](https://github.com/celery/celery/issues/2903#issuecomment-207005107). The patch I show there was untested at the time of writing but I've used it in production for months with Celery 3.x. 

Celery 4.x does not need it. Seems the issue was fixed in the code reshuffle.

markwingerd: Oh, I see. I completely missed that. Thanks. I do agree that adding socket_connect_timeout is unnecessary something should be done about the error. Would it be better to change line in make_connection in connection.py to explicitly pass the arguments?

`return self.connection_class(**self.connection_kwargs)`

lddubeau: I don't know whether that would work. The code I use in production and that I know works is the one I have in the comment I linked to above.

andymccurdy:
This seems to be an issue with Celery, not necessarily redis-py. Closing this.

PR #846, opened 2017-03-23
InterruptedError is not defined in Python 2

### LINKED ISSUE #845: redis._compat: global name 'InterruptedError' is not defined
In the [compat module](https://github.com/andymccurdy/redis-py/blob/master/redis/_compat.py), this code is broken:
```python
# For Python older than 3.5, retry EINTR.
if sys.version_info[0] < 3 or (sys.version_info[0] == 3 and
                               sys.version_info[1] < 5):
    # Adapted from https://bugs.python.org/review/23863/patch/14532/54418
    import socket
    import time
    import errno

    from select import select as _select

    def select(rlist, wlist, xlist, timeout):
        while True:
            try:
                return _select(rlist, wlist, xlist, timeout)
            except InterruptedError:
                continue
```

The problem is that it uses `InterruptedError`, but this exception only exists in Python 3, not Python 2.

SethMichaelLarson:
Fixes #845. Defined `InterruptedError` as `OSError` if not defined like in Python 2.

### DIFF
--- a/redis/_compat.py
+++ b/redis/_compat.py
@@ -1,6 +1,11 @@
 """Internal module for Python 2 backwards compatibility."""
 import sys
 
+try:
+    InterruptedError = InterruptedError
+except:
+    InterruptedError = OSError
+
 # For Python older than 3.5, retry EINTR.
 if sys.version_info[0] < 3 or (sys.version_info[0] == 3 and
                                sys.version_info[1] < 5):
@@ -15,8 +20,12 @@ if sys.version_info[0] < 3 or (sys.version_info[0] == 3 and
         while True:
             try:
                 return _select(rlist, wlist, xlist, timeout)
-            except InterruptedError:
-                continue
+            except InterruptedError as e:
+                # Python 2 does not define InterruptedError, instead
+                # try to catch an OSError with errno == EINTR == 4.
+                if hasattr(e, 'errno') and (e.errno == 4 or e.errno is None):
+                    continue
+                raise e
 
     # Wrapper for handling interruptable system calls.
     def _retryable_call(s, func, *args, **kwargs):


redis/_compat.py
```diff
@@ -15,8 +20,12 @@ def select(rlist, wlist, xlist, timeout):
         while True:
             try:
                 return _select(rlist, wlist, xlist, timeout)
-            except InterruptedError:
-                continue
+            except InterruptedError as e:
+                # Python 2 does not define InterruptedError, instead
+                # try to catch an OSError with errno == EINTR == 4.
+                if hasattr(e, 'errno') and (e.errno == 4 or e.errno is None):
```
andymccurdy: While the comment helps, this could be clearer testing: `e.errno == errno.EINTR`. I believe the `errno` module is in both Py2 and Py3.

Also, in what case does `e.errno is None` mean we should treat the error as an interrupt and try again?

redis/_compat.py
```diff
@@ -15,8 +20,12 @@ def select(rlist, wlist, xlist, timeout):
...
+                    continue
+                raise e
```
andymccurdy: If we just `raise` (without the `e`), we'll get the original traceback. I'm not 100% certain, but I believe on some Python versions `raise e` would raise the exception starting there rather than where the exception originally took place.

SethMichaelLarson: Sure, I'll switch this out.

redis/_compat.py
SethMichaelLarson: Might be able to remove the `errno is None` check. I don't know if the `InterruptedError` exception sets `errno` on Python 3 but it might, we can try it! And I thought about doing the import but I'm not sure if all platforms define `errno.EINTR`. I know Windows has a lot of flakiness in the `errno` module. I'll use a `getattr` to be safe.

SethMichaelLarson:
@andymccurdy Thanks for the quick review! I've implemented your review comments.

PR #848, opened 2017-03-27
fixed can't raise invalid expire time when set ex param is 0

hqy:
redis-cli
`set  py python ex 0` will raise invalid expire time in set 
but
redis-py call set function
`set('py', "python",  ex=0)` can't raise invalid expire time in set,  lead to persist the key 

I think this redis-py bug.


NoneGG:
That‘s true, the problem can be reproduced.

andymccurdy:
good catch. thanks!

PR #866, opened 2017-06-02
Add namespace support

Kazanz:
This PR attempts to address the namespace support being requested in [this issue](https://github.com/andymccurdy/redis-py/issues/12).

**Usage**
See the added section in the `README`

**Implementation**
Based on the suggestion by @AngusP namespacing is initiated by initializing your Redis instance with functions passed to the `add_namespace` and `rm_namespace` arguments.  Whenever a namespace is to be added/removed to/from a key the passed function will run with the key as the argument.  This gives the user control of the namespacing logic.

Detecting which arguments/responses are keys follows the suggestion by @andymccurdy of using a `keys_at` method being passed to the `execute_command` argument.  For each method an array of integers representing indexes of the args passed to `execute_command` is passed as a `keys_at` kwarg.

Namespace removal of responses from method's that return keys is handled by passing the provided `rm_namespace` function to the `execute_command` function who in turn passes it down to the `RESPONSE_CALLBACK` for that command, which handles the logic for namespace removal.

No state is maintained on the class instance, so as long as the user doesn't change the `add_namespace` and `rm_namespace` methods namespacing should be thread safe.

**Testing**

Duplicates of tests without namespacing are created and run with namespacing enabled using basic add and rm namespace functions.  There are tests for namespaced redis, pipelines, pubsubs, and luascripting.

These tests ensure that the keys_at param for each method has an accurate value passed, and that when namespaces are removed from responses they are accurate.  But...
I know Andy is cognizant of the time it takes to run tests, and these extra functions add a lot of increased test time.   So I am open to ideas for less strenuous testing, that still ensures good coverage but decreases test time.

**Performance**

I did a rudimentary performance test which produced the following results.

```python
In [1]: import time

In [2]: import redis

In [3]: def add_namespace(key):
    return 'ns:' + key
   ...:

In [4]: def rm_namespace(key):
    return key[3:]
   ...:

In [5]: r = redis.Redis()

In [6]: nsr = redis.Redis(add_namespace=add_namespace, rm_namespace=rm_namespace)

In [7]: def test(client, num):
    start = time.time()
    for i in range(num):
        _ = client.get('foo')
    return time.time() - start
   ...:

In [8]: test(r, 10000)
Out[8]: 9.901349067687988

In [9]: test(nsr, 10000)
Out[9]: 9.509238958358765

In [10]: def test(client, num):
    start = time.time()
    for i in range(num):
        _ = client.zadd('zfoo', a=1, b=2, c=3)
    return time.time() - start
   ....:

In [11]: test(r, 10000)

Out[11]: 3.7034590244293213

In [12]:

In [12]: test(nsr, 10000)
Out[12]: 4.358672142028809
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -674,6 +674,59 @@ that return Python iterators for convenience: `scan_iter`, `hscan_iter`,
     B 2
     C 3
 
+Namespacing
+^^^^^^^^^^^
+
+Namespacing keys is possible with the `StrictRedis`, `Redis`, `PubSub`,
+`StrictPipeline` and `Pipeline` objects using `add_namespace` and
+`rm_namespace` kwargs during instantiation.  `add_namespace` should be a
+function that takes a key argument and returns your namespaced key.
+`rm_namespace` function should take a namespace key argument and return
+your original key without the namespace. This allows you to create your own
+namespacing logic.  `Pipeline` and `PubSub` objects created via methods on
+a `Redis` or `StrictRedis` instance will receive the parent instance's
+namespace methods.
+
+.. code-block:: pycon
+
+    >>> import redis
+    >>>
+    >>> def add_namespace(key):
+    ...     return 'ns:{}'.format(key)
+    ...
+    >>> def rm_namespace(key):
+    ...     return key[3:]
+    ...
+    >>> r = redis.StrictRedis(host='localhost', port=6379, db=0,
+    ...                       add_namespace=add_namespace,
+    ...                       rm_namespace=rm_namespace)
+    ...
+    >>> # Sets key "ns:foo".
+    >>> r.set('foo', 'bar')
+    True
+    >>> # Gets key "ns:foo".
+    >>> r.get('foo')
+    'bar'
+
+Care must be taken when adding namespacing support to existing systems. The
+following caveats need be considered:
+
+- The `KEYS` command will namespace any provided pattern and default to the
+  return value of `add_namespace('*')`.  The `SCAN` command's `MATCH` arg
+  will default to `add_namespace('*')`.
+
+- There is no way to know if your existing data structures contain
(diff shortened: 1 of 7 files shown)


tony:
@Kazanz Rebase against @andymccurdy's master.

Quick link to *client.py*: https://github.com/Kazanz/redis-py/blob/84f254e6478a08185bdc3e479be53df66e57a181/redis/client.py (too big to preview)

For some bike shedding, I just wanted to be able to instantiate a class and have all keys for that command namespaced.

What I don't understand is a need for `rm_namespace` (maybe I just don't understand redis well enough to get it), since the user could just instantiate a new class, or alter/remove the namespace being used on the instantiated class themselves, rather than by a keyword argument on methods. While removing a namespace may seem helpful in some cases, I don't think it's as universal as the general need users have to work add namespaces.

Any thoughts on the implementation just being `namespace`?

@andymccurdy: Thoughts on this?

Kazanz:
@tony I have not intention of continuing to support this with the lack of support shown by @andymccurdy  over the last decade.

The reason for add_namespace and rm_namespace is so that the end user can control the namespace logic based on the keys going in.  This may be useful for back porting or more complicated environments.

It stems from a earlier conversation about the PR.  
If you want to take over this PR, you can fork my repo.

okomarov:
Such a shame this hasn't been merged...

Kazanz:
@okomarov The biggest shame in my opinion is the complete drop in communication from @andymccurdy after the constant back and forth.   It's not like I wasn't open to changing the implementation.  I did it happily over and over, and would have continued to, Andy after all has the most intimate knowledge with the code base, so it made sense to work on it until he was satisfied.  

But having a working implementation be completely ignored after months of back and forth to satisfy him is rather rude imo.

I see @AngusP has taken up the charge in the issue thread.  I hope he has better luck.

PR #867, opened 2017-06-15
Precalculate sha1 on Script initialization

bgreenberg-eb:
Currently, `Script` instances initializes their `sha` attribute to the empty string, which means that the first time a `Script` is called, it will attempt to `EVALSHA` the empty string. This is guaranteed to fail, and results in unnecessary round-trips to the server. We were seeing a lot of useless network traffic in our deployment because of this. The workaround is to manually set the `sha` attribute after initializing each `Script`, but this is not a very clean solution.

This change makes `Script` precalculate the SHA1 of of the script string, so that in the case where it is called on a connection to a server where the script is already registered, it saves itself an `EVALSHA` call and a `SCRIPT LOAD` call.

It also modifies the way `Script` interacts with `BasePipeline`. Previously, the `Script` would call `script_load_for_pipeline` prior to executing, which would call `SCRIPT LOAD` if the `Script.sha` was empty. That condition will never be `True` anymore, plus it is unnecessary because the pipeline calls `load_scripts` during `execute`. I removed that method and made `Script` directly add itself to the pipeline's `scripts` set so that the pipeline will register it during `load_scripts`.

I know there is already a PR for this same issue, but its approach is not quite right in my opinion, and it has been inactive since 2015.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2927,7 +2927,10 @@ class Script(object):
     def __init__(self, registered_client, script):
         self.registered_client = registered_client
         self.script = script
-        self.sha = hashlib.sha1(script.encode('utf-8')).hexdigest()
+        # Precalculate and store the SHA1 hex digest of the script.
+        # Encode the digest because it should match the return type of script_load, which
+        # in Python 3 is bytes.
+        self.sha = hashlib.sha1(script.encode('utf-8')).hexdigest().encode('utf-8')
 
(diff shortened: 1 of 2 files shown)


andymccurdy:
@bgreenberg-eb: Thanks for this. I altered your patch slightly to use the client's encoding options to get the byte-representation of the script before calculating the sha1. It's possible someone has unicode characters in one of their Lua scripts.

PR #872, opened 2017-06-22
Add a helper to check whether lock belongs to this thread or not

jonathansp:


jonathansp:
Not a good idea?

andymccurdy:
Hi, sorry for the delay. This same functionality was just merged from another PR as `lock.locked()`. Sorry for the long delay.

PR #877, opened 2017-07-17
change the error order of geo point

twz915:
`GEOADD` needs `lon`, `lat` in order, not the reverse

`GEOPOS` returns `longitude`, `latitude`


andymccurdy:
thanks!

PR #878, opened 2017-07-19
Improve if statement throwing KeyError

Mec-iS:
At line 210, if WITHSCORES is not specified within `execute_command()`, this line used to throw a `KeyError` because the relative key in `options` is not present. Solution: this PR add a check on the key 'withscores' before the key is accessed.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -207,7 +207,7 @@ def zset_score_pairs(response, **options):
     If ``withscores`` is specified in the options, return the response as
     a list of (value, score) pairs
     """
-    if not response or not options['withscores']:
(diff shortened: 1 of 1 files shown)


andymccurdy:
Your patch won't work if someone passes `withscores=False` because False != None. I think the following line is better:

`if not response or not options.get('withscores'):`

Mec-iS:
Yeah, right. Sorry, didn't have the time to take a look to the full trace and relative interfaces involved. Hope it helps anyway. It would be great to have it updated on PyPi also.

Mec-iS:
Submitted an update.

andymccurdy:
Thanks!

Issue #881, opened 2017-07-24
redis.srandmember returns one member when number = 0


a1exwang:
The document says:

> srandmember(name, number=None)[source]
If number is None, returns a random member of set name.
If number is supplied, returns a list of number random memebers of set name. Note this is only available when running Redis 2.6+.

But if I use redis.srandmember(key, 0), it returns one elememt.

I think there is a bug in redis/client.py:1647, 
```
number and [number] or []
```

andymccurdy:
thanks! merged.

PR #882, opened 2017-07-24
[bugfix] Fix srandmember(key, 0) returns 1 element bug

a1exwang:
#881


andymccurdy:
this makes sense. thanks!

PR #885, opened 2017-07-31
add note on readme file

nsantiago2719:


andymccurdy:
Thanks for making the note. I'd like to expand on it a bit within the docs. Users may be interested in the raw byte strings. Documenting that a user should always set `decode_responses=True` on Python 3 is IMO incorrect.

laixintao:
>  By default, the value returned if you are using Python3 are in raw byte strings.

There is a grammar mistake. 

>  By default, the value returned are bytes in python3.

Would be better.

andymccurdy:
Thanks! I've added a little extra info as well.

Issue #824, opened 2017-01-25
supporting maxidle connections


deepakverma:
Scenario:
I have a burst of load which causes ConnectionPool to scale up the connections to 100
After a while load is normal now, the load now uses 10 connections to fulfill the request. Pool still has 100 connections open.
Due to period of inactivity redis-server closed 90 connections.
Burst comes in again causing the Pool to scale up the connections.

New connections have to compete for CPU on client as well as redis-server where it;s already pegged due to load, causing connection timeouts.

I am thinking it would be useful if pool has a setting to configure maxidle connections so that ConnectionPool is already ready to fulfill the load

Issue #306, opened 2012-12-06
ConnectionPool doesn't reap timeout'ed connections


rslinckx:
The current ConnectionPool implementation doesn't disconnect redis connections until they are used again and found dead.

This means that if i have a huge peak of concurrent connections, then a few seconds later the redis server will close them for inactivity, but the tcp connection will still live within the process, and end up in CLOSE_WAIT state, which can cause a too many open files later on.

Maybe the release() method could check if there are any old connections and close them if appropriate (by storing a last_used timestamp along with each connection for example).

Or any other solution preventing a lot of CLOSE_WAIT connections until i restart my process. Any idea?

LotharSee:
I'm facing the same issue and I see it was already reported.

A simpler solution would be to provide a `connection_timeout` parameter (that should match the `timeout` attribute in Redis configuration) then we keep track of the last time we used each connection in the pool. If one is too old, we just close/drop it from the pool without raising error.

Another alternative would be to catch and hide these connection timeouts (again, behind a parameter such as `drop_timeout_conections` or whatever).

In both case, one difficulty is be to be able to catch exactly this kind of exception (timeout of a connection from the pool), and not to mix them with other `ConnectionError`.

harshadptl:
Same issue with me.

I think the get_connection method should check and weed out old connections.

pranjal5215:
```python
class CustomConnectionPool(redis.ConnectionPool):
    """ 
    Custom overridden ConnectionPool to handle server 
    timeouts. Check in get_coonection whether connection 
    is timed out from server, is so make a new one.
    """
    def __init__(self, **kwargs):
        super(CustomConnectionPool, self).__init__(**kwargs)

    def get_connection(self, command_name, *keys, **options):
        "Get a connection from the pool and check for a timeout."
        self._checkpid()
        try:
            connection = self._available_connections.pop()
        except IndexError:
            connection = self.make_connection()
        try:
            connection.send_command("PING")
            connection.read_response()
        except redis.ConnectionError:
            connection.disconnect()
            connection = self.make_connection()
        self._in_use_connections.add(connection)
        return connection
```



We tried to override get_connection() method of class redis.ConnectionPool. In case server has timed out the response PING raises ConnectionError, in such event we create a new connection and insert it back to pool

scosgrave:
Same issue here. I think the solution from @pranjal5215 looks pretty good, except for a few things. 

1.  I think you can move the send_command and read_response into the first try and have 2 excepts
[...]

PR #886, opened 2017-08-02
Connection management improvements.

btimby:
This PR addresses #306 and #824. We ran into these issues (many connections in CLOSE_WAIT state, and too many open connections following a load spike).

Specifically, I modified `ConnectionPool.get_connection()` so that it retrieves the connection at the head of the list of available connections. `release()` still appends to the tail, so all connections will be utilized in round-robin fashion. This change was made to support the other change to `get_connection()` which verifies a connection's "liveness". If a connection is not live, it is disconnected, and the next available is similarly tested. If no connections are usable, then a new one is created.

I also modified the `Connection.disconnect()` method so that if `shutdown()` errors (due to CLOSE_WAIT or other socket state) `close()` will still be called. I am not sure this is necessary, but I figured it was safer to ensure close is always called before the reference to the socket is deleted by setting `self._sock = None`.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -547,6 +547,9 @@ class Connection(object):
             return
         try:
             self._sock.shutdown(socket.SHUT_RDWR)
+        except (OSError, socket.error):
+            pass
+        try:
             self._sock.close()
         except socket.error:
             pass
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -57,6 +57,7 @@ SYM_CRLF = b('\r\n')
 SYM_EMPTY = b('')
 
 SERVER_CLOSED_CONNECTION_ERROR = "Connection closed by server."
+TCP_CLOSE_WAIT = 8
 
 
 class Token(object):
@@ -670,6 +671,27 @@ class Connection(object):
             output.append(SYM_EMPTY.join(pieces))
         return output
 
+    def validate(self):
+        # If we are not yet connected, this connection is valid.
+        if self._sock is None:
+            return
+
+        # Check that connection is usable. We check specifically for the
+        # CLOSE_WAIT state on the socket. The socket enters this state for
+        # example when the Redis server times out the connection due to
+        # idleness.
+        try:
+            state = ord(self._sock.getsockopt(
+                socket.SOL_TCP, socket.TCP_INFO, 1))
+
+            if state == TCP_CLOSE_WAIT:
+                raise ConnectionError()
+
+        except socket.error:
+            raise ConnectionError()
+
+        return True
+
 
 class SSLConnection(Connection):
     description_format = "SSLConnection<host=%(host)s,port=%(port)s,db=%(db)s>"
@@ -751,6 +773,9 @@ class UnixDomainSocketConnection(Connection):
         else:
             return "Error %s connecting to unix socket: %s. %s." % \
                 (exception.args[0], self.path, exception.args[1])
+
+    def validate(self):
+        return True
 
 
 FALSE_STRINGS = ('0', 'F', 'FALSE', 'N', 'NO')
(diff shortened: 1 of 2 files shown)


taylor-cedar:
I am running into issues with connections becoming idle and timing out. This would be very helpful to fix those issues.

nickwilliams-eventbrite:
+1 could use

lambdaq:
+1 plz merge

taylor-cedar:
This PR has a couple of major issues. I attempted to use it in production and ran into some bugs.

First is line https://github.com/smartfile/redis-py/blob/f9348968dec581830a32e8469228c74ce443bd88/redis/connection.py#L1010 . `self._created_connections` will eventually exceed `self.max_connections` because connection counts are not reset when refreshing a dead connection.

Second issue is https://github.com/smartfile/redis-py/blob/f9348968dec581830a32e8469228c74ce443bd88/redis/connection.py#L687 . `socket.TCP_INFO` is not supported on OSX, so this line throws an exception.

I updated the code and have been using it in production for 2 months without issue. Here is the diff for the improved version.

https://github.com/smartfile/redis-py/compare/master...cedar-team:master

ssjunior:
+1

btimby:
@taylor-cedar Good changes! Counting the connections rather than relying on a counter is better than the original method. Also, I work exclusively with Mac, so I am unable to test any other platform.

I do have a question about line 990 in your diff however, `connection` should never be in `self._available_connections` since it is popped on line 979 right? The only reason this `if` statement would ever evaluate to `True` is that the `connection` is duplicated within the list, which would indicate some other issue no? Did you run into this?

taylor-cedar:
@btimby You are correct about the `self._available_connections`. That is unnecessary.

btimby:
@taylor-cedar OK, makes sense. Seems like we would want your changes merged into this PR. Do you think you can make a PR against my fork? I can merge them in and issue another PR or update this one.

I could try to add your changes myself, but I would not be able to test the Mac specific parts, and I would probably mess it up.

TCTChadwell:
For limiting connections, have you guys had any luck using the `BlockingConnectionPool` class? http://redis-py.readthedocs.io/en/latest/#redis.BlockingConnectionPool

btimby:
@TCTChadwell for me it is not about limiting connections, but about handling timeouts properly. The default pool reuses the same connections over and over. When load spikes new connections are created to handle load (OK). But then when load reduces these connections stagnate and are timed out by the server. The connection pool never notices and never closes them (it does not attempt to use them again) so they remain in CLOSE_WAIT (NOT OK.)

If load increases again these stale connections throw application errors when they are used again. The pool does not check the connection before returning it for use.

This change causes all connections in the pool to be used, round robin. This keeps timeouts at bay and detects errors. It also culls extras and performs a liveness test (ping) before returning a connection to the application. Broken connections are closed and replaced (errors that arise from CLOSE_WAIT while closing are handled and the connection is fully closed).

I don't need to limit connections, my redis server can handle the load, but by default connections are not properly managed by this pool.

wsantos:
Guys what do we need to get this merged ? in the mean time @taylor-cedar do you mind to merge/rebase your fork ?

taylor-cedar:
@wsantos I can rebase my fork, but it seems like this library is abandoned. Nothing has changed since November. If @andymccurdy or anyone else with write access has any interest in merging this PR, I will happy to do the work necessary.

kfrendrich:
+1

Ondrysak:
+1

andymccurdy:
I've just released 3.2.0 which attempts to address the underlying problem in a different way. I tried to use the code in this PR but couldn't get it to accurately identify the state of the connection.

PR #887, opened 2017-08-03
update documents for set()

laixintao:
Hi,

The old function string quite confused me, it sounds like 

> If you set nx=True, redis create a new key when it not exists,
> If you set nx=False, redis doesn't create new key when it not exists.

https://redis.io/commands/set#options


andymccurdy:
thanks!

PR #888, opened 2017-08-10
fix problem when using geopos on unexistent hash member

categulario:


andymccurdy:
Should `GEOPOS` instead return just `None` (instead of `[None, None]`) in the case of a non-existent hash member? That's what redis-cli does according to: https://redis.io/commands/geopos. And it seems slightly easier to test for in calling code as well by simple saying:

```python
position = r.geopos('foo')
if position is None:
    # we don't know about this key
```

rather than:

```python
position = r.getpos('foo')
if position == [None, None]:
    # we don't know about this key
```

categulario:
I was thinking about querying for a list of members and you get an array of
responses according to the item being found or not, which is kind of
consistent

categulario:
Reading the GEOPOS documentation in redis.io says

> Non existing elements are reported as NULL elements of the array

The output from redis-cli is also an array of nils, which is consistent with the example they give in the same page.

andymccurdy:
I completely misinterpreted the GEOPOS command. You're completely correct. Merging now.

NoneGG:
The test case may need to be tested with newer version of redis.
In my dev env(with redis 3.2.1), `$0` will be returned if command like `geopos a b c` is sent.
But in my colleague's env(with redis 4.0), the test passed.

andymccurdy:
@NoneGG Nice catch. I just saw the note about this in the Redis 4.0 patch notes.

I've just committed a fix to the test suite.

PR #893, opened 2017-08-20
Add support for Python collections backed by Redis

ke4roh:
I needed to bring in Redis after I'd already written code to work with Python collections, so this adapter code was born, offering the ability to store arbitrary objects and dress up structures in Redis to support the standard operations of lists, sets, and dicts.  The library is designed to use the most efficient routines to perform each operation.  

Further, there is justification to offer similar drop-in replacements for the queue module, but I thought the basic collections and the TTL collections would be a good start.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,8 +34,30 @@ or from source:
 Getting Started
 ---------------
 
+For using Redis backed by collections
 .. code-block:: pycon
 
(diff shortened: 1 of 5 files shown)


ke4roh:
I'll fix the build and re-submit.

PR #902, opened 2017-09-12
fix<lock>: fix "expire may failed in some extreme situations and then lock failed"

yangluoshen:
- In function "do_acquirea"(in lock.py), if a client crash before
      "self.redis.pexpire" and after "self.redis.setnx", the key
"self.name" will store in redis with unexpired, another client couldn't
acquire this lock any more. That is unexpected.

Change-Id: I5aebf76ada669fa9267f562254c4123b3f256ab9

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -118,12 +118,15 @@ class Lock(object):
             mod_time.sleep(sleep)
 
     def do_acquire(self, token):
-        if self.redis.setnx(self.name, token):
-            if self.timeout:
-                # convert to milliseconds
-                timeout = int(self.timeout * 1000)
-                self.redis.pexpire(self.name, timeout)
-            return True
+        if self.timeout:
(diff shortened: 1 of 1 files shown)


yangluoshen:
No，I support this only in theory. I have some trouble in mocking these situations. I may upload testcases later.
Actually, I mean the SET operation and EXPIRE operation should be atomic. Separating this two operation may cause  CONSISTENT problems(if SET operation returns success, and then client crash, EXPIRE operation doesn't execute).

laixintao:
@yangluoshen 
> Actually, I mean the SET operation and EXPIRE operation should be atomic.

Agree with that, and your code seems good.

yangluoshen:
Thanks :)

andymccurdy:
This was fixed at some point in the past. Sorry for not updating this sooner.

PR #904, opened 2017-09-19
clientlist response, client_name value might contain the '=' character, in that case response parser will fail

swilly22:
Recreate
127.0.0.1:6379> CLIENT SETNAME roi=lipman
OK
127.0.0.1:6379> CLIENT LIST
id=2 addr=127.0.0.1:49481 fd=8 name=roi=lipman age=27 ...

redis-py

File "/opt/redislabs/lib/cnm/python/redis/client.py", line 609, in client_list
  File "/opt/redislabs/lib/cnm/python/redis/client.py", line 574, in execute_command
  File "/opt/redislabs/lib/cnm/python/redis/client.py", line 588, in parse_response
  File "/opt/redislabs/lib/cnm/python/redis/client.py", line 248, in parse_client_list
ValueError: dictionary update sequence element #3 has length 3; 2 is required

This PR splits on '=' only once,
such that client_name key will be "name" and the value would be "roi=lipman".

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -245,7 +245,8 @@ def bool_ok(response):
 def parse_client_list(response, **options):
     clients = []
     for c in nativestr(response).splitlines():
-        clients.append(dict([pair.split('=') for pair in c.split(' ')]))
+        # Values might contain '='
+        clients.append(dict([pair.split('=', 1) for pair in c.split(' ')]))
     return clients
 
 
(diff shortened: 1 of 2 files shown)


swilly22:
@laixintao 
Sure, added.

andymccurdy:
looks good to me! thanks!

PR #909, opened 2017-09-30
Make sentinel thread-safe during master failover

lalinsky:
In the current version, the sentinel code tries to close all
connections immediately after discovering there is a new master.
This is a problem in multi-threaded environment, because
neither `ConnectionPool.disconnect` nor `Connection.disconnect` are
thread-safe. If you call `SentinelConnectionPool.disconnect` after master
failover, that will close all connections that are potentially used
from other threads, causing all kinds of errors.

This change avoids that behavior by adding acquire/release checks, so
connections that don't belong the current master are never returned
to the pool and they are closed instead.

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -20,6 +20,7 @@ class SlaveNotFoundError(ConnectionError):
 class SentinelManagedConnection(Connection):
     def __init__(self, **kwargs):
         self.connection_pool = kwargs.pop('connection_pool')
+        self.to_be_disconnected = False
         super(SentinelManagedConnection, self).__init__(**kwargs)
 
     def __repr__(self):
@@ -56,12 +57,10 @@ class SentinelManagedConnection(Connection):
             return super(SentinelManagedConnection, self).read_response()
         except ReadOnlyError:
             if self.connection_pool.is_master:
-                # When talking to a master, a ReadOnlyError when likely
+                # When talking to a master, a ReadOnlyError likely
                 # indicates that the previous master that we're still connected
                 # to has been demoted to a slave and there's a new master.
-                # calling disconnect will force the connection to re-query
-                # sentinel during the next connect() attempt.
-                self.disconnect()
+                self.to_be_disconnected = True
                 raise ConnectionError('The previous master is now a slave')
             raise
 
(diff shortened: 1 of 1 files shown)


redis/sentinel.py
```diff
@@ -59,9 +59,6 @@ def read_response(self):
                 # When talking to a master, a ReadOnlyError when likely
                 # indicates that the previous master that we're still connected
                 # to has been demoted to a slave and there's a new master.
-                # calling disconnect will force the connection to re-query
-                # sentinel during the next connect() attempt.
-                self.disconnect()
```
lalinsky: The connection will get closed when it's returned to the pool.

redis/sentinel.py
```diff
@@ -97,14 +94,12 @@ def reset(self):
...
+        """Get the address of the current master"""
         master_address = self.sentinel_manager.discover_master(
             self.service_name)
         if self.is_master:
-            if self.master_address is None:
+            if master_address != self.master_address:
                 self.master_address = master_address
-            elif master_address != self.master_address:
-                # Master address changed, disconnect all clients in this pool
-                self.disconnect()
```
lalinsky: All old connections will get closed when attempting to fetch them from the pool. I could theoretically close connections from `_available_connections` here, but I don't know what would the thread-safety implications would be. The fact that this code depends on atomic operations in Python is confusing enough.

redis/sentinel.py
```diff
@@ -125,16 +120,36 @@ def rotate_slaves(self):
             pass
         raise SlaveNotFoundError('No slave found for %r' % (self.service_name))
 
-    def _checkpid(self):
```
lalinsky: The implementation in `ConnectionPool` is thread-safe and the `__init__()` call is not needed, `reset()` does everything necessary.

redis/sentinel.py
```diff
@@ -97,14 +94,12 @@ def reset(self):
         self.slave_rr_counter = None
 
     def get_master_address(self):
+        """Get the address of the current master"""
         master_address = self.sentinel_manager.discover_master(
             self.service_name)
         if self.is_master:
-            if self.master_address is None:
+            if master_address != self.master_address:
                 self.master_address = master_address
```
lalinsky: Related pull request - https://github.com/andymccurdy/redis-py/pull/847

PR #914, opened 2017-10-18
Enable pip cache in Travis CI

jdufresne:
Can speed up builds and reduce load on PyPI servers.

For more information, see:

https://docs.travis-ci.com/user/caching/#pip-cache


andymccurdy:
thanks!

PR #917, opened 2017-10-26
Optional retry on connection error

p1c2u:
Added `retry_on_connerror` as optional `Connection` argument. It allows user to handle `ConnectionError` by himself instead retry connection forever (currently). By default retry is `True` so it works the same like before.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -65,6 +65,14 @@ def dict_merge(*dicts):
     return merged
 
 
+def handle_connection_error(connection, e):
+    connection.disconnect()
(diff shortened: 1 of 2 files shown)


dejlek:
Apparently 3.3.x fixes this, but I am not 100% sure. Could @andymccurdy confirm? :)

andymccurdy:
3.3.x introduced health checks that prevent connections from going stale or being timed out by the server or other devices in your infrastructure. This should greatly improve the quality of life for users.

I prefer the health check strategy to this PR. Thanks for letting me know this issue was still opened.

PR #919, opened 2017-10-30
Add connection retrying with customizable retry delay

cwisecarver:
When failing over a primary node in AWS Elasticache redis-py wouldn't re-establish a connection. It would throw a ConnectionError. This adds a configurable number of retries with a configurable period to sleep between retries. It will throw the last caught exception after the retry count is exceeded.

I wrote a few tests and functionally tested it against one of our Elasticache servers. Let me know if there's anything else I can do to help it get merged. This would really cut-down on our service restarts.


cwisecarver:
@andymccurdy Do you think is a valuable addition or should this be handled in a custom Connection class? I've got both implementations done. We're just deciding if we should wait for this to be merged or use the custom class.

andymccurdy:
@cwisecarver Thanks for the contribution. Let me wrap up getting redis-py 3.0 out the door (tonight!) and I'll get back to this. Sorry for the delay.

PR #921, opened 2017-11-07
Allow socket type to be configured in Connection

garlicnation:
This change allows the client to force IPV4 or IPV6.


andymccurdy:
looks good. thanks!

PR #923, opened 2017-11-13
Handling with redis auth error

otherpirate:
Improving error handler in redis auth.

When a auth error occur, should raise an AuthenticationError
I think ResponseError is too much generic for this scenario.

Response error example: "Client sent AUTH, but no password is set"

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -559,7 +559,12 @@ class Connection(object):
         # if a password is specified, authenticate
         if self.password:
             self.send_command('AUTH', self.password)
-            if nativestr(self.read_response()) != 'OK':
+            try:
+                response = self.read_response()
+            except ResponseError as error:
+                raise AuthenticationError(str(error))
+
+            if nativestr(response) != 'OK':
                 raise AuthenticationError('Invalid Password')
 
         # if a database is specified, switch to it


pylemon:
I have a redis server with no password set up running on production environment.
Since I want to add password for that server and keep everything running at the mean time. 
I have to do following steps.

1. set up password in the codebase.
2. publish the code to production.
3. set up password for redis-server.

on step 2, on_connect method will thow out an error "Client sent AUTH, but no password is set" which I just want to ignore it.
Do you have any suggessions on this, instead of override default on_connect behavior to catch the exact ResponseError?

otherpirate:
Hi @pylemon 

Yeh, it's a problem I pass by when I did this implementation. 

I caught the error and check (try/except), if the problem is with a password, retry without the password.

The setup in Redis to enable password took a few seconds, after that, I did a new deploy of my application, without this logic.
So, I did not take a lot of time to choose this path because this code ran for a few minutes and got removed. 

I'm not sure about your scenario, but to me and my company, it's worked pretty well.

If it does not work for, what is the reason? maybe I still can help you :smile:

andymccurdy:
I committed a patch that extends the existing error parsing logic to also work with the different authentication errors. `AuthenticationError` is now a subclass of `ConnectionError` so that the connection is disconnected and cleaned up appropriately after receiving any of the various auth errors. This will go out in the next release. Thanks reporting this!

otherpirate:
Thanks for the update @andymccurdy 

Deleting branch ❤️

pylemon:
@otherpirate Thanks, problem solved ❤️

Issue #621, opened 2015-05-26
Lock with blocking_timeout and context manager trap


gjcarneiro:
Consider the following code:

``` python
with redis_client.lock("foo", blocking_timeout=2):
    # do stuff
```

What happens if the lock is already acquired by someone else for a long time and we expire the 2 second blocking timeout?  Well, redispy happily proceeds to execute the "do stuff" code without holding the lock!  This is dangerous.  I had a nasty bug in my code because of this.

Why this happens?  Because the `redis_client.lock("foo", blocking_timeout=2)` constructs a Lock object, which does `self.blocking_timeout = blocking_timeout` in the constructor.  When we `__enter__` the context manager, this code is executed:

``` python
    def __enter__(self):
        # force blocking, as otherwise the user would have to check whether
        # the lock was actually acquired or not.
        self.acquire(blocking=True)
        return self
```

Clearly the author was aware of a similar danger involving blocking=False.  However, following the code, we get to:

``` python
    def acquire(self, blocking=None, blocking_timeout=None):
        [...]
        if blocking_timeout is None:
            blocking_timeout = self.blocking_timeout
```

So, in the end, the acquire method runs with a blocking_timeout=2 parameter, which means it may return before acquiring the lock, returning False instead of True in this case.

In this case, I would expect `__enter__` to check the return value of self.acquire() and raise an exception if it is True.  I worked around it in my code, but others may fall into the same trap.  Please fix.

omenar:
How did you worked it around in your code?

gjcarneiro:
I made my own context manager without the bug.  Basically, check the return value of acquire and raise an exception if the lock couldn't be acquired.  Something like (simplified):

```python

class AlreadyLockedError(Exception): pass

class MyLock:

    def __init__(self, redis_client, key):
         self.lock = redis_client.lock(key, blocking_timeout=2)

    def __enter__(self):
        if self.lock.acquire():
             return self
        else:
             raise AlreadyLockedError

     def __exit__(self, type, value, tb):
         self.lock.release()

# example:
with MyLock("keyname"):
    do_something()
```

If the acquiring the lock times out, you receive an exception, instead of executing the do_something() code anyway.  Much safer this way.

nathan-hoad:
I've just discovered this, and I've done a similar thing to @gjcarneiro. Any reason why this hasn't been fixed yet? It makes using locks as context managers quite broken as it stands now.

jclark360:
[...]

PR #927, opened 2017-11-23
__enter__ will raise LockError if failed to acquire lock

chenfengyuan:
#621

1. It is easy to forget to check whether the lock is acquired(test `lock.local.token` value).
2. `Lock.__exit__` method always raise `LockError("Cannot release an unlocked lock")` if lock is not acquired.
3. If we want to use the origin version of `redis.lock`, we must check lock status in the `with statement suite` and wrapper a `try statement` outside the `with statement`.
like codes below:
```python
def foo(start, end):
    return_value = 0, ""
    try:
        for i in range(start, end):
            with redis_client.lock(str(i), blocking_timeout=2) as lock:
                if lock.local.token is None:
                    return_value = -1, "can not acquire {} lock".format(i)
                    return
     except LockError:
         return return_value
     # do other stuff
     return_value = 0, "success"
     return return_value
```
4. I suppose it is better to check wether lock is acquired in the `Lock.__enter__` method and raise `LockError` if necessary

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -79,10 +79,9 @@ class Lock(object):
             raise LockError("'sleep' must be less than 'timeout'")
 
     def __enter__(self):
-        # force blocking, as otherwise the user would have to check whether
-        # the lock was actually acquired or not.
-        self.acquire(blocking=True)
-        return self
+        if self.acquire(blocking=True):
+            return self
+        raise LockError("Cannot acquire lock")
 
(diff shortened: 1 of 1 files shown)


chenfengyuan:
https://github.com/andymccurdy/redis-py/issues/621

yotamofek:
This would really benefit us!

andymccurdy:
Fixed in redis-py 3.0

PR #941, opened 2017-12-25
Fix handling for empty arguments in commands that throw server-side error

tzickel:
Sometimes you pass an array to mget as an input, if the array is empty, instead of returning an empty array, mget throws an exception from the server (rightfully so) but it's not very pythonic.

I have lots of places in the code which call mget with an array that might or might not be empty, since in the end i just iterate over the result, adding an if not empty check before each mget call is superfluous (and if someone really wants to check it, he can).

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -28,6 +28,18 @@ from redis.exceptions import (
 SYM_EMPTY = b('')
 
 
+_empty = object()
+
+
+def is_empty(cmd):
+    return len(cmd) >= 2 and cmd[1] == _empty
+
+
+COMMAND_EMPTY_RETURN_VALUE = {
+    'MGET': []
+}
+
+
 def list_or_args(keys, args):
     # returns a single list combining keys and args
     try:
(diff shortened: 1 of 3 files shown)


itamarhaber:
Break early makes sense to me, and the change is sane. I don't know if the suggested is more pythonic, but it is certainly a nicer UX.

However, I'm worried about backward compatibility as an exception is currently thrown. I'd actually change the return statement to an exception, although I understand this does not address your pain.

andymccurdy:
This breaks pipelines. If you pass an empty list to MGET within a pipeline, you'll never see that response in the pipeline's execute because the MGET request won't get queued.

andymccurdy:
If it can be made to work with pipelines, I think returning an empty list is OK.

tzickel:
@andymccurdy something like this ? this is more generic, you can see the changes now are for MGET.

andymccurdy:
I created a different implementation #1052 

IMHO I think it's a little cleaner.

andymccurdy:
I merged #1052 as I preferred its implementation to this one. It only affected MGET and we'll need a separate PR to add support for other commands with similar behavior, such as DEL.

PR #943, opened 2017-12-28
Properly handle timezones (fixes #598)

### LINKED ISSUE #598: expireat behavior
`expireat` uses `time.mktime` to convert dattimes into epoch timestamps. But mktime makes no check if timezone infos are available and just assumes local timezone.
The confusing result is that if you use datetimes in UTC (which might be the programmers intuitive choice after reading http://redis.io/commands/expireat) or have to deal with many timezones and use aware datetimes, the expiry is wrong.
To solve this `expireat` should check for `tzinfo` and use it if available.

zeebonk:
Orignally this was done using time.mktime and datetime.timetuple but these
functions don't properly handle timezones:

- dt.timetuple creates a time_struct while completely ignoring the timezone of
  the datetime.
- time.mktime assumes the time_struct to be in local time, e.g. whatever the
  machine has set as timezone.

This would end up in bad timestamps when either the machine runs in a different
timezone or the datetime is in a different timezone than UTC.

Also included is a UTC tzinfo so that datetimes generated by redis-py can be
assigned the correct timezone.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1,11 +1,11 @@
 from __future__ import unicode_literals
 from itertools import chain
+import calendar
 import datetime
 import sys
 import warnings
 import time
 import threading
-import time as mod_time
 import hashlib
 from redis._compat import (basestring, bytes, imap, iteritems, iterkeys,
                            itervalues, izip, long, nativestr, safe_unicode)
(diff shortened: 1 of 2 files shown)


zeebonk:
I updated the branch with latest changes from master (awesome work the everyone did for on 3.0!) as I still think this is a relevant change. I can see how this change can break backwards compatibility for people who expect it to use local time. If the change as it stands can not be accepted for that specific reason, perhaps we can talk about different methods of implementation. For example: an opt-in argument than can be passed to the `Client` that triggers UTC usage everywhere and checks that `datetimes` passed to methods on the `Client` are always 'aware'.

PR #950, opened 2018-01-29
Prevent unclosed connections on timeout/connect error (ResourceWarning)

vtermanis:
This patch addresses two potential cases of `ResourceWarning` (e.g. when running interpreter with `-W all` argument) due to socket errors.

1. Fix retry behaviour (`retry_on_timeout`): Previously if retries were not enabled, a retry would still be performed **unless** the exception was `TimeoutError`.
2. When using hiredis module, the socket would not be closed when encountering an exception (e.g. disconnected from server or inability to connect)

**Note**: This does **not** address lack of tidy-up if retries are enabled, where the [retry performs an attempt without a try-except block](https://github.com/andymccurdy/redis-py/blob/3edccee4a18197da0436a6d2d33269190d07ed54/redis/client.py#L674-L675) (unlike the initial attempt).

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -669,7 +669,8 @@ class StrictRedis(object):
             return self.parse_response(connection, command_name, **options)
         except (ConnectionError, TimeoutError) as e:
             connection.disconnect()
-            if not connection.retry_on_timeout and isinstance(e, TimeoutError):
+            if not (connection.retry_on_timeout and
+                    isinstance(e, TimeoutError)):
                 raise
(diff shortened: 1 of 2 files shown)


andymccurdy:
Hi, sorry for the late reply. I've already merged another patch that fixes the retry_on_timeout logic.

As for the HiredisParser/socket issue, that's being addressed in #1108, though in a different way. The Connection classes own the socket and should be the only ones cleaning them up, not the parser classes.

PR #951, opened 2018-02-04
Fix attribute error

yakhira:


andymccurdy:
fixed in redis-py 3.0. thanks!

PR #953, opened 2018-02-11
	modified:   redis/client.py

gqdw:
support command: shutdown nosave with function shutdownnosave()


itamarhaber:
Hello @gqdw 

Thanks for the PR - regrettably, it has been lying unattended for a while.

'NOSAVE' is indeed missing. However, the implementation in #1042 looks to me as more consistent with overall API. What do you think?

andymccurdy:
Agreed. This feature will be implemented in #1042. Sorry for the lag.

PR #955, opened 2018-02-15
Fix: zadd raises unhandled exception - fixes #954

### LINKED ISSUE #954: zadd raises TypeError exception
When i try 
to update score of non-existing field like this:
`r.zadd('fooset', 'xx', 'INCR', 10, 'bar')`
OR 
to update score of existing field like this:
`r.zadd('fooset', 'xx', 'INCR', 10, 'zoo')`

ZADD returns None and method is raising
`TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'`

nitingupta89:
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -367,7 +367,7 @@ class StrictRedis(object):
             'SETBIT SETRANGE SINTERSTORE SREM STRLEN SUNIONSTORE ZADD ZCARD '
             'ZLEXCOUNT ZREM ZREMRANGEBYLEX ZREMRANGEBYRANK ZREMRANGEBYSCORE '
             'GEOADD',
-            int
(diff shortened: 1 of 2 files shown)


nitingupta89:
Fixes #954

nitingupta89:
@andymccurdy Please merge

andymccurdy:
This is because the existing `ZADD` command doesn't support any of the newer options because of the method signature. This will be fixed when the other ZADD options get added, but it'll be a backwards incompatible and need to wait for 3.0

andymccurdy:
Fixed in redis-py 3.0

PR #958, opened 2018-03-02
Convert boolean values to '0' and '1' in Encoder

ledmonster:
Currently, Encoder encodes boolean values to 'True' and 'False'. The words 'True' and 'False' are specific to python environment and not suitable as a value in redis. Current Encoder implementation seems not to 
consider about boolean input.

I changed the behavior to encode boolean values to '0' and '1'.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -108,6 +108,8 @@ class Encoder(object):
             return value.encoded_value
         elif isinstance(value, bytes):
             return value
+        elif isinstance(value, bool):
+            value = b(str(int(value)))
         elif isinstance(value, (int, long)):
             value = b(str(value))
         elif isinstance(value, float):
--- a/tests/test_encoder.py
+++ b/tests/test_encoder.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+from __future__ import with_statement
+import pytest
+import sys
+
+from redis.connection import Encoder, Token
+from redis._compat import b, long, unicode
+
+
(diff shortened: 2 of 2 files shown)


itamarhaber:
Hello @ledmonster 

> The words 'True' and 'False' are specific to python environment

True

> and not suitable as a value in redis.

False - why aren't they suitable? I agree 0 and 1 may be more compact (and also benefit from Redis' shared integers dict), but I disagree about suitability.

That said, while the change is interesting, it would break backward compatibility in the most horrible of ways so far down the road. I wonder if there is justification for that.

andymccurdy:
This seems like a bad idea for a variety of reasons. Most notably, Redis has no concept of boolean types. That means the server will return these values as strings and the client will not have enough information to know if they should be converted back into Python booleans, or for that matter, anything that evaluates to Python's `False`.

This conversion needs to happen in your app, prior to the value getting sent to the Redis client, and then again in your app after the client reads the value.

andymccurdy:
I don't see how storing '0'/'1' is any nicer for app code to deal with than 'True'/'False'. `bool('0')` still evaluates to `True` because it's a non-empty string.

This is something you could do when you're trying to store a boolean:

```python
>>> value = False
>>> redis_client.set('key', value or '')
>>> stored = bool(redis_client.get('key'))
>>> stored
False
```

ledmonster:
@andymccurdy Oops! I made a silly mistake. Sorry for bothering you.

PR #964, opened 2018-04-06
Documents rediss:// support in StrictRedis.from_url

remeika:
Documents the fact that StrictRedis.from_url can support the rediss:// protocol to create a TLS connection to Redis.

Fixes #964


andymccurdy:
Thanks!

Issue #970, opened 2018-04-24
Defaults


Lewiscowles1986:
Redis dev-rel just posted [this](https://opensource.com/article/18/4/how-build-hello-redis-with-python) over at OpenSource.com wondering if you'd be open to a PR so that vanilla local-machine settings are not necessary (looks like they are not now), and the client can work with environment vars instead of coding credentials.

Lewiscowles1986:
I am aware this isn't ideal if you have multiple applications that are not isolated or multiple sets of credentials in a single monolithic application. I'm quite sure that is a rare edge case.

Lewiscowles1986:
PR closure should have closed this. My bad for not closing

PR #971, opened 2018-04-24
Env var defaults

Lewiscowles1986:
Addresses #970


andymccurdy:
I don't believe this is a good default option. It's easy for users to implement this themselves in application code if they want to use env vars.

PR #976, opened 2018-05-13
Doc string support for redis standalone scheme rediss:// support in StrictRedis.from_url resolves #964

anandtripathi5:


itamarhaber:
This is actually a duplicate of #964, no?

anandtripathi5:
Yeah, I see now

PR #985, opened 2018-05-23
Fix parsing max_connections URL query string parameter

mmaslowskicc:
Previously ConnectionPool.from_url kept it as a string, causing a 'ValueError:
"max_connections" must be a positive integer'.


andymccurdy:
thanks!

PR #991, opened 2018-06-03
Remove from __future__ import with_statement

jdufresne:
All supported Python versions support the with statement.

Available since 2.5.

https://docs.python.org/3/whatsnew/2.5.html#pep-343-the-with-statement


jdufresne:
Already merged as 882f74a7b257e4eeee4afba2c1ddeb0173cb3cdd

PR #916, opened 2017-10-21
Drop Python 2.6 support

jdufresne:
Remove 2.6 workarounds. Update testing configurations. Use modern Python syntax where available:

* Use set literals
* Use dict comprehensions
* Use more generators
* Use `io` package for `BytesIO`
* Remove `from __future__ import with_statement`
* Use unicode literals throughout

Thanks Dustin Ingram for the original patch.

---

 This is a continuation of the work started by @di in #796. I have rebased the patch and addressed a couple other cleanups I saw. I'll happily close this PR if the other work continues.

PR #992, opened 2018-06-03
Pass generators to dict() and tuple() instead of coercing to a list

jdufresne:
Avoids storing temporary lists in memory.


zeebonk:
Instead of passing generators you could go a step further for lists, dicts and sets and use comprehensions.

jdufresne:
Thanks @zeebonk . That suggestiion is included in #916.

jdufresne:
Merged as d05afabcc038764d4edfe5040c4ce783da691724

PR #993, opened 2018-06-03
Remove Queue package workarounds for older unsupported Pythons

jdufresne:
Python 2.5 is neither supported nor tested by redis-py. Can remove
workaround.


jdufresne:
Merged as 3f1e817f24b4b51c1d1f17a7c1ef5a3bc5746751

Issue #605, opened 2015-04-06
Is MIGRATE implemented ? 


hostmaster:
Is command [MIGRATE](http://redis.io/commands/migrate) implemented ?

imankulov:
Apparently, it's not, but it's quite trivial to fall back to `execute_command`, with something like 

```python
>>> r.execute_command('migrate', '127.0.0.1', 6379, key, 0, 5000)
```

P.S. Some questions are worth replying even in more than a year :)

nashley:
Is there a technical reason why it's not implemented, or does someone just need to take the implement it?

Grokzen:
@andymccurdy This one can be closed as solved. #1000 talks about it and #1064 implemented it.

PR #896, opened 2017-09-06
Update client.py

landonsilla:
I am adding support for the MIGRATE command in this driver:
https://redis.io/commands/migrate
Thanks!

PR #1000, opened 2018-06-27
add migrate command

i0seph:
add migrate command

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -409,6 +409,7 @@ class StrictRedis(object):
             'HSCAN': parse_hscan,
             'INFO': parse_info,
             'LASTSAVE': timestamp_to_datetime,
+            'MIGRATE': bool_ok,
             'OBJECT': parse_object,
             'PING': lambda r: nativestr(r) == 'PONG',
             'RANDOMKEY': lambda r: r and r or None,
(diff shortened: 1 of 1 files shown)


itamarhaber:
Fixes #605, duplicate of #896.

itamarhaber:
Hello @i0seph 

Thanks for making this PR - `MIGRATE` is an important part of the Redis API, so this should be merged after a review.

redis/client.py
```diff
@@ -769,6 +770,27 @@ def lastsave(self):
         """
         return self.execute_command('LASTSAVE')
 
+    def migrate(self, host, port, key, destination_db, timeout, copy=False, replace=False, auth=None, keys=[]):
```
itamarhaber: I think it would be nicer if the signature would be keys, accepting either a string or a list of strings representing the names. Internally, you'll replace the 'key' with the single key, or an empty string and the `KEYS` keyword followed by the list.

redis/client.py
```diff
@@ -769,6 +770,27 @@ def lastsave(self):
...
+        """
+        Atomically transfer a key from a Redis instance to another one.
+        migrate(host, port, key, destination_db, timeout, copy=False, replace=False, auth=None, *keys)
+        """
+        remain_args = None
```
itamarhaber: Other functions in the lib use `args`, and then call something like self.execute('somecommand', *args). Would be nicer to be consistent.

andymccurdy:
I went ahead and implemented the changes in #1064. Thanks for letting us know migrate was missing!

PR #1004, opened 2018-07-09
Update from url func

dayiguizhen:
Let developer can chose BlockingConnectionPool when use from_url.


andymccurdy:
This is in my opinion beyond the scope of the client class. If you want a `BlockingConnectionPool` via an URL, you can pass the URL to `BlockingConnectionPool.from_url()` which you can then pass to the `connection_pool` argument of the client class.

itamarhaber:
Thanks for the PR @dayiguizhen - closing because this is out of scope.

PR #1007, opened 2018-07-10
Implement Lock.locked() check, as threading.Lock does

alanjds:
`threading.Lock` have a way to check if it is locked right now.

Implemented by checking if the `Lock.name` key is present in the redis.

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -127,6 +127,11 @@ class Lock(object):
             return True
         return False
 
+    def locked(self):
+        if self.redis.get(self.name) is not None:
+            return True
+        return False
+
     def release(self):
         "Releases the already acquired lock"
         expected_token = self.local.token
--- a/tests/test_lock.py
+++ b/tests/test_lock.py
@@ -15,7 +15,9 @@ class TestLock(object):
 
     def test_lock(self, sr):
         lock = self.get_lock(sr, 'foo')
+        assert lock.locked() is False
         assert lock.acquire(blocking=False)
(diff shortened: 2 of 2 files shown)


andymccurdy:
Thanks! I've slightly modified your patch to validate the token within redis and the lock object are the same.

alanjds:
Hum...

Question: A name locked on one machine should appear locked on another machine?

If yes (and I think so), you should not validate the token. Any token from another machine should be able to "lock my Lock". Right?

andymccurdy:
Checking the token is important to make sure the lock is still owned. If the lock is acquired, then the lock expires due to timeout, I don’t believe it should still be considered locked.

I don’t understand how this would interfere with the lock on other processes. Each process with the lock has its own token while the server is the source of truth.

If I’m missing something can you put together a short code snippet illustrating your problem?

alanjds:
I see. Maybe I had misunderstood the token system.

If an application on Server A got the lock, the Server B will see it as locked as well?

If so, I am ok.

andymccurdy:
Oh I see. I think we're talking about different things then.

I was under the impression that `.locked()` should verify whether the lock has been acquired by the current process.

If I understand your desire correctly, you want `.locked()` to indicate that the lock has been acquired by **any** process. Is that correct?

alanjds:
Yes, with comparable behaviours of `threading.Lock` and `multiprocessing.Lock`!

andymccurdy:
ok, 3.0.1 released with the locked() change. Thanks!

PR #1014, opened 2018-07-24
Add `.reacquire()` method to Lock

ikalnytskyi:
`Lock` class provides a method called `.extend()` to manage a TTL of the acquired lock. However, the method allows only to extend a timeout of existing lock by N seconds, there's no way you can reset a TTL to the timeout value you passed to this lock. There could be multiple use cases for such behaviour. For instance, one may want to use a lock to implement active/passive behaviour where only one process owns a lock and resets its TTL all over again until it dies. This commit adds a new method called `.reacquire()` to reset a TTL of the acquired lock back to the passed timeout value.

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -16,6 +16,7 @@ class Lock(object):
 
     lua_release = None
     lua_extend = None
+    lua_reacquire = None
 
     # KEYS[1] - lock name
     # ARGS[1] - token
@@ -46,6 +47,19 @@ class Lock(object):
             return 0
         end
         redis.call('pexpire', KEYS[1], expiration + ARGV[2])
+        return 1
+    """
+
+    # KEYS[1] - lock name
+    # ARGS[1] - token
+    # ARGS[2] - milliseconds
+    # return 1 if the locks time was reacquired, otherwise 0
+    LUA_REACQUIRE_SCRIPT = """
+        local token = redis.call('get', KEYS[1])
+        if not token or token ~= ARGV[1] then
+            return 0
+        end
+        redis.call('pexpire', KEYS[1], ARGV[2])
         return 1
     """
 
@@ -121,6 +135,9 @@ class Lock(object):
             cls.lua_release = client.register_script(cls.LUA_RELEASE_SCRIPT)
         if cls.lua_extend is None:
             cls.lua_extend = client.register_script(cls.LUA_EXTEND_SCRIPT)
+        if cls.lua_reacquire is None:
+            cls.lua_reacquire = \
+                client.register_script(cls.LUA_REACQUIRE_SCRIPT)
 
     def __enter__(self):
         # force blocking, as otherwise the user would have to check whether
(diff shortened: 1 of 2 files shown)


ikalnytskyi:
Hey @andymccurdy, do you want this patch hanging here? Or should I close this PR because you have no interest in this?

andymccurdy:
Hi @ikalnytskyi, sorry for the delay. I'm fine adding something like this but I don't think `reset` implies that the lock timeout is being renewed. 

What you're trying to do is very similar to `extend` with one exception: `extend` adds the `additional_time` to the existing time left. We could slightly modify `extend` instead of creating a new method. Perhaps something like:

`def extend(self, additional_time, add_to_existing_time=True)`

The `add_to_existing_time` flag controls whether the `additional_time` is added to the time remaining on the lock or not. For your specific use case, you'd call it like: `lock.extend(lock.timeout, add_to_existing_time=False)`.

ikalnytskyi:
@andymccurdy I think `.extend(add_to_existing_time=False)` should work just fine. :ok_hand:  I just find this a little bit confusing in terms that `extend` usually means you add something, not reset, and that's why in my opinion `.reset` gives you more context on what's going on.

Anyway, it's your call. I can update a PR, just let me know which way I should go? Adding a flag to `.extend`?

ikalnytskyi:
@andymccurdy renamed to `.reacquire()`. the new name sounds good to me.

redis/lock.py
```diff
@@ -214,3 +231,21 @@ def do_extend(self, additional_time):
...
+        if self.timeout is None:
+            raise LockError("Cannot reacquire a lock with no timeout")
+        return self.do_reacquire()
+
+    def do_reacquire(self):
+        timeout = int(self.timeout * 1000)
+        if not bool(self.lua_reacquire(keys=[self.name],
+                                       args=[self.local.token, timeout],
+                                       client=self.redis)):
+            raise LockError("Cannot reacquire a lock that's no longer owned")
```
andymccurdy: This should raise a `LockNotOwnedError`.

tests/test_lock.py
```diff
@@ -125,6 +125,34 @@ def test_extending_lock_no_longer_owned_raises_error(self, r):
...
+        assert lock.acquire(blocking=False)
+        with pytest.raises(LockError):
+            lock.reacquire()
+        lock.release()
+
+    def test_reacquiring_lock_no_longer_owned_raises_error(self, r):
+        lock = self.get_lock(r, 'foo')
+        assert lock.acquire(blocking=False)
+        r.set('foo', 'a')
+        with pytest.raises(LockError):
```
andymccurdy: This should expect a `LockNotOwnedError` to be raised.

tests/test_lock.py
```diff
@@ -125,6 +125,34 @@ def test_extending_lock_no_longer_owned_raises_error(self, r):
         with pytest.raises(LockNotOwnedError):
             lock.extend(10)
 
+    def test_reacquire_lock(self, r):
+        lock = self.get_lock(r, 'foo', timeout=10)
+        assert lock.acquire(blocking=False)
+        time.sleep(1)
```
andymccurdy: Instead of sleeping (and slowing down the test suite) could we manually `PEXPIRE foo 1000` after acquiring the lock initially? Then after the `reacquire` we should see `pttl foo` back in the 9000-10000 range.

ikalnytskyi:
@andymccurdy just updated the patch. could you please take a look one more time?

andymccurdy:
Looks good, thanks!

PR #1017, opened 2018-08-07
Enforce ssl_cert_reqs='required' by default

### LINKED ISSUE #1016: redis-py ssl support susceptible to MITM attacks by default
**Issue:** redis-py doesn't enforce hostname validation (Common Name nor Subject Alternative Name) by default when accepting a cert from a remote SSL terminator. This default behavior isn't compatible to accepted PEPs/RFCs and provides a dangerous sense of false security.

**Task:** Correct redis-py to validate certificates by default. IMHO this shouldn't be considered a breaking change as it simply reenforces the expected results when initiating a SSL connection.

**Additional research:**
- https://tools.ietf.org/html/rfc6125
- https://www.python.org/dev/peps/pep-0476/
- https://wiki.openssl.org/index.php/Hostname_validation

u2mejc:
This resolves #1016

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -496,7 +496,7 @@ class StrictRedis(object):
                  charset=None, errors=None,
                  decode_responses=False, retry_on_timeout=False,
                  ssl=False, ssl_keyfile=None, ssl_certfile=None,
-                 ssl_cert_reqs=None, ssl_ca_certs=None,
+                 ssl_cert_reqs='required', ssl_ca_certs=None,
                  max_connections=None):
         if not connection_pool:
             if charset is not None:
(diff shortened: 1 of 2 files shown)


u2mejc:
In testing this with a coworker, he pointed out that it's also not possible to use `ssl.wrap_socket` without passing the `ca_certs`. This makes the code rigid as it may not be possible the change cert providers without updating your python code.


Example: to accept the current default AWS Elasticache cert on ubuntu, you need to set: 
`r = redis.StrictRedis(host, ssl=True, ssl_ca_certs='/usr/share/ca-certificates/mozilla/Starfield_Class_2_CA.crt')`

andymccurdy:
Thanks. Merging this now.

edmorley:
Hi :-)

This was listed under "other changes" in the release notes rather than "backwards incompatible changes", which meant we missed this change when reviewing the upgrade, which resulted in breakage in our staging environment until rolled back.

Could the release notes be updated?

PR #1018, opened 2018-08-08
Fix parsing for INFO keys that include ':'

mzalimeni:
As of https://github.com/antirez/redis/commit/a81a92ca2ceba364f4bb51efde9284d939e7ff47 (first included in Redis 4), there is a new 'host:' pseudo-command. This caused the addition of an INFO commandstat key 'cmdstat_host:', which results in incorrect parsing of the INFO command into a Python dict by the `client.parse_info` function.

Before fix:
`{
...
'cmdstat_host': {':calls': 6, 'usec': 601, 'usec_per_call': 100.17}
...
}`

After fix:
`{
...
'cmdstat_host:': {'calls': 6, 'usec': 601, 'usec_per_call': 100.17}
...
}`

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -114,7 +114,8 @@ def parse_info(response):
     for line in response.splitlines():
         if line and not line.startswith('#'):
             if line.find(':') != -1:
-                key, value = line.split(':', 1)
+                # support keys that include ':' by using rsplit
+                key, value = line.rsplit(':', 1)
                 info[key] = get_value(value)
             else:
                 # if the line isn't splittable, append it to the "__raw__" key


andymccurdy:
thanks

oranagra:
@andymccurdy @itamarhaber @mzalimeni i think this fix must be done in a different way.
the redis info fields may contain `:` character, which will cause this fix to break.
two simple examples are `executable` and `config_file`.
e.g. if the user named his config file `prefix:suffix.conf` will cause the key to named `config_file:prefix` and value `suffix`.

but what's worse, when redis is running on an IPv6 system, the slave list may contains an IPv6 address, and it would cause redis-py to crash processing the INFO.

e.g. if the info field is: `slave0:ip=2001:db8:8714:3a90::12,port=12345,state=up,offset=10,lag=10`
the key would be named `slave0:ip=2001:db8:8714:3a90:`
and the value would be `['12', 'port=12345', 'state=up', 'offset=10', 'lag=10']`
and then the following line will crash
```k, v = item.rsplit('=', 1)```

oranagra:
IMHO, the proper fix for the `host:` "command" should be to explicitly check for `cmdstat_host:` string and give it a different handling (since that is the only case where a field name contains `:`)

mzalimeni:
@oranagra thanks for catching my mistake and putting up a new PR. Definitely should not have overlooked the possibility of colons in other INFO values.

PR #1019, opened 2018-08-13
 Allow pings in PubSub

krishan-carbon:
According to https://redis.io/topics/pubsub, “The commands that are allowed in the context of a subscribed client are SUBSCRIBE, PSUBSCRIBE, UNSUBSCRIBE, PUNSUBSCRIBE, PING and QUIT.”

According to https://redis.io/commands/ping, “If the client is subscribed to a channel or a pattern, it will instead return a multi-bulk with a "pong" in the first position and an empty bulk in the second position, unless an argument is provided in which case it returns a copy of the argument.”

The motivation for this is as follows:
We use redis-py in production for caching, key-value storage and pubsub. Some of our uses of pubsub are very sparse - publishers could potentially only publish every 3 days. We find, in about 1 out of 20 instances where this happens, that subscribers "stop listening" to a channel. Inspecting the local network we can see the channels are open but inspecting redis, it no longer is aware of these subscribers. Our hypothesis on this is that the connection has failed but since these instances of PubSub only subscribe and listen, nothing triggers the "reconnect" mechanism (and so `PubSub.on_connect` is never called). Redis recently (v3) implemented PINGs whilst subscribed which would give us the ability to routinely check the connection is still alive and reconnect if needed.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2521,6 +2521,7 @@ class PubSub(object):
         """
         Ping the Redis server
         """
+        message = '' if message is None else message
         return self.execute_command('PING', message)
 
     def handle_message(self, response, ignore_subscribe_messages=False):
--- a/redis/client.py
+++ b/redis/client.py
@@ -2517,6 +2517,12 @@ class PubSub(object):
             return self.handle_message(response, ignore_subscribe_messages)
         return None
 
+    def ping(self):
+        """
+        Ping the Redis server
+        """
+        return self.execute_command('PING')
+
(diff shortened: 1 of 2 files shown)


krishan-carbon:
PING 

... haha

krishan-carbon:
(added functionality to send a message with each ping)

krishan-carbon:
the tests seem to have failed to run on two versions, is there a way to restart them?

andymccurdy:
Everything here looks great. Thanks for the work on this. Merging now.

PR #1035, opened 2018-09-30
handle the case where geodist return -1

qingping209:
geodist may return none or float, response handler callback should be float_or_none


qingping209:
ping

andymccurdy:
Thanks.

PR #1036, opened 2018-10-05
Adds v5  new client subcommands

itamarhaber:
Specifically, `CLIENT ID` and `CLIENT UNBLOCK`.


itamarhaber:
@andymccurdy agreed - merging @RoeyPrat's PR to address that.

PR #1037, opened 2018-10-05
Adds support for ZPOPMAX, ZPOPMIN and their respective blocking variants

itamarhaber:


SupermanIsEvious:
@andymccurdy What is the release timeline for this feature? It's visible in the `latest` docs, but when installing the latest tag from PYPI these commands are not available yet. I would just like to know so that I can determine if I should implement my own version locally for now.

andymccurdy:
@SupermanIsEvious We'll be releasing a new version sometime next week. Thanks for your patience.

PR #1039, opened 2018-10-12
"while 1" --> "while True"

jeffwidman:
This is python, not C


andymccurdy:
Ya, I believe I read long ago that `while 1:` was faster in earlier versions of Python. I just tested locally and it looks like both are the same. Merging. Thanks!

PR #1040, opened 2018-10-17
supports for all streams operations

RoeyPrat:
@andymccurdy @nicois
this should give us support of all streams commands in redis 5.0
based on nicois work in https://github.com/andymccurdy/redis-py/pull/920, and commits from @itamarhaber.
also fixed some issues with tests, and changed CI to run on newer redis version.


andymccurdy:
Great, thanks. I’m visiting family for a few days and will be back on Wednesday. I’ll plan on reviewing/merging this then.

One observation already: looks like redis 5.0 is fully released now. Can we update the Travis config to pull the full release version and update the skip test decorators back to 5.0?

itamarhaber:
@andymccurdy enjoy the visit :)

RoeyPrat:
@andymccurdy updated travis config to redis5. 
this PR is not running travis by automation, but you can see the run [here](https://travis-ci.org/RoeyPrat/redis-py/builds/444006986)

redis/client.py
```diff
@@ -232,6 +232,61 @@ def int_or_none(response):
...
+    return result
+
+
+def parse_xclaim(response):
+    if isinstance(response, str):
+        return response
+    return stream_list(response)
+
+
+def multi_stream_list(response):
```
andymccurdy: There are several options for the return structure of `multi_stream_list`:

```python
# a list of of 2-item stream/messages lists
[
    [
        'stream-1',
        [
            [
                '1540456647578-0',
                {'key-1': 'value-1', 'key-2': 'value-2'}
            ],
            [
                '1540456647578-1',
                {'key-1': 'value-1', 'key-2': 'value-2'}
            ],
        ],
    ],
    [
        'stream-2',
        [
            [
                '1540456647578-0',
                {'key-1': 'value-1', 'key-2': 'value-2'}
            ],
            [
                '1540456647578-1',
                {'key-1': 'value-1', 'key-2': 'value-2'}
            ],
        ],
    ],
]

# this allows all list-based iterating....
for stream, messages in r.xread(streams={'stream-1': 0, 'stream-2': 0}):
    for message_id, message_data in messages:
        ....


# alternatively, the existing code represents this return value
# as a dict of streams to lists of message id: dict attr pairs
{
    'stream-1': [
        [
            '1540456647578-0',
            {'key-1': 'value-1', 'key-2': 'value-2'}
        ],
        [
            '1540456647578-1',
            {'key-1': 'value-1', 'key-2': 'value-2'}
        ],
    ],
    'stream-2': [
        [
            '1540456662589-0',
            {'key-1': 'value-1', 'key-2': 'value-2'}
        ],
        [
            '1540456662589-1',
            {'key-1': 'value-1', 'key-2': 'value-2'}
        ],
    ]
}

# iterating here requires the extra .items() call
for stream, messages in r.xread(streams={'stream-1': 0, 'stream-2': 0}).items()
    for message_id, message_data in messages:
        ....
# given redis returns null when there are no messages to read,
# you can't reliably index directly into the dictionary, which seems to
# take away much of the value of the dictionary
# this blows up with a TypeError if there's no message to read
# since xread()'s return value will be None.
r.xread(streams={'stream-1': 99999999999})['stream-1']
```

I think the first option might be the best choice here. I don't see any value from making the return value a dict of {stream: messages} pairs when you can't reliably index into the dictionary. It seems like this just forces users to type more. Am I missing something?

redis/client.py
```diff
@@ -1675,6 +1749,360 @@ def sunionstore(self, dest, keys, *args):
         args = list_or_args(keys, args)
         return self.execute_command('SUNIONSTORE', dest, *args)
 
+    # STREAMS COMMANDS
+    def xadd(self, _name, id='*', maxlen=None, approximate=True, **kwargs):
```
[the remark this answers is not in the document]
RoeyPrat: ok. changing it to "fields" and setting it as a must arg

redis/client.py
```diff
@@ -1675,6 +1749,360 @@ def sunionstore(self, dest, keys, *args):
...
+               prevent 'name' used in the kwargs
+        id: Location to insert this record. By default it is appended.
+        maxlen: truncate old stream members beyond this size
+        approximate: actual stream length may be slightly more than maxlen
+        **kwargs: key/value pairs to insert into the stream
+
+        """
+        pieces = []
+        if maxlen is not None:
+            if not isinstance(maxlen, int) or maxlen < 1:
```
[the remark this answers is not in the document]
RoeyPrat: ok

redis/client.py
```diff
@@ -1675,6 +1749,360 @@ def sunionstore(self, dest, keys, *args):
...
+                raise RedisError('XADD maxlen must be a positive integer')
+            pieces.append('MAXLEN')
+            if approximate:
+                pieces.append('~')
+            pieces.append(str(maxlen))
+        pieces.append(id)
+        for pair in iteritems(kwargs):
+            pieces.append(pair[0])
```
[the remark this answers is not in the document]
RoeyPrat: ok

redis/client.py
```diff
@@ -1675,6 +1749,360 @@ def sunionstore(self, dest, keys, *args):
...
+        start: first stream ID. defaults to '-',
+               meaning the earliest available.
+        finish: last stream ID. defaults to '+',
+                meaning the latest available.
+        count: if set, only return this many items, beginning with the
+               earliest available.
+        """
+        pieces = [start, finish]
+        if count is not None:
+            if not isinstance(count, int) or count < 1:
```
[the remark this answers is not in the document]
RoeyPrat: ok

redis/client.py
```diff
@@ -1675,6 +1749,360 @@ def sunionstore(self, dest, keys, *args):
...
+        pieces = [name, groupname, consumername, str(min_idle_time)]
+        pieces.extend(list(message_ids))
+
+        optional_ints = {idle: 'idle', time: 'time', retrycount: 'retrycount'}
+        for param, param_name in optional_ints.items():
+            if param is not None:
+                if not isinstance(param, int):
+                    raise RedisError("XCLAIM {} must be an integer"
+                                     .format(param_name))
+                pieces.append(str(param))
```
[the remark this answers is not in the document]
RoeyPrat: actually that is the int value. what i forgot to add was the param name...
renaming param to param_value, and adding param_name to pieces.

redis/client.py
```diff
@@ -232,6 +232,61 @@ def int_or_none(response):
     return int(response)
 
 
+def stream_key(response):
```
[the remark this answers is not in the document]
RoeyPrat: XADD removing unnecessary stream_key parse function

redis/client.py
```diff
@@ -232,6 +232,61 @@ def int_or_none(response):
...
+        return response
+    return stream_list(response)
+
+
+def multi_stream_list(response):
+    if response is None:
+        return None
+    result = dict()
+    for r in response:
+        result[r[0].decode('utf-8')] = stream_list(r[1])
```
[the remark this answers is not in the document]
RoeyPrat: ok.
multi_stream_list should use nativestr for compatibility

redis/client.py
RoeyPrat: ok. 
allow list based iterating on XREADGROUP results

redis/client.py
```diff
@@ -232,6 +232,76 @@ def int_or_none(response):
...
+    return stream_list(response)
+
+
+def parse_xreadgroup(response):
+    if response is None:
+        return None
+    return [[nativestr(r[0]), stream_list(r[1])] for r in response]
+
+
+def parse_xpending(response):
```
andymccurdy: Response callbacks can accept **options. These come from passing **kwargs to `execute_command`. In the `xtrim_range` function, you could pass a `parse_detail=True` to `execute_command`, then here say something like:

```python
if options.get('parse_detail', False):
    return parse_range_xpending(response)
```

redis/client.py
```diff
@@ -1675,6 +1764,323 @@ def sunionstore(self, dest, keys, *args):
...
+    def xpending(self, name, groupname):
+        """
+        Returns information about pending messages of a group.
+        name: name of the stream.
+        groupname: name of the consumer group.
+        """
+        return self.execute_command('XPENDING', name, groupname)
+
+    def xpending_range(self, name, groupname, start='-', end='+', count=-1,
+                       consumername=None):
```
coleifer: The count=-1 is something I had been using, but it is not supported going forward I don't believe. See:

https://github.com/antirez/redis/pull/5459#issuecomment-433929399

redis/client.py
```diff
@@ -1675,6 +1764,323 @@ def sunionstore(self, dest, keys, *args):
...
+        if count is not None:
+            if not isinstance(count, (int, long)) or count < 1:
+                raise RedisError('XREAD count must be a positive integer')
+            pieces.append(Token.get_token('COUNT'))
+            pieces.append(str(count))
+        if not isinstance(streams, dict) or len(streams) == 0:
+            raise RedisError('XREAD streams must be a non empty dict')
+        pieces.append(Token.get_token('STREAMS'))
+        pieces.extend(streams.keys())
+        pieces.extend(streams.values())
```
coleifer: Is it guaranteed that these will line up in earlier Python? Since dicts aren't *technically* ordered (at least until 3.7) it might be safer to be explicit here.

```python
keys, values = zip(*streams.items())
```

redis/client.py
```diff
@@ -1675,6 +1764,323 @@ def sunionstore(self, dest, keys, *args):
...
+         relative amount of milliseconds, it sets the idle time to a specific
+         Unix time (in milliseconds).
+        retrycount: optional integer. set the retry counter to the specified
+         value. This counter is incremented every time a message is delivered
+         again.
+        force: optional boolean, false by default. Creates the pending message
+         entry in the PEL even if certain specified IDs are not already in the
+         PEL assigned to a different client.
+        justid: optional boolean, false by default. Return just an array of IDs
+         of messages successfully claimed, without returning the actual message
```
coleifer: The redis docs mention some of these parameters are more targeted towards internal usage:

> The command has multiple options, however most are mainly for internal use in order to transfer the effects of XCLAIM or other commands to the AOF file and to propagate the same effects to the slaves, and are unlikely to be useful to normal users:

Do you think it's worthwhile exposing more than just name/group/consumer/min idle/id list?

andymccurdy:
> XREAD / XREADGROUP are kind of awkward...they supports multiple streams which can all have different IDs....

Currently I'm fine just support the single, explicit dict form. 

> Additionally, the response from the XREADs is kind of weird, because it returns _up to_ `count` rows from each stream passed in...The response is then read as a dict of `{stream: [list of messages], other_stream: [list of other stream messages]}`... which makes sense (and is what happens by default in my implementation). But it might also be nice if a single stream is passed-in to just skip the dict business and return the list? IDK...special cases and all that. I've implemented a high-level `TimeSeries` class that just interleaves the messages from multiple streams to form a single list with a total ordering and that's very handy, but not suitable for a low-level API like this.

@RoeyPrat initially implemented this has a dict with stream names as keys and a list of messages as the value. We talked a bit and found that having a list of (stream, message_list) pairs made for easier iteration. In practice, I can't see anyone using any of the dict functions. You (almost always) want to consume all messages returned to you, especially with `XREADGROUP`. The new implementation allows for:

```python
for stream, messages in r.xread(...):
    for message_id, fields in messages:
        ...
```

redis/client.py
andymccurdy: See https://docs.python.org/2/library/stdtypes.html#dict.items Specifically:

"If items(), keys(), values(), iteritems(), iterkeys(), and itervalues() are called with no intervening modifications to the dictionary, the lists will directly correspond."

andymccurdy: but ya, streams.items() seems better. I'll fix. thx.

coleifer:
> We talked a bit and found that having a list of (stream, message_list) pairs made for easier iteration. 

Brilliant, I must've misread the code. Probably no point in trying to interleave the messages to provide a total ordering, right?

Looking forward to getting these changes integrated into ol' walrus.

andymccurdy:
@coleifer It's early and I haven't had coffee yet. What do you mean by "interleaving the messages to provide a total ordering"?

itamarhaber:
I ass-u-me @coleifer means "order the messages by their ids"?

itamarhaber:
BTW congrats on this merge - hoping to see more action in this repo ;)

coleifer:
Yeap to both

coleifer:
I wondered about similar things...Redis return values can be awkward at times. My sense is that, since the message ID is discrete from any ID key/value within the message body, it makes sense to treat the message/body as a 2-tuple. It is also close to what the protocol is doing, so it should be easy to understand if you're coming from the protocol docs to the python client. Similarly with the list of (stream, message-list) for read responses. I get the sense that redis-py is first-and-foremost a straight up redis client, so any "interpretation" beyond basic translation into python types is left to the application developer (?). That's not to say you wouldn't add a higher-level `Stream` / whatever type -- similar to how pub/sub is handled -- but having the `x____` commands stick close to the protocol seems good. People can always manipulate the data how they want, since nothing is lost.

redis/client.py
```diff
@@ -1675,6 +1764,323 @@ def sunionstore(self, dest, keys, *args):
...
+        return self.execute_command('XREAD', *pieces)
+
+    def xgroup_create(self, name, groupname, id):
+        """
+        Create a new consumer group associated with a stream.
+        name: name of the stream.
+        groupname: name of the consumer group.
+        id: ID of the last item in the stream to consider already delivered.
+        """
+        return self.execute_command('XGROUP CREATE', name, groupname, id)
```
coleifer: There's a new option to make the streams, "MKSTREAM".

Walrus impl for comparison:

```python
    def xgroup_create(self, key, group, id='$', mkstream=False):
        """
        Create a consumer group.

        :param key: stream key -- must exist before creating a group if
            mkstream is ``False`` (default).
        :param group: consumer group name
        :param id: set the id of the last-received-message
        :param mkstream: create the stream automatically
        """
        cmd = ['XGROUP', 'CREATE', key, group, id]
        if mkstream:
            cmd.append('MKSTREAM')
        return self.execute_command(*cmd) == b'OK'
```

andymccurdy: Is this something in beta? I see no mention of it on https://redis.io/commands/xgroup

coleifer: https://github.com/antirez/redis/commit/cb27dd1a6

coleifer:
Thanks for this excellent patch. I removed all the overrides from my own lib here: 
ef25195a653cdd862fc64123e9686ea1480362ac

hell yeaaa: `7 files changed, 153 insertions(+), 653 deletions(-)`

redis/client.py
andymccurdy: added the MKSTREAM option in ff3bbdf. Thanks for the heads up.

coleifer:
Just a heads-up this command is currently in unstable and I'd assume it'll make its way into the next 5.x release: `XSETID <stream> <id>` for setting the max-id for a stream *outside* the context of a consumer group.

Note that the comment regarding the command arguments is actually incorrect. I've created a ticket for this here: https://github.com/antirez/redis/issues/5519

PR #1042, opened 2018-10-19
Adding a NOSAVE option to the SHUTDOWN command

### LINKED ISSUE #1041: command SHUTDOWN missing NOSAVE option
The SHUTDOWN command doesn't implement the NOSAVE option that Redis supports.  This option prevents Redis from saving the current in-memory state to the persistent files before a shutdown.

dwilliams-kenzan:
Fixes #1041 

Adding a NOSAVE option to the shutdown command.

~~Also added tests for the shutdown command and for the shutdown command with the NOSAVE option.  They're simple tests, but they wait for reconnection so other tests don't fail.~~

Reverted the tests back out as the Redis server doesn't recover on TravisCI.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -822,12 +822,12 @@ class StrictRedis(object):
         "Returns a list of slaves for ``service_name``"
         return self.execute_command('SENTINEL SLAVES', service_name)
 
-    def shutdown(self, option=None):
+    def shutdown(self, save=True):
         "Shutdown the server"
         try:
             args = ['SHUTDOWN']
-            if option is not None:
-                args.append(option)
+            if not save:
+                args.append('NOSAVE')
             self.execute_command(*args)
         except ConnectionError:
             # a ConnectionError here is expected
--- a/redis/client.py
+++ b/redis/client.py
@@ -822,10 +822,13 @@ class StrictRedis(object):
         "Returns a list of slaves for ``service_name``"
         return self.execute_command('SENTINEL SLAVES', service_name)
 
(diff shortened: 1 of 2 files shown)


redis/client.py
```diff
@@ -822,10 +822,13 @@ def sentinel_slaves(self, service_name):
         "Returns a list of slaves for ``service_name``"
         return self.execute_command('SENTINEL SLAVES', service_name)
 
-    def shutdown(self):
+    def shutdown(self, save=True):
         "Shutdown the server"
```
itamarhaber: It would be nice to have the `save` argument documented in the docstring

itamarhaber:
Hello @dwilliams-kenzan 

Thanks for catching this - I agree nosave should be supported. Please resolve my only comment - other than that looks good to merge. @andymccurdy - thoughts?

andymccurdy:
I just realized that Redis's `SHUTDOWN` command supports both `SAVE` and `NOSAVE` options. Given the parameter here is `save`, but it's really influencing the `NOSAVE` behavior, that seems confusing.

Instead, I think the code should look like this:

```python
def shutdown(self, save=False, nosave=False):
    if save and nosave:
        raise RedisError('SHUTDOWN save and nosave cannot both be set')
    ...
```

itamarhaber:
Makes sense.

dwilliams-kenzan:
I've update the code based on the feedback here.

andymccurdy:
Thanks! This looks good. Merging now.

PR #1045, opened 2018-11-02
Remove reference to deprecated easy_install

jdufresne:
easy_install is deprecated and its use is discouraged by PyPA:

https://setuptools.readthedocs.io/en/latest/easy_install.html

> Warning: Easy Install is deprecated. Do not use it. Instead use pip.

Follow upstream advice and only recommended supported tools.


andymccurdy:
Thanks. merged.

PR #1052, opened 2018-11-04
Fail gracefully with a default return value when 0 keys are are provided to a command expecting at least 1 key

andymccurdy:
This is an alternative implementation to #941 

Some commands like `MGET`, `DEL`, etc. accept 1 or more keys but fail if the user supplies 0 keys. From a user point of view, it would be nice if these commands simply returned an empty value when called with 0 keys. For example, `MGET` could return an empty list and `DEL` could return 0 indicating 0 keys were deleted. However, commands cannot simply return an empty value because of how they interact with pipelines.

This PR introduces a well named option `EMPTY_ERROR`. If a command detects one of these 0-key scenarios, it can pass the `EMPTY_ERROR` option with the default value that should be returned to the application. This includes the plumbing to work with pipelines.

Currently only `MGET` is implemented as an example. Other commands like `DEL` should also get this functionality if merged.

The question we need to answer is should we actually do this.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -26,6 +26,7 @@ from redis.exceptions import (
 )
 
 SYM_EMPTY = b('')
+EMPTY_ERROR = 'EMPTY_ERROR'
 
 
 def list_or_args(keys, args):
@@ -757,7 +758,12 @@ class StrictRedis(object):
 
     def parse_response(self, connection, command_name, **options):
         "Parses a response from the Redis server"
-        response = connection.read_response()
+        try:
+            response = connection.read_response()
+        except ResponseError:
+            if EMPTY_ERROR in options:
+                return options[EMPTY_ERROR]
+            raise
         if command_name in self.response_callbacks:
             return self.response_callbacks[command_name](response, **options)
         return response
@@ -1120,7 +1126,10 @@ class StrictRedis(object):
         Returns a list of values ordered identically to ``keys``
         """
         args = list_or_args(keys, args)
-        return self.execute_command('MGET', *args)
+        options = {}
+        if not args:
+            options[EMPTY_ERROR] = []
+        return self.execute_command('MGET', *args, **options)
 
     def mset(self, *args, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


RoeyPrat:
@andymccurdy I like this feature, but it gives me pause.

On one hand, I believe this is easier for the dev, perhaps even more pythonic.
It also matches the part of my brain that still works in SQL ([select * from X where Y is in []](https://www.w3schools.com/sql/sql_in.asp)).

On the other hand, this is a serious behaviour change from Redis.
I'm not that worried about backwards compatibility, though that also might be an issue.

I thought it would be nice to have two classes, one that is more pythonic, and easier to use, and one that strictly does what redis does. Then I noticed that we kind of already have that (Redis vs StrictRedis).

IMHO StrictRedis should be just that - strict, while features like this should be implemented elsewhere.
Maybe Redis class is a good fit for this, maybe not. If you think we need a new class for this, let's call it LiberalRedis or something.

andymccurdy:
@RoeyPrat 

re: Backwards Compatibility: It seems like now is the time to get these kinds of changes in. There will be several compatibility issues with v3 and I'd prefer to get all this kind of stuff in at the same time.

re: StrictRedis vs. Redis: This is a longer discussion, but my hope is that we remove the StrictRedis/Redis distinction for v3. Some history: StrictRedis was created because I initially implemented some commands that deviated away from the server's expected syntax. This lead to lots of questions/bug reports, so I created StrictRedis to adhere to the server's command syntax and kept the Redis class for backwards compatibility. Sadly, this didn't really fix the bug report/question problem -- there's still quite a few bug reports about ZADD/SETEX/etc. because some people continue to use the Redis class.

The docs have encouraged the use of StrictRedis for years now. So I'd like to remove the Redis class, rename StrictRedis -> Redis, and create a StrictRedis alias of Redis. People using StrictRedis today will see no difference and continue to function as they do today without any changes. People using the Redis class will be affected if they use the small group of commands that it overrides. Providing better documentation on this stuff should help reduce the issues going forward.

All this is to say that my feeling has evolved into favoring user happiness and being "more pythonic" over strictly adhering to what the server expects, provided that we:

1. Have great documentation so users know what to expect
2. Provide enhancements over the server's behavior that are lossless, meaning these types of changes are only adding syntactic sugar or create an experience that's superior to the server without hiding any information that the user might need or want.

IMO, this change meets the lossless requirement. It's a better user experience and providing the default return value doesn't hide information or require the need for any special case code.

RoeyPrat:
@andymccurdy Ok. that sounds like a good plan.

tzickel:
@RoeyPrat the constructor could also have another flag for throwing exceptions (like before) or silently handling this.

Also, might be time to start mapping which other commands like MGET need to be patched like this

PR #635, opened 2015-07-15
change exists to variodic

charsyam:
A days ago, redis change exists command can be variodic in unstable branch.

so I send pr for this :)

```
set a 123
set b  123
exists a b
:2
```

Issue #766, opened 2016-07-31
Add multiple keys support on EXISTS


kmcheung12:
Redis [EXISTS](http://redis.io/commands/exists) support multiple keys

As of [https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L939](https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L939) The python binding currently does not .

Suggest changing 'def exists(self, name):' to 'def exists(self, *name):'

thuck:
Looks like this change would break some code.
The EXISTS is considered a bool in the RESPONSE_CALLBACKS; it should be moved to an int, but this would break anything using ```r.exists(key) is True```.

Also there is  ```__contains__ = exists ```; and this should always return a bool; maybe this can be handled like this:
```
def __contains__(self, *name):
        size = len(names)
        size_from_redis = self.exists(*names)
        if size == size_from_redis:
            return True

        return False
```
@andymccurdy what do you think about the breaking compatibility? Or do you have any idea how to avoid it?

kute:
it could be done by yourself.

    def exists(self, *keys):
        return [redis.exists(key) for key in keys]

or

    def exists(self, *keys):
        with redis.pipeline() as pipe:
             for key in keys:
                 pipe.exists(key)
             return pipe.execute()

dgouldin:
I'm not sure if this will be helpful to anybody else, but since I didn't need backwards compatibility with the `is` operator, I created a client subclass that does what I want:

```
from redis.client import Redis as BaseRedis


class Redis(BaseRedis):
    RESPONSE_CALLBACKS = dict(
        BaseRedis.RESPONSE_CALLBACKS,
        EXISTS=int,
    )

    def exists(self, *names):
        return self.execute_command('EXISTS', *names)
```

```
>>> client = Redis()
>>> client.exists('foo', 'bar', 'baz')
0
>>> client.set('foo', 1)
True
>>> client.exists('foo', 'bar', 'baz')
1
>>> 'foo' in client
True
>>> 'bar' in client
False
```

alvinr:
+1 for this, although EXISTS command returns the *number* of keys that exists, its not a boolean.

```Since Redis 3.0.3 it is possible to specify multiple keys instead of a single one. In such a case, it returns the total number of keys existing. Note that returning 1 or 0 for a single key is just a special case of the variadic usage, so the command is completely backward compatible.```

ShaiWavesAudio:
[...]

PR #1053, opened 2018-11-04
update EXISTS but keep it backward compatible

tzickel:
EXISTS can accept varadic number of keys and return how many of them exists.


tzickel:
Hopefully should close #635 #766 as well.

andymccurdy:
I added this just now. Thanks!

PR #1055, opened 2018-11-06
Improve performence of transactions / pipeline requests which involve large chunks of data.

tzickel:
A reissue of #597 based on current master.


andymccurdy:
thanks!

PR #1054, opened 2018-11-05
Support for bitfields

itamarhaber:
WIP - addresses #926 

```python
from redis import StrictRedis, BitField

r = StrictRedis()

r.delete('bfchain')
b = r.bitfield('bfchain').set(8, 0, 1).get(8, 0)
print(b.execute())  # [0, 1]

for i in range(2):
    name = 'bfex{}'.format(i)
    r.delete(name)
    b = BitField()
    b.set(4, 0, i+1)
    b.incrby(2, 1, 2, positional=True)
    print(b.execute(client=r, name=name))  # [0, 3] and [0, 0]

r.delete('bfctx')
with r.bitfield('bfctx') as b:
    b.set(2, 0, 3, signed=False)
    b.overflow(fail=True)
    b.incrby(2, 0, 2)
    print(b.execute())  # [0, None]

```

Signed-off-by: Itamar Haber <itamar@redislabs.com>

PR #1056, opened 2018-11-07
Bitfields

### LINKED ISSUE #926: how to use bitfield command?
how to use bitfield command? Its good to have help doc to explain how to use all redis commands.

andymccurdy:
Shamelessly stole (with consent) @coleifer's bitfield implementation. TBD we should decide if we want to support the slice operations Charles made that look like: https://github.com/coleifer/walrus/blob/617d2efafd3f2021054c228c2e1a1bfcf15f575b/walrus/tests/containers.py#L924

Fixes #926 and provide an alternate implementation to #1054.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1005,6 +1005,13 @@ class StrictRedis(object):
             raise RedisError("Both start and end must be specified")
         return self.execute_command('BITCOUNT', *params)
 
+    def bitfield(self, key):
+        """
+        Return a BitFieldOperation instance to conveniently construct one or
+        more bitfield operations on ``key``.
+        """
+        return BitFieldOperation(self, key)
+
     def bitop(self, operation, dest, *keys):
         """
         Perform a bitwise operation using ``operation`` between ``keys`` and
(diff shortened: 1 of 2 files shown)


andymccurdy:
@coleifer ping, just an FYI. Feel free to comment/participate.

itamarhaber:
@andymccurdy - very nicely done, thanks! I don't see any shame in stealing from the best.

I don't completely agree with overflow being inside incrby, I'm a strict Redis geek after all, but I can live with that :)

As for slicing, I liked @coleifer's suggestion. My pains are the inability (Python's) to support field signage and/or overflowage with it, but perhaps that can be resolved via careful documentation.

coleifer:
righteous! Bear in mind, too, that I stole my initial idea from @itamarhaber's patch, so we're all standing on eachother's shoulders here.

itamarhaber:
I stole mine from Andy's pipeline - a shining proof of [Escherian Relativity](https://en.m.wikipedia.org/wiki/Relativity_(M._C._Escher)).

andymccurdy:
@itamarhaber I added an `overflow()` method just for you :) Also you can now set a `default_overflow` when instantiating a bitfield. This default gets re-applied after the bitfield operation is executed, so if the same instance is used again, the default behavior is still there.

I haven't added slicing yet because of the signage problem. We can always add it later and I don't consider it critical to merging this.

itamarhaber:
Brilliant + I learned about the `extend` list method :)

PR #1057, opened 2018-11-07
CLIENT PAUSE and  type argument in client_list

RoeyPrat:


redis/client.py
```diff
@@ -785,8 +786,20 @@ def client_kill(self, address):
...
+    def client_list(self, _type=None):
+        """
+        Returns a list of currently connected clients.
+        If type of client specified, only that type will be returned.
+        :param _type: optional. one of the client types (normal, master,
+         replica, pubsub)
+        """
         "Returns a list of currently connected clients"
+        if _type is not None:
+            if _type not in ['normal', 'master', 'replica', 'pubsub']:
```
itamarhaber: Perhaps `if str(_type).lower() not in ['normal', 'master', 'replica', 'pubsub']:`?

andymccurdy: Agreed with Itamar.

itamarhaber: Resolved via 0adcc53

itamarhaber:
@andymccurdy you have a clear path to happiness.

PR #1063, opened 2018-11-12
Removes `sudo` from README

### LINKED ISSUE #858: Using sudo with pip is hazardous, remove the recommendation
Mixing pip package management with system package management (apt-get etc) can cause significant issues with the entire Python installation.  It is not uncommon for people to come into #python with damaged installations after having used sudo with pip.  I presume the same goes for the other global installation methods.  The general recommendation is to use virtualenv's for isolation.

Anyways, please consider removing the recommendation of using sudo from the readme.

Cheers,
-kyle

itamarhaber:
Fixes #858

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,13 @@ Installation
 redis-py requires a running Redis server. See `Redis's quickstart
 <https://redis.io/topics/quickstart>`_ for installation instructions.
 
+redis-py can be installed using `pip` similar to other Python packages. Do not use `sudo`
+with `pip`. It is usually good to work in a
+[virtualenv](https://virtualenv.pypa.io/en/latest/) or
+[venv](https://docs.python.org/3/library/venv.html) to avoid conflicts with other package
+managers and Python projects. For a quick introduction see
+[Python Virtual Environments in Five Minutes](bit.ly/py-env).
+
 To install redis-py, simply:
 
 .. code-block:: bash
--- a/README.rst
+++ b/README.rst
@@ -20,13 +20,13 @@ To install redis-py, simply:
 
 .. code-block:: bash
 
-    $ sudo pip install redis
+    $ pip install redis
 
 or from source:
 
 .. code-block:: bash
 
-    $ sudo python setup.py install
+    $ python setup.py install
 
 
 Getting Started
--- a/README.rst
+++ b/README.rst
@@ -18,10 +18,10 @@ redis-py requires a running Redis server. See `Redis's quickstart
 
 redis-py can be installed using `pip` similar to other Python packages. Do not use `sudo`
 with `pip`. It is usually good to work in a
-[virtualenv](https://virtualenv.pypa.io/en/latest/) or
-[venv](https://docs.python.org/3/library/venv.html) to avoid conflicts with other package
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -20,13 +20,13 @@ To install redis-py, simply:
...
 
-    $ sudo pip install redis
+    $ pip install redis
 
 or from source:
 
 .. code-block:: bash
 
-    $ sudo python setup.py install
+    $ python setup.py install
```
altendky: You are actually supposed to use pip for pretty much all installation.  A local install might be `pip install .` with `.` being the working directory or replaced with whatever path to the directory holding `setup.py`.  Depending how granular you like to be, this could certainly be argued to be a separate change.

README.rst
```diff
@@ -20,13 +20,13 @@ To install redis-py, simply:
 
 .. code-block:: bash
 
-    $ sudo pip install redis
+    $ pip install redis
```
altendky: I think it would be good to provide an explicit mention of `virtualenv` and `venv`.  Maybe something like...

> Redis can be installed using pip similar to other Python packages.  Do not use `sudo` with `pip`.  It is usually good to work in a [virtualenv](https://virtualenv.pypa.io/en/latest/) or [venv](https://docs.python.org/3/library/venv.html) to avoid conflicts with other package managers and Python projects.  For a quick introduction see [Python Virtual Environments in Five Minutes](bit.ly/py-env).

Hopefully someday we have an overall better system for handling all this and a canonical description of how to handle installation, but for now we just have to steer people away from the things that come naturally.  Really, system pip mostly shouldn't be used at all in normal Linux setups (and Windows though it's not quite as problematic there).

README.rst
```diff
@@ -16,17 +16,24 @@ Installation
 redis-py requires a running Redis server. See `Redis's quickstart
 <https://redis.io/topics/quickstart>`_ for installation instructions.
 
+redis-py can be installed using `pip` similar to other Python packages. Do not use `sudo`
+with `pip`. It is usually good to work in a
+[virtualenv](https://virtualenv.pypa.io/en/latest/) or
+[venv](https://docs.python.org/3/library/venv.html) to avoid conflicts with other package
+managers and Python projects. For a quick introduction see
+[Python Virtual Environments in Five Minutes](bit.ly/py-env).
+
```
altendky: Sorry, I was lazy and wrote that in GitHub comment markdown, not RST...  just a minute and I'll translate what you've got here.

altendky:
```rst
redis-py can be installed using ``pip`` similar to other Python packages. Do not use ``sudo``
with ``pip``. It is usually good to work in a `virtualenv`_ or `venv`_ to avoid conflicts
with other package managers and Python projects. For a quick introduction see
`Python Virtual Environments in Five Minutes`_.

.. _`virtualenv`: https://virtualenv.pypa.io/en/latest/
.. _`venv`: https://docs.python.org/3/library/venv.html
.. _`Python Virtual Environments in Five Minutes`: https://bit.ly/py-env
```

itamarhaber: #facepalm :) Should have seen that coming myself

README.rst
altendky: Was this being left for fixing later?

PR #1077, opened 2018-11-18
Re-fix the recently broken INFO parsing, see #1018

oranagra:
The value part of the info line may contains : in many cases,
most importantly an IPv6 slave address, may cause the parser to crash.

see see #1018 for details


andymccurdy:
I don't like hardcoding key names, but I don't immediately see another way. Thanks.

PR #1083, opened 2018-11-21
Don't destroy data in GEORADIUS commands

bbayles:
Re: https://github.com/andymccurdy/redis-py/issues/1082, this PR changes how the responses for `georadius` and `georadiusbymember` are handled, such that they don't destroy information when the "member" contains non-unicode data.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -378,6 +378,17 @@ def parse_cluster_nodes(response, **options):
     return dict(_parse_node_line(line) for line in raw_lines)
 
 
+def text_or_binary(response):
+    """
+    Return the response as text data if possible (decoded as utf-8), otherwise
+    return it unmodified.
+    """
+    try:
+        return response.decode('utf-8')
+    except UnicodeDecodeError:
+        return response
+
+
 def parse_georadius_generic(response, **options):
     if options['store'] or options['store_dist']:
         # `store` and `store_diff` cant be combined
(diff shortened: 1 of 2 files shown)


redis/client.py
```diff
@@ -392,7 +403,7 @@ def parse_georadius_generic(response, **options):
     if not options['withdist'] and not options['withcoord']\
             and not options['withhash']:
         # just a bunch of places
-        return [nativestr(r) for r in response_list]
+        return [text_or_binary(r) for r in response_list]
```
andymccurdy: We can just return the `response_list` here since it should be decoded or not based on the `decode_responses` flag.

redis/client.py
```diff
@@ -402,7 +413,7 @@ def parse_georadius_generic(response, **options):
 
     # zip all output results with each casting functino to get
     # the properly native Python value.
-    f = [nativestr]
+    f = [text_or_binary]
```
andymccurdy: This could simply be a noop call. Something like: `f = [lambda x: x]`

redis/client.py
bbayles: Ah, magic! That does do the correct thing.

redis/client.py
bbayles: Quite right.

bbayles:
@andymccurdy - many thanks for the quick response.

I've zapped the decoding from `parse_georadius_generic` - you're right; this works well.

Looking forward to a release with this in it!

andymccurdy:
Great, thanks for this!

PR #1090, opened 2018-11-29
Call "eval" when script not loaded

limen:
"eval" would cache the script automatically.

build status  https://travis-ci.org/limen/redis-py/builds/461202367

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -3497,11 +3497,7 @@ class Script(object):
         try:
             return client.evalsha(self.sha, len(keys), *args)
         except NoScriptError:
-            # Maybe the client is pointed to a differnet server than the client
-            # that created this instance?
-            # Overwrite the sha just in case there was a discrepancy.
-            self.sha = client.script_load(self.script)
-            return client.evalsha(self.sha, len(keys), *args)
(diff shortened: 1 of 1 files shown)


limen:
The "SCRIPT LOAD" operation is of O(N) time complexity and would be cumbersome if the script was very long.
I see the Script.__init__ caculate the SHA of script, so no need to worry.

andymccurdy:
> The "SCRIPT LOAD" operation is of O(N) time complexity and would be cumbersome if the script was very long.

`EVAL` also needs to load the script. There's no performance savings here.

> The init method of the Script class caculate the SHA, so no need to worry.

Yes, however, if for some reason the client side method of computing the SHA is different than the server side then the `NoScriptError` would be encountered on every execution and the script object would never optimize itself. Loading the script separately and retrieving the SHA from the server ensures that subsequent calls to `EVALSHA` succeed as intended.

Again, I don't know of any case where this has happened, but there's no downside to doing it this way and it prevents errors in case there is such an issue.

andymccurdy:
I see no upside to only trusting the client here. No code simplification and no performance improvement.

PR #1095, opened 2018-12-03
Extend lock error for not owned special case

harlowja:
Using the locking routines, it is useful to be able to
distingush a generic lock error from a one that is related
to the lock not being owned anymore (without doing string
checks); this allows say a lock extension thread to attempt
to re-acquire the lock in this case (vs just dying).

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -1,7 +1,7 @@
 import threading
 import time as mod_time
 import uuid
-from redis.exceptions import LockError
+from redis.exceptions import LockError, LockErrorNotOwned
 from redis.utils import dummy
 
 
@@ -190,7 +190,8 @@ class Lock(object):
         if not bool(self.lua_release(keys=[self.name],
                                      args=[expected_token],
                                      client=self.redis)):
-            raise LockError("Cannot release a lock that's no longer owned")
+            raise LockErrorNotOwned("Cannot release a lock"
+                                    " that's no longer owned")
 
     def extend(self, additional_time):
         """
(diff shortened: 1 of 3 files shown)


redis/lock.py
```diff
@@ -2,6 +2,7 @@
 import time as mod_time
 import uuid
 from redis.exceptions import LockError
+from redis.exceptions import LockErrorNotOwned
```
andymccurdy: This can be imported on the line above along with LockError.

andymccurdy:
Looks good, thanks!

harlowja:
Thanks, this will allow me to get rid of:

```
                try:
                    self.conn_lock.extend(refresh_after)
                except redis_lock.LockError as e:
                    # This sucks, we need to make this better... by having
                    # a real exception type that denotes this case...
                    #
                    # When https://github.com/andymccurdy/redis-py/pull/1095
                    # merges and is released we should just use that instead.
                    g_m = e.message
                    if g_m == "Cannot extend a lock that's no longer owned":
                        LOG.info("Lock has been lost, attempting"
                                 " to re-acquire it")
                        if not self.conn_lock.acquire(blocking=False):
                            LOG.warning("Failed lock re-acquisition,"
                                        " shutting down", exc_info=True)
                            self.dead.set()
                    else:
                        LOG.warning("Failed lock extension, shutting down",
                                    exc_info=True)
                        self.dead.set()
```

in a local codebase.

andymccurdy:
FYI I renamed the new exception to `LockNotOwnedError`. It's more consistent with the other exception names in redis-py.

PR #1098, opened 2018-12-04
Add client kill with filter

theodesp:
Signed-off-by: Theo Despoudis

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -818,9 +818,11 @@ class Redis(object):
             if not isinstance(skipme, bool):
                 raise DataError("CLIENT KILL skipme must be a bool")
             if skipme:
-                args.extend((Token.get_token('SKIPME'), Token.get_token('YES')))
+                args.extend((Token.get_token('SKIPME'),
+                             Token.get_token('YES')))
             else:
-                args.extend((Token.get_token('SKIPME'), Token.get_token('NO')))
+                args.extend((Token.get_token('SKIPME'),
+                             Token.get_token('NO')))
         if _id is not None:
             args.extend((Token.get_token('ID'), _id))
         if addr is not None:
--- a/redis/client.py
+++ b/redis/client.py
@@ -808,13 +808,12 @@ class Redis(object):
         client_types = ('normal', 'master', 'slave', 'pubsub')
         yes_no = ('yes', 'no')
         for index in range(0, len(filter_options), 2):
-            option = filter_options[index]
-            value = filter_options[index + 1]
-            key = str(option).lower()
+            key = str(filter_options[index]).lower()
+            value = str(filter_options[index + 1]).lower()
             if key not in filter_types:
                 raise DataError("CLIENT KILL <filter> must be one of %r" % (
                     filter_types,))
-            if key == 'type' and option not in client_types:
+            if key == 'type' and value not in client_types:
                 raise DataError("CLIENT KILL TYPE <value> must be one of %r" % (
                     client_types,))
             if key == 'skipme' and value not in yes_no:
--- a/redis/client.py
+++ b/redis/client.py
@@ -798,39 +798,37 @@ class Redis(object):
 
     def client_kill_filter(self, _id=None, _type=None, addr=None, skipme=None):
         """
-        Disconnects the client using a variety of filter options
-        :param _id: Allows to kill a client by its unique ID field
-        :param _type: Kill by type. Where type is one of normal,
-        master, slave and pubsub
-        :param addr: This is exactly the same as the client_kill
-        :param skipme: If yes, then he client calling the command
-        will not get killed.
-        No will have the effect of also killing the client calling the command.
+        Disconnects client(s) using a variety of filter options
+        :param id: Kills a client by its unique ID field
+        :param type: Kills a client by type where type is one of 'normal',
+        'master', 'slave' or 'pubsub'
+        :param addr: Kills a client by its 'address:port'
+        :param skipme: If True, then the client calling the command
+        will not get killed even if it is identified by one of the filter
+        options. If skipme is not provided, the server defaults to skipme=True
         """
-        filter_options = ()
+        args = []
         if _type is not None:
             client_types = ('normal', 'master', 'slave', 'pubsub')
             if str(_type).lower() not in client_types:
                 raise DataError("CLIENT KILL type must be one of %r" % (
                                 client_types,))
-            filter_options = filter_options + (Token.get_token('TYPE'), _type)
+            args.extend((Token.get_token('TYPE'), _type))
         if skipme is not None:
-            yes_no = ('yes', 'no')
-            if str(skipme).lower() not in yes_no:
-                raise DataError("CLIENT KILL skipme yes/no must be one of %r"
-                                % (yes_no,))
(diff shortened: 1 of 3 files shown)


theodesp:
Code reviews welcome

tests/conftest.py
```diff
@@ -47,6 +47,15 @@ def r(request, **kwargs):
     return _get_client(redis.Redis, request, **kwargs)
 
 
+@pytest.fixture()
+def r3(request, **kwargs):
+    return [
```
theodesp: create a bunch of clients to kill

redis/client.py
```diff
@@ -790,6 +790,38 @@ def client_kill(self, address):
...
+        :param filter_options: a tuple or list of filter options with the following format:
+        (filter, value, filter, value,...) or
+        [filter, value, filter, value,...]
+        """
+        if not isinstance(filter_options, (list, tuple)) or not filter_options:
+            raise DataError("CLIENT KILL <filter> <value> ... ... <filter> <value>" +
+                            "must be a non empty list or "
+                            "tuple to execute")
+        if len(filter_options) % 2 != 0:
+            raise DataError("CLIENT KILL <filter> <value> requires a filter and a value pair. Got %r" % (
```
theodesp: DataError messages are open for review

redis/client.py
```diff
@@ -790,6 +790,39 @@ def client_kill(self, address):
...
+            if key not in filter_types:
+                raise DataError("CLIENT KILL <filter> must be one of %r" % (
+                    filter_types,))
+            if key == 'type' and value not in client_types:
+                raise DataError("CLIENT KILL TYPE <value> "
+                                "must be one of %r" % (client_types,))
+            if key == 'skipme' and value not in yes_no:
+                raise DataError("CLIENT KILL SKIPME <value> "
+                                "must be one of %r" % (yes_no,))
+        return self.execute_command('CLIENT KILL', *filter_options)
```
andymccurdy: The filter types (addr, id, type, skipme) need to be sent as Tokens or bytes so that they don't get encoded. e.g., `Token.get_value('addr')`

tests/test_commands.py
```diff
@@ -95,6 +95,43 @@ def test_client_setname(self, r):
         assert r.client_setname('redis_py_test')
         assert r.client_getname() == 'redis_py_test'
 
+    @skip_if_server_version_lt('2.4.0')
+    def test_client_kill(self, r, r3):
+        clients = r.client_list()
+        r.client_kill(clients[1].get('addr'))
```
andymccurdy: Let's check the return value and make sure it's a `bool` set to `True`

tests/test_commands.py
```diff
@@ -95,6 +95,43 @@ def test_client_setname(self, r):
         assert r.client_setname('redis_py_test')
         assert r.client_getname() == 'redis_py_test'
 
+    @skip_if_server_version_lt('2.4.0')
+    def test_client_kill(self, r, r3):
+        clients = r.client_list()
```
andymccurdy: An assert here to prove there are at least 4 connected test clients would help illustrate that the test is actually killing one of them.

Note that exact counts here will fail if additional non-test runner clients are connected to the Redis server. For example, if someone is connected to their local Redis server via redis-cli or a Python repl, that connection will be counted in the client list. We either need to filter those out so that we're only counting clients created by the test runner.

tests/test_commands.py
```diff
@@ -95,6 +95,43 @@ def test_client_setname(self, r):
...
+        r.client_kill(clients[1].get('addr'))
+        clients = r.client_list()
+        assert len(clients) == 3
```
andymccurdy: Same note here as above. This needs to count only connected test runner clients.

theodesp:
Cool I will look at the review tomorrow and update the PR

tests/test_commands.py
theodesp: Hey, this is quite tricky. This is what I've done:
```
      @skip_if_server_version_lt('2.4.0')
    def test_client_kill(self, r, r2):
        r.client_setname('c1')
        r2[0].client_setname('c2')
        r2[1].client_setname('c3')
        clients = r.client_list()
        own_clients = [client for client in clients 
                       if client.get('name') in ['c1', 'c2', 'c3']]
        assert len(own_clients) == 3
        resp = r.client_kill(clients[1].get('addr'))
        assert isinstance(resp, bool) and resp is True
        clients = r.client_list()
        own_clients = [client for client in clients 
                       if client.get('name') in ['c1', 'c2', 'c3']]
        assert len(own_clients) == 2
```

For some reason, I cannot use the r.client_id() as I get an error from redis and I have to somehow filter them out from the external ones.
What do you think?

itamarhaber: @theodesp what error? note that client_id is a v5 feature, so you need to have the proper version installed.

theodesp: It looks to me something is wrong in the server or something.
```
➜ uname -a
Darwin C02PV2X1FVH6 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64
➜ redis-server --version
Redis server v=5.0.2 sha=00000000:0 malloc=libc bits=64 build=25bb354a43384ae5
➜ ps aux | grep redis-server
itspare          70596   0.0  0.2  4335252  26340   ??  SN   Mon12pm   1:26.63 redis-server *:6379
➜ redis-cli
127.0.0.1:6379> client ID
(error) ERR Syntax error, try CLIENT (LIST | KILL | GETNAME | SETNAME | PAUSE | REPLY)

Comment (itamarhaber) [reply tests/test_commands.py]: Nope - your running redis-server is not-obviously older (I can tell from the error message). Do `INFO SERVER` from the the cli to see the real version, then `SHUTDOWN` and restart.

Comment (theodesp) [reply tests/test_commands.py]: A yes I had to restart the server as I did recently upgrade it. Thank you!

However, I would like  to avoid using r.client_id() in this test case as I will have to pump the min server requirements.

Comment (itamarhaber) [reply tests/test_commands.py]: Sure thing - was just interested in clarifying the error thingie.

Comment (andymccurdy) [reply tests/test_commands.py]: @theodesp You can use `client_setname` and `client_getname` to do the filtering. That was added in 2.6.9. I'd just set each client to have a known prefix like `'redis-py-test-...'`.

It looks like you're kind of already doing that in the code you copied above. But you need to use `r.client_getname` to read back the result.

File: redis/client.py [hunk 7]
```diff
@@ -471,7 +471,7 @@ class Redis(object):
         {
             'CLIENT GETNAME': lambda r: r and nativestr(r),
             'CLIENT ID': int,
-            'CLIENT KILL': bool_ok,
+            'CLIENT KILL': lambda r: int or nativestr(r) == 'OK',
```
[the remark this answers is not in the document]
andymccurdy: The docs say that the filters are combined via logical AND. So currently it's not very useful to pass multiple filters.

theodesp:
@andymccurdy Pr updated

theodesp:
@andymccurdy Test added

andymccurdy:
Thanks @theodesp. I made a few small changes and am merging now.

PR #1101, opened 2018-12-07
Use tox "extras" option to install package extras

jdufresne:
For details, see:

https://tox.readthedocs.io/en/latest/config.html#conf-extras

Removes the need to duplicate the version string in tox.ini. Allow setup.py to be the single source of truth.


andymccurdy:
happy to commit this. to educate me, what's the difference?

Issue #1065, opened 2018-11-14
How can I preheat ConnectionPool?


ethe:
`ConnectionPool`'s initializer does not create connections immediately, in some cases this strategy  causes some problem, so is there anyway to preheat connection pool (create some IDLE connections in pool initializing)?

andymccurdy:
Short answer, there's no automatic way to do this. You could do this manually by acquiring N number of connections, running a common on each of them, then releasing them all back to the connection pool.

tzickel:
What problems does it cause ?

tzickel:
Not sure how it's related to your problem ? If you preheat tons of connections at start won't that just make it crash sooner ?

ethe:
> Not sure how it's related to your problem ? If you preheat tons of connections at start won't that just make it crash sooner ?

The problem is, if there is no connection at the beginning, server's performance decreased heavily (cause connection pool should make connection before every request, it is a heavy overhead).

ethe:
Lots of Redis driver accepts optional `max_idle_connections` param let user choose if keep connections in idle time (includes beginning). Yeah there is a manual way to do this in redis-py (call `make_connection` I think), but I think it is more convenience if this feature is integrated in driver.

theodesp:
Hey how can we implement this? Having a initial capacity property that will invoke the make_connection so many times and keep that number in the _available_connections ?

antsbean:
> Short answer, there's no automatic way to do this. You could do this manually by acquiring N number of connections, running a common on each of them, then releasing them all back to the connection pool.

why not do add max_idle_connections  feature

PR #1104, opened 2018-12-08
Add overflow-type connection pool

nerdvegas:
https://github.com/andymccurdy/redis-py/issues/1103

### Description of change

Added `OverflowConnectionPool` and corresponding tests.

### DIFF
--- a/redis/__init__.py
+++ b/redis/__init__.py
@@ -1,6 +1,7 @@
 from redis.client import Redis, StrictRedis
 from redis.connection import (
     BlockingConnectionPool,
+    OverflowConnectionPool,
     ConnectionPool,
     Connection,
     SSLConnection,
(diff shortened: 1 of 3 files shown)


andymccurdy:
I don't see why this needs to be within redis-py. You can subclass `ConnectionPool` in your own code base.

In general I think promoting an overflow strategy is a bad idea. I can see the reasoning for logging/warnings about the number of connections in use or being created. However, allowing the number of connections to grow indefinitely could have bad consequences.

theodesp:
I'm not sure why we should keep an overflow pool and not just increase the max_connections in the plain  ConnectionPool. It makes little sense and it's  a little bit confusing.

nerdvegas:
I'm not clear on why this is regarded as a bad idea when it's the default
pool behaviour for HTTP connections in urllib3, arguably a very wide used
lib. Is there some difference in the situation that makes this
inappropriate for redis connections?

My case is fairly typical I would think... 99% of the time I need a low
connection count (50 perhaps), but in theory there can be bursts of traffic
to my app that could bring that to 1000 or so. Without an overflowing conn
pool, I would need something like a 1000-max conn blocking pool, leaving
the majority of conns idle for the majority of the time.

Thx
A

andymccurdy:
The biggest reason is that Redis has a `maxclients` setting. If you exceed it you'll start getting errors. It's generally a good idea to configure your application servers to not exceed that (in aggregate) so that you can avoid this situation. An overflow strategy is directly counter to this.

I would rather see an effort to monitor and purge stale connections from a connection pool. In such a setup, you could set `max_connections` to something very high which would continue to be a hard cap. Connections that remain idle for some threshold of time would get closed and purged. And perhaps also a new setting, `min_connections`, that prevents the last used N connections from being purged.

nerdvegas:
I do agree that a pool that removes idle connections after a timeout would
be the ideal. I may have a stab at this soonish if I get the time.

Thx
A

nerdvegas:
Thinking further on this. Removing idle conns after a timeout is somewhat
arbitrary (what should that timeout be?) I think it'd be better to have an
overflow pool that has a min_connections and max_connections. Ie, anything
> min_connections is created on demand, and deleted when released, but the
pool still blocks on max_connections. This would be simpler also. Thoughts?

A

itamarhaber:
That's how https://github.com/gomodule/redigo/blob/master/redis/pool.go behaves - `MaxIdle` and `MaxActive`.

BTW, `MaxIdle` is similar to #1065

andymccurdy:
Closing this as I think there are better strategies to improve the connection pool, some of which were discussed in this thread.

PR #1105, opened 2018-12-17
Fix #764 - sub-unsub-resub caused PubSub() to forget the channel

### LINKED ISSUE #764: Sub-unsub-resub causes redispy to forget the channel
We recently came across some strange behaviour in our web servers, and tracked the problem down to a race condition in redispy, which is triggered when you unsubscribe and quickly resubscribe to a channel.

I can trivially reproduce it with this:
```
pubsub.subscribe("channel1")
# pubsub.channels is {"channel1": None}
time.sleep(1)
pubsub.unsubscribe("channel1")
pubsub.subscribe("channel1")
time.sleep(1)
# pubsub.channels is {} and pubsub.subscribed is False!!
```
pubsub.subscribe() synchronously adds the channel to pubsub.channels, however pubsub.unsubscribe() doesn't: it waits for the redis server to asynchronously send an unsubscribe message.

So, what happens is, the unsubscribe request is sent, subscribe() readds the channel to the dict, but then the unsubscribe message is returned from the server and the newly-added entry is removed from the dict. Both our application and the redis server end up thinking the channel is subscribed, however redispy has forgotten about it (meaning our callback function becomes unregistered and we stop receiving messages). In theory, the same can happen with the pattern calls.

A possible fix is to just not wait for the server to respond, e.g:
```
class RedisPyPubSubFix(PubSub):
	_sentinel = object()

	def __init__(self, redis_client, **kwargs):
		super(RedisPyPubSubFix, self).__init__(redis_client.connection_pool, **kwargs)

	def unsubscribe(self, *args):
		resp = super(RedisPyPubSubFix, self).unsubscribe(*args)
		if not args:
			# Unsub(empty) is unsub all
			self.channels.clear()
		else:
			for arg in args:
				self.channels.pop(self.encode(arg), None)
		return resp

	def punsubscribe(self, *args):
		resp = super(RedisPyPubSubFix, self).punsubscribe(*args)
		if not args:
			self.patterns.clear()
		else:
			for arg in args:
				self.patterns.pop(self.encode(arg), None)
		return resp

	def handle_message(self, response, ignore_subscribe_messages=False):
		"""
		Parses a pub/sub message. If the channel or pattern was subscribed to
		with a message handler, the handler is invoked instead of a parsed
		message being returned.
		"""
		message_type = nativestr(response[0])
		if message_type == 'pmessage':
			message = {
				'type': message_type,
				'pattern': response[1],
				'channel': response[2],
				'data': response[3]
			}
		else:
			message = {
				'type': message_type,
				'pattern': None,
				'channel': response[1],
				'data': response[2]
			}

		if message_type in self.PUBLISH_MESSAGE_TYPES:
			# if there's a message handler, invoke it
			if message_type == 'pmessage':
				handler = self.patterns.get(message['pattern'], self._sentinel)
			else:
				handler = self.channels.get(message['channel'], self._sentinel)
			if handler is self._sentinel:
				# We're no longer subbed to the channel, so we
				# don't care about the message.
				return None
			if handler:
				handler(message)
				return None
		else:
			# this is a subscribe/unsubscribe message. ignore if we don't
			# want them
			if ignore_subscribe_messages or self.ignore_subscribe_messages:
				return None

		return message
```

marc1n:
Fix #764

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2903,7 +2903,9 @@ class PubSub(object):
             self.connection_pool.release(self.connection)
             self.connection = None
         self.channels = {}
+        self.pending_unsubscribe_channels = set()
         self.patterns = {}
+        self.pending_unsubscribe_patterns = set()
 
     def close(self):
         self.reset()
@@ -2913,6 +2915,8 @@ class PubSub(object):
         # NOTE: for python3, we can't pass bytestrings as keyword arguments
         # so we need to decode channel/pattern names back to unicode strings
         # before passing them to [p]subscribe.
+        self.pending_unsubscribe_channels.clear()
+        self.pending_unsubscribe_patterns.clear()
         if self.channels:
             channels = {}
             for k, v in iteritems(self.channels):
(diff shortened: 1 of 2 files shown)


andymccurdy:
Looks good. I restarted the broken job which had nothing to do with your code. Assuming it passes I'll merge this when it's done. Thanks!

Issue #1085, opened 2018-11-21
PythonParser vs HiredisParser on_disconnect behavior


popravich:
PythonParser's `on_disconnect` implementation is inconsistent with HiredisParser implementation (or vice versa):
```python
class PythonParser(...):
    def on_disconnect(self):
        "Called when the socket disconnects"
        if self._sock is not None:
            self._sock.close()
            self._sock = None
        if self._buffer is not None:
            self._buffer.close()
            self._buffer = None
        self.encoder = None
```
and
```python
class HiredisParser(...):
    def on_disconnect(self):
        self._sock = None
        self._reader = None
        self._next_response = False
```
Why does the PythonParser closes the `_sock` object?
By doing this the subsequent `shutdown()` and `close()` in `Connection.disconnect` does not make any sense, in fact it shutdown on closed socket raises error which is ignored.

I can submit a PR but please tell me what place to fix? (HiredisParser/PythonParser/shutdown)

PS: this issue causes other issues in other repos (celery/kombu#954, celery/celery#3898)

theodesp:
I suspect it's because it's not covered with test cases. Feel free to provide a PR with coverage that does not break compatibility.

popravich:
I'm sorry, but I don't understand, what a PR should cover?
Just add a test to show that PythonParser closes the Connection's socket and HiredisParser does not?

andymccurdy:
You're right, the inconsistency is an error. From briefly scanning this, it seems wrong that `PythonParser` is closing the socket in `on_disconnect`.

Are you proposing that we simply remove the `self._sock.close()` from `PythonParser.on_disconnect`?

popravich:
> Are you proposing that we simply remove the `self._sock.close()` from `PythonParser.on_disconnect`?

Yes, this, at least, will make "on disconnect" behaviour similar regardless of parser used in connection.

andymccurdy:
@popravich I agree. If you want to create a PR that removes the `socket.close()` I'll gladly merge it.

I'm also open to addressing the fork issues you're seeing in kombu/celery. Just let me know.

popravich:
Hi, @andymccurdy,
I've just updated related PR with the fix, please take a look.

PR #1108, opened 2018-12-19
Fix PythonParser.on_disconnect not to close socket before Connection

popravich:
### Pull Request check-list

- [x] Does `$ python setup test` pass with this change (including linting)?

### Description of change

Related to #1085 

I have added failing tests to show inconsistencies between `HiredisParser` and `PythonParser`
`on_disconnect` implementations.
When `hiredis` is installed (and `HiredisParser` is used implicitly)
connection can not be securily shared between process forks because if it gets closed in
one of forks other processes can receive `BrokenPipe` errors because of operations on `shutdown` socket.

### DIFF
--- a/tests/test_multiprocessing.py
+++ b/tests/test_multiprocessing.py
@@ -1,0 +1,127 @@
+import pytest
+import multiprocessing
+import contextlib
+
+from redis.connection import Connection, ConnectionPool
+
+
+@contextlib.contextmanager
+def exit_callback(callback, *args):
+    try:
+        yield
+    finally:
+        callback(*args)
+
+
+class TestMultiprocessing(object):
+    # Test connection sharing between forks.
+    # See issue #1085 for details.
+
+    def test_connection(self):
+        conn = Connection()
+        assert conn.send_command('ping') is None
+        assert conn.read_response() == b'PONG'
+
+        def target(conn):
+            assert conn.send_command('ping') is None
+            assert conn.read_response() == b'PONG'
+            conn.disconnect()
+
+        proc = multiprocessing.Process(target=target, args=(conn,))
+        proc.start()
+        proc.join(3)
+        assert proc.exitcode is 0
+
+        # Check that connection is still alive after fork process has exited.
+        conn.send_command('ping')
+        assert conn.read_response() == b'PONG'
+
+    def test_close_connection_in_main(self):
+        conn = Connection()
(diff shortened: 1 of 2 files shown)


tests/test_multiprocessing.py
```diff
@@ -0,0 +1,130 @@
...
+
+        ev = multiprocessing.Event()
+        proc = multiprocessing.Process(target=target, args=(conn, ev))
+        proc.start()
+
+        conn.disconnect()
+        ev.set()
+
+        proc.join(3)
+        assert proc.exitcode is 1
```
andymccurdy: Why is the exitcode `1` here? In all the other tests exitcode is `0`. Am I missing something?

tests/test_multiprocessing.py
```diff
@@ -0,0 +1,130 @@
...
+            assert conn.read_response() == b'PONG'
+            conn.disconnect()
+
+        proc = multiprocessing.Process(target=target, args=(conn,))
+        proc.start()
+        proc.join(3)
+        assert proc.exitcode is 0
+
+        # Check that connection is still alive after fork process has exited.
+        with pytest.raises(ConnectionError):
```
andymccurdy: The comment on the previous line says the connection should still be alive after the fork has exited. Why are we expecting a `ConnectionError` here then?

If `ConnectionError` is in fact expected, let's change the comment to indicate such and remove the `assert` statements within the `pytest.raises` block. Just looking at this I don't know if we're actually testing that `conn.send_command('ping') is None` and `conn.read_response() == b'PONG'` or if those lines are just there to trigger the `ConnectionError`. If they're just triggering the `ConnectionError`, there's no need to assert on them.

tests/test_multiprocessing.py
```diff
@@ -0,0 +1,130 @@
...
+
+        proc = multiprocessing.Process(target=target, args=(pool,))
+        proc.start()
+        proc.join(3)
+        assert proc.exitcode is 0
+
+        # Check that connection is still alive after fork process has exited.
+        conn = pool.get_connection('ping')
+        with exit_callback(pool.release, conn):
+            with pytest.raises(ConnectionError):
```
andymccurdy: Same comment here from `test_connection` re: the comment and the asserts within the `pytest.raises` block.

andymccurdy:
I tweaked your patch a bit and merged it. Thanks!

Issue #772, opened 2016-08-17
PythonParser vs. HiredisParser inconsistent when decode_responses=True


ms7s:
Hi, I just noticed that `PythonParser` and `HiredisParser` implementations differ when you use `decode_responses=True` and the data returned from Redis is not decodable by the selected encoding.

`PythonParser` raises `ValueError` while `HiredisParser` (respectively `hiredis-py`) returns the original byte string instead in this case.

You can see the difference at [redis-py connection.py:296](https://github.com/andymccurdy/redis-py/blob/2ca3ffdc2102ecca1e7e419428706c29684ab38e/redis/connection.py#L296) versus [hiredis-py reader.c:78](https://github.com/redis/hiredis-py/blob/5cfdf41df274167cd399556821752459e43c316f/src/reader.c#L78)

amw:
I have just noticed the same thing.

andymccurdy:
This is the same problem as #1006. Closing this and will discuss over there.

PR #1109, opened 2018-12-19
when decoding a byte stream of pickled data, return the byte stream..…

### LINKED ISSUE #1006: Differences in parsing with decode_responses=True, with and without hiredis
I ran into a similar problem as #809: I have a redis connection using `decode_repsonses=True`, but I am also using a 3rd-party package (RQ) that pickles some data and deep in that library I get `UnicodeDecodeError`s. I'm kind of stuck because of that library's use of pickled data.

However, I thought to install [hiredis-py](https://github.com/redis/hiredis-py) to see if it handled parsing differently and it does.

So, with this:

```python
import redis
import pickle
rdb = redis.StrictRedis(decode_responses=True)
data = pickle.dumps(dict(hello='world'))
rdb.set('foo', data)
rdb.get('foo')
```

With hiredis installed I see:

```python
b'\x80\x03}q\x00X\x05\x00\x00\x00helloq\x01X\x05\x00\x00\x00worldq\x02s.'
```

Without hiredis installed, I get:

```python
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```

Looking at [hiredis-py code](https://github.com/redis/hiredis-py/blob/master/src/reader.c#L77-L84), I see the equivalent of this:

```python
if encoding:
   try:
      data = data.decode(encoding)
   except ValueError:
      pass
```

I'm wishing redis-py did the same. I'd love to see the following change to [connection.py](https://github.com/andymccurdy/redis-py/blob/master/redis/connection.py#L122-L126):

```python
def decode(self, value, force=False):
   "Return a unicode string from the byte representation"
   if (self.decode_responses or force) and isinstance(value, bytes):
      try:
         value = value.decode(self.encoding, self.encoding_errors)
      except ValueError:
         # Ignore encoding and return bytes
         pass
   return value
```

Considering using hiredis comes highly recommended, I'd love to see this base difference in parsing resolved.

Thanks!

LCHarold:
…. this makes the behavior consistent whether we're using PythonParser or HiredisParser

### Pull Request check-list

- [ ] Does `$ python setup test` pass with this change (including linting)?

### Description of change

when decoding a byte stream of pickled data, return the byte stream... this makes the behavior consistent whether we're using PythonParser or HiredisParser

this commit fixes #1006  #772

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -129,7 +129,11 @@ class Encoder(object):
     def decode(self, value, force=False):
         "Return a unicode string from the byte representation"
         if (self.decode_responses or force) and isinstance(value, bytes):
-            value = value.decode(self.encoding, self.encoding_errors)
+            try:
+                value = value.decode(self.encoding, self.encoding_errors)
+            except ValueError:
+                # Ignore encoding and return bytes
+                pass
         return value
 
 
--- a/tests/test_encoding.py
+++ b/tests/test_encoding.py
@@ -1,4 +1,6 @@
 from __future__ import unicode_literals
+
+import pickle
 import pytest
 import redis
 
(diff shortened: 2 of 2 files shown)


andymccurdy:
If you're dealing with pickled (or other binary) data you should be using a connection that is created with `decode_responses=False`.

PR #1110, opened 2018-12-21
Remove outdated distutils fallback

jdufresne:
Modern Python environments have setuptools or pip installed. The distutils fallback is unnecessary.


andymccurdy:
I like deleting code. Thanks!

PR #1111, opened 2018-12-28
Remove unnecessary compat shim for 'bytes'

jdufresne:
Both Python 2.7 & Python 3 have the types bytes. On Python 2.7, it is an alias for the type `str`, same as what was previously defined in `_compat.py`.


andymccurdy:
good catch, thanks!

PR #1112, opened 2019-01-02
Add support for specifying the token when acquiring a lock

dave-nm:
- [X] Does `$ python setup test` pass with this change (including linting)? (pytest passes but I'm not sure how to run the linting)
- [ ] Does travis tests pass with this change (enable it first in your forked repo and wait for the travis build to finish)? (I don't know how to do this)
- [X] Is the new or changed code fully tested? (Added test for the bug fix and the new feature)
- [X] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? (I updated the documentation of the function and didn't see anything else that I'm supposed to do)

### Description of change
*  Allowed `token` to be specified in `Lock.acquire()`
* Fixed `Lock.locked()` to verify the value of the token

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -193,17 +193,18 @@ class Lock(object):
             return True
         return False
 
-    def locked(self, current=False):
+    def locked(self):
         """
         Returns True if this key is locked by any process, otherwise False.
-
-        ``current`` specifies if the lock check needs to be this lock
         """
-        value = self.redis.get(self.name)
-        if current:
-            return self.local.token is not None and value == self.local.token
-        else:
-            return value is not None
+        return self.redis.get(self.name) is not None
+
+    def owned(self):
+        """
+        Returns True if this key is locked by this lock, otherwise False.
+        """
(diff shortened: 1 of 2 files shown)


daveisfera:
Are you ok if I add a parameter that defaults to `True` to maintain the current behavior and if set to `False` will do the new behavior?

daveisfera:
I added a commit with what that would look like.

andymccurdy:
I think it would be better to have a separate method to indicate whether the current lock instance owns the lock. Perhaps `owned()`?

daveisfera:
Oh sorry, did I set this up to merge into a branch instead of `master`? Can you merge it into `master`?

daveisfera:
oh ok. thanks for merging this. We're looking forward to the 3.1.0 release so we can dump our homegrown lock, now that this one has `reacquire()` and the ability to specify `token`.

andymccurdy:
ok, it's been merged to master. Thanks!

PR #1114, opened 2019-01-07
Add READONLY and READWRITE commands

theodesp:
### Pull Request check-list

- [x] Does `$ python setup test` pass with this change (including linting)?

### Description of change
Fixes https://github.com/andymccurdy/redis-py/issues/1050
Added 2 commands READONLY and READWRITE that are relevant to the cluster context

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -8,7 +8,7 @@ import time
 
 from redis._compat import (unichr, ascii_letters, iteritems, iterkeys,
                            itervalues, long)
-from redis.client import parse_info, bool_ok
+from redis.client import parse_info
 from redis import exceptions
 
 from .conftest import skip_if_server_version_lt, skip_if_server_version_gte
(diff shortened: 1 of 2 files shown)


tests/test_commands.py
```diff
@@ -1639,6 +1639,19 @@ def test_cluster_slaves(self, mock_cluster_resp_slaves):
         assert isinstance(mock_cluster_resp_slaves.cluster(
             'slaves', 'nodeid'), dict)
 
+    @skip_if_server_version_lt('3.0.0')
+    def test_readwrite(self, r):
+        assert bool_ok(r.readwrite())
```
andymccurdy: Why can we not just use `assert r.readwrite()`? Shouldn't the `bool_ok` response callback have already been called?

theodesp: @andymccurdy Sorry for the delay. PR updated

gyermolenko:
seems like a glitch from Travis. Maybe job should be restarted.

andymccurdy:
I took care of the conflicts and merged this. Thanks!

PR #1124, opened 2019-01-17
implement CLIENT REPLY ON|OFF|SKIP

### LINKED ISSUE #1059: CLIENT REPLY is not supported
OFF and SKIP may be tricky if we rely on replies

LCHarold:
### Pull Request check-list

- [ ] Does `$ python setup test` pass with this change (including linting)?

### Description of change

this commit resolves #1059
implement CLIENT REPLY ON|OFF|SKIP

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -947,7 +947,7 @@ class Redis(object):
         """
         Controls whether server will wait for a reply after sending command.
         ON means standard request-reply behavior, waiting for reply.
-        OFF means fire-and-forget behavior, request sent without waiting for reply.
+        OFF means fire-and-forget behavior, request sent not waiting for reply.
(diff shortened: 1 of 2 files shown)


andymccurdy:
Thanks for authoring this. Unfortunately due to the current design of redis-py, this is going to be a bit more complicated.

The state of the client reply status needs to be stored on individual connection objects, not on the client instance itself. A client instance may use many different connection instances via the connection pool through its life. If a client sends `CLIENT REPLY OFF` on connection-1, there's no guarantee that the subsequent command will be sent on the same connection.

Further, storing state on the client instance causes problems if the instance is shared between threads.

For this to be effective and useful, we need to create a way that a user can specify which connection to use for a set of commands.

PR #1139, opened 2019-02-25
Add a eventlet specific selector in case eventlet has monkey patched select

gogglesguy:
Please see https://github.com/andymccurdy/redis-py/issues/1136

### DIFF
--- a/redis/selector.py
+++ b/redis/selector.py
@@ -1,3 +1,4 @@
+import sys
 import errno
 import select
 from redis.exceptions import RedisError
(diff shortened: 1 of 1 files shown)


gogglesguy:
After testing our eventlet patch (see https://github.com/eventlet/eventlet/issues/55) the last few days, I don't think we should apply this to redis-py. The correct thing to do is to fix eventlet. (gevent btw works fine with redis-py and no patching is needed).

andymccurdy:
Interesting. Were you still seeing errors after applying your patch?

PR #1146, opened 2019-03-11
Make .lastsave() / timestamp_to_datetime() tzaware

bsolomon1124:
### Description of change

Adds UTC `.tzinfo` to the result of `timestamp_to_datetime()` and `.lastsave()` by extension.

Since Redis always returns a Unix ts, we have enough knowledge to be explicit and attach a timezone to the result rather than to keep it naive.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -45,15 +45,37 @@ def list_or_args(keys, args):
     return keys
 
 
+if sys.version_info[0] == 2:
+    _zero = datetime.timedelta(0)
+
+    class UTC(datetime.tzinfo):
+        """UTC (Python 2 compat)."""
+
+        def utcoffset(self, dt):
+            return _zero
+
+        def tzname(self, dt):
+            return "UTC"
+
+        def dst(self, dt):
+            return _zero
+
+    utc = UTC()
+    del UTC
+
+else:
+    utc = datetime.timezone.utc
+
+
 def timestamp_to_datetime(response):
-    "Converts a unix timestamp to a Python datetime object"
+    "Converts a Unix timestamp to a tz-aware Python datetime object"
     if not response:
         return None
     try:
         response = int(response)
     except ValueError:
         return None
-    return datetime.datetime.fromtimestamp(response)
+    return datetime.datetime.fromtimestamp(response, tz=utc)
 
 
 def string_keys_to_dict(key_string, callback):


zeebonk:
Hey @bsolomon1124, I noticed your PR because [I've been waiting on feedback on a similar PR myself](https://github.com/andymccurdy/redis-py/pull/943), which addresses a timezones related bug:

`fromtimestamp` doesn't work as you'd expect. Effectively it does two things:
1. Convert the unix timestamp to a UTC datetime without tzinfo
2. Convert the datetime from UTC to your local timezone, still without tzinfo.

This behaviour, combined with the `replacec(tz...)`, means that when your machine is running in Europe/Berlin time, the resulting datetime is actually Europe/Berlin time but tagged as being UTC!

What you want to use in this case is `utcfromtimestamp`.

bsolomon1124:
Ah thanks @zeebonk , you're totally right, oversight on my part - squashed bad commit.  I've actually stuck with just `.fromtimestamp()` but now included `tz`, which (admittedly this is a microptimization) looks to be about 3x faster than `.utcfromtimestamp()`.

```
In [3]: %timeit datetime.datetime.fromtimestamp(t, tz=datetime.timezone.utc)                                                                                                                                                                                 
1.04 µs ± 4.51 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit datetime.datetime.utcfromtimestamp(t).replace(tzinfo=datetime.timezone.utc)                                                                                                                                                                  
2.57 µs ± 23.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [5]: datetime.datetime.fromtimestamp(t, tz=datetime.timezone.utc)                                                                                                                                                                                         
Out[5]: datetime.datetime(2019, 4, 9, 15, 37, 52, 208586, tzinfo=datetime.timezone.utc)

In [6]: datetime.datetime.utcfromtimestamp(t).replace(tzinfo=datetime.timezone.utc)                                                                                                                                                                          
Out[6]: datetime.datetime(2019, 4, 9, 15, 37, 52, 208586, tzinfo=datetime.timezone.utc)
```

zeebonk:
@bsolomon1124 I wasn't aware of the `tz` argument on `fromtimestamp`, nice one!

PR #1153, opened 2019-03-18
Unregister socket from ready_poller

bmerry:
### Pull Request check-list

### Description of change

It looks like there was a typo in PollSelector.close that made it
unregister from the read_poller twice instead of once from each poller.

This code was never actually executed, because Connection.disconnect didn't close the poller. I added that so that the modified code gets tested.


andymccurdy:
Thanks for catching this!

PR #1154, opened 2019-03-18
Fix PollSelector.close to unregister socket from read_poller and ready_poller

gogglesguy:
This was probably never called anyway, but if it ever does.


gogglesguy:
Duplicate

andymccurdy:
@gogglesguy Thanks. I've merged the other PR.

Issue #503, opened 2014-06-24
Pass connection info through environment to test suite


jbergstroem:
The test suite assumes that a redis is running on localhost:6379. This is not optimal for a few reasons, the most important one being that it might actually destroy/mess with vital data.

One way to avoid this would be to pass connection information through the environment, or use a config.py within the tests folder for passing the info to the test cases.

gyermolenko:
It would be convenient (at least for local development) to be able to run tests against particular Redis server, to compare test runs from two servers.
I.e. start server from docker as `docker run --rm -p 6301:6379 redis:5.0.0` and run  
`pytest --redis-url=redis://localhost:6301 tests`

I can create a PR to discuss this idea further.

andymccurdy:
We now have a way to specify a connection string with the `--redis-url` option. That seems to satisfy the needs of this issue.

PR #1170, opened 2019-06-03
WIP: Add redis_url command line argument for tests

gyermolenko:
It is convenient to be able to run tests against particular redis-server, like
`$ pytest tests/test_commands.py --redis-url="redis://localhost:6371/9"`

It should work if you have several running redises, e.g.
`$ docker run --rm -p 6371:6379 redis:5.0.1`
`$ docker run --rm -p 6375:6379 redis:5.0.5`

Unfortunately these changes do not cover situation when redis version is  < 5, because of 
`get_info()` and `skip_if_` helpers.

Please let me know what you think.

Also: ref #503

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,25 +5,57 @@ from mock import Mock
 from distutils.version import StrictVersion
 
 
-_REDIS_INFO = {}
+REDIS_INFO = {}
+default_redis_url = "redis://localhost:6379/9"
 
 
-def get_info(**kwargs):
-    params = {'host': 'localhost', 'port': 6379, 'db': 9}
-    params.update(kwargs)
-    key = '%s:%s' % (params['host'], params['port'])
-    if key not in _REDIS_INFO:
-        client = redis.Redis(**params)
-        _REDIS_INFO[key] = client.info()
-        client.connection_pool.disconnect()
-    return _REDIS_INFO[key]
+def pytest_addoption(parser):
+    parser.addoption('--redis-url', default=default_redis_url,
+                     action="store",
+                     help="Redis connection string,"
+                          " defaults to `%(default)s`")
 
 
-def _get_client(cls, request=None, **kwargs):
-    params = {'host': 'localhost', 'port': 6379, 'db': 9}
-    params.update(kwargs)
-    client = cls(**params)
-    client.flushdb()
+def _get_info(redis_url):
+    client = redis.Redis.from_url(redis_url)
+    info = client.info()
+    client.connection_pool.disconnect()
+    return info
+
+
+def pytest_sessionstart(session):
+    redis_url = session.config.getoption("--redis-url")
+    info = _get_info(redis_url)
+    version = info["redis_version"]
+    arch_bits = info["arch_bits"]
+    REDIS_INFO["version"] = version
+    REDIS_INFO["arch_bits"] = arch_bits
+
+
+def skip_if_server_version_lt(min_version):
+    redis_version = REDIS_INFO["version"]
+    check = StrictVersion(redis_version) < StrictVersion(min_version)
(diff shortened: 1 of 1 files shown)


andymccurdy:
@gyermolenko Could we simply make `get_info()` accept the `r` fixture, which would then use the connection provided by the url? That would then allow all the `skip_*` decorators to work.

gyermolenko:
At the moment `get_info()` has a couple of weird traits: 
- uses hardcoded connection parameters
- relies on `client.info()` which theoretically can break after some changes in the lib
- uses global variable `_REDIS_INFO` 

This is why I'm not sure about it. I'll try to come up with smth different here or in another PR

andymccurdy:
If `get_info()` used the `r` fixture, we could eliminate the hardcoded connection params. And it only uses the global `_REDIS_INFO` to cache the result after the first lookup. I think it's safe to assume we're only talking to a single Redis server in the existing test suite.

gyermolenko:
once `get_info` uses `r` it should become either test or fixture.
But it is used in skip-calculating functions directly.

For skips there are other approaches.
Separate module, markers, `pytest_collection_modifyitems`.

andymccurdy:
I'm fine merging a patch that allows the connection info to be specified on the command line. But the current patch fails when specifying a connection string that's different than localhost:6379 as get_info will still be looking at the localhost.

We need to fix that before this can be merged IMO.

gyermolenko:
Added example on how we can infer redis version on start and use it in tests.
If you are ok with that - I will update PR a little bit.

andymccurdy:
Sure, this looks good to me. I didn't know about `pytest_sessionstart()`. Seems like it's a great fit for what we need.

I assume we can continue to use the `skip_if_server_version_lte` decorator and just update its implementation. It's slightly more convenient than the more verbose `pytest.mark.skipif` + conditional check.

And we'll also need to make sure the server's `arch_bits` value is contained within `REDIS_INFO` and update the `skip_unless_arch_bits` decorator as well.

Thanks!

gyermolenko:
Ok, done.
Worth noting:
I've run test suite against 5.0.1 (passed) and 4.0.0 - `test_monitor` failed because `CLIENT ID` was introduced in 5.0.0 and there is no skip.

gyermolenko:
I would also like to propose the variant of `r` fixture with "setup / yield / teardown" instead of  `_get_client` version with finalizer.
```python3
@pytest.fixture
def r(pytestconfig):
   redis_url = pytestconfig.getoption("--redis-url")
   client = redis.Redis.from_url(redis_url)
   yield client
   client.flushdb()
   client.connection_pool.disconnect()
```

gyermolenko:
Testing with different versions of redis 
(`pytest tests/ --redis-url="redis://localhost:637x/9" -rs`)

5.0.0 - ok
5.0.1 - ok
5.0.2 - ok
5.0.3 - 1 fail at `tests/test_pubsub.py:315: TestPubSubMessages.test_unicode_channel_message_handler`
5.0.4 - ok
5.0.5 - ok

andymccurdy:
Looks good to me. I'm fine with the setup/yield/teardown generator style fixture if you want to change it.

re: the 5.0.3 pubsub test failure -- is this something that's consistently reproducible on that version? I haven't seen it before.

gyermolenko:
> I'm fine with the setup/yield/teardown generator style fixture if you want to change it.
will probably do that in separate PR

> the 5.0.3 pubsub test failure -- is this something that's consistently reproducible on that version?
hmmm , sorry. Just rerun it 5 times - no errors. A bit strange

andymccurdy:
Sounds good. I'll merge once Travis finishes.

andymccurdy:
Thanks for this!

Issue #1140, opened 2019-02-26
3.2.0 Error while reading from socket: ('Connection closed by server.',)


LucyWengCSS:
**Version:**
Python: 3.6.7
Redis: 3.2.7 (Azure Redis)
Redis-py: 3.2.0
Django: 2.1.1

**Description**: 
Hi Experts, 

Our service met a similar issue with the issue #1127 3.1.0 causing intermittent connection closed by server error. By reviewing the whole discussion of the issue #1127, we upgraded redis-py to the version 3.2.0 and the issue has been mitigated but still happening. Due to the Azure Redis server will close the connections which are idle more than 10 mins, and the default redis-py behavior is to not close connections, recycling them when possible, could you please suggest how to avoid the exception "Redis ConnectionError: Error while reading from socket: ('Connection closed by server.',)" on our product?

**Configuration settings:**
{'default': {'BACKEND': 'django_redis.cache.RedisCache',
  'LOCATION': 'redis://xx.x.x.xxx:6379/0',
  'TIMEOUT': 60,
  'OPTIONS': {'DB': 0,
   'SOCKET_TIMEOUT': 120,
   'SOCKET_CONNECT_TIMEOUT': 30,
   'COMPRESSOR': 'django_redis.compressors.zlib.ZlibCompressor',
   'IGNORE_EXCEPTIONS': True,
   'REDIS_CLIENT_KWARGS': {'socket_keepalive': True},
   'PASSWORD': 'xxxxxxxxxxxxxxx='}},
 'cachalot': {'BACKEND': 'django_redis.cache.RedisCache',
  'LOCATION': 'redis://xx.x.x.xxx:6379/1',
  'TIMEOUT': 60,
  'OPTIONS': {'DB': 1,
   'SOCKET_TIMEOUT': 120,
   'SOCKET_CONNECT_TIMEOUT': 30,
   'COMPRESSOR': 'django_redis.compressors.zlib.ZlibCompressor',
   'IGNORE_EXCEPTIONS': True,
   'REDIS_CLIENT_KWARGS': {'socket_keepalive': True},
   'PASSWORD': 'xxxxxxxxxxxxxxxxxxxxxx='}},
 }

**Error Message:**
Redis ConnectionError: Error while reading from socket: ('Connection
closed by server.',)

Thanks a lot.

andymccurdy:
[...]

PR #1175, opened 2019-06-19
Fix issue #1140 - reconnect on ConnectionError while executing command

ziollek:
### Description of change

Fix issue #1140 - which was caused by mistake made in commit: https://github.com/andymccurdy/redis-py/commit/7860995ce4226ee0ca82d2eefa4b8bd5c3943a75
where logic was totally changed by moving parenthesis. After mentioned change client **dont** reconnect to redis server if any ConnectionError was raised while command was executing - this PR reverts this bug.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -809,8 +809,7 @@ class Redis(object):
             return self.parse_response(connection, command_name, **options)
         except (ConnectionError, TimeoutError) as e:
             connection.disconnect()
-            if not (connection.retry_on_timeout and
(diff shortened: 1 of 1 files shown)


andymccurdy:
We don't want to blindly reconnect in the event of a `ConnectionError`. Doing so can lead to data duplication, data loss or data corruption.

We never wanted to automatically reconnect on `ConnectionError`. Prior to version 3.1.0, there was a bug that did so however. 3.1.0+ fixed that issue in the commit you cited.

Issue #1191, opened 2019-07-15
xclaim raises an Exception if the item is not there


thomdask:
**Version**: 3.2.1

**Platform**: Python 3.7.3 on OS X 10.14.5

**Description**: Traceback is 
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ttt/Desktop/axs/lib/python3.7/site-packages/redis/client.py", line 2048, in xclaim
    return self.execute_command('XCLAIM', *pieces, **kwargs)
  File "/Users/ttt/Desktop/axs/lib/python3.7/site-packages/redis/client.py", line 775, in execute_command
    return self.parse_response(connection, command_name, **options)
  File "/Users/ttt/Desktop/axs/lib/python3.7/site-packages/redis/client.py", line 795, in parse_response
    return self.response_callbacks[command_name](response, **options)
  File "/Users/ttt/Desktop/axs/lib/python3.7/site-packages/redis/client.py", line 272, in parse_xclaim
    return parse_stream_list(response)
  File "/Users/ttt/Desktop/axs/lib/python3.7/site-packages/redis/client.py", line 258, in parse_stream_list
    return [(r[0], pairs_to_dict(r[1])) for r in response]
  File "/Users/ttt/Desktop/axs/lib/python3.7/site-packages/redis/client.py", line 258, in <listcomp>
    return [(r[0], pairs_to_dict(r[1])) for r in response]
TypeError: 'NoneType' object is not subscriptable
```

The problem is that the response for XCLAIM can be a list of Nones if the items do not exist in the stream (because it has been trimmed) 

To replicate you can do the following:
```
import redis
r = redis.Redis(decode_responses=True)

r.xgroup_create("STREAM", "CGROUP", id="$", mkstream=True)
sid = r.xadd("STREAM", {"item": 0})
r.xreadgroup("CGROUP", "C1", {"STREAM": ">"})
r.xadd("STREAM", {"item": 3}, maxlen=1, approximate=False)

# this line raises the above exception
r.xclaim("STREAM", "CGROUP", "C2", 0, [sid])
```

PR #1192, opened 2019-07-15
Handle removed claimed messages without an exception

thomdask:
### Pull Request check-list

### Description of change
Modify the parse stream list to be able to handle the list of None values and return any items that are there (if any) while keeping the existing behavior

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -287,6 +287,8 @@ def parse_stream_list(response):
     for r in response:
         if r is not None:
             data.append((r[0], pairs_to_dict(r[1])))
+        else:
+            data.append(None)
     return data
 
 
--- a/redis/client.py
+++ b/redis/client.py
@@ -288,7 +288,7 @@ def parse_stream_list(response):
         if r is not None:
             data.append((r[0], pairs_to_dict(r[1])))
         else:
-            data.append(None)
+            data.append((None, None))
     return data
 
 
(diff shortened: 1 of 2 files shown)


RoeyPrat:
this is a fix for #1191

RoeyPrat:
@thomdask Thank you for looking into this and bringing up a solution. 
I'll let @andymccurdy review this, but please add a unit test.
I opened to your branch https://github.com/thomdask/redis-py/pull/1 with a unit test based on your description in the issue.

redis/client.py
```diff
@@ -283,7 +283,13 @@ def nativestr_or_none(response):
 def parse_stream_list(response):
     if response is None:
         return None
-    return [(r[0], pairs_to_dict(r[1])) for r in response]
+    data = []
+    for r in response:
+        if r is not None:
+            data.append((r[0], pairs_to_dict(r[1])))
+        else:
+            data.append(None)
```
andymccurdy: I think this should be: `data.append((None, None))`

This would allow users to iterate over XCLAIM results like:

```python
for message_id, message in r.xclaim(...):
    ...
```

thomdask: Alright. Makes sense to me too. Thank you

andymccurdy:
Thanks! This has been merged in 5faed95b3b248d7d076298afd357a042cf916757

PR #1201, opened 2019-08-09
fix monitor() return format [0 lua] parse error

kukey:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -3034,7 +3034,7 @@ class Monitor(object):
         c_info = client_info.split(':', 1)
         if(len(c_info) == 2):
             client_address = c_info[0]
-            client_port = int(c_info[1])
+            client_port = c_info[1]
             return {
                 'time': float(command_time),
                 'db': int(db_id),
--- a/redis/client.py
+++ b/redis/client.py
@@ -3002,7 +3002,6 @@ class Monitor(object):
     next_command() method returns one command from monitor
     listen() method yields commands from monitor.
     """
-    #monitor_re = re.compile(r'\[(\d+) (.+):(\d+)\] (.*)')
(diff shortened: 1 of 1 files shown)


andymccurdy:
What command causes the error that you’re fixing? Can you add a test to prove it works?

kukey:
1.Redis monitor command.
redis monitor source code like this:
`void replicationFeedMonitors(client *c, list *monitors, int dictid, robj **argv, int argc) {
    listNode *ln;
    listIter li;
    int j;
    sds cmdrepr = sdsnew("+");
    robj *cmdobj;
    struct timeval tv;

    gettimeofday(&tv,NULL);
    cmdrepr = sdscatprintf(cmdrepr,"%ld.%06ld ",(long)tv.tv_sec,(long)tv.tv_usec);
    if (c->flags & CLIENT_LUA) {
        cmdrepr = sdscatprintf(cmdrepr,"[%d lua] ",dictid);
    } else if (c->flags & CLIENT_UNIX_SOCKET) {
        cmdrepr = sdscatprintf(cmdrepr,"[%d unix:%s] ",dictid,server.unixsocket);
    } else {
        cmdrepr = sdscatprintf(cmdrepr,"[%d %s] ",dictid,getClientPeerId(c));
    }`

redis-py pasre lua client format [%d lua] will cause regular expression error。

2.yes，I tested，and I already use it online.

andymccurdy:
Ah, I see now. The client info section varies based on the client type. I've altered your commit a bit so that we can keep the same dictionary structure for all types of clients. I introduced a new key `client_type` which will be one of `('tcp', 'unix', 'lua')`. `client_address` and `client_port` are always present and their values vary based on the `client_type`.

I released version 3.3.8 to PyPI with these changes. Thanks for reporting this and for the PR!

Issue #1187, opened 2019-07-09
support ACL


RoeyPrat:
ACL SETUSER
ACL LIST
ACL GETUSER
ACL CAT

IAmATeaPot418:
Hi Roey,

I think we also need to include ACL DELUSER here. I know reset could be used to basically make it useless but think this is a lot more intuitive and people would use it. What do you think?

Not sure if we want to include every command since you have what appears to be the bread and butter of ACLs but what about the other ACL subcommands. They would definitely be useful for development at least.

 * ACL HELP  -- provides help information (listed below) 
 * ACL LOAD  -- Reload users from the ACL file.
 * ACL LIST  -- Show user details in config file format.
 * ACL USERS  -- List all the registered usernames.
 * ACL CAT [<category>]   -- List available categories // -- List commands inside category.
 * ACL SETUSER <username> ... acl rules ... -- Create or modify a user
 * ACL DELUSER <username> [...] -- Delete a list of users."
 * ACL GETUSER <username> - Get the user details.
 * ACL GENPASS - Generate a secure password
 * ACL WHOAMI - Return the current connection username.

RoeyPrat:
Dear @IAmATeaPot418 
Yeah, we should probably do `ACL DELUSER` as well, and other ones you mentioned.

andymccurdy:
ACL support has been merged into master.

PR #1216, opened 2019-09-30
ACL Support

andymccurdy:
#1187 

* Redis and Connection classes now accept a username argument.
* URLs now look for and parse usernames.
* If a username is set, it is sent to the AUTH command when a connection
  is established.
* NoPermissionError is raised in the event of a NOPERM response


andymccurdy:
@itamarhaber @RoeyPrat @IAmATeaPot418 I've implemented all the ACL commands in this PR. If anyone has time and/or interest, I'd appreciate another set of eyes to make sure I didn't miss anything.

The only thing I'm questioning is whether `ACL LIST` should return the string representation of the ACL (as it is currently implemented) or if I should attempt to parse it out into a dictionary similar to what `ACL GETUSER` returns.

I chose to keep it as the string representation because someone might want to use this to build an ACL file and currently `ACL LIST` is the only way to get that representation. But I'm open to changing it if you guys think it should be represented differently.

itamarhaber:
I'd call the user parser because a) you already implemented `parse_acl_user` and b)

> someone might want to use this to build an ACL file

This "someone" sounds like an advanced user who would probably be ok using `execute_command('ACL', 'LIST')` for that purpose.

andymccurdy:
ACL Support has been merged into master.

PR #1220, opened 2019-10-08
Add type filter to scan functions

netocp:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/__init__.py
+++ b/redis/__init__.py
@@ -18,7 +18,16 @@ from redis.exceptions import (
     RedisError,
     ResponseError,
     TimeoutError,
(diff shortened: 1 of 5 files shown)


andymccurdy:
Thanks, this looks really good. My only question is whether the client should check the `_type` against the 6 you have hardcoded. Redis modules provides the ability for module authors to create their own custom data types which wouldn't be scan-able with the current check.

I think we should remove the check and then I'll merge this.

andymccurdy:
I went ahead removed the type checking and merged to master. This will go out in the next release. Thanks!

PR #1222, opened 2019-10-09
Fix SSL read timeouts in Python 2.7

zbristow:
### Pull Request check-list

### Description of change

The ssl module in Python 2.7 raises timeouts as ssl.SSLError instead of socket.timeout. When these timeouts are encountered, the error will be re-raised as socket.timeout so it is handled appropriately by the connection.

This issue was discovered when using Redis as a result backend for [Celery](http://www.celeryproject.org/). When celery was waiting for results in Redis, its connection would time out causing the operation to fail. This only occurred when using secure connections; it works fine with non-secure connections.

### DIFF
--- a/redis/_compat.py
+++ b/redis/_compat.py
@@ -61,6 +61,34 @@ else:  # Python 3.5 and above automatically retry EINTR
         return sock.recv_into(*args, **kwargs)
 
 if sys.version_info[0] < 3:
+    # In Python 3, the ssl module raises socket.timeout whereas it raises
+    # SSLError in Python 2. For compatibility between versions, ensure
+    # socket.timeout is raised for both.
+    import functools
+
+    try:
+        from ssl import SSLError as _SSLError
+    except ImportError:
+        class _SSLError(Exception):
+            """A replacement in case ssl.SSLError is not available."""
+            pass
+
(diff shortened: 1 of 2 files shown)


andymccurdy:
Do we also need to wrap `socket.connect`? Establishing a connection can raise a `socket.timeout` error before `recv` is ever called. What exception is raised when establishing a secure connection trips the `socket_timeout` timeout?

zbristow:
> Do we also need to wrap `socket.connect`? Establishing a connection can raise a `socket.timeout` error before `recv` is ever called. What exception is raised when establishing a secure connection trips the `socket_timeout` timeout?

I tested this out using netcat to force a timeout and sure enough, an `ssl.SSLError` timeout was raised during the handshake.

I decided to find all the possible "timed out" error messages in the [ssl module](https://github.com/python/cpython/blob/2.7/Modules/_ssl.c) and updated the wrapper to raise `socket.timeout` when it encountered one of the messages. I then added some additional compatibility wrappers to catch all the other paths.

I didn't want to amend the existing commit with these changes without checking if that was acceptable; let me know if you'd want me to do that.

andymccurdy:
Awesome, thanks. This looks great. I've merged and released 3.3.9 to PyPI.

PR #1224, opened 2019-10-10
Fix SSL regression introduced in 3.3.9

zbristow:
### Pull Request check-list

### Description of change

The wrapper introduced to handle SSL timeout errors in Python 2.7 incorrectly assumed that instances of SSLError would always have a string as the first element in their args. The safer approach is to check the message attribute on the error.

This was caught when the `ssl` module raised an error that contained an errno as it's first element, which was not seen in my original testing. I've confirmed that this patch preserves the timeout handling and fixes the regression.

### DIFF
--- a/redis/_compat.py
+++ b/redis/_compat.py
@@ -98,7 +98,7 @@ if sys.version_info[0] < 3:
             try:
                 return func(*args, **kwargs)
             except _SSLError as e:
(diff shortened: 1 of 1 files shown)


andymccurdy:
Merged and pushed 3.3.10 to PyPI. Thanks!

PR #1225, opened 2019-10-11
check exception.args rather than exception.message

andymccurdy:
@harpener @zbristow Let me know if this makes sense to you guys. This is the same logic that the deprecated `message` property uses.


andymccurdy:
Great. Released this as 3.3.11

PR #1230, opened 2019-10-21
Adding hashed passwords to ACL implementation

IAmATeaPot418:
Adding ACLs with hashes and tests for ACLs with hashes. 

Submitted a couple of PRs but messed them both up. This one should be good. Let me know what you think.


IAmATeaPot418:
@andymccurdy - This is to add hash support for #1216

Ignore the shameful retesting. I gave up and ran the tests locally to get it to pass.

This one is finally pythonic. Let me know if you have any issues.

andymccurdy:
You can split a string across multiple lines like this:

```python
foo = ('part 1 of the string '
           'part 2 of the string '
           'notice no comma between these strings '
           'otherwise this would be a tuple')
```

tests/test_commands.py
```diff
@@ -166,10 +166,21 @@ def teardown():
...
         assert len(r.acl_getuser(username)['passwords']) == 1
-
-    @skip_if_server_version_lt('6.0.0')
-    def test_acl_list(self, r, request):
-        username = 'redis-py-user'
+        # Resets and tests hash value "variable HP" set properly.
+        HP = '5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8'
+        assert r.acl_setuser(username, enabled=True, reset=True,
+                             add_passwords=[HP])
+        assert acl['passwords'] == [HP]
```
andymccurdy: I don't understand how the `acl` dict is updated here to include the `HP` hash. Did you forget to call `acl_getuser(...)` after the most recent `acl_setuser()`?

Or is `HP` the hash of "pass1" and it's lingering around from the previous `acl_getuser()`?

tests/test_commands.py
```diff
@@ -166,10 +166,21 @@ def teardown():
...
+        # test remove_passwords for hash removal
+        assert r.acl_setuser(username, enabled=True, reset=True,
+                             add_passwords=[HP, 'pass1'])
+        assert len(r.acl_getuser(username)['passwords']) == 2
+        assert r.acl_setuser(username, enabled=True,
+                             remove_passwords=[HP])
+        assert len(r.acl_getuser(username)['passwords']) == 1
+        @skip_if_server_version_lt('6.0.0')
```
andymccurdy: You mis-indented the `test_acl_list` function and decorator. Unindent and add a space there please.

tests/test_commands.py
IAmATeaPot418: HP is the SHA-256 hash of the word 'password'. Since Redis will now respect hashes as entered passwords. What this is doing is defining the variable to use in order to pass in the password and retesting that the hash remains a hash and does not transform into another like it would if you used 'password' in the password instead.

tests/test_commands.py
IAmATeaPot418: @andymccurdy - should be all set with the latest commits. LMK what you think.

IAmATeaPot418:
@andymccurdy - when you get a chance I've refactored based on your feedback. LMK what you think.

tests/test_commands.py
```diff
@@ -166,6 +166,20 @@ def teardown():
         assert r.acl_setuser(username, enabled=True,
                              remove_passwords=['pass2'])
         assert len(r.acl_getuser(username)['passwords']) == 1
+        # Resets and tests that hashed values are set properly.
+        hashed_password = ('5e884898da28047151d0e56f8dc629'
+                           '2773603d0d6aabbdd62a11ef721d1542d8')
+        assert r.acl_setuser(username, enabled=True, reset=True,
+                             add_hashes=[hashed_password])
+        assert acl['passwords'] == [hashed_password]
```
andymccurdy: Did you run these tests locally and not skip them? I can't imagine this line passing as the `acl` dict hasn't been refreshed since the `acl_setuser()` call above it. I think you need an `acl_getuser()` here.

FWIW Travis isn't running these tests because it doesn't have Redis Unstable.

IAmATeaPot418: I did run these locally but not with a few minor modifications for the username.

acl = r.acl_getuser(username) so its calling it with the variable in this.

redis/client.py
```diff
@@ -1009,13 +1021,27 @@ def acl_setuser(self, username, enabled=False, nopass=False,
...
             # to be specified as a simple string or a list
             add_passwords = list_or_args(add_passwords, [])
             for password in add_passwords:
                 pieces.append('>%s' % password)
 
+        if add_hashes:
+            # as most users will have only one password, allow remove_passwords
+            # to be specified as a simple string or a list
+            add_hashes = list_or_args(add_hashes, [])
+            for password in remove_hashes:
```
andymccurdy: This should be `for password in add_hashes:`

andymccurdy:
I just ran the tests against Redis Unstable and am getting a failure on the getuser/setuser test. Are you sure you're not running against a 5.0 server where these tests will be skipped?

redis/client.py
IAmATeaPot418: Just pushed this one.

andymccurdy:
Merged this into the "acl" branch. Thanks!

PR #1234, opened 2019-10-24
Pipeline.execute() doesn't short-circuit on empty command stack if it is watching keys

### LINKED ISSUE #1233: execute() doesn't send EXEC if no commands are queued
**Version**: redis 5.0.4, redis-py 3.3.6
**Platform**: Python 3.6.8 on Ubuntu 18.04
**Description**: When calling `execute()` on a pipeline (created with `transaction=True`), the EXEC is never sent if no commands are queued. While at first glance this makes sense as an optimization (why send an empty pipeline to the server?), in fact the EXEC might still have a purpose: to check if any of the WATCHed keys have since been modified.

For example, under normal conditions:
```
pipe = r.pipeline()
pipe.watch('foo')
r.set('foo', 'bar')
pipe.multi()
pipe.get('bar')
pipe.execute()  # raises WatchError
```
However, if we don't run the `pipe.get('bar')`, the EXEC never gets sent:
```
pipe = r.pipeline()
pipe.watch('foo')
r.set('foo', 'bar')
pipe.multi()
pipe.execute()  # returns []
```

It's easy to see that this is caused by this short-circuit check on line 3673 of client.py:
https://github.com/andymccurdy/redis-py/blob/fa0b0392c86126cae1a264197dddab647fa37821/redis/client.py#L3670-L3674
Which, based on the git blame, looks like it was done based on the discussion here: https://github.com/andymccurdy/redis-py/issues/362

But it seems to me that if `pipe.watch()` has been called, we can't short circuit and must run the EXEC to see if the transaction succeeded or failed.

brianmaissy:
Pipeline.execute() no longer short-circuits on empty command stack (eturns an empty list without running EXEC) if it is watching keys.

Closes #1233

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -3670,7 +3670,7 @@ class Pipeline(Redis):
     def execute(self, raise_on_error=True):
         "Execute all the commands in the current pipeline"
         stack = self.command_stack
-        if not stack:
+        if not stack and not self.watching:
             return []
         if self.scripts:
             self.load_scripts()
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -191,6 +191,19 @@ class TestPipeline(object):
 
             assert not pipe.watching
 
+    def test_watch_failure_in_empty_transaction(self, r):
+        r['a'] = 1
+        r['b'] = 2
+
+        with r.pipeline() as pipe:
+            pipe.watch('a', 'b')
+            r['b'] = 3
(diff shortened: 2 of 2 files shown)


andymccurdy:
This looks great, thanks. Merged into master. It'll go out in the next release.

brianmaissy:
Awesome, thanks! Very impressed by your response time, thanks for being a great maintainer!

PR #1240, opened 2019-11-09
Add equality test on Redis client and conn pool

brainix:
Currently, there's no `__eq__()` method defined on the `Redis` or
`ConnectionPool` classes.  Therefore, no two instances of either of
these classes will ever be equal.

I have a use case where it would be nice for one Redis client instance
to be equal to another if they have the same connection kwargs; that is,
if they're connected to the same Redis database.

### Pull Request check-list

### Description of change

Add an `__eq__()` method on the `Redis` and `ConnectionPool` classes that return true if they have the same connection kwargs; that is, if they're connected to the same Redis database.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -707,6 +707,9 @@ class Redis(object):
     def __repr__(self):
         return "%s<%s>" % (type(self).__name__, repr(self.connection_pool))
 
+    def __eq__(self, other):
+        return self.connection_pool == other.connection_pool
+
     def set_response_callback(self, command, callback):
         "Set a custom Response Callback"
         self.response_callbacks[command] = callback
--- a/redis/client.py
+++ b/redis/client.py
@@ -708,7 +708,10 @@ class Redis(object):
         return "%s<%s>" % (type(self).__name__, repr(self.connection_pool))
 
     def __eq__(self, other):
-        return self.connection_pool == other.connection_pool
+        return (
+            isinstance(other, self.__class__)
+            and self.connection_pool == other.connection_pool
+        )
 
     def set_response_callback(self, command, callback):
         "Set a custom Response Callback"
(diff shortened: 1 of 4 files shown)


redis/client.py
```diff
@@ -707,6 +707,12 @@ def __init__(self, host='localhost', port=6379,
     def __repr__(self):
         return "%s<%s>" % (type(self).__name__, repr(self.connection_pool))
 
+    def __eq__(self, other):
+        return (
```
brainix: Do you think it's better written as-is, or using duck typing as so?

    try:
        return self.connection_pool == other.connection_pool
    except AttributeError:
        return False

redis/connection.py
```diff
@@ -1044,6 +1044,12 @@ def __repr__(self):
             repr(self.connection_class(**self.connection_kwargs)),
         )
 
+    def __eq__(self, other):
+        return (
```
brainix: Do you think it's better written as-is, or using duck typing as so?

    try:
        return self.connection_kwargs == other.connection_kwargs
    except AttributeError:
        return False

Now that I think about it, I'm leaning towards duck typing...

andymccurdy:
Looks good to me, thanks!

andymccurdy:
@brainx After thinking about this a bit more, I now believe this isn't the right way to go. Just because two connection pools share the same configuration doesn't make them identical. Clearly they manage a separate set of connections (to the same server). It seems wrong to say `pool_a == pool_b` when `pool_a` will never give me one of `pool_b`'s connections or when I can call `pool_a.disconnect()` but `pool_b`'s connections are still active.

I'd tentatively be OK with saying `client_a == client_b` if they share the same connection pool instance. I'm not sure if that helps you at all.

I'm going to revert this change for now. If you really need this functionality it's trivial to compare `r1.connection_pool.connection_kwargs == r2.connection_pool.connection_kwargs`.

brainix:
@andymccurdy - Thanks for the heads up!  That works for me.

PR #1242, opened 2019-11-11
Drop support for end-of-life Python 3.4

jdufresne:
Python 3.4 has been EOL since 2019-03-18.

Removing from the test matrix helps reduce testing resources.

For additional details on unsupported Python versions, see:

https://devguide.python.org/devcycle/#end-of-life-branches


andymccurdy:
Thanks!

PR #1243, opened 2019-11-11
Unify tox and Travis test matrix

jdufresne:
tox can now always be used as the test entry point and allows for
contributors to easily test the full test matrix locally.

Starting with setuptools v41.5.0 (27 Oct 2019) the `setup.py test`
command is formally deprecated. It now emits a warning on use. Prior to
this version, it use was discouraged by the setuptools developers.

Now, always use tox as the one true testing entry point. It handles
multiple version of Python and Redis.

The Travis CI configuration was reworked to use tox to install
dependencies and run tests in the same manner that contributors would do
locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,31 +1,38 @@
-dist: xenial
-sudo: false
 language: python
 cache: pip
-python:
-  - 3.8
-  - 3.7
-  - 3.6
-  - 3.5
-  - 3.4
-  - 2.7
+matrix:
+  include:
+    - env: TOXENV=pycodestyle
+    - python: 2.7
+      env: TOXENV=py27-plain
+    - python: 2.7
+      env: TOXENV=py27-hiredis
+    - python: 3.4
+      env: TOXENV=py34-plain
+    - python: 3.4
+      env: TOXENV=py34-hiredis
+    - python: 3.5
+      env: TOXENV=py35-plain
+    - python: 3.5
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,30 +1,31 @@
...
-install:
-  - pip install -e .
-  - "if [[ $TEST_PYCODESTYLE == '1' ]]; then pip install pycodestyle; fi"
-  - "if [[ $TEST_HIREDIS == '1' ]]; then pip install hiredis; fi"
-  - "if [[ $TEST_PYCODESTYLE != '1' ]]; then pip install codecov coverage; fi"
-script: "if [[ $TEST_PYCODESTYLE == '1' ]]; then pycodestyle --repeat --show-source --exclude=.venv,.tox,dist,docs,build,*.egg,redis_install .; else coverage run setup.py test; fi"
-after_success:
-  - "if [[ $TEST_PYCODESTYLE != '1' ]]; then codecov; fi"
+services:
+  - redis-server
```
andymccurdy: While it would be great to rely on Travis's internal redis-server, we want to explicitly control the Redis server vesion. We often build redis-py features against beta versions of the Redis server prior to it being released.

jdufresne:
That makes sense. I've restored the custom Redis version in the latest revision.

andymccurdy:
This looks great, thanks!

PR #1244, opened 2019-11-12
Remove deprecated license_file from setup.cfg

jdufresne:
Starting with wheel 0.32.0 (2018-09-29), the "license_file" option is
deprecated.

The wheel will continue to include LICENSE, it is now included
automatically:

https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file
https://wheel.readthedocs.io/en/stable/news.html

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,6 +4,3 @@ exclude = .venv,.tox,dist,docs,build,*.egg
 
 [bdist_wheel]
 universal = 1
-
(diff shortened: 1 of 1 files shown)


andymccurdy:
You giveth (62fd0923f1a673ad72e10545c73bb734773b14e9) and you taketh away. Thanks for both :)

PR #1245, opened 2019-11-12
Compare numbers using '==' instead of 'is'

jdufresne:
While numbers are frequently singletons as an optimization, it is not
guaranteed by the language specification. Fixes flake8 error:

    F632 use ==/!= to compare str, bytes, and int literals


andymccurdy:
Yup, makes sense. Thanks!

PR #1249, opened 2019-11-18
Redis client get method EMTPY_RESPONSE in comment.

jeffgamedev:
Edit Redis client get method comment to accurately describe return value.

Multiple cases were encountered on production servers where Redis was returning 'EMPTY_RESPONSE' instead of None, which was not expected, and resulted in exceptions. Changes were made to our Redis wrapper to expect this, but the base client library comments still did not accurately describe the values that could be returned. This change amends the comments to describe that get method can either return you the value, None, _or_ the EMPTY_RESPONSE string.

This change should help future developers understand that the get method can return them a value that is neither their stored value or None.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1336,7 +1336,8 @@ class Redis(object):
 
     def get(self, name):
         """
-        Return the value at key ``name``, or None if the key doesn't exist
+        Return the value at key ``name``, or None if the key doesn't exist,
+        or "EMPTY_RESPONSE" string on timeout.
         """
         return self.execute_command('GET', name)
 


andymccurdy:
I don't see any possible way that `get()` could return `'EMPTY_RESPONSE'` (unless that's the value of the key you're trying to `GET` from the server). And this is most certainly not intentional.

Could you elaborate on what you're doing that's returning this value please?

jeffgamedev:
> I don't see any possible way that `get()` could return `'EMPTY_RESPONSE'` (unless that's the value of the key you're trying to `GET` from the server). And this is most certainly not intentional.
> 
> Could you elaborate on what you're doing that's returning this value please?

Maybe It wasn't returning `EMPTY_RESPONSE`, but it was returning the default option that is set to EMPTY_REPONSE? It was an edge case that occurred when redis would timeout, and a string was returned.

https://github.com/andymccurdy/redis-py/blob/656b551eb247c491976646092a71320407c0fa03/redis/client.py#L863 

The case is that a string was returned, even though we never set a string to this value. We were expecting a dictionary or None, yet a string was returned. It is true that I haven't been able to log what the exact value was yet. I can't find where options[EMPTY_RESPONSE] is set, as that appears to be what is returned in the case of a timeout.

I can understand your skepticism of this pull request.

jeffgamedev:
> > I don't see any possible way that `get()` could return `'EMPTY_RESPONSE'` (unless that's the value of the key you're trying to `GET` from the server). And this is most certainly not intentional.
> > Could you elaborate on what you're doing that's returning this value please?
> 
> Maybe It wasn't returning `EMPTY_RESPONSE`, but it was returning the default option that is set to EMPTY_REPONSE? It was an edge case that occurred when redis would timeout, and a string was returned.
> 
> https://github.com/andymccurdy/redis-py/blob/656b551eb247c491976646092a71320407c0fa03/redis/client.py#L863
> 
> The case is that a string was returned, even though we never set a string to this value. We were expecting a dictionary or None, yet a string was returned. It is true that I haven't been able to log what the exact value was yet. I can't find where options[EMPTY_RESPONSE] is set, as that appears to be what is returned in the case of a timeout.
> 
> I can understand your skepticism of this pull request.

It seems probably I misunderstood what I was reading in the code if in fact EMPTY_RESPONSE is never returned. Is there a default option set for the EMPTY_RESPONSE options? As I see that is returned in the case of timeout. I'm unable to find it.

jeffgamedev:
I will investigate further and find exactly what the production servers are returning. Thank you for your time!

andymccurdy:
The `EMPTY_RESPONSE` behavior allows a handful of commands to return sane, “empty’ values rather than raising an error due to inadequate user input.

For instance, `mget()` expects a list of keys. If the user passes an empty list of keys, the Redis server raises an error. While these commands can validate the input, they cannot simply short circuit the call to the server due to the way pipelines are implemented. So MGET (for instance) notices the user supplied zero keys and sets the EMPTY_RESPONSE option to an empty list. The request/response goes to the server, catches the error is raises and then returns the EMPTY_RESPONSE value.

Only a handful of commands implement this behavior and `get()` is not one of them. Which is why I’m confused about the return value you’re seeing.

You say you’re expecting a dictionary or a None value, so I suspect you’re not actually talking about `get()` as `get()` never returns a dictionary.

jeffgamedev:
> The `EMPTY_RESPONSE` behavior allows a handful of commands to return sane, “empty” values rather than raising an error due to inadequate user input.
> 
> For instance, `mget()` expects a list of keys. If the user passes an empty list of keys, the Redis server raises an error. While these commands can validate the input, they cannot simply short circuit the call to the server due to the way pipelines are implemented. So MGET (for instance) notices the user supplied zero keys and sets the EMPTY_RESPONSE option to an empty list. The request/response goes to the server, catches the error is raises and then returns the EMPTY_RESPONSE value.
> 
> Only a handful of commands implement this behavior and `get()` is not one of them. Which is why I’m confused about the return value you’re seeing.
> 
> You say you’re expecting a dictionary or a None value, so I suspect you’re not actually talking about `get()` as `get()` never returns a dictionary.

Thanks, I appreciate your time and help. I'm hunting down edge cases so that is why things are unclear. I'll establish logging to find out exactly what is happening and report back if it is relevant. Thank you!

PR #1255, opened 2019-12-20
reduce len call

Deneby67:
reduce len call and up perfomance


andymccurdy:
Thanks. We definitely don't want to cache `buffer_length` as that changes every iteration of the loop. But I've merged the `arg_length` argument optimization.

PR #1257, opened 2020-01-06
pubsub: handle disconnection while checking if we are connected

gjcarneiro:
### Pull Request check-list

### Description of change

After upgrade from redis-py 2.x to 3.3.11, I was getting these tracebacks in a loop:

```
2020-01-06 15:32:36,519 ERROR Problem reading redis queues
  File "/code/molly/pricefeed/pubsubutils.py", line 217, in run
    msg = self.pubsub.get_message()
  File "/env/lib/python3.6/site-packages/redis/client.py", line 3297, in get_message
    response = self.parse_response(block=False, timeout=timeout)
  File "/env/lib/python3.6/site-packages/redis/client.py", line 3183, in parse_response
    if not block and not conn.can_read(timeout=timeout):
  File "/env/lib/python3.6/site-packages/redis/connection.py", line 695, in can_read
    return self._parser.can_read(timeout)
  File "/env/lib/python3.6/site-packages/redis/connection.py", line 404, in can_read
    raise_on_timeout=False)
  File "/env/lib/python3.6/site-packages/redis/connection.py", line 416, in read_from_socket
```

I monkey patched it in my app, but this PR should do something equivalent upstream.

The problem is that I was getting ConnectionError in `conn.can_read(timeout=timeout)`, and there is no code to handle it. parse_response does handle this error, but later, when it calls `self._execute()`.

Because the exception is not handled, the system doesn't recover from the disconnection, on its own.  With the patch, it will recover on its own.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -3439,8 +3439,12 @@ class PubSub(object):
 
         self.check_health()
 
-        if not block and not conn.can_read(timeout=timeout):
-            return None
+        try:
+            if not block and not conn.can_read(timeout=timeout):
+                return None
+        except ConnectionError:
+            conn.disconnect()
+            conn.connect()
         response = self._execute(conn, conn.read_response)
 
         if conn.health_check_interval and \


andymccurdy:
We don't want redis-py to handle `ConnectionError`s like this. With this patch any message published between the time that the client was disconnected and the time it was reconnected is silently lost with no notification to the app that anything happened.

If your disconnects are related to connection timeouts (e.g., infrastructure between the client and the server timing out idle connections), use the `health_check_interval` option included in redis-py 3.3+.

For any other disconnect, I believe the client/app should be A) made aware that there is an issue and B) be able to respond to it in a way that's suitable for the specific app.

gjcarneiro:
I tried setting `health_check_interval=10`, but it doesn't help.  When I restart the redis server, the app gets into a reconnection loop and never recovers.

It's a fair comment, but I'm confused, the docs say:

> PubSub objects remember what channels and patterns they are subscribed to. In the event of a disconnection such as a network error or timeout, the PubSub object will re-subscribe to all prior channels and patterns when reconnecting.

So, I know how to catch the ConnectionError exception.  But how do I tell the PubSub object to "reconnect"?  The `reset()` method (or its alias `close()`) forgets the channels I had subscribed.   Should I peek inside the `pubsub.connection` attribute?  Is it part of public API?...

andymccurdy:
If you're using `health_check_interval` you shouldn't have to reconnect manually. I just tested locally using this:

```python
import redis
import time

r = redis.Redis(health_check_interval=10, socket_timeout=2)
p = r.pubsub()
p.subscribe('foo')
while True:
    try:
        print(p.get_message())
    except redis.ConnectionError:
        print('Error...')
    time.sleep(0.5)
```

While the client code is running, I killed my Redis server. The client starts printing "Error...". Obviously in a real app you'd probably want to do something more meaningful.

When I turn the Redis server back on, the client loop automatically reconnects, though it might take up to `check_health_interval` seconds to fully reconnect and resubscribe to the channels/patterns it was previously subscribed to.

gjcarneiro:
In my case it doesn't reconnect, even with health_check_interval=10, I'm not sure why.  But, OK, I am happy with the design, at least, and will have to further debug why health_check_interval is not having effect in my code.  Cheers.

PR #1262, opened 2020-01-23
Change order of tasks in connection reset

### LINKED ISSUE #1138: KeyError on self._in_use_connections.remove(connection)
**Version**: 2.10.5
**Platform**: Python 2.7, Ubuntu 
This is a very rare occurrence on a very high traffic server but I receive a key error occasionally from here:
`self._in_use_connections.remove(connection)`
https://github.com/andymccurdy/redis-py/blob/2.10.5/redis/connection.py#L913
I haven't been able to reproduce this in testing or root out the cause.

Theories: 
`self._available_connections` is a `list` so you could  `append` in the same connection more then once. Then when you try to `self._in_use_connections.add(connection)` 
https://github.com/andymccurdy/redis-py/blob/2.10.5/redis/connection.py#L898
You get no error for using the same connection. Until you try to release `_in_use_connections` is a `set` so were the `add()` multiple times desen't raise an error `remove()` will.

mdellweg:
This fixes a race condition, where a task can be returned an old
connection instead of waiting, while another one already started the reset.

Thanks for the analysis done by @yht804421715 and @gmbnomis .

fixes #1138

### Pull Request check-list

- [ ] <s>Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?</s>

### Description of change

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -1084,10 +1084,11 @@ class ConnectionPool(object):
         )
 
     def reset(self):
-        self.pid = os.getpid()
         self._created_connections = 0
         self._available_connections = []
         self._in_use_connections = set()
+        # Setting the pid signals, that we have done all the work here.
+        self.pid = os.getpid()
         self._check_lock = threading.Lock()
 
     def _checkpid(self):


andymccurdy:
This is replicated by #1270 which addresses several other thread-safety issues with the ConnectionPool.

PR #1263, opened 2020-01-23
lock.py: do not raise error on large sleep

clslgrnc:
### Pull Request check-list

### Description of change

I do not see why `sleep` should be smaller than `timeout`: I can be perfectly fine with waiting a long time (`sleep`) to acquire a lock for doing something that will be done in a short time (`timeout` only applies once the lock is acquired).
On the other hand I could see why `sleep` should be smaller than `blocking_timeout`. I suggest another solution: do not sleep if blocking_timeout will have expired when you wake up.

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -124,8 +124,6 @@ class Lock(object):
         self.thread_local = bool(thread_local)
         self.local = threading.local() if self.thread_local else dummy()
         self.local.token = None
-        if self.timeout and self.sleep > self.timeout:
-            raise LockError("'sleep' must be less than 'timeout'")
         self.register_scripts()
 
     def register_scripts(self):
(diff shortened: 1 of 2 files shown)


andymccurdy:
The change to calculate whether the next loop iteration exceeds `blocking_timeout` is great. Thanks!

In general, most users should use a small `sleep` value to get as much work done as possible. Large sleep times risk that a lock will stay unlocked for extended periods of time while all the processes are sleeping.

peterzdeb:
The new calculation of blocking timeout looks not correct to me. On large numbers of `sleep`  (>`blocking_timeout`) we will always have the wait time close to zero - it won't be `sleep + blocking_timeout`. Is it expected?

clslgrnc:
Why do you expect wait time to be `sleep + blocking_timeout` ?
`acquire` tries to acquire the lock every `sleep` up to `blocking_timeout`.
It stops as soon as the lock has been acquired or `blocking_timeout` is reached (or would be reached while sleeping).

If `sleep > blocking_timeout` we try to acquire the lock then return immediately since sleeping would lead to blocking more than `blocking_timeout`.

The current behavior is block at most `blocking_timeout`.
I do not think that changing to "try at least for `blocking_timeout`" would be good.

One possible improvement would be to compute the last sleep to reach `blocking_timeout`.

Issue #1265, opened 2020-01-23
Support memoryviews instead of explicitly requiring data to be bytes type


Cody-G:
**Version**: master
**Platform**: Python 3.6.1 Ubuntu 18.04
**Description**: memoryview objects are currently unsupported

Since Python's [memoryview](https://docs.python.org/3/library/stdtypes.html#memoryview) objects support Python's [buffer protocol](https://docs.python.org/3/c-api/buffer.html#bufferobjects) I think it makes sense for redis-py to accept memoryviews instead of explicitly requiring bytes types.  This can eliminate an unnecessary memory copy incurred by calling `tobytes()` on a memoryview.  In our case this is useful for interfacing with libraries like [pyarrow](https://arrow.apache.org/docs/python/).  This is just a one-line change since Python sockets already support memoryviews, I'll make a PR shortly.

dpipemazo:
PR in question for where this feature would be useful:

https://github.com/elementary-robotics/atom/pull/234

We send high-frequency, low-latency, sometimes-serialized data through redis streams. Thanks for considering!

PR #1266, opened 2020-01-23
Support memoryview encoding (as a no-op)

Cody-G:
### Pull Request check-list

- [(not sure) ] Does `$ tox` pass with this change (including linting)?

### Description of change

Supports passing memoryviews to redis-py as outlined in #1265

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -92,7 +92,7 @@ SENTINEL = object()
 
 
 class Encoder(object):
-    "Encode strings to bytes and decode bytes to strings"
+    "Encode strings to bytes-like and decode bytes-like to strings"
 
     def __init__(self, encoding, encoding_errors, decode_responses):
         self.encoding = encoding
@@ -100,8 +100,8 @@ class Encoder(object):
         self.decode_responses = decode_responses
 
     def encode(self, value):
-        "Return a bytestring representation of the value"
-        if isinstance(value, bytes):
+        "Return a bytestring or bytes-like representation of the value"
+        if isinstance(value, bytes) or isinstance(value, memoryview):
             return value
         elif isinstance(value, bool):
             # special case bool since it is a subclass of int
@@ -122,7 +122,7 @@ class Encoder(object):
         return value
 
     def decode(self, value, force=False):
-        "Return a unicode string from the byte representation"
+        "Return a unicode string from the bytes-like representation"
         if (self.decode_responses or force) and isinstance(value, bytes):
             value = value.decode(self.encoding, self.encoding_errors)
         return value
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -738,6 +738,12 @@ class Connection(object):
             raise response
         return response
 
+    def _maybe_to_bytes(self, arg):
+        if isinstance(arg, memoryview):
+            return arg.tobytes()
+        else:
+            return arg
+
     def pack_command(self, *args):
         "Pack a series of arguments into the Redis protocol"
         output = []
@@ -767,7 +773,7 @@ class Connection(object):
             else:
                 buff = SYM_EMPTY.join(
                     (buff, SYM_DOLLAR, str(arg_length).encode(),
-                     SYM_CRLF, arg, SYM_CRLF))
+                     SYM_CRLF, self._maybe_to_bytes(arg), SYM_CRLF))
         output.append(buff)
         return output
 
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -2399,6 +2399,19 @@ class TestRedisCommands(object):
         # xread starting at the last message returns an empty list
         assert r.xread(streams={stream: m2}) == []
 
+        # memoryviews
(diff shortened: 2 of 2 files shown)


andymccurdy:
Looks like Py27 is having an issue with this.

Also can we move the test to the test_encoding.py module? I'd prefer a simple test that persists the a `memoryview` value to a Redis string and verifies a subsequent `get()` returns the correct sequence of bytes. No need to complicate it with Redis streams.

Cody-G:
>Looks like Py27 is having an issue with this.

Oops!  I'm not well-practiced maintaining python version compatibility but I'll give it a shot.

>Also can we move the test to the test_encoding.py module? I'd prefer a simple test that persists the a memoryview value to a Redis string and verifies a subsequent get() returns the correct sequence of bytes. No need to complicate it with Redis streams.

Yes this makes sense, I had a feeling it wasn't the best place for the test code.

andymccurdy:
@Cody-G Does `memoryview.tobytes()` cause a memcpy of the underlying data? If so I thought that's what this was trying to avoid.

If `memoryview.tobytes()` does not copy the memory, then I'd argue that `Encoder.encode()` should special case `memoryview` objects and convert them to `bytes`.

If `memoryview.tobytes()` DOES copy the memory and `b''.join((my_memory_view, ...))` DOES NOT copy the memory, then it seems like we'd only need to call `.tobytes()` in Python 2.7. In which case we could have a simple function in _compat.py that's a noop in Python 3 but calls `.tobytes()` in Python 2.

Cody-G:
Tests are now passing on all Python versions.  I'm not sure why, but the python 2.7 tests seem to execute [this](https://github.com/andymccurdy/redis-py/blob/09bf90f961fce9322dc2aa54d6f68e8eecfe9c87/redis/connection.py#L773-L776) branch but I think the other python versions tests do not.  The problem occurs when we try to do a string join on a `memoryview`.  I see that happening in other places in the code, like [this line](https://github.com/andymccurdy/redis-py/blob/09bf90f961fce9322dc2aa54d6f68e8eecfe9c87/redis/connection.py#L791) and [this line](https://github.com/andymccurdy/redis-py/blob/09bf90f961fce9322dc2aa54d6f68e8eecfe9c87/redis/connection.py#L802), so my concern is that in many cases we'll end up incurring a memory copy anyway, and we won't get the benefit of using a `memoryview`.  My main goal is that, at least for large arguments, they can make it all the way to the [socket sendall](https://github.com/andymccurdy/redis-py/blob/09bf90f961fce9322dc2aa54d6f68e8eecfe9c87/redis/connection.py#L686) without being copied in any way.  Would it be acceptable to reduce the number of string joins further?  I realize that all of this chunking/joining was implemented to improve network performance, and I don't want to compromise that, but I wonder if we can compromise somehow.

Cody-G:
Oh I think we posted at the same time, let me read your post.

Cody-G:
>Does memoryview.tobytes() cause a memcpy of the underlying data?
Yes it does.  With my last commit I compromised on that for small objects (less than the hard-coded size limit)

>b''.join((my_memory_view, ...)) DOES NOT copy the memory
I was pretty sure this does copy the memory, but let me double check that

Cody-G:
`b''.join((my_memory_view, ...))` does in fact make a copy.  It seems if we want to [use memoryviews and send small chunks over the socket](https://julien.danjou.info/high-performance-in-python-with-zero-copy-and-the-buffer-protocol/) the recommendation is to use memoryview slices, which are zero-copy.  That makes me wonder if redis-py could just wrap all bytestrings in memoryviews (because it's free anyway) and then always handle the chunking with slices.  We could still join very small arguments at the cost of a copy, but those small arguments aren't the problem from a memory standpoint anyway.

andymccurdy:
It looks like this PR always calls `memoryview.tobytes()` when it encounters a `memoryview`. Isn't this what we're trying to avoid?

It seems like the way to do this is to make a special case for `memoryview` in the `pack_command` loop and append the `memoryview` to the `output` list without converting it. Something like:

```python
for arg in imap(self.encoder.encode, args):
    arg_length = len(arg)
    if len(buff) > buffer_cutoff or arg_length > buffer_cutoff or isinstance(arg, memoryview):
        # same code that exists currently
```

This will append the current buffer plus the length of the memoryview to the output list, and then append the memoryview directly without a copy.

Note that I haven't actually tested this. If you could that would be quite helpful. Thanks!

Cody-G:
> It looks like this PR always calls memoryview.tobytes() when it encounters a memoryview. Isn't this what we're trying to avoid?

Indeed it is, and this PR in its current form fails to avoid it.  I wasn't sure how much to respect the current optimization of joining small strings before sending over the socket.  In [our company fork](https://github.com/elementary-robotics/redis-py/pull/1/files) I took the more aggressive approach of never converting memoryviews to bytes, which is I think what you're suggesting.  If you're ok with that approach I can update this PR / make a new one next week.  Would definitely like to get this in somehow so we don't need to maintain our own fork.

Cody-G:
Superseded by #1285

PR #1270, opened 2020-01-27
better thread-safety for ConnectionPool

### LINKED ISSUE #906: connection pool in multi-thread is not thread safe
when multi thread share one instance of connectionpool, the self._in_use_connections and self._available_connections and the self._created_connections are not thread safe.

i do not see lock with these variables, so the connection pool may create more connections than the max_connections variable in the ConnectionPool class

the only using of lock is the reset() method, but this method is for checking the change of pid, but not the id of thread.

thanks  !!

### LINKED ISSUE #1138: KeyError on self._in_use_connections.remove(connection)

### LINKED ISSUE #1178: Connection is not thread-safe
**Version**: redis==2.10.6

**Platform**:  Python 3.7 on  Ubuntu 18.04

**Description**: 
I encountered the same question with the link: [https://github.com/celery/celery/issues/4363](https://github.com/celery/celery/issues/4363)

I reproduce the error by inserting time.sleep() in Connection.connect which is trying to initialize self._sock, then start two thread, make both pass self._sock existence check and run into Connection._on_connect (apprently Connection.connnect without any thread-lock).

So when one thread in on_connect() call  `read_response`, it maybe get another thread's response, then raise the error.
 

    def on_connect(self):
        "Initialize the connection, authenticate and select a database"
        self._parser.on_connect(self)

        # if a password is specified, authenticate
        if self.password:
            self.send_command('AUTH', self.password)
            if nativestr(self.read_response()) != 'OK':
                raise AuthenticationError('Invalid Password')

        # if a database is specified, switch to it
        if self.db:
            self.send_command('SELECT', self.db)
            res = self.read_response()
            if nativestr(res) != 'OK':
                raise ConnectionError('Invalid Database')

Is this a bug or something I don't get?

andymccurdy:
This PR implements better thread safety for `ConnectionPool`.

Fixes #1138 
Fixes #1178
Fixes #906 
Fixes #1262


redis/connection.py
```diff
@@ -1085,49 +1095,96 @@ def __eq__(self, other):
 
     def reset(self):
         self.pid = os.getpid()
```
gmbnomis: Apparently, this code still exhibits the race condition described in https://github.com/andymccurdy/redis-py/issues/1138#issuecomment-512813988 and addressed in PR https://github.com/andymccurdy/redis-py/pull/1262.  

The PR does not change the behavior (w.r.t. issue #1138) for me. Moving this line to the end of the method does help.

redis/__init__.py
```diff
@@ -33,9 +34,24 @@ def int_or_str(value):
 VERSION = tuple(map(int_or_str, __version__.split('.')))
 
 __all__ = [
-    'Redis', 'StrictRedis', 'ConnectionPool', 'BlockingConnectionPool',
-    'Connection', 'SSLConnection', 'UnixDomainSocketConnection', 'from_url',
-    'AuthenticationError', 'BusyLoadingError', 'ConnectionError', 'DataError',
-    'InvalidResponse', 'PubSubError', 'ReadOnlyError', 'RedisError',
-    'ResponseError', 'TimeoutError', 'WatchError'
+    'Redis',
```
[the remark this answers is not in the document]
andymccurdy: I agree.

redis/connection.py
andymccurdy: @gmbnomis Do you have a brief code sample that can repro what you're seeing?

andymccurdy: I see what's going on now. You're right, setting the PID should happen as the very last call as the PID check isn't protected via a lock for performance reasons.

andymccurdy: @gmbnomis I pushed the change to this PR. Would you mind testing this branch in your setup and see if it fixes your issue? I'd like to get this out in the next release.

gmbnomis: @andymccurdy I gave it a spin and I can't reproduce the problem anymore. Thank you for fixing the issue!

andymccurdy: Excellent! Thanks for helping to figure out what was happening and the testing effort! I'm going to merge this and release redis-py 3.4.0 now.

PR #1271, opened 2020-01-29
Make hset support multiple field/value pairs.

laixintao:
### Pull Request check-list

### Description of change

1. `hset` accept multiple field/value s now.
2. `hmset` marked as deprecated.

close https://github.com/andymccurdy/redis-py/issues/1269

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,8 @@
       in 3.4.0. This ended up being a bad idea as two separate connection
       pools be considered equal yet manage a completely separate set of
       connections.
+    * HSET command now can accept multiple pairs. HMSET has been marked as 
+      deprecated now. Thanks to @laixintao #1271
 * 3.4.0
     * Allow empty pipelines to be executed if there are WATCHed keys.
       This is a convenient way to test if any of the watched keys changed
--- a/redis/client.py
+++ b/redis/client.py
@@ -3024,6 +3024,7 @@ class Redis(object):
         Set key to value within hash ``name`` for each corresponding
         key and value from the ``mapping`` dict.
         """
+        warnings.warn(DeprecationWarning('Use hset'))
         if not mapping:
             raise DataError("'hmset' with 'mapping' of length 0")
         items = []
--- a/redis/client.py
+++ b/redis/client.py
@@ -2999,7 +2999,8 @@ class Redis(object):
     def hset(self, name, key=None, value=None, mapping=None):
         """
         Set ``key`` to ``value`` within hash ``name``,
-        Use kwargs to set multiple key/value pairs for a hash ``name``.
+        Use ``mappings`` keyword args to set multiple key/value pairs
+        for a hash ``name``.
         Returns the number of fields that were added.
         """
         if not key and not mapping:
--- a/redis/client.py
+++ b/redis/client.py
@@ -2996,12 +2996,21 @@ class Redis(object):
         "Return the number of elements in hash ``name``"
         return self.execute_command('HLEN', name)
 
-    def hset(self, name, key, value):
+    def hset(self, name, key=None, value=None, **pairs):
         """
         Set ``key`` to ``value`` within hash ``name``
         Returns 1 if HSET created a new field, otherwise 0
         """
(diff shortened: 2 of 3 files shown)


andymccurdy:
Thanks for making this PR. I'd prefer to not use keyword args here -- this was a big mistake with ZADD from the 2.x days and 3.x ended up breaking a bunch of user code. Instead, lets make it look like:

`def hset(self, name, key=None, value=None, mapping=None)`

Also we should adjust the comment so that it mentions that the return value is the number of fields added to the hash (as the redis docs suggest).

laixintao:
@andymccurdy Hi thanks for review, updated. PTAL again :)

laixintao:
ping @andymccurdy

redis/client.py
```diff
@@ -2996,12 +2996,22 @@ def hlen(self, name):
         "Return the number of elements in hash ``name``"
         return self.execute_command('HLEN', name)
 
-    def hset(self, name, key, value):
+    def hset(self, name, key=None, value=None, mapping=None):
         """
-        Set ``key`` to ``value`` within hash ``name``
-        Returns 1 if HSET created a new field, otherwise 0
+        Set ``key`` to ``value`` within hash ``name``,
+        Use kwargs to set multiple key/value paris for a hash ``name``.
```
Kawaamai:
```suggestion
        Use kwargs to set multiple key/value pairs for a hash ``name``.
```

You probably want to update this comment to reflect the use of `mapping` rather than kwargs.

laixintao: Thanks, I updated.

andymccurdy:
Thanks, looks great!

elcolumbio:
I like this commit and it is such an important interface.
I have a learning question.
When would we consider to break backwards compatibility in favor of this:
`hset(self, name, mapping)`

andymccurdy:
We could consider a backwards incompatible change when we increment a major version number. For example, 3.x -> 4.x.

redis/client.py
```diff
@@ -2996,12 +2996,23 @@ def hlen(self, name):
         "Return the number of elements in hash ``name``"
         return self.execute_command('HLEN', name)
 
-    def hset(self, name, key, value):
+    def hset(self, name, key=None, value=None, mapping=None):
         """
-        Set ``key`` to ``value`` within hash ``name``
-        Returns 1 if HSET created a new field, otherwise 0
+        Set ``key`` to ``value`` within hash ``name``,
+        Use ``mappings`` keyword args to set multiple key/value pairs
```
igorsobreira, line 3002: Small typo here: `s/mappings/mapping`

laixintao: Thanks, but it's already fixed in master: https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L3037

CHANGES
```diff
@@ -14,6 +14,8 @@
       in 3.4.0. This ended up being a bad idea as two separate connection
       pools be considered equal yet manage a completely separate set of
       connections.
+    * HSET command now can accept multiple pairs. HMSET has been marked as 
+      deprecated now. Thanks to @laixintao #1271
```
ala-ableton, line 18: @laixintao this unfortunately doesn't mention that it requires at least Redis 4.0.0. We got bitten by this when updating `redis-py` (from 3.4.1 to 3.5.2) in a system that runs Redis 3.2.4.

@andymccurdy I couldn't find any information about which version(s) of Redis is supported by `redis-py`. Did I miss something?

andymccurdy: @ala-abelton This is completely backwards compatible with older versions of Redis server. You can still use the `key` and `value` that were present in all previous versions of redis-py. We could add a comment to the function that `mapping` requires Redis server 4.0 or greater.

redis-py works with all known versions of Redis server. Obviously you need a version of Redis server that supports the commands and command options you're working with.

ala-ableton: @andymccurdy sorry, I misread the diff. It seems that our issue didn't come from `redis-py`, but from https://github.com/rq/rq. Thanks for the quick reply!

PR #1281, opened 2020-02-08
Stop hiding errors that occur inside __del__ methods

jdufresne:
If an exception occurs inside the `__del__` method, it should be reported to the developer. Not doing so could hide bugs.

Python automatically handles exceptions inside `__del__` methods, for example:

    class A:
        def __del__(self):
            1 / 0

    A()
    print("after del")

Results in the output:

    $ python3 ~/blah.py
    Exception ignored in: <function A.__del__ at 0x7fbbf2bbfc20>
      File "/home/jon/test.py", line 3, in __del__
        1 / 0
    ZeroDivisionError: division by zero
    after del

From this example, we can see the bug was not hidden and the code after `__del__` still executed.

### Pull Request check-list

- <s>[ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?</s> N/A

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -3370,13 +3370,10 @@ class PubSub(object):
         self.reset()
 
     def __del__(self):
-        try:
-            # if this object went out of scope prior to shutting down
-            # subscriptions, close the connection manually before
-            # returning it to the connection pool
-            self.reset()
-        except Exception:
-            pass
+        # if this object went out of scope prior to shutting down
+        # subscriptions, close the connection manually before
+        # returning it to the connection pool
+        self.reset()
 
     def reset(self):
         if self.connection:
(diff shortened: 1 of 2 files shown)


andymccurdy:
I too do not like hiding these errors. I'm going to merge this but I'm concerned about what corner cases might be exposed in real world code that the test suite doesn't adequately cover. If the next release causes significant issues for folks I may end up reversing this.

jdufresne:
I understand, thanks. The only side effect should be output on stderr, no breakage or change in functionality.  If you do decide to revert it, can you please CC me and I'll see if I can address it more directly?

andymccurdy:
Sure, sounds good. Thanks!

PR #1285, opened 2020-02-08
Support memoryview encoding/decoding as a no-op, calling tobytes() on…

Cody-G:
…ly when decode_responses=True

### Pull Request check-list

### Description of change

Supports passing memoryviews to redis-py as first outlined in #1265 and discussed further in #1266

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -775,9 +775,8 @@ class Connection(object):
             # to avoid large string mallocs, chunk the command into the
             # output list if we're sending large values or memoryviews
             arg_length = len(arg)
-            if len(buff) > buffer_cutoff \
-                or arg_length > buffer_cutoff \
-                or isinstance(arg, memoryview):
+            if (len(buff) > buffer_cutoff or arg_length > buffer_cutoff
+                    or isinstance(arg, memoryview)):
                 buff = SYM_EMPTY.join(
                     (buff, SYM_DOLLAR, str(arg_length).encode(), SYM_CRLF))
                 output.append(buff)
@@ -800,9 +799,8 @@ class Connection(object):
         for cmd in commands:
             for chunk in self.pack_command(*cmd):
                 chunklen = len(chunk)
-                if buffer_length > buffer_cutoff \
-                    or chunklen > buffer_cutoff \
-                    or isinstance(chunk, memoryview):
+                if (buffer_length > buffer_cutoff or chunklen > buffer_cutoff
+                        or isinstance(chunk, memoryview)):
                     output.append(SYM_EMPTY.join(pieces))
                     buffer_length = 0
                     pieces = []
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -805,7 +805,7 @@ class Connection(object):
                     buffer_length = 0
                     pieces = []
 
-                if chunklen > self._buffer_cutoff:
+                if chunklen > buffer_cutoff or isinstance(chunk, memoryview):
                     output.append(chunk)
                 else:
                     pieces.append(chunk)
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -94,7 +94,7 @@ SENTINEL = object()
 
 
 class Encoder(object):
-    "Encode strings to bytes and decode bytes to strings"
+    "Encode strings to bytes-like and decode bytes-like to strings"
 
     def __init__(self, encoding, encoding_errors, decode_responses):
         self.encoding = encoding
@@ -102,8 +102,8 @@ class Encoder(object):
         self.decode_responses = decode_responses
 
     def encode(self, value):
-        "Return a bytestring representation of the value"
-        if isinstance(value, bytes):
+        "Return a bytestring or bytes-like representation of the value"
+        if isinstance(value, bytes) or isinstance(value, memoryview):
             return value
         elif isinstance(value, bool):
             # special case bool since it is a subclass of int
@@ -124,9 +124,12 @@ class Encoder(object):
         return value
 
     def decode(self, value, force=False):
-        "Return a unicode string from the byte representation"
-        if (self.decode_responses or force) and isinstance(value, bytes):
-            value = value.decode(self.encoding, self.encoding_errors)
+        "Return a unicode string from the bytes-like representation"
+        if (self.decode_responses or force):
+            if isinstance(value, memoryview):
+                value = value.tobytes()
+            if isinstance(value, bytes):
+                value = value.decode(self.encoding, self.encoding_errors)
         return value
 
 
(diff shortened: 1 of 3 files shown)


redis/connection.py
```diff
@@ -94,16 +94,16 @@
...
     def __init__(self, encoding, encoding_errors, decode_responses):
         self.encoding = encoding
         self.encoding_errors = encoding_errors
         self.decode_responses = decode_responses
 
     def encode(self, value):
-        "Return a bytestring representation of the value"
-        if isinstance(value, bytes):
+        "Return a bytestring or bytes-like representation of the value"
+        if isinstance(value, bytes) or isinstance(value, memoryview):
```
jdufresne: You can pass multiple types to `isinstance`

```suggestion
        if isinstance(value, (bytes, memoryview)):
```

redis/connection.py
```diff
@@ -124,9 +124,12 @@ def encode(self, value):
         return value
 
     def decode(self, value, force=False):
-        "Return a unicode string from the byte representation"
-        if (self.decode_responses or force) and isinstance(value, bytes):
-            value = value.decode(self.encoding, self.encoding_errors)
+        "Return a unicode string from the bytes-like representation"
+        if (self.decode_responses or force):
```
jdufresne:
```suggestion
        if self.decode_responses or force:
```

redis/connection.py
andymccurdy: @jdufresne's suggestion makes sense here.

redis/connection.py
andymccurdy: @jdufresne's suggestion also makes sense here.

redis/connection.py
```diff
@@ -795,7 +799,8 @@ def pack_commands(self, commands):
         for cmd in commands:
             for chunk in self.pack_command(*cmd):
                 chunklen = len(chunk)
-                if buffer_length > buffer_cutoff or chunklen > buffer_cutoff:
+                if (buffer_length > buffer_cutoff or chunklen > buffer_cutoff
+                        or isinstance(chunk, memoryview)):
```
andymccurdy: I agree that this needs to happen here, but I believe we also need to special case `memoryview`'s below. Likely we'd want to change your line 808 to:

`if chunklen > self._buffer_cutoff or isinstance(chunk, memoryview):`

As it is now, `memoryview` instances will get appended to `pieces` which will eventually get copied via `SYM_EMPTY.join(pieces)`. I haven't actually tested this so please let me know if I'm overlooking something.

tests/test_encoding.py
```diff
@@ -11,12 +12,37 @@ class TestEncoding(object):
...
+            decode_responses=False,
+        )
+
+    def test_simple_encoding(self, r, r_no_decode):
         unicode_string = unichr(3456) + 'abcd' + unichr(3421)
         r['unicode-string'] = unicode_string
         cached_val = r['unicode-string']
         assert isinstance(cached_val, unicode)
         assert unicode_string == cached_val
+        r_no_decode['unicode-string'] = unicode_string.encode('utf-8')
```
andymccurdy: Can we make this added portion its own test for simplicity?

tests/test_encoding.py
```diff
@@ -11,12 +12,37 @@ class TestEncoding(object):
...
+        assert unicode_string == cached_val.decode('utf-8')
+
+    def test_memoryview_encoding(self, r, r_no_decode):
+        unicode_string = unichr(3456) + 'abcd' + unichr(3421)
+        unicode_string_view = memoryview(unicode_string.encode('utf-8'))
+        r['unicode-string-memoryview'] = unicode_string_view
+        cached_val = r['unicode-string-memoryview']
+        assert isinstance(cached_val, unicode)
+        assert unicode_string == cached_val
+        r_no_decode['unicode-string-memoryview'] = unicode_string_view
```
andymccurdy: Again, can we make the no-decode version its own test for simplicity?

redis/connection.py
```diff
@@ -124,9 +124,12 @@ def encode(self, value):
         return value
 
     def decode(self, value, force=False):
-        "Return a unicode string from the byte representation"
-        if (self.decode_responses or force) and isinstance(value, bytes):
-            value = value.decode(self.encoding, self.encoding_errors)
+        "Return a unicode string from the bytes-like representation"
+        if (self.decode_responses or force):
+            if isinstance(value, memoryview):
```
andymccurdy: Did you run into a case where you needed this? Are we ever actually decoding `memoryview` instances?

Cody-G: I don't think I ran into a case in the tests where a `memoryview` gets decoded.  I think I got a bit spooked by lines like [this](https://github.com/andymccurdy/redis-py/blob/4f0eed6c7b8cc2ad02eed71b3097b8f07a87ee84/redis/client.py#L1083) and [this](https://github.com/andymccurdy/redis-py/blob/4f0eed6c7b8cc2ad02eed71b3097b8f07a87ee84/redis/client.py#L1090), but on a closer look those lines shouldn't cause a problem unless the user passes a `memoryview` kwarg -- and the docs instruct them to pass strings.  Maybe it just felt right to make sure that `decode(encode(x))` always works, but it seems it's not essential, happy to remove this if you prefer.

redis/connection.py
Cody-G: Agreed, will change.  I'll also implement the subclass that emits a warning as you suggested, to make sure we're catching this.

Cody-G:
Thanks for the review!  I think most of the changes are straightforward.  I like the idea of subclassing `memoryview`, but I found out that's not possible.  Quoting from the [docs](https://docs.python.org/3/c-api/memoryview.html): "It is not currently allowed to create subclasses of memoryview."  So I think we can't use a custom class to test that we're never calling `tobytes()`, since the code relies on `isinstance(x, memoryview)`.  What we really care about is whether the argument is still a `memoryview` by the time it gets sent over the socket by `sendall` [here](https://github.com/andymccurdy/redis-py/blob/4f0eed6c7b8cc2ad02eed71b3097b8f07a87ee84/redis/connection.py#L703), so testing the return value of `pack_command()` should be sufficient (it gets called [here](https://github.com/andymccurdy/redis-py/blob/4f0eed6c7b8cc2ad02eed71b3097b8f07a87ee84/redis/connection.py#L723)).

Cody-G:
This seems ready for another review.  To summarize the changes, everything was straightforward except:

- I couldn't subclass `memoryview` to make a warning-emitting view.  But I think the tests to `pack_command` and `pack_commands` are sufficient to test that `tobytes()` doesn't get called (i.e. that we're saving a memory copy)
- It's undecided whether we should allow `Encoder.decode` to work with memoryviews (see my comment for details).

I also noticed that the [argument length-checking and joining code](https://github.com/andymccurdy/redis-py/blob/5750b4c7b221133a219fe64e17914b4a1a9903f5/redis/connection.py#L799-L812) in `pack_commands` is redundant with that of `pack_command`.

andymccurdy:
Thanks @Cody-G. This has been merged.

PR #1287, opened 2020-02-15
Move to more explicit, deterministic resource management

jdufresne:
This removes `__del__` methods. Now, mishandled resource management will emit a `ResourceWarning`. These warnings should not be ignored. They highlight code that are not closing resources explicitly nor deterministically. Library users should use `.close()`, `.disconnect()` or
context managers to resolve these warnings.

This follows the Python semantics of resources such as file objects and sockets. For example:

    $ python3 -W always
    >>> f = open('/tmp/foo', 'w')
    >>> del f
    __main__:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/foo' mode='w' encoding='UTF-8'>
    ResourceWarning: Enable tracemalloc to get the object allocation traceback

Even with these warnings, the resources are still closed by Python when their reference count goes to zero, but now library users can notice the mishandling and fix it.

The previous use of `__del__` comes with a number of warnings. For example, CPython does not guarantee `__del__` is called on exit:

https://docs.python.org/3/reference/datamodel.html#object.__del__

> It is not guaranteed that `__del__()` methods are called for objects that still exist when the interpreter exits.

There are other warnings with using `__del__()`:

> Warning
>
> Due to the precarious circumstances under which `__del__()` methods are invoked, exceptions that occur during their execution are ignored, and a warning is printed to sys.stderr instead. In particular:
>
> `__del__()` can be invoked when arbitrary code is being executed, including from any arbitrary thread. If `__del__()` needs to take a lock or invoke any other blocking resource, it may deadlock as the resource may already be taken by the code that gets interrupted to execute `__del__()`.
>
> `__del__()` can be executed during interpreter shutdown. As a consequence, the global variables it needs to access (including other modules) may already have been deleted or set to `None`. Python guarantees that globals whose name begins with a single underscore are deleted from their module before other globals are deleted; if no other references to such globals exist, this may help in assuring that imported modules are still available at the time when the `__del__()` method is called.

All warnings emitted during tests have been cleaned up.

### Pull Request check-list

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -881,9 +881,7 @@ class Redis(object):
 
     def __exit__(self, exc_type, exc_value, traceback):
         self.close()
-
-    def __del__(self):
-        self.close()
+        self.connection_pool.disconnect()
 
     def close(self):
         conn = self.connection
@@ -3410,12 +3408,6 @@ class PubSub(object):
     def __exit__(self, exc_type, exc_value, traceback):
         self.reset()
 
-    def __del__(self):
-        # if this object went out of scope prior to shutting down
-        # subscriptions, close the connection manually before
-        # returning it to the connection pool
-        self.reset()
-
     def reset(self):
         if self.connection:
             self.connection.disconnect()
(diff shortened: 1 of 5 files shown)


jdufresne:
If you think any docs should be updated with this change, please let me know and I'll modify them.

PR #1301, opened 2020-03-05
Don't send DISCARD after ExecAbortError in pipeline

### LINKED ISSUE #1300: Pipeline incorrectly calls DISCARD after EXEC in case of EXECABORT error
**Version**:
Redis-py 3.4.1
Redis 5.0.7

**Platform**:
Python 3.8.2
Mac OS 10.13.6

**Description**:
Executing an explicit transaction containing an invalid command results in a response error, masking the original parsing error.

Example:
```
from redis import Redis
redis = Redis(...)

with redis.pipeline() as pipe:
    pipe.multi()
    pipe.execute_command('INCR', 'a', 'b', 'c')
    pipe.execute()
```

Expected:
```
redis.exceptions.ResponseError: Command # 1 (INCR a b c) of pipeline caused error: wrong number of arguments for 'incr' command
```

Actual:
```
redis.exceptions.ResponseError: DISCARD without MULTI
```

nickgaya:
### Description of change

Don't send `DISCARD` after `ExecAbortError` when executing a `MULTI`/`EXEC` transaction in the `Pipeline` class.

The `EXECABORT` error type was added in Redis 2.6.5 and is returned from an `EXEC` command to indicate that the transaction was aborted due to an invalid command.  It is not necessary to call `DISCARD` after this error, and doing so causes a "DISCARD without MULTI" error.

Fixes #1300 

### Pull Request check-list

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -3898,8 +3898,6 @@ class Pipeline(Redis):
         try:
             response = self.parse_response(connection, '_')
         except ExecAbortError:
-            if self.explicit_transaction:
-                self.immediate_execute_command('DISCARD')
             if errors:
                 raise errors[0][1]
             raise sys.exc_info()[1]
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -172,6 +172,21 @@ class TestPipeline(object):
             assert pipe.set('z', 'zzz').execute() == [True]
             assert r['z'] == b'zzz'
 
+    def test_parse_error_raised_transaction(self, r):
+        with r.pipeline() as pipe:
+            pipe.multi()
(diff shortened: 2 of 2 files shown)


andymccurdy:
Merged in 07fec7e18996b6b3b4f30ec9636b88c9b287ece5. Thanks for finding this!

I had to do some digging to validate this was a good change. I was actually involved in the original discussion with antirez back in 2012 when this code was added. The existing code worked against a branch he was developing to fix an underlying problem. I shipped the redis-py code, then a few days later in the same thread he changed the behavior of the server to automatically discard before replying with EXECABORT. I apparently missed that change in direction.

PR #1303, opened 2020-03-10
Feature: let lock.extend ignore the remaining ttl.

laixintao:
### Pull Request check-list

### Description of change

Lock.extend now has a new option: keep_remaining, default to False, which is identical with the old Lock.extend; When set to True, extend will set lock's ttl exactly the additional time.

A common use case would be, let's see we have 4 process and just want one to keep running, others standby. When running process is broken or offline, one of the other 3 can takeover. In order to to that, we use a lock to make sure only one prcess is running, and this process extends  lock every 8 seconds before a timeout(10 seconds).

If without the new option `keep_remaining`(not sure if it's a good arg name that makes itself clear), the `tll` will get longer and longer, failover time will be unacceptable long.

Currently there are many workarounds (see https://github.com/andymccurdy/redis-py/issues/629 for example), but I think it would be better for redis-py to support this.

close https://github.com/andymccurdy/redis-py/issues/629

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -16,6 +16,7 @@ class Lock(object):
 
     lua_release = None
     lua_extend = None
+    lua_extend_to = None
     lua_reacquire = None
 
     # KEYS[1] - lock name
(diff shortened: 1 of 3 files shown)


redis/lock.py
```diff
@@ -50,6 +51,26 @@ class Lock(object):
...
+    # return 1 if the locks time was extended, otherwise 0
+    LUA_EXTEND_TO_SCRIPT = """
+        local token = redis.call('get', KEYS[1])
+        if not token or token ~= ARGV[1] then
+            return 0
+        end
+        local expiration = redis.call('pttl', KEYS[1])
+        if not expiration then
+            expiration = 0
+        end
```
laixintao: I checked the document: https://redis.io/commands/pttl

I am not sure when would `pttl key` be `nil` or `false`? Is this check necessary?

laixintao:
@andymccurdy What do you think?

laixintao:
Sounds good, I will make the change. Don't worry I will test the lua script :)

laixintao:
Please take a look on this, when you got time :) @andymccurdy

andymccurdy:
Thanks. I changed the arg name (again) and merged this in d1279291d68c77709d69fe13aa7ff6d912d98ce4

Issue #1128, opened 2019-01-30
Connections are left in an invalid state on interrupt


Chronial:
**Version**: Starting with 3.0.0, introduced by this commit: 04784cb2ecb0847198aaf64102e5129868c17da6

Thanks to the github related feature, I noticed that this is in fact a duplicate of #360 from 2013 – it was fixed back then in 80baa99. But 04784cb2ecb0847198aaf64102e5129868c17da6 (which was released as part of 3.0.0) accidentally reversed that change, so the bug is back.

## How to reproduce
```py
r = redis.Redis(host='localhost', port=6379, db=0)
try:
    print("Press Ctrl-C now")
    r.brpop(['nonexist'])
finally:
    print("This hangs:")
    r.set('status', 'down')
    print("We never get here")
```

andymccurdy:
This was fixed, closing.

Issue #360, opened 2013-06-26
Connections can end up in inconsistent state due to except Exception: instead of bare except:


leifkb:
The changelog for 2.7.3 mentions all bare `except:`s have been changed to `except Exception:`. Unfortunately, this can sometimes cause connections to be released to the pool in an inconsistent state.

Here's a script that demonstrates the problem using gevent timeouts: https://gist.github.com/leifkb/5871666 And a patch: https://gist.github.com/leifkb/5871691

PR #1319, opened 2020-04-10
Simplify exception handlers

jdufresne:
Use the "as" keyword to capture the exception in a variable instead of
sys.exc_info().

Re-raise exception with the bare "raise" syntax.

Avoid "# noqa: E722" by catching BaseException, which includes all
exceptions including SystemExit.

### Pull Request check-list

### DIFF
--- a/benchmarks/command_packer_benchmark.py
+++ b/benchmarks/command_packer_benchmark.py
@@ -13,8 +13,7 @@ class StringJoiningConnection(Connection):
             self.connect()
         try:
             self._sock.sendall(command)
-        except socket.error:
-            e = sys.exc_info()[1]
+        except socket.error as e:
             self.disconnect()
             if len(e.args) == 1:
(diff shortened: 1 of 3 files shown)


andymccurdy:
@jdufresne If I recall correctly, the except clauses that were empty (with the #noqa E722) for a specific reason. #328, #360 and #1128 have more details.

jdufresne:
Yeah, that makes sense.

The old version you linked to used `except Exception` which does not catch `SystemExit` or `KeyboardInterrupt`. But `BaseException` is the base class of _all_ exceptions and so is equivalent to a bare `except:` but it also signals to the reader (and tools) that all exceptions are intentionally caught and it is not simply an oversight due to an omission.

So, IMO, I this version is still preferable as it follows community conventions for catching all exceptions and avoids the need to squelch the tool. WDYT?

andymccurdy:
Oh, gotcha. `BaseException` makes sense then.

PR #1320, opened 2020-04-10
Simplify by removing unnecessary class dummy

jdufresne:
Was simply an alias of class object(). Use that instead.

### Pull Request check-list

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -2,7 +2,6 @@ import threading
 import time as mod_time
 import uuid
 from redis.exceptions import LockError, LockNotOwnedError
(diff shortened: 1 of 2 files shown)


andymccurdy:
The `dummy` class exists because instances of `object` prevent attribute assignment. The line immediately following assigns to the `token` attribute of the object. That will fail on an instance of `object`.

jdufresne:
Thanks for pointing this out. I did not know about this aspect of `object()` classes.

I added a test that would have caught this earlier in #1326

andymccurdy:
Perfect, thanks.

PR #1324, opened 2020-04-11
Fix str/bytes mixup in PythonParser.read_response()

jdufresne:
Calling str() on a bytes object can result in a BytesWarning being
emitted and usually indicates a mixup between byte and string handling.

Now, in the event of an invalid RESP response, use the repr value of the
raw response in the exception message.

Can further simplify the bytes/str handling by comparing the first byte
as a bytes object instead of converting it to str. The bytes literal is
available on all supported Pythons. This removes the need for the
compatibility function, byte_to_chr().

### Pull Request check-list

### DIFF
--- a/redis/_compat.py
+++ b/redis/_compat.py
@@ -143,9 +143,6 @@ if sys.version_info[0] < 3:
     def next(x):
         return x.next()
 
-    def byte_to_chr(x):
-        return x
-
     unichr = unichr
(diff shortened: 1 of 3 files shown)


andymccurdy:
Makes sense to me. Also slightly increases performance as we're no longer casting the first character from bytes to string.

PR #1330, opened 2020-04-18
[Optimize] Rename ambiguous param name in func eval

unbeatablekb:
### Pull Request check-list

### Description of change

I think params here maybe a little ambiguous for people who are not familiar to luajit in redis. 

So I open this pr : )

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -3119,28 +3119,28 @@ class Redis(object):
     def cluster(self, cluster_arg, *args):
         return self.execute_command('CLUSTER %s' % cluster_arg.upper(), *args)
 
-    def eval(self, script, numkeys, *keys_and_args):
+    def eval(self, script, keys_num, *keys_and_args):
         """
-        Execute the Lua ``script``, specifying the ``numkeys`` the script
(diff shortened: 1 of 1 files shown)


andymccurdy:
`keys_num` seems worse to me. If we really wanted to be more verbose, `number_of_keys` would be better.

However, I don't see why we need to do this. It potentially breaks backwards compatibility for anyone passing this value as a named argument. Additionally, I assume that nearly everyone who uses Lua scripting with redis-py uses the builtin `Script` object.

unbeatablekb:
Thanks for your comment : )

I read docmenation and code again after receive your cooment. Useing `Script` object is a good idea since I had nearly no concept about it.

I will close this pr soon.

`It potentially breaks backwards compatibility for anyone passing this value as a named argument.` I still don't understand it. Do you mean keep `numkeys`  stay coordinate with `keys_and_args` ?

andymccurdy:
I don’t believe we should change the name of an argument unless we have a very good reason for doing so.

unbeatablekb:
Thanks.

Issue #1337, opened 2020-05-06
hset for key=0 raises redis.exceptions.DataError


AleksMat:
Hi,

In the new `redis-py` version `3.5.0` I have noticed some additional checks in `Redis.hset` method. Because of these checks the following corner case now fails:
```
Redis().hset('my-name', 0, 'my-value')
```
raising an error `redis.exceptions.DataError: 'hset' with no key value pairs`.

The problem is in [this line](https://github.com/andymccurdy/redis-py/blob/3.5.0/redis/client.py#L3041) where `not key` check is made.

Potential solutions:
 - Changing `not key` into `key is None` check. This covers case `key=0` but it doesn't raise an error if `key=''` where it probably should?
- Changing `not key` into `key is None or not str(key) `. I think this should cover all cases but it's an additional casting operation.

Let me know if any of these options makes sense. I would be happy to submit a PR for this.

For now, I'll just make sure that I always cast `key` argument into string in my code to avoid this issue.

andymccurdy:
It appears that your first suggestion (comparing `key` to `None`) is sufficient. It's actually legal in Redis to have an empty string  as a key within a hash. A PR would be great, thanks!

```python
In [3]: r.hset('my-hash', mapping={'foo': 'bar'})                                                                                                                                 
Out[3]: 1

# manually sending commands to a connection to avoid the `hset` checks
In [5]: conn = r.connection_pool.get_connection('_')
In [6]: conn.send_command('HSET', 'my-hash', '', 'empty value')
In [7]: conn.read_response()                                                                                                                                                         
Out[7]: 1

# notice the empty string key...
In [8]: r.hgetall('my-hash')                                                                                                                                                      
Out[8]: {b'foo': b'bar', b'': b'empty value'}

```

PR #1341, opened 2020-05-09
Fixed hset method for key=0

AleksMat:
This PR is a fix for issue #1337. I have also made a small fix in `hset` docs string and added 2 additional checks in unit tests.

Let me know if there is anything else to do here.


andymccurdy:
Looks good to me. Thanks!

PR #1360, opened 2020-07-01
Support for loading, unloading and listing Redis Modules

RoeyPrat:
### Pull Request check-list

### Description of change
This is a continuation of @vamc19 's PR: https://github.com/andymccurdy/redis-py/pull/808

### DIFF
--- a/redis/exceptions.py
+++ b/redis/exceptions.py
@@ -57,6 +57,10 @@ class NoPermissionError(ResponseError):
     pass
 
 
+class ModuleError(RedisError):
+    pass
+
+
 class LockError(RedisError, ValueError):
     "Errors acquiring or releasing a lock"
     # NOTE: For backwards compatability, this class derives from ValueError.
--- a/redis/exceptions.py
+++ b/redis/exceptions.py
@@ -57,7 +57,7 @@ class NoPermissionError(ResponseError):
     pass
 
 
-class ModuleError(RedisError):
+class ModuleError(ResponseError):
     pass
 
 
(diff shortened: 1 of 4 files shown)


RoeyPrat:
tested manually with this script:
```
import urllib.request
import redis
import os
import stat

urllib.request.urlretrieve(
    "https://github.com/RedisJSON/RedisJSON/releases/download/v1.0.4/rejson.so",
    "rejson.so")
st = os.stat("/home/vagrant/redis-py/rejson.so")
os.chmod("/home/vagrant/redis-py/rejson.so", st.st_mode | stat.S_IEXEC)
r = redis.Redis()
r.module_load("/home/vagrant/redis-py/rejson.so")
print(r.module_list())
r.module_unload('ReJSON')
```
the unload action fails because rejson has commands, and you can't unload a module with commands.

redis/exceptions.py
```diff
@@ -57,6 +57,10 @@ class NoPermissionError(ResponseError):
     pass
 
 
+class ModuleError(RedisError):
```
andymccurdy, line 60: `ModuleError` should inherit from `ResponseError` rather than `RedisError`.

RoeyPrat: OK

RoeyPrat:
This code is not fully tested, and this fails the codecov checks.
The test idea we've had requires compiling a Redis module, which will make the testing a bit too complex for what this is worth.
Perhaps after we move to docker test env we can revisit the test.
For now I'm merging as-is @andymccurdy

PR #1365, opened 2020-07-08
Dockerize

abrookins:
### Pull Request check-list

### Description of change

Move from using Vagrant to build and run a dev environment and tests to using Docker.


abrookins:
@RoeyPrat Do you anticipate wanting a "debug" or "test" type of container that has the redis-py code -- for manual or automated testing purposes? In my local testing with this setup, I've been running `tox` outside of a container, and the tests communicate with the different redis nodes that are running in containers.

andymccurdy:
I think it would be nice to provide a container that executes the tox suite. If a contributor only needs docker installed to execute the full test suite, that seems like a significant improvement.

andymccurdy:
Also, it would be really nice if we can specify the source Redis docker image in a single location. Perhaps we could have a multistage build with a single Dockerfile that references the official Redis docker image. All the other Dockerfiles that you have now can be based on the intermediate one. This way if we want to upgrade to Redis 6.1, we would only need to change a that one intermediate file.

abrookins:
Good point! I’ll make those changes. -A

abrookins:
@andymccurdy Hey, Andy! I read through the docs on multi-stage builds, and they all seemed to refer to situations a little different from this one. In the end I took the approach of building a base image that all the Dockerfiles use, pointed at the "latest" tag. So we would bump the base build and then rebuild the other images, which would pull the new base image.

What do you think about that? I also made some test changes to handle the fact that, in a mult-container setup, each container has its own hostname and IP.

Some of these commands get pretty complex, so I added a Makefile to contain them.

Makefile
```diff
@@ -0,0 +1,13 @@
+.PHONY: base
+
+build:
+	docker build -t redis-py-base docker/base
+	docker-compose down
```
abrookins, line 5: Not sure this is really necessary.

tests/conftest.py
```diff
@@ -14,14 +14,19 @@
...
+DEFAULT_REDIS_MASTER_HOST = "localhost"
 
 
 def pytest_addoption(parser):
-    parser.addoption('--redis-url', default=default_redis_url,
+    parser.addoption('--redis-url', default=DEFAULT_REDIS_URL,
                      action="store",
                      help="Redis connection string,"
                           " defaults to `%(default)s`")
+    parser.addoption('--redis-master-host', default=DEFAULT_REDIS_MASTER_HOST,
```
andymccurdy, line 26: Is there ever a reason where `redis-master-host` would not point to the same host that's defined in the `redis-url`? I can't think of one.

Assuming there isn't a valid reason, we should infer the master hostname from the `redis-url` argument rather than introducing a new setting.

tests/conftest.py
```diff
@@ -156,6 +161,11 @@ def mock_cluster_resp_slaves(request, **kwargs):
     return _gen_cluster_mock_resp(r, response)
 
 
+@pytest.fixture(scope="module")
```
andymccurdy, line 164: Does the `master_host` value ever change from module to module? If not, scoping this to `'session'` seems more appropriate.

abrookins: It shouldn't, no - updated!

tests/conftest.py
abrookins: Probably not. Fixed!

abrookins:
@andymccurdy What do you think about me adding a CONTRIBUTING.rst file, or a new section to README.rst, with a walk-through of how to set up your dev environment and run the tests?

andymccurdy:
@abrookins That would be great. CONTRIBUTING sounds like the appropriate place. We can add a link in the README to the CONTRIBUTING page.

abrookins:
@andymccurdy This is ready for review.

- Added CONTRIBUTING.rst
- Refactored the travis config to use Docker

andymccurdy:
The CONTRIBUTING doc looks great. Thanks!

I played around with the setup locally and have some comments/questions.

1. I generally have a `build` directory in my redis-py directory that `twine` uses to store artifacts to upload to PyPI. As such, `make build` didn't do anything because the build directory was there. Do we need to add `build` to the Makefile's `PHONY` target?

2. Do users really need to run the `make build` target manually? It seems like we could use the `--build` flag to `docker-compose up` like:

```
build:
        docker build -t redis-py-base docker/base

dev:
        build
        docker-compose up -d --build
```

3. Something is wrong with the tox config regarding PyPy builds. The py-plain, py-hiredis, py3-plain and py3-hiredis envs are defaulting to the most recent vanilla Python 2 and Python 3 versions rather than the PyPy versions. Prior to this PR, we used the Travis env matrix to explicitly specify the PyPy builds with `TOXENV=pypy-plain`. Off hand I'm not sure what the best way is of specifying this solely in tox.ini. We could likely add 4 sections (pypy-plain, pypy-hiredis, pypy3-plain, pypy3-hiredis) to the tox.ini file like this (though I haven't tested it):

```
[testenv:pypy-plain]
basepython = pypy

[testenv:pypy-hiredis]
basepython = pypy

[testenv:pypy3-plain]
basepython = pypy3

[testenv:pypy3-hiredis]
basepython = pypy3
```

4. We should specify `make clean` target (or something similar) to call `docker-compose down`, shutting down all the containers.

5. docker-compose services offer a `depends_on` list which specified other services the current service depends on. Should we make use of this to make sure the master starts up before the slave? And maybe that the sentinels start up after both? I don't know how resilient Redis to variability in startup times.

abrookins:
Very good points.

1. Yes, the PHONY specification was out of date. Will fix.

2. I think you're right and we can drop this target.

3. I'll test that out.

4. Ah yes, good point!

5. We should start them in the correct order for sure -- I'll get that better set up.

Makefile
```diff
@@ -0,0 +1,16 @@
...
+base:
+	docker build -t redis-py-base docker/base
+
+dev: base
+	docker-compose up -d --build
+
+test: dev
+	docker-compose run test find . -type f -name '*.py[co]' -delete -o -type d -name __pycache__ -delete
+	docker-compose run test util/wait-for-it.sh master:6379 -- tox -e py3-plain -- --redis-url=redis://master:6379/9
+	docker-compose run test rm -rf .tox
```
abrookins, line 12: So... if you run `make test`, your `.tox` folder gets messed up (files owned by root). I'm not sure what to do about this, but it means that if you run `make test` and then later try to run `tox` locally (outside of a Docker container), the command will fail.

My solution? Nuke the `.tox` folder from the container-side after a test run.

tests/test_connection_pool.py
```diff
@@ -122,25 +129,28 @@ def test_connection_pool_blocks_until_timeout(self):
...
+                             connection_kwargs=connection_kwargs)
         c1 = pool.get_connection('_')
 
         def target():
             time.sleep(0.1)
             pool.release(c1)
 
-        Thread(target=target).start()
         start = time.time()
+        Thread(target=target).start()
```
abrookins, line 147: This test is failing repeatedly in all interpreters, from within a container and without, because the timing is sometimes wrong. (It's inconsistent.) I think it's because we start the thread before starting the timer? This version of the test hasn't failed yet.

Makefile
andymccurdy: It might be easier to just COPY/ADD the redis-py project into the test container rather than using the VOLUME mount. The .pyc files can easily be prevented from getting copied with a .dockerignore file. It would also prevent someone from modifying a python file mid-test execution and having that being reflected via the VOLUME mount.

abrookins: That would help with test file cleanup for sure. Downside - we need to build the test image each time we run the command?

andymccurdy: That should be fast though assuming the COPY is at the end of the Dockerfile. All the previous commands will be cached.

tox.ini
```diff
@@ -1,6 +1,9 @@
+[pytest]
+addopts = -s
```
abrookins, line 2: I added this so that people can drop into `pdb` within tests.

abrookins:
@andymccurdy Ok, I think I addressed your last points. Did I understand your question about the `build` target?

abrookins:
Last remaining problem here is getting codecov to work -- I'm not sure what the right environment variable it's looking for is, and @andymccurdy had some ideas to make this work.

docker-entry.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+# This is the entrypoint for "make clean". It invokes Tox. If running
```
abrookins, line 3: Small nit - it's "make test" right?

abrookins:
I probably reduced coverage by changing the tests to pass in connection details -- I should be able to fix that.

abrookins:
codecov from tox seems to work! 🥳 

I'm a little confused about how running coverage with multiple environments interacts with the "coverage" file. I'm trying some options to make sure that we're combining the output of all the runs when we eventually run `codecov`.

abrookins:
Ok, this is better. Now we get a text-based coverage report when we run tests, and we produce the XML that `codecov` can then use. Also, the coverage is taken separately by interpreter and then merged into a single coverage file, which should give a better representation.

PR #1373, opened 2020-07-27
WIP - Improve test coverage

abrookins:
### Pull Request check-list

### Description of change

Leaving this in Draft mode while I add test coverage, so I can see the result in Codecov.

It seems that Codecov ignores reports that fall under the required percentage, so we're in an odd place now -- having merged some PRs that were under that target percentage.

Because it hasn't recorded any of the coverage reports from PRs since then, Codecov now lacks a 'base report' against which to compare open PRs, and continues to report 92% coverage in master -- which is untrue now, but was true the last time it saw a PR that passed the coverage requirement and which was merged.

My goal is to get us back to the target coverage percent with this PR.

### DIFF
--- a/docker-entry.sh
+++ b/docker-entry.sh
@@ -15,5 +15,5 @@ if [ -z ${TRAVIS-} ]; then
 fi
 
 # use the wait-for-it util to ensure the server is running before invoking Tox
-util/wait-for-it.sh ${REDIS_MASTER} -- tox -- --redis-url=redis://"${REDIS_MASTER}"/9 --is-docker
+util/wait-for-it.sh ${REDIS_MASTER} -- tox -- --redis-url=redis://"${REDIS_MASTER}"/9 --docker
 
--- a/docker-entry.sh
+++ b/docker-entry.sh
@@ -15,5 +15,5 @@ if [ -z ${TRAVIS-} ]; then
 fi
 
 # use the wait-for-it util to ensure the server is running before invoking Tox
-util/wait-for-it.sh ${REDIS_MASTER} -- tox -- --redis-url=redis://"${REDIS_MASTER}"/9
+util/wait-for-it.sh ${REDIS_MASTER} -- tox -- --redis-url=redis://"${REDIS_MASTER}"/9 --is-docker
 
--- a/docker-entry.sh
+++ b/docker-entry.sh
@@ -15,4 +15,4 @@ if [ -z ${TRAVIS-} ]; then
 fi
 
 # use the wait-for-it util to ensure the server is running before invoking Tox
-util/wait-for-it.sh ${REDIS_MASTER} -- tox -- --redis-url=redis://"${REDIS_MASTER}"/9 --docker
+util/wait-for-it.sh ${REDIS_MASTER} -- tox -- --redis-url=redis://"${REDIS_MASTER}"/9 --ssl-port=${REDIS_MASTER_SSL_PORT} --docker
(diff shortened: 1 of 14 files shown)


docker-entry.sh
```diff
@@ -15,5 +15,4 @@ if [ -z ${TRAVIS-} ]; then
 fi
 
 # use the wait-for-it util to ensure the server is running before invoking Tox
-util/wait-for-it.sh ${REDIS_MASTER} -- tox -- --redis-url=redis://"${REDIS_MASTER}"/9
-
+util/wait-for-it.sh ${REDIS_MASTER} -- tox -- --redis-url=redis://"${REDIS_MASTER}"/9 --ssl-port=${REDIS_MASTER_SSL_PORT} --docker
```
abrookins, line 18: Not sure about this approach.

We should test both the SSL and non-SSL connections. We can run both on the same master, but using different ports.

docker/base/Dockerfile
```diff
@@ -1 +1,7 @@
 FROM redis:6.0.6-buster
+
+RUN apt update && apt install openssl
+
+COPY gen-test-certs.sh /tests/gen-test-certs.sh
```
abrookins, line 5: This is a bit of a PITA. Takes a while to generate these test certs when you have to build the base image. But this way they won't expire.

tests/conftest.py
```diff
@@ -23,6 +23,12 @@ def pytest_addoption(parser):
                      action="store",
                      help="Redis connection string,"
                           " defaults to `%(default)s`")
+    parser.addoption('--docker', default=False, action="store_true",
```
abrookins, line 26: I'm using this to decide of the slave host is "slave" (docker environment) or the same host as the master -- based on the --redis-url parameter.

This doesn't quite feel right but making it more correct (according to my own view) would require more pytest options, like --slave-url.

tests/test_sentinel.py
```diff
@@ -76,6 +92,16 @@ def sentinel(request, cluster):
     return Sentinel([('foo', 26379), ('bar', 26379)])
 
 
+@pytest.fixture()
+def live_sentinel(is_docker):
+    if is_docker:
```
abrookins, line 97: This is another place I use the 'docker' pytest flag -- and here again I guess an alternative is more pytest options?

tests/test_pubsub.py
```diff
@@ -545,3 +545,27 @@ def test_get_message_with_timeout_returns_none(self, r):
...
+
+
+class TestPubSubWorkerThread(object):
+    def setup_method(self, method):
+        self.message = None
+
+    def message_handler(self, message):
+        self.message = message
+
+    def test_run_in_thread(self, r):
```
abrookins, line 557: Interesting thing: this test passes, but we leave around some error text in the test output:

```
  /redis-py/tests/test_commands.py:378: ResourceWarning: unclosed <socket.socket fd=5, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('172.18.0.7', 35596)>

    r.ping()
```

The test seems to be using this object as it's designed to be used, so I'm not sure this is important.

tests/test_commands.py
```diff
@@ -345,6 +345,38 @@ def test_client_kill_filter_by_addr(self, r, r2):
...
+                                for client in clients])
+        assert 'redis-py-c1' in clients_by_name
+
+        client_id = clients_by_name['redis-py-c1'].get('id')
+        resp = r.client_kill_filter(_id=client_id, skipme=False)
+        assert resp >= 1
+
+        # This client should have been killed!
+        with pytest.raises(exceptions.ConnectionError):
+            r.ping()
```
abrookins, line 378: Interesting thing: this test passes, but we leave around some error text in the test output:

```
  /redis-py/tests/test_commands.py:378: ResourceWarning: unclosed <socket.socket fd=5, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('172.18.0.7', 35596)>

    r.ping()
```

PR #1376, opened 2020-07-31
Fix check_health not work after the first tick

flisky:
### Pull Request check-list

### Description of change

We found that `health_check_interval` not fully work.
It turns out `next_health_check` always updated in `read_response`, so `time() > self.next_health_check` never comes to true after the first time.

The new added test case `test_health_check_invoked_after_interval` covers this. However, I'm not very clear with the health check under `pipeline`, and please tell me if I break anything? Thanks!

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -704,6 +704,8 @@ class Connection(object):
                 if nativestr(self.read_response()) != 'PONG':
                     raise ConnectionError(
                         'Bad response from PING health check')
+            else:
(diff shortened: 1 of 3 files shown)


andymccurdy:
`next_health_check` gets updated in `read_response` because getting any response back for any command proves that the connection is healthy.

flisky:
Aha, sorry for my misunderstanding, and thank you for the clarification!

PR #1378, opened 2020-08-05
Add a method to get Redis version

ShipraShalini:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1316,6 +1316,9 @@ class Redis(object):
         else:
             return self.execute_command('INFO', section)
 
+    def version(self):
+        return self.info('Server')['redis_version']
+
     def lastsave(self):
         """
         Return a Python datetime object representing the last time the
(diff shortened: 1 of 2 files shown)


andymccurdy:
Hi, thanks for the PR. Two things:

* Users can already get this information from `INFO` (as you do here). Why do we need a redundant method to get the server version?

* The current implementation won't work with pipelines. Client methods cannot format the response of a command in any way. We use response callbacks to format responses so that they can be invoked separately of the client methods.

RoeyPrat:
@ShipraShalini - as an example of a response callback see [parse_info](https://github.com/andymccurdy/redis-py/blob/f001927f91ba3c22cb892bbf1e39b34fd47693bc/redis/client.py#L123).
you can use `parse_info` in a new callback, `parse_version`, and add it to the [RESPONSE_CALLBACKS](https://github.com/andymccurdy/redis-py/blob/f001927f91ba3c22cb892bbf1e39b34fd47693bc/redis/client.py#L627).
 
I'm ignoring the issue of why we need this method when we have the `info` command, but if we want to do this, that is how.

PR #1383, opened 2020-08-22
Change pipeline behaviour to autoexecute upon exit of the `with` block

eoghanmurray:
A behaviour I very much expected!

I've a similar open pull request for aredis at https://github.com/NoneGG/aredis/pull/168
The author there suggested that redis-py is the template being followed, so I thought it might be good to open it up for discussion here also to see if there are any objections?

The idea is that if you are just setting a few values in a transaction, you'd expect them all to be completed upon exiting the `with` block (you don't care about the results in this case):

```
        with r.pipeline() as pipe:
            pipe.set('d', 'val')
            pipe.set('d1', 'val2')
            pipe.set('d1.1', 'val3')
```

I didn't find any documentation relating to pipelines, but didn't look all that hard.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -3874,7 +3874,7 @@ class Pipeline(Redis):
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.reset()
+        self.execute()  # also resets
 
     def __del__(self):
         try:
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -51,6 +51,15 @@ class TestPipeline:
             pipe.execute()
             assert len(pipe) == 0
 
+    def test_pipeline_autoexecute(self, r):
+        with r.pipeline() as pipe:
+            # Fill 'er up!
+            pipe.set('d', 'd1').set('e', 'e1').set('f', 'f1')
+            assert len(pipe) == 3
+
+        # exiting with block calls execute() and reset(), so empty once again
+        assert len(pipe) == 0
+
     def test_pipeline_no_transaction(self, r):
         with r.pipeline(transaction=False) as pipe:
             pipe.set('a', 'a1').set('b', 'b1').set('c', 'c1')


eoghanmurray:
> Sometimes you do care about the reply values from commands

Yes I mused about setting a `pipe.results` attribute; but that might be a distraction and unneeded?

eoghanmurray:
To play devil's advocate, could there be code in the wild written as follows?

    with r.pipeline() as pipe:
         pipe.some_command()
         if some_condition:
             pipe.execute()
         else:
             pass  # don't execute

But I'm struggling to come up with what `some_condition` would be (as you don't get any results/errors until the pipe is executed).

Also presumably they'd be calling `pipe.reset()` explicitly instead of letting a `pass` take care of it?

eoghanmurray:
Prompted by bot ...

Re. the 'access to results' problem, I think @Fogapod is saying that this change request is not preventing you from doing the following:

```
    with r.pipeline() as pipe:
        pipe.set('d', 'val')
        pipe.set('d1', 'val2')
        pipe.set('d1.1', 'val3')
        pipe.get('d1.1', 'val3')            
        res = pipe.execute()
```
It does however prevent the following
```
    with r.pipeline() as pipe:
        pipe.set('d', 'val')
        pipe.set('d1', 'val2')
        pipe.set('d1.1', 'val3')
        pipe.get('d1.1', 'val3')            
    res = pipe.execute()  # would be empty
```
Is the lack of the latter a concern (or is it even possible in terms of the __exit__ being already called?)

If it is, we could store the result in the pipe, such that a later `pipe.execute()` returns the stored result.

jaklinger:
Just a passing visitor, but I noticed this PR as I've also encountered this problem. One way to satisfy all parties in a light-touch way would be to modify [redis.utils.pipeline](https://github.com/andymccurdy/redis-py/blob/master/redis/utils.py#L23):

i.e. change

```python
@contextmanager
def pipeline(redis_obj):
    p = redis_obj.pipeline()
    yield p
    p.execute()
```

to

```python
@contextmanager
def pipeline(redis_obj):
    p = redis_obj.pipeline()
    try:
        yield p
    finally:
        p.execute()
```

and then encourage users who want this behaviour to change their code from:

```python
with r.pipeline() as pipe:
    ...
```

to

```python
from redis.utils import pipeline
with pipeline(r) as pipe:
    ...
```

Obviously it's light-touch enough that I don't really need this PR, but perhaps it would be useful for others.

Issue #717, opened 2016-03-10
pubsub.run_in_thread throws exceptions in case of lost redis connection


baranyaib90:
I started the thread for keyspace events:

```
self._pubsub = self._redis.pubsub()
self._pubsub.psubscribe(**{'__keyspace@0__:*': self._redis_notification})
self._pubsub_thread = self._pubsub.run_in_thread(sleep_time=0.5)
```

If the connection broke with redis (redis process stopped), lot of exceptions appeared.
And I do not know, how my main thread could handle this (or get notified about it).

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 129, in _read_from_socket
    raise socket.error(SERVER_CLOSED_CONNECTION_ERROR)
OSError: Connection closed by server.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/redis/client.py", line 2165, in _execute
    return command(*args)
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 577, in read_response
    response = self._parser.read_response()
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 238, in read_response
    response = self._buffer.readline()
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 168, in readline
    self._read_from_socket()
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 143, in _read_from_socket
    (e.args,))
redis.exceptions.ConnectionError: Error while reading from socket: ('Connection closed by server.',)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 439, in connect
    sock = self._connect()
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 494, in _connect
    raise err
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 482, in _connect
    sock.connect(socket_address)
ConnectionRefusedError: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.4/threading.py", line 920, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.4/dist-packages/redis/client.py", line 2345, in run
    timeout=sleep_time)
[...]

PR #1395, opened 2020-09-03
Add optional exception handler to PubSub.get_message

adeora:
### Pull Request check-list

### Description of change

This PR adds an optional exception handler to `PubSub.get_message`. The intended use case is to solve problems of the sort reported in #717 , where exceptions thrown by `get_message` through the `PubSubWorkerThread` go totally uncaught.


adeora:
@andymccurdy I intend on adding tests/doc changes to this PR, but first I wanted to make a minimal PR with the code changes to discuss the implementation. What do you think of the addition of the `exception_handler` to `get_message`?

I opted to add the handler to `get_message` rather than solely to the `PubSubWorkerThread` because I thought it would be nice for `get_message` to support the same behaviors as the `PubSubWorkerThread`, but I am open to suggestions on a better way to do this.

andymccurdy:
Thanks for looking into this. Several thoughts:

1. I don't think `get_message()` needs to support this. It's trivial to wrap `get_message()` with your own try/except which is a pattern that any Python programmer understands. I don't see the need to complicate things there. I do believe something like this is needed for the `PubSubWorkerThread`.

2. I think the exception handler function should be passed additional arguments. I could see a user wanting access to the exception object (you're already passing this), the `pubsub` object (perhaps they want to re-connect or re-subscribe or something) and the `PubSubWorkerThread` object (`self`, perhaps a user wants to stop the thread?).

3. I'm assuming that exception handler's return value is ignored. I think that makes sense, especially if we're providing the additional args that allow the exception handler to control the thread and pubsub objects.

4. We should catch `BaseException` rather than `Exception`. `BaseException` is Python's root exception class and a few system-level exceptions (e.g., `KeyboardInterrupt`) are derived from it rather than `Exception`. Catching `BaseException` will ensure that the exception handler is run in all cases.

adeora:
@andymccurdy All of that makes sense to me, thanks for those suggestions. I particularly like passing the `pubsub` object and the `PubsubWorkerThread` object to the exception handler in addition to the exception itself. 

I updated the PR.

andymccurdy:
@adeora Looks great. Let's add a test and a quick blurb in the docs.

adeora:
Will do. I don't see any existing tests for the `PubSubWorkerThread` (but maybe I'm missing them); is there some way in particular you want that structured?

andymccurdy:
Hm, you're right. I don't see any tests for the worker thread either.

My initial thought would be something like this (I haven't tested this, but it's probably close):

```python
def test_pubsub_worker_thread_exception_handler(r):
    event = threading.Event()
    
    def exception_handler(ex, pubsub, thread):
        thread.stop()
        event.set()

    pubsub = r.pubsub()
    pubsub.subscribe({'foo': lambda m: m})
    with mock(pubsub, 'get_message', side_effect=Exception('error')):
        pubsub_thread = pubsub.run_in_thread(exception_handler=exception_handler)

    assert event.wait(timeout=1.0)
    assert not pubsub_thread.is_alive()  
```

I'm not sure how the mock will work across threads, so that may need a little tweaking.

adeora:
Hi @andymccurdy  - apologies for the delay here. I just pushed 2 commits that, respectively, include a test addition and a brief blurb in the docs explaining the exception handler.

andymccurdy:
This looks great. Thanks!

adeora:
@andymccurdy  awesome, thanks for merging!

I am very excited to start using this fix in production; is there another release of redispy on the horizon? I'm not super familiar with what your release schedule looks like.

superg28:
Has this merge been added to the pypi repo yet?

andymccurdy:
Not yet. I'll get a release out soon

margars:
Hi,
Could you say when this feature will b released?

adeora:
Hi @andymccurdy - just wanted to quickly follow up on the question of a release date. Do you have a sense of when a new version of `redispy` might be released? My team would love to be able to start using this fix in prod.

Thanks!

mevinbabuc:
@joshuakoh1 Did the above method work out for you ?

PR #1402, opened 2020-09-24
🕰️ Use monotonic clock in Lock (and tests)

coiax:
### Pull Request check-list

### Description of change

During a call to `acquire()`, if the call is `blocking` and has a
`blocking_timeout` set, it uses `time.time()` calls to determine when to
give up attempting to acquire the lock.

However, since `time.time()` is marked as "adjustable", it is possible
for it to go backwards or forwards at a rate other than 1 second per
second, meaning the spinloop may exit earlier or later than expected.

By changing the implementation to use `time.monotonic()`, which is
guaranteed to never go backwards, and not be affected by system clock
updates, this potential problem is fixed.

For the same reason, some time dependent lock tests have also been
changed to use `time.monotonic()`.

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -186,14 +186,14 @@ class Lock:
             blocking_timeout = self.blocking_timeout
         stop_trying_at = None
         if blocking_timeout is not None:
-            stop_trying_at = mod_time.time() + blocking_timeout
+            stop_trying_at = mod_time.monotonic() + blocking_timeout
         while True:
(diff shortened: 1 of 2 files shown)


coiax:
In regards to

>  Is the new or changed code fully tested?

I have honestly no idea how to set up a scenario to test the difference between `time.time` and `time.monotonic`. Suggestions welcome.

RoeyPrat:
Thanks @coiax !

PR #1412, opened 2020-10-26
Added GET argument to SET command

2014BDuck:
### Pull Request check-list

- [X] Does travis tests pass with this change (enable it first in your forked repo and wait for the travis build to finish)?

### Description of change

Issue: https://github.com/andymccurdy/redis-py/issues/1411

Added GET argument to SET command, new in Redis 6.2: https://redis.io/commands/set

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -557,6 +557,7 @@ def parse_set_result(response, **options):
         return response
     return response and str_if_bytes(response) == 'OK'
 
+
 class Redis:
     """
     Implementation of the Redis protocol.
--- a/redis/client.py
+++ b/redis/client.py
@@ -544,6 +544,19 @@ def parse_module_result(response):
     return True
 
 
+def parse_set_result(response, **options):
+    """
+    Handle SET result since GET argument is available since Redis 6.2.
+    Parsing SET result into:
+    - BOOL
+    - String when GET argument is used
+    """
+    if options.get('get'):
+        # Redis will return a getCommand result.
+        # See `setGenericCommand` in t_string.c
+        return response
+    return response and str_if_bytes(response) == 'OK'
+
 class Redis:
     """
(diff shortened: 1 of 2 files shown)


2014BDuck:
To handle this, the `read_response()` method in `connection.py` may need to change a little bit.
When redis return `'+OK'` or `'$2'` (and a "OK" string value), they will be parsed into `b'OK'` in `read_response()`. 

What do you think

andymccurdy:
This looks good. I'd like to wait to merge until Redis 6.2 is released so that we can include it in the unit tests.

2014BDuck:
Hmm but it's not working now. There is no way to tell the difference between 'OK' (bool value) and 'OK' (string value, return when ''get'' param is set to True) in `client.py`.
```
r.set(1,'OK')     # we got b'OK' response
r.set(1,'foo',get=True)  # we got b'OK' again and will be parsed as bool value
```

this must be solve if we want to add 'get' argument to the set command. Will take a look tomorrow

andymccurdy:
This should be pretty easy to solve.

```python
def set(self, ..., get=False):
    options = {}
    ...
    if get:
        pieces.append(b'GET')
        options['get'] = True

    return self.execute_command(*pieces, **options)
```

Then we make a response callback for `SET`:

```python
def parse_set(response, **options):
    if options.get('get'):
        # return the string
    # return the boolean
```

2014BDuck:
haha allright. I will deal it in this way. I thought it should be more concise before, without passing any extra argument.

will catch it up tomorrow.

andymccurdy:
This looks good. Thanks! Tagging this as Redis 6.2 and will merge once the server is released.

chayim:
Thank you for your contribution @2014BDuck. I just merged, and validated with redis server 6.2.5. Approving, and merging into master.

PR #1417, opened 2020-11-04
Add more documentation about encoding of strings

felipou:
Add some more documentation about encoding of strings (input and output) in the README.

https://github.com/andymccurdy/redis-py/issues/1410


andymccurdy:
Thanks!

PR #1420, opened 2020-11-09
use github actions instead of travis-CI

RoeyPrat:
### Pull Request check-list

### Description of change


andymccurdy:
This had failed initially, but then passed when I retired it. Not sure if this was a freak error in github actions or an error in the action itself.

In the action yaml, it looks like this only runs against master. We had Travis setup to run on a push/pr to any branch. Is that not a good idea?

RoeyPrat:
> This had failed initially, but then passed when I retired it. Not sure if this was a freak error in github actions or an error in the action itself.
> 
> In the action yaml, it looks like this only runs against master. We had Travis setup to run on a push/pr to any branch. Is that not a good idea?

It's set to run on push to master branch or PR pointed at master branch.

Issue #1434, opened 2020-12-15
Redis 6.2 New Commands Check List


2014BDuck:
Thanks for wanting to report an issue you've found in redis-py. Please delete this text and fill in the template below.  
It is of course not always possible to reduce your code to a small test case, but it's highly appreciated to have as much data as possible. Thank you!

**Version**: Redis 6.2

**Platform**: N/A

**Description**: Check Redis 6.2 RC [Release Note](https://github.com/redis/redis/blob/b8c67ce41b51247c02a639929e4fab20189afa1e/00-RELEASENOTES) and update missing commands

Commands new in Redis 6.2:
- [ ] Add SMISMEMBER command that checks multiple members [(#7615)](https://github.com/redis/redis/pull/7615)
- [x] Add ZMSCORE command that returns an array of scores [(#7593)](https://github.com/redis/redis/pull/7593)
- [ ] Add LMOVE and BLMOVE commands that pop and push arbitrarily [(#6929)](https://github.com/redis/redis/pull/6929)
- [ ] Add RESET command that resets client connection state[(#7982)](https://github.com/redis/redis/pull/7982)
- [ ] Add COPY command that copies keys [(#7953)](https://github.com/redis/redis/pull/7953)
- [ ] Add ZDIFF and ZDIFFSTORE commands [(#7961)](https://github.com/redis/redis/pull/7961)
- [ ] Add ZINTER and ZUNION commands [(#7794)](https://github.com/redis/redis/pull/7794)
- [ ] Add GEOSEARCH/GEOSEARCHSTORE commands for bounding box spatial queries [(#8094)](https://github.com/redis/redis/pull/8094)
- [x] Add GET parameter to SET command, for more powerful GETSET [(#7852)](https://github.com/redis/redis/pull/7852)
- [ ] Add exclusive range query to XPENDING [(#8130)](https://github.com/redis/redis/pull/8130)
- [ ] Add exclusive range query to X[REV]RANGE [(#8072)](https://github.com/redis/redis/pull/8072)
- [ ] Add GT and LT options to ZADD for conditional score updates [(#7818)](https://github.com/redis/redis/pull/7818)
- [ ] Add CLIENT INFO and CLIENT LIST for specific ids [(#8113)](https://github.com/redis/redis/pull/8113)
- [ ] Add IDLE argument to XPENDING command [(#7972)](https://github.com/redis/redis/pull/7972)
[...]

PR #1437, opened 2020-12-21
#1434 Added support for ZMSCORE new in Redis 6.2 RC

2014BDuck:
### Pull Request check-list

- [X] Does travis tests pass with this change (enable it first in your forked repo and wait for the travis build to finish)?

### Description of change
Added support for zmscore command.
https://github.com/andymccurdy/redis-py/issues/1434
https://github.com/andymccurdy/redis-py/issues/1435

```
>>> r.zmscore(key="test", members=["member1", "member2", "invalid_member"])
[1.0, 2.0, None]
```


adamserafini:
One comment: I think the new test introduced here never runs on CI because the base docker image for redis needs to bumped to something >= 6.2: https://github.com/andymccurdy/redis-py/blob/master/docker/base/Dockerfile#L1

so the the test is currently always skipped because `@skip_if_server_version_lt('6.1.240')` will be true.

2014BDuck:
> One comment: AFAICT, I think the new test introduced here never runs on CI because the base docker image for redis needs to bumped to something >= 6.2: https://github.com/andymccurdy/redis-py/blob/master/docker/base/Dockerfile#L1
> 
> The test is currently always skipped because `@skip_if_server_version_lt('6.1.240')` will be true.

Yes I see. So when I wrote the test, I removed the decorator and run it locally(tox) with version 6.2RC. 
But you are right I think we need to update the docker image thing for Codecov testing.

chayim:
[This PR](https://github.com/andymccurdy/redis-py/pull/1536) does that. once it's merged we should re-run this flow.

redis/client.py
```diff
@@ -3281,6 +3287,22 @@ def zunionstore(self, dest, keys, aggregate=None):
...
+        Returns the scores associated with the specified members
+        in the sorted set stored at key.
+
+        ``members`` should be a list of the member name.
+        Return type is a list of score.
+
+        If the member does not exist, a None will be returned
+        in corresponding position.
+        """
+        if not isinstance(members, list) or len(members) < 1:
```
chayim, line 3301:
```suggestion
        if not members:
```

redis/client.py
```diff
@@ -3281,6 +3287,22 @@ def zunionstore(self, dest, keys, aggregate=None):
         """
         return self._zaggregate('ZUNIONSTORE', dest, keys, aggregate)
 
+    def zmscore(self, key, members):
```
chayim, line 3290:
```suggestion
    def zmscore(self, key, members=[]):
```

chayim:
Hi @2014BDuck thanks for your submission! Everything passes with the current unit tests.

redis/client.py
2014BDuck: Sorry for the late reply.

I am ok with a default value proposal (for compatible reason). But not sure if using a mutable type is fine. This contains risk that the default value might be modified.

So maybe I should change it with a default value None and modify the function body accordingly?

redis/client.py
2014BDuck: So we should check if it's None or not and if it's a list type data?

2014BDuck:
@chayim Hi sorry for the late reply. I would prefer using an immutable type as a default value, what do you think?

redis/client.py
chayim: According to the [zmscore docs](https://redis.io/commands/zmscore) ZMSCORE requires a minimum of one member. This either means that we either:

1. Validate that there's a non-zero length list (as above).
2. Validate that the item passed in is a string or a dictionary.

I'd rather we not overload the members argument, and instead always pass in a list. For the simple case, the list has one item (i.e ['foo']). Validating that we were passed a list I think is overkill, as the docs and the function signature are indicative of the requirement.

chayim:
I think members should merely be required, and we shouldn't do a check on
it. It should be a list.  For what it's worth in the future I'm pretty pro
type hints, but we're not there.

--c

2014BDuck:
@chayim OK then. I am going to add a list-type default value now.

2014BDuck:
Ah sorry for the change. I am still thinking if we could just keep `None` as a default value.

So we can stick with the same validation:
```
# user pass an empty list / empty string / didn't pass member / etc
if not members:
    raise DataError()
```

And, use empty list as default won't stop communicating with redis server when user pass, for example "asdfasdf" or a dict contain something.

So the code is clean, and no risk for the default value being modified, same effect as we use empty list as default.

@chayim

chayim:
For ZMSCORE, you have to have at a minimum one member
<https://redis.io/commands/zmscore>. and it's a list of members. We should
pass in a list, in order to not allow for overloadable constructs. A list
of length one, is still a list.

2014BDuck:
> For ZMSCORE, you have to have at a minimum one member <https://redis.io/commands/zmscore>. and it's a list of members. We should pass in a list, in order to not allow for overloadable constructs. A list of length one, is still a list.

Yes it requires a minimum one member so we don't need a default value, **as setting a `[]` as default won't have different behaviour as the current one**. Like I said it still won't pass the validation. And if user don't pass a list with length >= 1, the suggested changes won't help.

So the only thing it can help is to tell user that it should be a list. However user can notice this requirement in the function doc.

chayim:
Two difference things are at play.  The first is that yes - we should
document accordingly.  Secondly, given that the only viable input is a
list, we should accept a list. If you want to validate the minimum length
of the list, to support the client use - I can support that! However, we
can't accept a default value of None, as the command doesn't support that.

2014BDuck:
Yes I see. 

So **currently** the default value is removed, which means user must pass `members` in. 
```
# user cannot do
r.zmscore("key")

# user can do:
r.zmscore("key", ["a", "b"])             # case 0
r.zmscore("key", "invalid_value")        # case 1
r.zmscore("key", [])                     # case 2
r.zmscore("key", {"invalide": "kv"})     # case 3
```

> we should accept a list.

**And with a `[]` as default value, user can still call this method with all 4 cases' input.** The `[]` is not helpful in limiting users' input.

Currently, this method can do:
- Users must send a `members` argument.

Can't:
- Limiting users to send a list input (The possible way is to validate by `isinstance`, which is suggested to change at the very beginning)

My point is, the default value won't affect any validation logic in the function body. No matter what default value we set, users can pass anything they want. So why not just don't set the default value and require an argument to be passed in. And we will have 2 different ways to handle the rest:
1. validate if it's a list with item: `if isinstance(members, list) and len(members) > 0:`.
2. A less solid validation: `if members:` and hand it over to Redis server. (Currently using)

I think either way we don't need the default value.

chayim:
There are really only two cases present - the right items being specified, or not. The problem with validating the existence of a list, is that it doesn't prevent a user from passing in an invalid list - hence my being against that valiadation.  

Take as an example:

```
x = ['a', 'b']
y = ['a', 'b', 999, {'hello': 'world'}
z = ['a', None]
assert isinstance(x, list)
assert isinstance(y, list)
assert isinstance(z, list)
```

etc.

In this example, True will always be accurate, as everything is a list. I can appreciate the isinstance check on a list, but I wouldn't do anything beyond that. The use will receive a ResponseError should they provide invalid input to redis. Were we to sanitize the inputs in this function, we'd be responsible for raising a DataError (or to be fair an AttributeError) exception in all cases. I'd rather allow redis to do what it does and surface accordingly.

The following is a valid case, and I could understand generally doing this. However, I feel that [type hints](https://www.python.org/dev/peps/pep-0484/) are a better answer. 

```
x = ['some','list','of','things']
if not isinstance(x, list):
    raise AttributeError("x must be a list.")
```

2014BDuck:
OK, I am more clear about the responsibility of redis-py and redis server now. 

But I can't agree with using a mutable object as a default value. It's not recommended in any way and I never do this. Type hint could do this, but using `[]` as default value is just absolutely a different (and not secure) thing.

So for the function signature part: Use `None` as default(not preferred as we discussed) or just don't set a default value. Since user need to understand that they have to pass a list, I think we should keep the current code without a default value.

And for the body part, I've removed the `isinstance()` validation for a long time.  I think we could just be specific on the function signature now if you have a better idea.

2014BDuck:
By the way, we have warnings for unresolved reference on the `Commands` class (for the `execute_command` method) after the refactor in https://github.com/andymccurdy/redis-py/pull/1534 . Is it awarded and planed to fix?

chayim:
Hey @2014BDuck I'll take a look at the warning in another issue - or happily accept a PR. As always, thanks for this contribution, it's great to see!

PR #1460, opened 2021-03-14
Add the getdel command available from redis 6.2.0

pcart-grandjean:
### Pull Request check-list

### Description of change

Coming with **Redis 6.2** is the new command [GETDEL](https://redis.io/commands/getdel). This pull request allows to also have it in this Python interface.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,7 @@ from urllib.parse import urlparse
 # redis 6 release candidates report a version number of 5.9.x. Use this
 # constant for skip_if decorators as a placeholder until 6.0.0 is officially
 # released
-REDIS_6_VERSION = '5.9.0'
+REDIS_6_VERSION = '6.2.0'
 
 
 REDIS_INFO = {}
--- a/tests/test_pubsub.py
+++ b/tests/test_pubsub.py
@@ -8,7 +8,7 @@ import redis
 from redis.exceptions import ConnectionError
 
 from .conftest import _get_client
-from .conftest import skip_if_server_version_lt, skip_if_server_version_gte
+from .conftest import skip_if_server_version_lt
 
 
 def wait_for_message(pubsub, timeout=0.1, ignore_subscribe_messages=False):
@@ -474,7 +474,7 @@ class TestPubSubSubcommands:
     def test_pubsub_channels(self, r):
         p = r.pubsub()
         p.subscribe('foo', 'bar', 'baz', 'quux')
-        for i in range(4):
+        for _ in range(4):
             assert wait_for_message(p)['type'] == 'subscribe'
         expected = [b'bar', b'baz', b'foo', b'quux']
         assert all([channel in r.pubsub_channels() for channel in expected])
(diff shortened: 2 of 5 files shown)


pcart-grandjean:
OK, I have been lazy: I deactivated the unit test that failed. Travis is OK, but of course Codecov is not happy. The thing is that I am not sure about this _test_pubsub_worker_thread_exception_handler_ unit test. I will try and understand, but if you have some information that could help, I am really interested.

tests/conftest.py
```diff
@@ -10,7 +10,7 @@
 # redis 6 release candidates report a version number of 5.9.x. Use this
 # constant for skip_if decorators as a placeholder until 6.0.0 is officially
 # released
-REDIS_6_VERSION = '5.9.0'
+REDIS_6_VERSION = '6.2.0'
```
2014BDuck, line 13: This is not correct. `REDIS_6_VERSION` actually means "Since which version Redis6 feature introduced". Please check the following usage.
`
@skip_if_server_version_lt(REDIS_6_VERSION)
`

So you cannot change it to `6.2.0`

tests/test_pubsub.py
```diff
@@ -474,7 +474,7 @@ class TestPubSubSubcommands:
     def test_pubsub_channels(self, r):
         p = r.pubsub()
         p.subscribe('foo', 'bar', 'baz', 'quux')
-        for i in range(4):
+        for _ in range(4):
```
2014BDuck, line 477: Good catch! But personally speaking, I don't prefer irrelevant changes being included in this PR. What do you think @andymccurdy

2014BDuck (commented):
Hi! Thank you for your PR. I think the constant value should not be modified. WDYT

tests/conftest.py
pcart-grandjean: You are right, this is a poor work around. The issue is that I do not understand why the tests are failing with Redis 6.2. All I want here is to add the _getdel_ verb that is now available in Redis. I would appreciate a lot some help to resolve the Unit Test issues of that PR. If not I will cancel it and wait for somebody else to add _getdel_ in another PR.

tests/test_pubsub.py
pcart-grandjean: I am very new contributing in open source project. So please forgive my silly question. Are you saying here that resolving this Pylint warning should not be done in this PR because it is not its main focus? I would have thought we should use every opportunity to reduce the technical debt, especially if it is clear that it will not break anything.

chayim:
@pcart-grandjean Thank you for the contribution! There's some good news / bad news. The good news is that we have getdel support in master. But the bad news is that I hadn't seen this PR when that happened. My apologies. I'm going to close this PR as a result - and again, my apologies.

On the other hand, the smaller changes you were discussing earlier with @2014BDuck [like this one](https://github.com/andymccurdy/redis-py/pull/1460#discussion_r628783370) we'd love to see in redis-py.  Anything that improves quality is always welcome!

pcart-grandjean:
Hello I am very interested to use the `redis` 4 that includes the `getdel`. But when I install it with pip I only get version 3.5.3. If I want the `getdel`, I need to install the `redis` package from the sources. What's the plan to have a new version on Pypi?

chayim:
@pcart-grandjean A new version of redis-py is coming in the near future! However, until then, will doing a pip install of master at a hash suffice?

`pip install git+git://github.com/andymccurdy/redis-py.git@<the hash of your choice>`

PR #1467, opened 2021-04-06
Updated to return multiple elements

david-1013:
Updated to return multiple elements 

 https://redis.io/commands/lpop: "By default, the command pops a single element from the beginning of the list. When provided with the optional count argument, the reply will consist of up to count elements, depending on the list's length."


chayim:
Closing, already present.

PR #1477, opened 2021-04-27
Add a client info from func for parse_slowlog_get

JasonDevops:
### Pull Request check-list

### Description of change
Then I collection slow-log of redis, 
I find that the logs lack source client info, like this：
```python
r = redis.StrictRedis()
r.config_set(
    "slowlog-max-len", self.slowlog_max_log
)
r.config_set(
    "slowlog-log-slower-than", slowlog_ms * 10
)
s = r.slowlog_get()
print(s) 
```

response
```python
[{'id': 17, 'start_time': 1619514582, 'duration': 23, 'command': b'CONFIG SET slowlog-max-len 1000'}, {'id': 16, 'start_time': 1619513328, 'duration': 35, 'command': b'SLOWLOG GET'}
```

We need get client info from slow-log，for redisc-cli：
image

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -404,7 +404,8 @@ def parse_slowlog_get(response, **options):
         'id': item[0],
         'start_time': int(item[1]),
         'duration': int(item[2]),
-        'command': space.join(item[3])
+        'command': space.join(item[3]),
+        'client': item[4]
     } for item in response]
 
 


Grokzen:
Ain't those fields redis-server 4.0 and above only? Taking this code and running this slowlog method on anything before that will then break as it is not written in backwards compatible fashion

PR #1483, opened 2021-05-12
Remove blocking behaviour from context manager __enter__

aenima:
### Pull Request check-list

### Description of change

Removing unneccessary fixed blocking behaviour in Lock's context manager. Using a lock in a context manager now pays respect to the _blocking_ setting passed to the constructor instead of being ignored. This looks like a leftover from the transition from 2.x to 3.x. 

The comment in the Lock's context manager enter method was also somewhat misleading: The user does know if the lock was or wasn't acquired by the raised exception, so there really is no need in overwriting the blocking behaviour and sacrificing flexibility.

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -149,9 +149,7 @@ class Lock:
                 client.register_script(cls.LUA_REACQUIRE_SCRIPT)
 
     def __enter__(self):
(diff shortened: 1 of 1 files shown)


andymccurdy:
Looks good to me. Thanks!

PR #1419, opened 2020-11-09
sentinel: Add SentinelManagedSSLConnection

AbdealiJK:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `$ tox` pass with this change (including linting)?
- [ ] Does travis tests pass with this change (enable it first in your forked repo and wait for the travis build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change
Create a simple class which taken in both:
 - SentinelManaged
 - SSLConnection

in cases where the sentinel is being used with a TLS enabled redis
setup.

Issue #1306, opened 2020-03-12
SentinelManagedConnection needs a SSLSentinelManagedConnection to complement SSLConnection


nickwilliams-eventbrite:
**Version**: 3.4.1

**Platform**: 2.7.17, 3.5.9, and 3.7.6

**Description**: With Redis 6, Redis provides TLS support directly. This means that replicas and Sentinels can communicate with Redis over TLS. When Sentinel communicates with Redis over TLS, it gives its clients the TLS address for the Redis master and replicas when the master and replica address are requested. However, the `Sentinel` client in this library contains only `SentinelManagedConnection`, and `SentinelManagedConnection` does not understand TLS. It extends `Connection` directly.

So, if I configure my Sentinel like this:

```
self._sentinel = redis.sentinel.Sentinel(
    hosts,
    sentinel_kwargs={'socket_connect_timeout': 5.0, 'socket_timeout': 5.0},
    **{
        'ssl_ca_certs': '/srv/run/tls/ca.crt',
        'ssl_certfile': '/srv/run/tls/redis.crt',
        'ssl_keyfile': '/srv/run/tls/redis.key',
    },
)
```

I get a `TypeError` that `SentinelManagedConnection` does not have keyword arguments `ssl_ca_certs`, etc. But if I configure it like this, without SSL info:

```
self._sentinel = redis.sentinel.Sentinel(
    hosts,
    sentinel_kwargs={'socket_connect_timeout': 5.0, 'socket_timeout': 5.0},
    **{},
)
```

I get SSL handshake errors on the Redis server and unexpected-disconnect errors on the client (because the server is expecting a TLS handshake and the client is not).

I was able to solve this problem very simply. I created this class:

```
class SSLSentinelManagedConnection(redis.sentinel.SentinelManagedConnection, redis.SSLConnection):
    def __init__(self, **kwargs):
        self.connection_pool = kwargs.pop('connection_pool')  # replicate the init code from SentinelManagedConnection
        redis.SSLConnection.__init__(self, **kwargs)  # we cannot use super() here because it is not first in the MRO
```

And then I configured my Sentinel like this:

```
self._sentinel = redis.sentinel.Sentinel(
    hosts,
    sentinel_kwargs={'socket_connect_timeout': 5.0, 'socket_timeout': 5.0},
    **{
        'connection_class': SSLSentinelManagedConnection,
        'ssl_ca_certs': '/srv/run/tls/ca.crt',
[...]

PR #1484, opened 2021-05-15
Allow ssl connection to redis when use sentinel

jsecchiero:
I have used this https://github.com/aio-libs/aioredis-py/pull/953 to fix this #1306.
Dunno if fits with your repo but at least fix my issue

There is also this PR #1419 opened long ago but no one merged, the approach is slightly different

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -2,7 +2,7 @@ import random
 import weakref
 
 from redis.client import Redis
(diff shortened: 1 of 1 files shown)


panteparak:
any update on this? I'm also waiting for this PR

chayim:
Superseeded by #1419.

PR #1488, opened 2021-05-27
typo

ibrag8998:


chayim:
@ibrag8998 Thanks for your contribution, but this isn't a mistake. GETSET sets the [value atomically](https://redis.io/commands/GETSET) (i.e. in an atomic operation) as opposed to automatically (with no direct human control),

PR #1491, opened 2021-06-08
Replace uses of time.time with time.monotonic

bmerry:
### Pull Request check-list

### Description of change

The latter is unaffected by changes to the system clock and hence is
more appropriate for measuring time elapsed between two events, such as
health checks.

Since the initial value is not specified, I set the initial
next_health_check to -1 instead of 0 in case the first call returns 0.
It's unlikely in any real environment, but might be possible if time is
being mocked out.

This change was already done for redis.lock in e635130.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -3649,7 +3649,8 @@ class PubSub:
                 'pubsub connection not set: '
                 'did you forget to call subscribe() or psubscribe()?')
 
-        if conn.health_check_interval and time.time() > conn.next_health_check:
+        if (conn.health_check_interval
+                and time.monotonic() > conn.next_health_check):
             conn.send_command('PING', self.HEALTH_CHECK_MESSAGE,
                               check_health=False)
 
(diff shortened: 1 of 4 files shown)


bmerry:
I haven't added any new tests since I'm assuming the existing tests should cover the modified code. No docs are added since there are no API changes. If user code does examine `Connection.next_health_check` is will observe a difference, but I don't know if that's really considered to be public API.

bmerry:
Not sure why the Travis build of the PR failed. The [Travis run on my fork](https://travis-ci.com/github/bmerry/redis-py/builds/228373433) built fine.

chayim:
@bmerry Would you mind if @dvora-h picked this up from here?

bmerry:
Sure. I'm on holiday for a while, and also moving to a different project which doesn't use redis, so I probably won't have any further involvement in redis-py.

chayim:
> Sure. I'm on holiday for a while, and also moving to a different project which doesn't use redis, so I probably won't have any further involvement in redis-py.

Well, I'm glad you're on holiday, but sad that you likely won't be contributing to redis-py. Here's to hoping

PR #1494, opened 2021-06-17
Add retry mechanism with backoff

nbraun-amazon:
* Add the Backoff abstract base class with four backoff strategies
  based on https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
* Add the Retry class which is constructed from
  a backoff strategy and a maximum number of retries, through the `retry` parameter

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1,4 +1,5 @@
 from itertools import chain
+import copy
 import datetime
 import warnings
 import time
@@ -752,7 +753,13 @@ class Redis:
                  ssl_cert_reqs='required', ssl_ca_certs=None,
                  ssl_check_hostname=False,
                  max_connections=None, single_connection_client=False,
-                 health_check_interval=0, client_name=None, username=None):
+                 health_check_interval=0, client_name=None, username=None,
+                 retry=None):
+        """
+        Initialize a new Redis client.
+        To specify a retry policy, first set `retry_on_timeout` to `True`
+        then set `retry` to a valid `Retry` object
+        """
         if not connection_pool:
             if charset is not None:
                 warnings.warn(DeprecationWarning(
@@ -772,6 +779,7 @@ class Redis:
                 'encoding_errors': encoding_errors,
                 'decode_responses': decode_responses,
                 'retry_on_timeout': retry_on_timeout,
+                'retry': copy.deepcopy(retry),
                 'max_connections': max_connections,
                 'health_check_interval': health_check_interval,
                 'client_name': client_name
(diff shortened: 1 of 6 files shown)


adamserafini:
My 2 cents: obviously interested to hear what the core maintainers think, but I don't think this belongs in `redis-py` library. IMO this breaks the 'do one thing and one thing well' principal. 

I would rather see 'back-off and retry' as a standalone library I can use for any network cal instead of bundling

For example, here is an example of a standalone library covering this exact same functionality which I can already use with `redis-py` (or anything else) if I choose to do so: https://pypi.org/project/retry/

I have no particular say on the matter (this is just a drive by comment) but I would vote to close this PR.

nbraun-amazon:
Thanks for your comment. This is a small change, and it has a specific API with callbacks on failures as well, which the library you mentioned does not support, as far as I can tell. Moreover, I don't think the retry with backoff mechanisms warrants adding an external dependency to `redis-py`.

adamserafini:
Just to clarify, I don’t suggest to add external dependency to redis-py for retry functionality.

I am suggesting that ‘retry’ is a concern that can remain outside of this library, we do not need to build it into redis-py.

abrookins:
Not a maintainer, but I like this PR. This allows a uniform retry policy that applies to all command executions, without requiring someone to figure out how to hook in an external retry library. Moreover, this allows configuring full jitter backoff. From my view, this PR makes redis-py more reliable for more people.

redis/client.py
```diff
@@ -3606,21 +3626,21 @@ def execute_command(self, *args):
...
-            # previously listening to
-            return command(*args, **kwargs)
+    def _disconnect_raise_connect(self, conn, error):
+        conn.disconnect()
+        if not (conn.retry_on_timeout and isinstance(error, TimeoutError)):
+            raise error
+        conn.connect()
+
+    def _execute(self, conn, command, *args, **kwargs):
+        # Connect manually upon disconnection. If the Redis server is down,
```
abrookins, line 3636: This looks like it should be a docstring rather than a line comment.

nbraun-amazon:
Ping on this PR @andymccurdy. Please let me know your feedback. Thanks!

redis/client.py
nbraun-amazon: @abrookins thanks, resolved. Please note I simply moved the comment from the previous location in the code when I moved the code to this new function.

chayim:
First @nbraun-amazon I love the approach - it's elegant and easy to understand. Things are tidy, nicely documented, and high quality. It also solves my personal, #1 issue.

I'm torn. I really love _do one thing and do it well_. It's the reason I love Unix. At the same time, it seems like python libraries in general as they mature move from that model to an extended version of it. I'd vote for inclusion. If anything, toggling retry to default False and friends, solve these concerns. I'm :+1: @andymccurdy ?

redis/client.py
```diff
@@ -772,6 +779,7 @@ def __init__(self, host='localhost', port=6379,
                 'encoding_errors': encoding_errors,
                 'decode_responses': decode_responses,
                 'retry_on_timeout': retry_on_timeout,
+                'retry': copy.deepcopy(retry),
```
chayim, line 782: nit: Since retry is added after client_name and username, I'd place it there in the DeprecationWarning.

nbraun-amazon: Not sure what you mean. What are we deprecating here.

chayim (changes requested):
Can we change the comment format to docstrings for classes and functions when not please (i.e EqualJitteBackoff, DecorrelatedJitterBackoff..) Similarly, can we docstring functions like reset and compute.

Overall, I'm a proponent of merging - but I'd like to hear from @andymccurdy

redis/client.py
chayim: This was a result of GH folding. Ignore. Resolving.

nbraun-amazon:
> Can we change the comment format to docstrings for classes and functions when not please (i.e EqualJitteBackoff, DecorrelatedJitterBackoff..) Similarly, can we docstring functions like reset and compute.
> 
> Overall, I'm a proponent of merging - but I'd like to hear from @andymccurdy

I addressed your comments. Thanks for your feedback.

abrookins:
I'm in favor of merging this PR. It creates a more reliable, batteries-included client out of the box, which I like. It does "one thing," being Redis connection and command management, better than before, IMO.

chayim (approved):

PR #1496, opened 2021-06-19
Fix some typos.

enjoy-binbin:
Hi @andymccurdy I found some typos and i try to fix thoes.
Please take a look for me when you have time. Thanks

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -400,7 +400,7 @@
 * 2.10.2
     * Added support for Hiredis's new bytearray support. Thanks
       https://github.com/tzickel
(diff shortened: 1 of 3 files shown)


chayim:
@enjoy-binbin Since you're making language changes, what do you think about running aspell against the docs and codebase in general?  If you have a Linux, aspell check **FILENAME** helps.

enjoy-binbin:
@chayim Hi
Those errors are what I found when i looking at the pipeline implementation.
I think that would be great if we can check the spell error.

Actually we can create a CI action to check, like https://github.com/redis/redis/blob/unstable/.github/workflows/spell-check.yml

chayim (approved):

PR #1502, opened 2021-07-09
Add NoReplicasError exception

LD250:
### Pull Request check-list

### Description of change

Add NoReplicasError exception.

During redis failover server can send `NOREPLICAS Not enough good replicas to write.` response. Redis server does not accept redis write command in this case. See: https://github.com/redis/redis/blob/ec582cc7ad0706e252bc905822226e49f4c4d0e4/src/server.c#L4118

So it is safe to retry.

We would like to catch this exception and retry write command.


redis/exceptions.py
```diff
@@ -61,6 +61,10 @@ class ModuleError(ResponseError):
     pass
 
 
+class NoReplicasError(ResponseError):
```
chayim, line 64: Please add docstrings

tests/test_connection_pool.py
```diff
@@ -525,6 +525,16 @@ def test_read_only_error(self, r):
         with pytest.raises(redis.ReadOnlyError):
             r.execute_command('DEBUG', 'ERROR', 'READONLY blah blah')
 
+    @skip_if_server_version_lt('2.8.8')
```
chayim, line 528: Are you sure this was in 2.8.8? I think it's 2.8.21 based on the [git blame](https://github.com/redis/redis/commit/14ff572482cfe02c5ac61a1816cb8099d458e499) and tag release dates.

chayim (changes requested):
Thanks for your contribution, this is awesome! Please see the request for a docstring. The unit test item isn't critical, I was just wondering if you knew. Thanks again :+1:

PR #1523, opened 2021-07-26
add idle to xpending

AvitalFineRedis:
### Pull Request check-list

### Description of change

add IDLE option to XPENDING command

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -2736,32 +2736,32 @@ class Redis:
         idle: available from  version 6.2. filter entries by their
         idle-time, given in milliseconds (optional).
         """
+        if {min, max, count} == {None}:
+            if idle is not None or consumername is not None:
+                raise DataError("if XPENDING is provided with idle time"
+                                " or consumername, it must be provided"
+                                " with min, max and count parameters")
+            return self.xpending(name, groupname)
+
         pieces = [name, groupname]
-        if min is not None or max is not None or count is not None:
-            if min is None or max is None or count is None:
-                raise DataError("XPENDING must be provided with min, max "
-                                "and count parameters, or none of them. ")
-            # idle
-            if idle is not None:
-                if int(idle) and idle < 0:
-                    raise DataError("XPENDING idle must be a integer >= 0")
-                pieces.extend(['IDLE', str(idle)])
-            # count
-            if count is not None:
-                if int(count) and count < 0:
-                    raise DataError("XPENDING count must be a integer >= 0")
-                pieces.extend([min, max, str(count)])
-        if idle is not None and \
-                (min is None or max is None or count is None):
-            raise DataError("if XPENDING is provided with idle time,"
-                            " it must be provided with min, max and"
-                            " count parameters")
-        if consumername is not None:
-            if min is None or max is None or count is None:
-                raise DataError("if XPENDING is provided with consumername,"
-                                " it must be provided with min, max and"
-                                " count parameters")
-            pieces.append(consumername)
+        if min is None or max is None or count is None:
+            raise DataError("XPENDING must be provided with min, max "
+                            "and count parameters, or none of them.")
+        # idle
+        try:
+            if int(idle) < 0:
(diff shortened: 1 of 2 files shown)


AvitalFineRedis:
@chayim

redis/client.py
```diff
@@ -2733,15 +2733,26 @@ def xpending_range(self, name, groupname, min, max, count,
...
         consumername: name of a consumer to filter by (optional).
+        idle: available from  version 6.2. filter entries by their
+        idle-time, given in milliseconds (optional).
         """
         pieces = [name, groupname]
         if min is not None or max is not None or count is not None:
             if min is None or max is None or count is None:
                 raise DataError("XPENDING must be provided with min, max "
                                 "and count parameters, or none of them. ")
+            if idle is not None:
```
chayim, line 2744:
```suggestion
            try:
                if count < -1:
                    raise DataError("XPENDING count must be an integer >=1 -1")
                pieces.extend(['IDLE', str(idle)])
            except TypeError:
                pass
```

redis/client.py
```diff
@@ -2733,15 +2733,26 @@ def xpending_range(self, name, groupname, min, max, count,
...
+                if not isinstance(idle, int) or count < -1:
+                    raise DataError("XPENDING idle must be a integer >= -1")
+                pieces.extend(['IDLE', str(idle)])
             if not isinstance(count, int) or count < -1:
                 raise DataError("XPENDING count must be a integer >= -1")
-            pieces.extend((min, max, str(count)))
+            pieces.extend([min, max, str(count)])
+        if idle is not None:
```
chayim, line 2751:
```suggestion
        if idle is not None and (min is None or max is None or count is None):
        	raise DataError("If XPENDING is provided with idle time,"
                            " it must be provided with min, max and"
                            " count parameters")
 ```

File: tests/test_commands.py [hunk 3]
```diff
@@ -2562,6 +2562,27 @@ def test_xpending_range(self, r):
...
+
+        # read 1 message from the group with each consumer
+        r.xreadgroup(group, consumer1, streams={stream: '>'}, count=1)
+        r.xreadgroup(group, consumer2, streams={stream: '>'}, count=1)
+
+        response = r.xpending_range(stream, group,
+                                    min='-', max='+', count=5)
+        assert len(response) == 2
+        response = r.xpending_range(stream, group,
+                                    min='-', max='+', count=5, idle=1000)
```
chayim, line 2583: It would be good to test the DataError explicitly here.

chayim (changes requested):
Thanks @AvitalFineRedis - take a look at the changes.

redis/client.py
AvitalFineRedis: Does not handle the case then `idle` is not an integer.

redis/client.py
```diff
@@ -2733,21 +2733,35 @@ def xpending_range(self, name, groupname, min, max, count,
...
-            pieces.append(consumername)
+        if min is None or max is None or count is None:
+            raise DataError("XPENDING must be provided with min, max "
+                            "and count parameters, or none of them.")
+        # idle
+        try:
+            if int(idle) < 0:
+                raise DataError("XPENDING idle must be a integer >= 0")
+            pieces.extend(['IDLE', idle])
+        except TypeError:
```
AvitalFineRedis, line 2755: What if the user uses a float number?@chayim 
It will pass the int test and redis server will raise an error: `redis.exceptions.ResponseError: value is not an integer or out of range`

chayim: According to the redis docs IDLE should be an [int](https://redis.io/commands/XPENDING).  This means you have to catch ValueError as well to catch the int(str("5.5")) case.

```suggestion
        except (TypeError, ValueError):
```

AvitalFineRedis: But I don't want to catch it...
I want to raise a Value error in case of int(5.5)

AvitalFineRedis: Ok, I think I will leave it as before, and the command will throw a ResponseError

chayim (approved):

PR #1543, opened 2021-08-16
better owned function check

Mograbi:
there is some cases where the local token is not bytes so this check fails.
so we need to convert it to bytes also and not only the stored_token

### Pull Request check-list

### Description of change

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -222,8 +222,13 @@ class Lock:
         if stored_token and not isinstance(stored_token, bytes):
             encoder = self.redis.connection_pool.get_encoder()
             stored_token = encoder.encode(stored_token)
-        return self.local.token is not None and \
-            stored_token == self.local.token
+        local_token = self.local.token
+        # in case local token is not stored as bytes
+        if self.local.token and not isinstance(self.local.token, bytes):
+            encoder = self.redis.connection_pool.get_encoder()
+            local_token = encoder.encode(self.local.token)
+        return local_token is not None and \
+            stored_token == local_token
 
     def release(self):
         "Releases the already acquired lock"


chayim:
@Mograbi thanks for the contribution! Do you think you could highlight what those cases are where the token check fails, and maybe even provide an associated unit test? I'd like to be able to see that we have a failed token, and that this addresses that, prior to looking to introduce this change.

Mograbi:
@chayim , in my code I was using self.local.token as string. this check only passes in case I use Python2
in python3 I got all sort of problems with owned because of that, and since inside the function it's written that the comparison should be done in bytes I think we should also encode the local.token.
what do you think ?

andymccurdy:
@Mograbi Are you setting the token manually somehow? The only place within redis-py that sets the token value is here: https://github.com/andymccurdy/redis-py/blob/master/redis/lock.py#L190. That value comes from a few lines above where the token is always encoded to `bytes`. So I'm confused how `self.local.token` could ever be a string unless you're setting it manually.

Mograbi:
@andymccurdy so here is the scenario that I have problems with:

- create a Lock and lock it with value X (indicating that user X has the lock).
- get an instance to the same lock (here in __init__ the local.token is None) with the same user X
- if you try to do acquire it returns False, so that means do_acquire returned False and local.token is still None.
- now I want  to check If the current user X has that lock.
- owned fails because the local.token is None
- so I needed to modify the local.token manually (set it to string 'X')

took me a while to figure out what was the problem

andymccurdy:
It seems like you're using the lock in an unintended way. `Lock.owned()` is used to determine if the current lock instance has successfully acquired the lock. But it seems like you're trying to use it to determine if "User X" has acquired the lock, potentially in some other process or thread.

If it's important to your use case to know which user currently holds the lock, I'd do something like this:

```python
r = redis.Redis(...)

with r.lock('my-lock'):
    # set the 'lock-owned-by' key to the `current_user`, which probably should be some unique identifier of the user.
    r.set('lock-owned-by', current_user)

    # do other stuff here
    # ...

    # delete the 'lock-owned-by' key as the lock is about to be released.
    r.delete('lock-owned-by')
```

Now you can just check the value of the 'lock-owned-by' key in another process or thread to determine which user (if any) has the lock.

Issue #1546, opened 2021-08-19
Unsupported Redis Commands as of 6.2.0


chayim:
The following is the list of released redis commands that redis-py does not support. This is a tracking issue, so that we can link back, and check off each item as support is added, or the decision is made to node support them. Thanks to @AvitalFineRedis for compiling this list by going through each redis command. For the curious, the full up-to-date command list can always be found [here](https://github.com/redis/redis-doc/blob/master/commands.json).

- [x] ACL DELUSER - support deletion of only one user (can't have a list of users)
- [x] ACL GENPASS - Does not support optional parameter `bits`
- [x] ACL HELP - not supported
- [x] AUTH - not supported
- [x] BGSAVE - Does not support optional parameter `SCHEDULE` (version >= 3.2.2)
- [x] CLIENT KILL - does to support `USER` filter
- [x] CLIENT LIST - support only single client-id
- [x] CLIENT GETREDIR - [PR](https://github.com/redis/redis-py/pull/1623)
- [x] CLIENT PAUSE - #1549 (need to add test)
- [x] CLIENT REPLY - not supported
- [x] CLIENT TRACKING - #1612 
- [x] CLIENT TRACKINGINFO- not supported
- [x] COMMAND (get an array of Redis commands details) - not supported
- [x] COMMAND COUNT  - not supported
- [x] COMMAND GETKEYS - #1738
- [x] COMMAND INFO  - #1609 
- [x] DEBUG SEGFAULT - not supported
- [x] DISCARD - not supported
- [x] FAILOVER - not supported (version >= 6.2.0)
- [x] GEOADD - [NX|XX] [CH] options not supported (version >= 6.2.0) #1605
- [x] GEORADIUS and GEORADIUSBYMEMBER -  The `ANY` option for `COUNT` is not supported #1606
- [x] GEOSEARCH
- [x] GEOSEARCHSTORE
- [x] HELLO - not supported
- [x] LULWUT - #1568 
- [x] LPUSHX - cant have a list.
- [x] MEMORY DOCTOR-  #1608
- [x] MEMORY HELP- #1608 
- [x] MEMORY MALLOC-STATS- not supported
- [x] MIGRATE - `AUTH2` not supported #1616
- [x] MODULE LOAD - args option not supported
- [x] QUIT- not supported
- [x] RESET- not supported
- [x] RESTORE - does not support `IDLETIME` and `FREQ` params
- [x] ROLE- not supported #1610 
[...]

PR #1548, opened 2021-08-19
Support MINID and LIMIT on XADD

AvitalFineRedis:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/commands.py
+++ b/redis/commands.py
@@ -1646,10 +1646,10 @@ class Commands:
         name: name of the stream
         fields: dict of field/value pairs to insert into the stream
         id: Location to insert this record. By default it is appended.
-        maxlen: truncate old stream members beyond this size
+        maxlen: truncate old stream members beyond this size. Can't be specify with minid.
+        minid: the minimum id in the stream to query. Can't be specify with maxlen.
         approximate: actual stream length may be slightly more than maxlen
         nomkstream: When set to true, do not make a stream
-        minid: the minimum id in the stream to query
         limit: specifies the maximum number of entries to retrieve
         """
         pieces = []
--- a/redis/commands.py
+++ b/redis/commands.py
@@ -1680,32 +1680,6 @@ class Commands:
             pieces.extend(pair)
         return self.execute_command('XADD', name, *pieces)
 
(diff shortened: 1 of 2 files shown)


AvitalFineRedis:
#1546

redis/commands.py
```diff
@@ -1649,6 +1649,8 @@ def xadd(self, name, fields, id='*', maxlen=None, approximate=True,
         maxlen: truncate old stream members beyond this size
         approximate: actual stream length may be slightly more than maxlen
         nomkstream: When set to true, do not make a stream
+        minid: the minimum id in the stream to query
+        limit: specifies the maximum number of entries to retrieve
         """
         pieces = []
         if maxlen is not None:
```
chayim, line 1656: Maxlen and minid are mutually exclusive. This would allow us to specify both in the command.
https://redis.io/commands/XADD

redis/commands.py
```diff
@@ -1658,6 +1660,14 @@ def xadd(self, name, fields, id='*', maxlen=None, approximate=True,
             if approximate:
                 pieces.append(b'~')
             pieces.append(str(maxlen))
+        if minid is not None:
+            pieces.append(b'MINID')
+            if approximate:
+                pieces.append(b'~')
```
chayim, line 1666: IMHO approximate should be False, by default. But, according to the docs it looks like equals (=) or (~) can both be specified explicitly. Can we test case both?

chayim (changes requested):
See inlines. Thanks @AvitalFineRedis

redis/commands.py
AvitalFineRedis: I tested it in L2449. 
They need to specify a threshold, so I can't merge them into one variable anyway,.

redis/commands.py
AvitalFineRedis: I don't understand...
I can't change the default because we don't want to break anything, 
And not specifying the `~` sign is equivalent to adding `=`.

I can change it to:
            ```
if approximate:
                pieces.append(b'~')
            else:
                pieces.append(b'=')
```

PR #1578, opened 2021-09-27
Fix garbage collection deadlock

emorozov:
### Pull Request check-list

For a long time we experienced weird lockups of the Celery package when using Redis backend (even when using redis backend only for results).

After weeks of debugging, I've found a culprit: `PubSub` objects create a circular reference to self via `connection.register_connect_callback`. This means that `PubSub` objects are not garbage collected immediately, when their refcount decreases to zero, they're collected asynchronously using generational GC.

Also, `PubSub` defines a destructor that releases a connection from pool. Pool uses threading.Lock when managing connections. If some pool method grabs the lock and at the same time `PubSub` object is garbage collected (which is pretty likely as our celery instances lock up at least once a week), `PubSub` destructor deadlocks forever waiting for a lock that will be never released.

I've decided that the only solution to this problem is to avoid creating circular references to `PubSub` objects.

Test look weird and I'm worried that it is a bit fragile, but I failed to devise a scheme that would cause Python garbage collector to run at exact moment in time when the pool is inside critical section. So the test run enough cycles to cause a problem. But different interpreters/different gc settings may cause a test to finish without locking up. I don't know how to solve that.

It runs on CPython 3.9 though: without weakref changes, the test will hang forever until pytest-timeout will cancel it.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -1,5 +1,6 @@
 flake8>=3.9.2
 pytest==6.2.5
+pytest-timeout==2.0.1
 tox==3.24.4
 tox-docker==3.1.0
 invoke==1.6.0
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -10,6 +10,7 @@ import os
 import socket
 import threading
 import warnings
+import weakref
 
 from redis.exceptions import (
     AuthenticationError,
@@ -562,7 +563,7 @@ class Connection:
             pass
 
     def register_connect_callback(self, callback):
-        self._connect_callbacks.append(callback)
+        self._connect_callbacks.append(weakref.WeakMethod(callback))
 
     def clear_connect_callbacks(self):
         self._connect_callbacks = []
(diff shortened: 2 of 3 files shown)


emorozov:
Corresponding celery PR: https://github.com/celery/celery/pull/6969
Initially I thought that the problem is caused by Celery code.

chayim:
I think this makes sense. I've validated these tests as much as I can both with and without your changes, and I've tried on machines that should produce some deadlocks (really, really slow arm boxen).  Mind merging in the latest master into your branch. Specifically you need to get your dependencies into the dev_dependencies and out of tox, and then I'll merge it in - probably on Sunday.

Thanks  a tonne!

chayim:
Hi @emorozov we're close! The changes you made to tox.ini, really belong in the dev_requirements.txt. Your tox.ini shouldn't need modifications,

emorozov:
Hello @chayim ,
I've moved the dependency from tox.ini to dev_requirements.txt

chayim:
Thank you @emorozov, I really appreciate it. Let's get this into the next release candidate. Merging!

PR #1598, opened 2021-10-08
client_list do not use mutable default argument

Andrew-Chen-Wang:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/commands.py
+++ b/redis/commands.py
@@ -349,7 +349,7 @@ class Commands:
         """
         return self.execute_command('CLIENT INFO')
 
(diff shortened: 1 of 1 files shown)


Andrew-Chen-Wang:
@chayim I'll close this since the reasoning is more stylistic in this scenario, but, given a function where you write ``def run(a_list=[]): a_list.append(None); print(a_list)``, if you execute ``run`` multiple times without passing in an argument, the number of elements inside ``a_list`` increases. In this case, this doesn't happen, so it's more stylistic.

PR #1602, opened 2021-10-13
Add support to consumername in `xpending_range` 

AvitalFineRedis:
### Pull Request check-list

### Description of change

Add support to consumername parameter in `xpending_range`

### DIFF
--- a/redis/commands.py
+++ b/redis/commands.py
@@ -2072,6 +2072,9 @@ class Commands:
             pieces.extend([min, max, count])
         except TypeError:
             pass
+        # consumername
+        if consumername:
+            pieces.append(consumername)
 
         return self.execute_command('XPENDING', *pieces, parse_detail=True)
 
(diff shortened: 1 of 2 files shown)


chayim (changes requested):
Can you review xpending in general? It doesn't feel like the [redis command](https://redis.io/commands/XPENDING) at all. I think the confusion comes from the previous implementation.

Please see what you can do to collapse the two...

redis/commands.py
```diff
@@ -2034,18 +2034,21 @@ def xpending(self, name, groupname):
...
-                       consumername=None, idle=None):
+    def xpending_range(self, name, groupname, idle=None,
+                       min=None, max=None, count=None,
+                       consumername=None):
         """
         Returns information about pending messages, in a range.
+
         name: name of the stream.
         groupname: name of the consumer group.
+        idle: available from  version 6.2. filter entries by their
```
chayim, line 2045: @AvitalFineRedis I don't see idle-time in the function signature, but I do see it in the documentation.

AvitalFineRedis: The third param

coleifer:
Thanks for the fix @AvitalFineRedis

PR #1635, opened 2021-10-21
Enable floating parameters in SET (ex and px)

### LINKED ISSUE #1576: error in set when ex is a float is confusing
**Version**: redis-py 3.5.3

**Platform**: Python 3.7.3 (default, Jan 22 2021, 20:04:44) [GCC 8.3.0] on linux

**Description**:
```py
>>> import redis
>>> r = redis.Redis(host='localhost', port=6379, db=0)
>>> r.set('foo', 'bar', ex=10.0)
  File "/redis/client.py", line 1801, in set
    return self.execute_command('SET', *pieces)
  File "/redis/client.py", line 901, in execute_command
    return self.parse_response(conn, command_name, **options)
  File "/redis/client.py", line 915, in parse_response
    response = connection.read_response()
  File "/redis/connection.py", line 756, in read_response
    raise response
redis.exceptions.ResponseError: value is not an integer or out of range
```
This exception is confusing, because it doesn't specify which value is supposed to be an integer. There is a parameter to set called "value", but that's not what caused the exception. It would be better to validate that ex is an integer and raise something like `TypeError: ex must be int or timedelta, not "float"`.

AvitalFineRedis:
### Pull Request check-list

### Description of change

Closes #1576 
Now it is possible to use floating numbers without getting an error.
One thing I have in mind - If the floating number is not rounded, i.e., 10.5, the command will use the integer 10.

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -1104,6 +1104,8 @@ class TestRedisCommands:
         assert r['a'] == b'1'
         assert 0 < r.pttl('a') <= 10000
         assert 0 < r.ttl('a') <= 10
+        with pytest.raises(exceptions.DataError):
+            assert r.set('a', '1', px=10.0)
 
     @skip_if_server_version_lt('2.6.0')
     def test_set_px_timedelta(self, r):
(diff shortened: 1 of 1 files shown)


redis/commands.py
```diff
@@ -1173,13 +1173,21 @@ def set(self, name, value,
         if ex is not None:
             pieces.append('EX')
             if isinstance(ex, datetime.timedelta):
-                ex = int(ex.total_seconds())
-            pieces.append(ex)
+                ex = ex.total_seconds()
+            try:
```
chayim, line 1177: What about:

if not isinstance(ex, int):
    raise DataError("something something").

Repeat

PR #1660, opened 2021-10-28
Adding RedisCluster client to support Redis Cluster Mode

### LINKED ISSUE #931: Redis Cluster Support?
I've been looking into getting Celery to work with redis clusters.

Kombu has a PR that's a few months old to implement it's own transport/backend for redis-cluster, but it doesn't seem like it's going to get merged.

I was wondering if this library would ever support redis clusters such that it's transparent to the end user whether you're connecting to a redis clusters or a standalone redis instance.

Is there any chance of redis-py supporting redis clusters going forward?

barshaul:
### Pull Request check-list

- [ V] Does `$ tox` pass with this change (including linting)?
- [V ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ V] Is the new or changed code fully tested?
- [ V] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change

closes #931 

Added support for cluster mode and provides a client for
[Redis Cluster](<https://redis.io/topics/cluster-tutorial>).

The cluster client is based on [redis-py-cluster](https://github.com/Grokzen/redis-py-cluster)
by Grokzen, with a lot of added and 
changed functionality.

**Create RedisCluster:**

Connecting redis-py to the Redis Cluster instance(s) is easy.
RedisCluster requires at least one node to discover the whole cluster nodes,
and there is multiple ways of creating a RedisCluster instance:

- Use the 'host' and 'port' arguments:

``` pycon
    >>> from redis.cluster import RedisCluster as Redis
    >>> rc = Redis(host='localhost', port=6379)
    >>> print(rc.get_nodes())
    [[host=127.0.0.1,port=6379,name=127.0.0.1:6379,server_type=primary,redis_connection=Redis<ConnectionPool<Connection<host=127.0.0.1,port=6379,db=0>>>], [host=127.0.0.1,port=6378,name=127.0.0.1:6378,server_type=primary,redis_connection=Redis<ConnectionPool<Connection<host=127.0.0.1,port=6378,db=0>>>], [host=127.0.0.1,port=6377,name=127.0.0.1:6377,server_type=replica,redis_connection=Redis<ConnectionPool<Connection<host=127.0.0.1,port=6377,db=0>>>]]
```
- Use Redis URL:

``` pycon
    >>> from redis.cluster import RedisCluster as Redis
    >>> rc = Redis.from_url("redis://localhost:6379/0")
```

- Use ClusterNode(s):

``` pycon
    >>> from redis.cluster import RedisCluster as Redis
    >>> from redis.cluster import ClusterNode
    >>> nodes = [ClusterNode('localhost', 6379), ClusterNode('localhost', 6378)]
    >>> rc = Redis(startup_nodes=nodes)
```

When a RedisCluster instance is being created it first attempts to establish a
connection to one of the provided startup nodes. If none of the startup nodes
are reachable, a 'RedisClusterException' will be thrown.
After a connection to the one of the cluster's nodes is established, the
RedisCluster instance will be initialized with 3 caches:
a slots cache which maps each of the 16384 slots to the node/s handling them,
a nodes cache that contains ClusterNode objects (name, host, port, redis connection)
for all of the cluster's nodes, and a commands cache contains all the server
supported commands that were retrieved using the Redis 'COMMAND' output.

RedisCluster instance can be directly used to execute Redis commands. When a
command is being executed through the cluster instance, the target node(s) will
be internally determined. When using a key-based command, the target node will
be the node that holds the key's slot.
Cluster management commands or other cluster commands have predefined node
group targets (all-primaries, all-nodes, random-node, all-replicas), which are
outlined in the command’s function documentation.
For example, ‘KEYS’ command will be sent to all primaries and return all keys
in the cluster, and ‘CLUSTER NODES’ command will be sent to a random node.
Other management commands will require you to pass the target node/s to execute
the command on.

``` pycon
    >>> # target-nodes: the node that holds 'foo1's key slot
    >>> rc.set('foo1', 'bar1')
    >>> # target-nodes: the node that holds 'foo2's key slot
    >>> rc.set('foo2', 'bar2')
    >>> # target-nodes: the node that holds 'foo1's key slot
    >>> print(rc.get('foo1'))
    b'bar'
    >>> # target-nodes: all-primaries
    >>> print(rc.keys())
    [b'foo1', b'foo2']
    >>> # target-nodes: all-nodes
    >>> rc.flushall()
```

**Specifying Target Nodes:**

As mentioned above, some RedisCluster commands will require you to provide the
target node/s that you want to execute the command on, and in other cases, the
target node will be determined by the client itself. That being said, ALL
RedisCluster commands can be executed against a specific node or a group of
nodes by passing the command kwarg `target_nodes`.
The best practice is to specify target nodes using RedisCluster class's node
flags: PRIMARIES, REPLICAS, ALL_NODES, RANDOM. When a nodes flag is passed
along with a command, it will be internally resolved to the relevant node/s.
If the nodes topology of the cluster changes during the execution of a command,
the client will be able to resolve the nodes flag again with the new topology
and attempt to retry executing the command.

``` pycon
    >>> from redis.cluster import RedisCluster as Redis
    >>> # run cluster-meet command on all of the cluster's nodes
    >>> rc.cluster_meet(Redis.ALL_NODES, '127.0.0.1', 6379)
    >>> # ping all replicas
    >>> rc.ping(Redis.REPLICAS)
    >>> # ping a specific node
    >>> rc.ping(Redis.RANDOM)
    >>> # ping all nodes in the cluster, default command behavior
    >>> rc.ping()
    >>> # execute bgsave in all primaries
    >>> rc.bgsave(Redis.PRIMARIES)
```

You could also pass ClusterNodes directly if you want to execute a command on a
specific node / node group that isn't addressed by the nodes flag. However, if
the command execution fails due to cluster topology changes, a retry attempt
will not be made, since the passed target node/s may no longer be valid, and
the relevant cluster or connection error will be returned.

``` pycon
    >>> node = rc.get_node('localhost', 6379)
    >>> # Get the keys only for that specific node
    >>> rc.keys(node)
    >>> # get Redis info from a subset of primaries
    >>> subset_primaries = [node for node in rc.get_primaries() if node.port > 6378]
    >>> rc.info(subset_primaries)
```

In addition, you can use the RedisCluster instance to obtain the Redis instance
of a specific node and execute commands on that node directly. The Redis client,
however, cannot handle cluster failures and retries.

``` pycon
    >>> cluster_node = rc.get_node(host='localhost', port=6379)
    >>> print(cluster_node)
    [host=127.0.0.1,port=6379,name=127.0.0.1:6379,server_type=primary,redis_connection=Redis<ConnectionPool<Connection<host=127.0.0.1,port=6379,db=0>>>]
    >>> r = cluster_node.redis_connection
    >>> r.client_list()
    [{'id': '276', 'addr': '127.0.0.1:64108', 'fd': '16', 'name': '', 'age': '0', 'idle': '0', 'flags': 'N', 'db': '0', 'sub': '0', 'psub': '0', 'multi': '-1', 'qbuf': '26', 'qbuf-free': '32742', 'argv-mem': '10', 'obl': '0', 'oll': '0', 'omem': '0', 'tot-mem': '54298', 'events': 'r', 'cmd': 'client', 'user': 'default'}]
    >>> # Get the keys only for that specific node
    >>> r.keys()
    [b'foo1']
```

**Multi-key commands:**

Redis supports multi-key commands in Cluster Mode, such as Set type unions or
intersections, mset and mget, as long as the keys all hash to the same slot.
By using RedisCluster client, you can use the known functions (e.g. mget, mset)
to perform an atomic multi-key operation. However, you must ensure all keys are
mapped to the same slot, otherwise a RedisClusterException will be thrown.
Redis Cluster implements a concept called hash tags that can be used in order
to force certain keys to be stored in the same hash slot, see 
[Keys hash tag](https://redis.io/topics/cluster-spec#keys-hash-tags).
You can also use nonatomic for some of the multikey operations, and pass keys
that aren't mapped to the same slot. The client will then map the keys to the
relevant slots, sending the commands to the slots' node owners. Non-atomic
operations batch the keys according to their hash value, and then each batch is
sent separately to the slot's owner.

``` pycon
    #  Atomic operations can be used when all keys are mapped to the same slot
    >>> rc.mset({'{foo}1': 'bar1', '{foo}2': 'bar2'})
    >>> rc.mget('{foo}1', '{foo}2')
    [b'bar1', b'bar2']
    # Non-atomic multi-key operations splits the keys into different slots
    >>> rc.mset_nonatomic({'foo': 'value1', 'bar': 'value2', 'zzz': 'value3')
    >>> rc.mget_nonatomic('foo', 'bar', 'zzz')
    [b'value1', b'value2', b'value3']
```

**Cluster PubSub:**

When a ClusterPubSub instance is created without specifying a node, a single
node will be transparently chosen for the pubsub connection on the
first command execution. The node will be determined by:
 1. Hashing the channel name in the request to find its keyslot
 2. Selecting a node that handles the keyslot: If read_from_replicas is
    set to true, a replica can be selected.
    
*Known limitations with pubsub:*

Pattern subscribe and publish do not work properly because if we hash a pattern 
like fo* we will get a keyslot for that string but there is a endless   
possibilities of channel names based on that pattern that we can’t know in 
advance. This feature is not limited but the commands is not recommended to use 
right now. 
See [redis-py-cluster documentaion](https://redis-py-cluster.readthedocs.io/en/stable/pubsub.html) 
 for more.

``` pycon
    >>> p1 = rc.pubsub()
    # p1 connection will be set to the node that holds 'foo' keyslot
    >>> p1.subscribe('foo')
    # p2 connection will be set to node 'localhost:6379'
    >>> p2 = rc.pubsub(rc.get_node('localhost', 6379))
```

**Read Only Mode**

By default, Redis Cluster always returns MOVE redirection response on accessing 
a replica node. You can overcome this limitation and scale read commands with 
READONLY mode. 

To enable READONLY mode pass read_from_replicas=True to RedisCluster 
constructor. When set to true, read commands will be assigned between the
primary and its replications in a Round-Robin manner. 

You could also enable READONLY mode in runtime by running readonly() method,  
or disable it with readwrite().

``` pycon
    >>> from cluster import RedisCluster as Redis
    # Use 'debug' mode to print the node that the command is executed on
    >>> rc_readonly = Redis(startup_nodes=startup_nodes, 
                    read_from_replicas=True, debug=True)
    >>> rc_readonly.set('{foo}1', 'bar1')
    >>> for i in range(0, 4):
            # Assigns read command to the slot's hosts in a Round-Robin manner
    >>>     rc_readonly.get('{foo}1')
    # set command would be directed only to the slot's primary node
    >>> rc_readonly.set('{foo}2', 'bar2')
    # reset READONLY flag
    >>> rc_readonly.readwrite()
    # now the get command would be directed only to the slot's primary node
    >>> rc_readonly.get('{foo}1')
```

See [Redis Cluster tutorial](https://redis.io/topics/cluster-tutorial) and
[Redis Cluster specifications](https://redis.io/topics/cluster-spec)
to learn more about Redis Cluster.


docker/base/create_cluster.sh
```diff
@@ -0,0 +1,21 @@
+#! /bin/bash
+mkdir -p /nodes
+echo -n > /nodes/nodemap
+for PORT in $(seq 16379 16384); do
+  mkdir -p /nodes/$PORT
+  if [[ -e /redis.conf ]]; then
```
chayim, line 6: Why would /redis.conf be on the filesystem? If we want to include an override - I'd argue for just using the repo.

docker/cluster/redis.conf
```diff
@@ -0,0 +1,3 @@
+# Redis Cluster config file will be shared across all nodes.
+# Dont pass node-unique arguments (e.g. port, dir).
```
chayim, line 2:
```
Don't
```

docker/base/create_cluster.sh
```diff
@@ -0,0 +1,21 @@
...
+  else
+    touch /nodes/$PORT/redis.conf
+  fi
+  cat << EOF >> /nodes/$PORT/redis.conf
+port $PORT
+daemonize yes
+logfile /redis.log
+dir /nodes/$PORT
+EOF
+  redis-server /nodes/$PORT/redis.conf
```
chayim, line 17: Suggestion: Validate that the redis-server comes up properly:

```
if [ $? -ne 0 ]; then
    echo "Redis failed to start, exiting."
    exit 3
fi

File: docker/base/create_cluster.sh [hunk 4]
```diff
@@ -0,0 +1,21 @@
+#! /bin/bash
+mkdir -p /nodes
+echo -n > /nodes/nodemap
```
chayim, line 3: Why echo without a newline? Perhaps you can achieve the same thing with touch?

```
touch /nodes/nodemap
```

docker/base/create_cluster.sh
chayim: Perhaps we should also run redis with nohup?

docker/base/create_cluster.sh
```diff
@@ -0,0 +1,21 @@
...
+  cat << EOF >> /nodes/$PORT/redis.conf
+port $PORT
+daemonize yes
+logfile /redis.log
+dir /nodes/$PORT
+EOF
+  redis-server /nodes/$PORT/redis.conf
+  echo 127.0.0.1:$PORT >> /nodes/nodemap
+done
+echo yes | redis-cli --cluster create $(seq -f 127.0.0.1:%g 16379 16384) --cluster-replicas 1
```
chayim, line 20: This is awesome! :rocket:

tasks.py
```diff
@@ -40,7 +40,10 @@ def tests(c):
     """Run the redis-py test suite against the current python,
     with and without hiredis.
     """
+    print("Starting Redis tests")
     run("tox -e plain -e hiredis")
+    print("Starting RedisCluster tests")
+    run("tox -e plain -e hiredis -- --redis-url=redis://localhost:16379/0")
```
chayim, line 46: Can we allow for an environment override. Currently we allow the override in pyenv, and tox - I'm thinking something like:

```
    REDIS_URL = os.getenv("REDIS_URL", "redis://localhost:16379/0")
    run("tox -e plain -e hiredis -- --redis-url={}".format(REDIS_URL)
```

redis/exceptions.py
```diff
@@ -84,3 +84,67 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
     were sent to the AUTH command
     """
     pass
+
+
+class RedisClusterException(Exception):
```
chayim, line 89: Docstring please

redis/exceptions.py
```diff
@@ -84,3 +84,67 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
...
+    pass
+
+
+class ClusterError(RedisError):
```
chayim, line 93: Docstring please

redis/exceptions.py
```diff
@@ -84,3 +84,67 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
...
+    pass
+
+
+class ClusterDownError(ClusterError, ResponseError):
```
chayim, line 97: Docstring please

redis/exceptions.py
```diff
@@ -84,3 +84,67 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
...
+        """should only redirect to master node"""
+        self.args = (resp,)
+        self.message = resp
+        slot_id, new_node = resp.split(' ')
+        host, port = new_node.rsplit(':', 1)
+        self.slot_id = int(slot_id)
+        self.node_addr = self.host, self.port = host, int(port)
+
+
+class TryAgainError(ResponseError):
```
chayim, line 124: Docstring please

redis/exceptions.py
```diff
@@ -84,3 +84,67 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
...
+
+    def __init__(self, *args, **kwargs):
+        pass
+
+
+class ClusterCrossSlotError(ResponseError):
```
chayim, line 130: Docstring please

redis/exceptions.py
```diff
@@ -84,3 +84,67 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
...
+    message = "Keys in request don't hash to the same slot"
+
+
+class MovedError(AskError):
+    pass
```
chayim, line 135: Docstring please

redis/exceptions.py
```diff
@@ -84,3 +84,67 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
...
+
+
+class MasterDownError(ClusterDownError):
```
chayim, line 138: Docstring please

tox.ini
```diff
@@ -74,6 +74,21 @@ image = redisfab/lots-of-pythons
 volumes =
     bind:rw:{toxinidir}:/data
 
+[docker:redis_cluster]
+name = redis_cluster
+image = barshaul/redis-py:6.2.6-cluster
+healtcheck_cmd = python -c "import socket;print(True) if 0 == socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect_ex(('127.0.0.1',16379)) else False"
```
chayim, line 80: If we're going to use all of these ports - perhaps a healthcheck that checks the entire cluster is up is more useful...

redis/commands/__init__.py
```diff
@@ -2,9 +2,13 @@
 from .redismodules import RedisModuleCommands
 from .helpers import list_or_args
 from .sentinel import SentinelCommands
+from .cluster import ClusterCommands
```
chayim, line 5: Let's sort the imports

redis/commands/__init__.py
```diff
@@ -2,9 +2,13 @@
...
+from .parser import CommandsParser
 
 __all__ = [
     'CoreCommands',
+    'ClusterCommands',
```
chayim, line 10: Let's sort the exports

redis/crc.py
```diff
@@ -0,0 +1,28 @@
...
+    """Calculate key slot for a given key.
+    :param key - bytes
+    :param bucket - int
+    """
+    start = key.find(b"{")
+    if start > -1:
+        end = key.find(b"}", start + 1)
+        if end > -1 and end != start + 1:
+            key = key[start + 1: end]
+    return crc16(key) % bucket
```
chayim, line 28: Since crc16(key) isn't used elsewhere in the codebase, this could change?

```
return crc_hqx(data,0) % bucket
```

chayim: Maybe be kind to the future reader and comment a link to: https://redis.io/topics/cluster-spec

Specifically the piece in the doc: 'HASH_SLOT = CRC16(key) mod 16384' is probably of interest.

redis/client.py
```diff
@@ -475,8 +476,13 @@ def _parse_node_line(line):
 
 
 def parse_cluster_nodes(response, **options):
-    raw_lines = str_if_bytes(response).splitlines()
-    return dict(_parse_node_line(line) for line in raw_lines)
+    """
```
chayim, line 479: Awesome. Thank you!

redis/client.py
```diff
@@ -824,7 +847,7 @@ def __init__(self, host='localhost', port=6379,
                  ssl_check_hostname=False,
                  max_connections=None, single_connection_client=False,
                  health_check_interval=0, client_name=None, username=None,
-                 retry=None):
+                 retry=None, redis_connect_func=None):
```
chayim, line 850: Note to self: This is really getting out of control. Far too many items in init now.

redis/connection.py
```diff
@@ -567,6 +570,9 @@ def register_connect_callback(self, callback):
     def clear_connect_callbacks(self):
         self._connect_callbacks = []
 
+    def set_parser(self, parser_class):
```
chayim, line 573: Docstring.

tests/conftest.py
```diff
@@ -28,6 +30,12 @@ def pytest_addoption(parser):
                           " with loaded modules,"
                           " defaults to `%(default)s`")
 
+    parser.addoption('--cluster-nodes', default=default_cluster_nodes,
```
chayim, line 33: **Nit:** Since all our added options start with redis, might as well do --redis-cluster-nodes to match.

tests/conftest.py
```diff
@@ -41,14 +49,37 @@ def pytest_sessionstart(session):
     info = _get_info(redis_url)
     version = info["redis_version"]
     arch_bits = info["arch_bits"]
+    cluster_enabled = info["cluster_enabled"]
     REDIS_INFO["version"] = version
     REDIS_INFO["arch_bits"] = arch_bits
+    REDIS_INFO["cluster_enabled"] = cluster_enabled
```
chayim, line 55: Awesome

tests/conftest.py
```diff
@@ -41,14 +49,37 @@ def pytest_sessionstart(session):
...
     redismod_url = session.config.getoption("--redismod-url")
     info = _get_info(redismod_url)
     REDIS_INFO["modules"] = info["modules"]
 
+    if cluster_enabled:
+        cluster_nodes = session.config.getoption("--cluster-nodes")
+        wait_for_cluster_creation(redis_url, cluster_nodes)
+
+
+def wait_for_cluster_creation(redis_url, cluster_nodes, timeout=20):
```
chayim, line 67: Docstring on new function please.

tests/conftest.py
```diff
@@ -41,14 +49,37 @@ def pytest_sessionstart(session):
...
+    now = time.time()
+    timeout = now + timeout
```
chayim, line 69: Mislabelled variable. Timeout is the amount of time to wait (i.e in seconds), whereas this is something liek a wait until.... This affects the while loop as well.

tests/conftest.py
```diff
@@ -41,14 +49,37 @@ def pytest_sessionstart(session):
...
+          format(cluster_nodes))
+    while now < timeout:
+        try:
+            client = redis.RedisCluster.from_url(redis_url)
+            if len(client.get_nodes()) == cluster_nodes:
+                print("All nodes are available!")
+                break
+        except RedisClusterException:
+            pass
+        time.sleep(1)
```
chayim, line 80: At some point, IMHO you probably want to raise a RedisClusterException if things don't't come up and we've made it all the way through the while loop.

tests/conftest.py
```diff
@@ -100,10 +142,14 @@ def _get_client(cls, request, single_connection_client=True, flushdb=True,
         redis_url = request.config.getoption("--redis-url")
     else:
         redis_url = from_url
-    url_options = parse_url(redis_url)
-    url_options.update(kwargs)
-    pool = redis.ConnectionPool(**url_options)
-    client = cls(connection_pool=pool)
+    if REDIS_INFO["cluster_enabled"]:
```
chayim, line 145: **Nit** we should default to the non-cluster case, then the cluster case - since it's currently the larger use case.

tests/conftest.py
```diff
@@ -116,7 +162,10 @@ def teardown():
                     # just manually retry the flushdb
                     client.flushdb()
             client.close()
-            client.connection_pool.disconnect()
+            if REDIS_INFO["cluster_enabled"]:
+                client.disconnect_connection_pools()
```
chayim, line 166: Same - assume the non-clustered case as the default. But this disconnect is *GREAT*

redis/commands/parser.py
```diff
@@ -0,0 +1,108 @@
...
+            keys_pos = list(range(command['first_key_pos'], last_key_pos + 1,
+                                  command['step_count']))
+            keys = [args[pos] for pos in keys_pos]
+
+        return keys
+
+    def _get_moveable_keys(self, redis_conn, *args):
+        try:
+            pieces = []
+            cmd_name = args[0]
```
chayim, line 71: pieces
cmd_name

Belong outside of this try/catch block - epespecially since you're only catching a ResponseError.

redis/commands/parser.py
```diff
@@ -0,0 +1,108 @@
...
+            for arg in cmd_name.split():
```
chayim, line 72: Concatenate the lists directly, since pieces is a list.

```
pieces = pieces + cmd_name.split()
```

chayim (changes requested):
Thank you for this amazing contribution!

This PR solidly crashed my browser, everywhere apparently. I'm going to do this in more steps. I still need to read through the list below, and play with the code. For now, the current comments are a great start!

* redis/cluster.py
* redis/commands/cluster.py
* redis/commands/core.py
* redis/client.py
* README.md

redis/client.py
```diff
@@ -475,8 +476,13 @@ def _parse_node_line(line):
 
 
 def parse_cluster_nodes(response, **options):
-    raw_lines = str_if_bytes(response).splitlines()
-    return dict(_parse_node_line(line) for line in raw_lines)
+    """
+    @see: http://redis.io/commands/cluster-nodes  # string
```
chayim, line 480:
```
https://redis.io/commands/cluster-nodes
```
etc

redis/client.py
```diff
@@ -515,6 +521,21 @@ def parse_geosearch_generic(response, **options):
     ]
 
 
+def parse_command(response, **options):
+    commands = {}
+    for command in response:
+        cmd_dict = {}
+        cmd_name = str_if_bytes(command[0])
+        cmd_dict['name'] = cmd_name
+        cmd_dict['arity'] = str_if_bytes(command[1])
```
chayim, line 530: artiy should be a number

docker/base/create_cluster.sh
barshaul: we're starting redis with --daemonize, do you think we should also add nohup?

tasks.py
```diff
@@ -40,7 +40,24 @@ def tests(c):
     """Run the redis-py test suite against the current python,
     with and without hiredis.
     """
-    run("tox -e plain -e hiredis")
+    print("Starting Redis tests")
+    run("tox -e '{redis,cluster}'-'{plain,hiredis}'")
+
+
+@task
+def redis_tests(c):
```
chayim, line 48: This change means you'll need to update the CONTRIBUTING.md as well.

redis/exceptions.py
```diff
@@ -84,3 +84,105 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
...
+    command execution TTL
+    """
+    pass
+
+
+class ClusterDownError(ClusterError, ResponseError):
+    """
+    Error indicated CLUSTERDOWN error received from cluster.
+    By default Redis Cluster nodes stop accepting queries if they detect there
+    is at least an hash slot uncovered (no available node is serving it).
```
chayim, line 108: is at least a hash slot not covered

redis/exceptions.py
```diff
@@ -84,3 +84,105 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
...
+    This way if the cluster is partially down (for example a range of hash
+    slots are no longer covered) all the cluster becomes, eventually,
```
chayim, line 110: slots are no longer covered) the entire cluster eventually becomes unavailable.

redis/exceptions.py
```diff
@@ -84,3 +84,105 @@ class AuthenticationWrongNumberOfArgsError(ResponseError):
...
+    def __init__(self, resp):
+        self.args = (resp,)
+        self.message = resp
+
+
+class AskError(ResponseError):
+    """
+    Error indicated ASK error received from cluster.
+    When a slot is set as MIGRATING, the node will accept all queries that are
+    about this hash slot, but only if the key in question exists, otherwise the
```
chayim, line 123: .... all queries that pertain to this hash slot.

redis/commands/core.py
```diff
@@ -12,14 +12,7 @@
...
 
-class CoreCommands:
-    """
-    A class containing all of the implemented redis commands. This class is
-    to be used as a mixin.
-    """
-
-    # SERVER INFORMATION
-
+class AclCommands:
```
chayim, line 15: Ongoing nit for these classes. If we're going to separate commands by class (which I'm in favour of), then the classes need docstrings.

redis/commands/core.py
```diff
@@ -1813,6 +1837,8 @@ def sunionstore(self, dest, keys, *args):
         args = list_or_args(keys, args)
         return self.execute_command('SUNIONSTORE', dest, *args)
 
+
+class StreamsCommands:
```
chayim, line 1841: StreamCommands

redis/commands/core.py
chayim: Also - while you're here, if we're going to split up the classes, rather than have ~3.5K worth of classes in a single file, let's split the files up (i.e redis.commands.acl contains ACLCommands.

chayim: In cases where something is an acryonym (i.e ACL) we should use ACLCommands. In cases (i.e pubusb) where it's a word, we should stick to what you've done and use PubSubCommands.

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+
+    def __del__(self):
+        self.close()
+
+    def disconnect_connection_pools(self):
+        for node in self.get_nodes():
+            if node.redis_connection:
+                node.redis_connection.connection_pool.disconnect()
```
chayim, line 486: Further into the disconnect code I see that we pass on OSError, but there could be client disconnects here - since this is the furthest things are exposed.

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        # Update the index
+        self.primary_to_idx[primary] = (server_index + 1) % list_size
+        return server_index
+
+    def reset(self):
+        self.primary_to_idx.clear()
+
+
+class NodesManager:
+    def __init__(self, startup_nodes, from_url=False,
```
chayim, line 963: Shouldn't from_url come first? If you think about it, it's likely that a user passes a string-based connection and then wants to populate the rest, as opposed to start_nodes and a from_url. 
If we take it a step further, perhaps both have no value, then it's checked and we raise an ValueError if things aren't set accordingly.

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        self.read_load_balancer = LoadBalancer()
+        if lock is None:
+            lock = threading.Lock()
+        self._lock = lock
+        self.initialize()
+
+    def get_node(self, host=None, port=None, node_name=None):
+        if node_name is None and (host is None or port is None):
+            warnings.warn(
+                "get_node requires one of the followings: "
```
chayim, line 984: get_node requires one of the following

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+                "1. node name "
+                "2. host and port"
+            )
+            return None
+        if host is not None and port is not None:
```
chayim, line 989: Can we do something to reduce the condition checks. First we're checking the node name and the host/port combination. Then for none, then the host.

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        """
+        if self._moved_exception:
+            with self._lock:
+                if self._moved_exception:
+                    self._update_moved_slots()
+
+        if self.slots_cache.get(slot) is None or \
+                len(self.slots_cache[slot]) == 0:
+            raise SlotNotCoveredError(
+                'Slot "{0}" not covered by the cluster. '
```
chayim, line 1043: 'Slot "{0}" is not covered by the cluster. '

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+                '"require_full_coverage={1}"'.format(
+                    slot, self._require_full_coverage)
+            )
+
+        if read_from_replicas:
```
chayim, line 1048: if read_from_replicas is True

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+                )
+
+        # at least one node should have cluster-require-full-coverage yes
+        return any(node_require_full_coverage(node)
+                   for node in cluster_nodes.values())
+
+    def check_slots_coverage(self, slots_cache):
+        # Validate if all slots are covered or if we should try next
+        # startup node
+        for i in range(0, REDIS_CLUSTER_HASH_SLOTS):
```
chayim, line 1109:
```
if slots_cache not in range(0, REDIS_CLUSTER_HASH_SLOTS):
    return False
return True
```

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+                    **self.connection_kwargs,
+                )
+
+    def create_redis_node(self, host, port, **kwargs):
+        if self.from_url:
+            # Create a redis node with a costumed connection pool
+            kwargs.update({"host": host})
+            kwargs.update({"port": port})
+            connection_pool = ConnectionPool(**kwargs)
+            r = Redis(
```
chayim, line 1132: Sug: remove the connection_pool assignment and initialize the class with it
```
r = Redis(connection_pool=ConnectionPool(**kwargs))
```

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+                    copy_kwargs = copy.deepcopy(kwargs)
+                    copy_kwargs.update({"decode_responses": True})
+                    copy_kwargs.update({"encoding": "utf-8"})
+                    r = self.create_redis_node(
+                        startup_node.host, startup_node.port, **copy_kwargs)
+                    self.startup_nodes[startup_node.name].redis_connection = r
+                cluster_slots = r.execute_command("CLUSTER SLOTS")
+                startup_nodes_reachable = True
+            except (ConnectionError, TimeoutError):
+                continue
```
chayim, line 1171: Good example of where I'd like the logging restored - or at least added. A user has no idea otherwise.

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        kwargs = self.connection_kwargs
+        for startup_node in self.startup_nodes.values():
+            try:
+                if startup_node.redis_connection:
+                    r = startup_node.redis_connection
+                else:
+                    # Create a new Redis connection and let Redis decode the
+                    # responses so we won't need to handle that
+                    copy_kwargs = copy.deepcopy(kwargs)
+                    copy_kwargs.update({"decode_responses": True})
```
chayim, line 1163: You can mix updates in a single statement

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+                    'server: {0}. error: {1}'.format(
+                        startup_node, message)
+                )
+
+            # If there's only one server in the cluster, its ``host`` is ''
+            # Fix it to the host in startup_nodes
+            if (len(cluster_slots) == 1
+                    and len(cluster_slots[0][2][0]) == 0
+                    and len(self.startup_nodes) == 1):
+                cluster_slots[0][2][0] = startup_node.host
```
chayim, line 1200: Please add comment as to why [0][2][0] ==

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+            raise RedisClusterException(
+                "Redis Cluster cannot be connected. Please provide at least "
+                "one reachable node. "
+            )
+
+        # Create Redis connections to all nodes
+        self.create_redis_connections(list(tmp_nodes_cache.values()))
+
+        fully_covered = self.check_slots_coverage(tmp_slots)
+        if not fully_covered:
```
chayim, line 1261: Combine the ifs
```
if not fully_covered and self._require_full_coverage
```

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+                        len(self.slots_cache), REDIS_CLUSTER_HASH_SLOTS)
+                )
+            else:
+                # The user set require_full_coverage to False.
+                # In case of full coverage requirement in the cluster's Redis
+                # configurations, we will raise an exception. Otherwise, we may
+                # continue with partial coverage.
+                # see Redis Cluster configuration parameters in
+                # https://redis.io/topics/cluster-tutorial
+                if not self._skip_full_coverage_check and \
```
chayim, line 1277: This needs to change as a result of ^^^^

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        self.populate_startup_nodes(self.nodes_cache.values())
+
+    def close(self):
+        for node in self.nodes_cache.values():
+            if node.redis_connection:
+                node.redis_connection.close()
+
+    def reset(self):
+        if self.read_load_balancer is not None:
+            self.read_load_balancer.reset()
```
chayim, line 1302: It's probably cheaper to run a try/catch here if the read_load_balancer is None instead. Python's try/catch is cheaper. This also follows the python rule "ask for forgiveness, not permission".

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+         2. Selecting a node that handles the keyslot: If read_from_replicas is
+            set to true, a replica can be selected.
+        """
+        self.node = None
+        connection_pool = None
+        if host is not None and port is not None:
+            node = redis_cluster.get_node(host=host, port=port)
+            self.node = node
+        if node is not None:
+            if not isinstance(node, ClusterNode):
```
chayim, line 1330: You can just check isinstance, even with node not being None. This will make this more performant.

```
if isinstance(node, ClusterNode) is not ClusterNode:
    raise DataError("'node' must be a ClusterNode")
```

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+    def __nonzero__(self):
+        "Pipeline instances should  always evaluate to True on Python 2.7"
+        return True
+
+    def __bool__(self):
+        "Pipeline instances should  always evaluate to True on Python 3+"
+        return True
+
+    def execute_command(self, *args, **kwargs):
+        """
```
chayim, line 1456: Missing docstring

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        """
+        return self.pipeline_execute_command(*args, **kwargs)
+
+    def pipeline_execute_command(self, *args, **options):
+        """
```
chayim, line 1461: Missing docstring

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        """
+        self.command_stack.append(
+            PipelineCommand(args, options, len(self.command_stack)))
+        return self
+
+    def raise_first_error(self, stack):
+        """
```
chayim, line 1468: Missing docstring

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        """
+        for c in stack:
+            r = c.result
+            if isinstance(r, Exception):
+                self.annotate_exception(r, c.position + 1, c.args)
+                raise r
+
+    def annotate_exception(self, exception, number, command):
+        """
```
chayim, line 1477: Missing docstring

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        """
+        cmd = ' '.join(map(safe_str, command))
+        msg = 'Command # %d (%s) of pipeline caused error: %s' % (
+            number, cmd, exception.args[0])
+        exception.args = (msg,) + exception.args[1:]
+
+    def execute(self, raise_on_error=True):
+        """
```
chayim, line 1485: Missing docstring

redis/cluster.py
```diff
@@ -0,0 +1,1884 @@
...
+        """
+        stack = self.command_stack
+
+        if not stack:
+            return []
+
+        try:
```
chayim, line 1492: Shouldn't send_cluster_commands operate based on the value of the stack... so the if statement above belongs in here?

chayim (changes requested):
Comments inline!

redis/cluster.py
barshaul: It will not give us the required result. We want to check if all slots from 0 to REDIS_CLUSTER_HASH_SLOTS are found in slots_cache, what you suggests mean that if we'll have slots not in the above range we'll get false, but we can have only one slot in slots_cache and it will return true...

tasks.py
barshaul: updated

redis/cluster.py
barshaul: added a try except

redis/cluster.py
barshaul: If the user creates an instance with redisCluster.from_url(), then the url is being parsed and processed to a startup node in the RedisCluster __init__ method, which then passes the startup node to the nodeManager.
So, no matter what - the nodes manager excepts to get startup_nodes. from_url bool is only there to indicate if we need to create all the other cluster's Redis nodes with a costumed connection pool. See NodesManager::create_redis_node

Issue #1664, opened 2021-11-01
pexpire api inconsistency


WisdomPill:
Thanks for wanting to report an issue you've found in redis-py. Please delete this text and fill in the template below.  
It is of course not always possible to reduce your code to a small test case, but it's highly appreciated to have as much data as possible. Thank you!

**Version**: What redis-py and what redis version is the issue happening on? potentially all

**Platform**: What platform / version? (For example Python 3.5.1 on Windows 7 / Ubuntu 15.10 / Azure) potentially all 

**Description**: Description of your issue, stack traces from errors and code that reproduces the issue

`expire` and `pexpire` behave differently like many commands that return 0 or 1, the response is parsed into a boolean.

In the case of `expire` it is parsed as boolean, while with `pexpire` not, even if they are basically doing the same.

Thanks to @puittenbroek who discovered the issue in https://github.com/jazzband/django-redis/pull/565#discussion_r740043060

It is a very quick fix, I would love to start contributing fixing it.

chayim:
@WisdomPill As always, thanks a tonne for reporting this - including the helpful link to the django-redis issue. 

Uhoh, be careful what you wish for! Seriously though, I assigned it to you, since you asked ;). redis/client.py would use the changes in the RESPONSE_CALLBACKS. Those two commands don't currently have int mappings.

I'm hip deep in tests and other things. But - if you think it'll take more than a week, I'm happy to grab it back from you. Otherwise, frankly, I love sharing the load. Bring it!!!

PR #1665, opened 2021-11-01
Added boolean parsing to PEXPIRE and PEXPIREAT

WisdomPill:
### Pull Request check-list

### Description of change

Fixes: #1664 

`pexpire` and `pexpire_at` will behave like their non p* counterpart parsing the return value into a boolean

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -618,7 +618,8 @@ class Redis(RedisModuleCommands, CoreCommands, object):
     """
     RESPONSE_CALLBACKS = {
         **string_keys_to_dict(
-            'AUTH COPY EXPIRE EXPIREAT PEXPIRE PEXPIREAT HEXISTS HMSET LMOVE BLMOVE MOVE '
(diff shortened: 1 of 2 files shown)


WisdomPill:
@chayim Where should I put the documentation?

I guess some more tests need to have `is True` and `is False` instead of the pain assert which succeeds if a number 0 or 1 is returned.

chayim:
The docs IMHO don't need to change since we're fixing a bug, not changing behaviour. But when merging I'll add the breakingchange label, so that we do the right thing by the community.
Once changes are in and tests pass, we're copacetic.

WisdomPill:
May I add `is True` in other tests testing for boolean value?

chayim:
@WisdomPill I'm a big fan of the is True/is False/is None checks as opposed to the older style. I think it's more readable - and this always tends to be my PR comment.  As for this, just awesome - thank you so much. Merging in!

WisdomPill:
I will open a PR to address it then

PR #1671, opened 2021-11-08
Export Sentinel, and SSL like other classes

terencehonles:
### Pull Request check-list

### Description of change

Now that `Sentinel` needs to be used specifically instead of via `Redis`, this change exports the class at the top level like the `Redis` client is, and also allows the shorthand of `ssl=True` to set the connection class to the `SentinelManagedSSLConnection`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,12 +365,24 @@ Connecting redis-py to the Sentinel instance(s) is easy. You can use a
 Sentinel connection to discover the master and slaves network addresses:
 
 ``` pycon
->>> from redis.sentinel import Sentinel
+>>> from redis import Sentinel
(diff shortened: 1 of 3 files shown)


terencehonles:
> Now that `Sentinel` needs to be used specifically instead of via `Redis`, this change exports the class at the top level like the `Redis` client is

The export may not be the issue I'm actually trying to fix. It looks like the project I'm contributing to does import `Sentinel` from the full path. The `ssl` change is complimentary to #1419, and I can back out the re-export if needed. I'm going to continue trying to track down why the code I'm looking at is breaking with the more recent changes.

chayim:
Thanks @terencehonles I fully agree, we should be exposing this. I hope it fixes your bug. Thank you so much for the contribution!

PR #1721, opened 2021-11-17
removing hiredis warning

chayim:


pcmoritz:
Would it also make sense to remove the warning in https://github.com/redis/redis-py/pull/1721/files#diff-45b6ba995a827b484d3ef97cb806d336bf058ff9678b04cf371200ea53e584d0L76 ? We have been running into problems with hiredis on Apple Silicon (M1) and Windows, so for us it is preferable to run the redis module without hiredis.

PR #1730, opened 2021-11-21
Fixing deprecating distutils (PEP 632) 

### LINKED ISSUE #1729: Python 3.10 distutils Deprecation Warning
**Version**: Redis-py 4.0.1

**Platform**: Mac OS

**Description**: In Python 3.10 there is a deprecation warning: `DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives`, which is raised from using distutils in the Redis connection module: https://github.com/redis/redis-py/blob/master/redis/connection.py#L1. 

This breaks our test suite, and we can silence the warning for now, but perhaps it makes sense to migrate.

chayim:
### Pull Request check-list

### Description of change

fixes #1729


elderlabs:
Would there be any benefit to using [setuptools](https://pypi.org/project/setuptools/) version comparison [implementation](https://www.python.org/dev/peps/pep-0386/#setuptools), as many projects already require it to install other deps, instead of adding `packaging` as a requirement for this sole purpose?

chayim:
I thought about it - as I definitely prefer to **not** have any dependencies (in fact, I intend to remove the deprecation dep in the future) for this type of library.  There are two issues here:

1. PEP 632 specifically refers to packaging (find, we can ignore this)
2. Setuptools version comparison is broken for our needs- in that it fails to parse the third period (part of PEP-0440)

Here's an example of #2
```
from pkg_resources import parse_version as V
v456 = V('4.5.6')
v457prev = V('4.5.7preview')
v457prev2 = V('4.5.7.preview')
v457 = V('4.5.7')

v457prev > v456
>>> True

v457 > v456
>>> True

v457prev2 > v456
>>> False
```

Issue #1720, opened 2021-11-17
Race condition in handling health checks for pub-sub


bmerry:
**Version**: What redis-py and what redis version is the issue happening on?
redis-py 4.0.0, redis 6.2.6 (have also reproduced a failure in redis-py 3.5.3, although with a different failure mode)

**Platform**: What platform / version? (For example Python 3.5.1 on Windows 7 / Ubuntu 15.10 / Azure)
Python 3.8, Ubuntu 20.04

**Description**: 
`PubSub.execute_command` checks `self.subscribed` before deciding whether to issue a health check, but if we've issued an unsubscribe to the last channel but not yet received the unsubscribe confirmation to the server, `self.subscribed` will still be true. The response to that (a string "PONG") can then be misinterpreted as a pub-sub response. This shows up as `get_message` returning `{'type': 80, 'pattern': None, 'channel': 79, 'data': 78}`, where 80, 79, 78 are the ASCII codes of P, O, N (they're integers because indexing a `bytes` gives integers).

With redis-py 3.5.3 I think I hit the opposite problem, where a 'PING redis-py-health-check' is issued thinking that the connection has subscriptions, and expecting a multi-bulk response ["ping", "redis-py-health-check"], but getting just "redis-py-health-check" back on the wire because there are no subscriptions. I *think* I once saw the same thing happen with redis-py 4.0.0, but I've been unable to reproduce it since and I have been jumping between the versions to compare, so I may be mistaken.

Here is some minimal code that illustrates the issue. It doesn't reliably reproduce it - I have to run it in a loop and can take 5-20 times to reproduce. When it does, it prints the following. I haven't dug into the backtrace and it might be just something in my own code - it's the message with numbers in the fields that is the smoking gun.

I think both cases can be dealt with by ignoring responses that are a single string instead of a multi-bulk when trying to parse a pubsub message, although it is a bit hacky.

```python
#!/usr/bin/env python3

import threading
import time

import redis


def poll(ps):
    while True:
        message = ps.get_message(timeout=5)
        if message is not None:
            print(message)
        else:
            break


def main():
    r = redis.Redis.from_url("redis://localhost", health_check_interval=1)
    ps = r.pubsub()
    ps.subscribe("foo")
    poller = threading.Thread(target=poll, args=(ps,))
    poller.start()
    ps.unsubscribe("foo")
    time.sleep(3)
    ps.subscribe("baz")
    poller.join()


main()
```

Output:
```
[...]

PR #1733, opened 2021-11-21
Changed the PubSub's health check command to be performed only on the…

barshaul:
… first command execution.

### Pull Request check-list

- [V ] Does `$ tox` pass with this change (including linting)?
- [V ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ V] Is the new or changed code fully tested?
- [ V] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change

Closing #1720:

I reproduced the PubSub bug described in #1720 and was able to find the RC.
When the PubSub's 'execute_command' method is being called it passes a 'health_check' bool to determine if it needs to run a health check. The 'health_check' value is set to **not self.subscribed**, which checks if the pubsub instance has any items in the channels/patterns lists. That means, that we perform a health check within the execute_command function only if we are not yet subscribed. All subsequent commands, after the first subscription, should be executed without performing a health check, since the channels/patterns list is no longer empty.
The pubsub's 'get_messages()' method can be used to poll published messages after a pubsub instance has been created. If a poller thread is created (thread that waits on get_message()) it will listen on the same socket as the pubsub execute_command is listening to when it performs a health check. Hence, we should not send a healthcheck using the pubsub execute_command function after the poller thread is initiated, since then it will be racing the poller thread to read the response from the socket.

In the example in #1720 we see the following flow:

1. PubSub instance is being created
2. Channel 'foo' is being subscribed - health check is performed since self.channels is empty
3. 'PONG' is received 
4. A poller thread is being started, looping on 'get_messages()'
5. The poller thread polls the 'subscribe' response
6. Channel 'foo' is being unsubscribed, health check is not being performed since self.channels still contains 'foo'. 'foo' is removed from self.channels
7. The poller thread polls the 'unsubscribe' response
8. Channel 'baz' is being subscribed - health check is performed since self.channels is empty again
9. The poller thread tries to poll the 'subscribe' response and gets the 'PONG' response instead
10. The health check waits for a PONG response that has already been received by the poller thread, and is therefore timed out

Therefore, we shouldn't use self.channels and self.patterns to determine whether a health check needs to be executed, but we should have another variable to indicate whether this is the first command execution, and if so, to run a health check. 

However, a poller thread may be started before subscribing to a channel, e.g. :

1. ps = r.pubsub()
2. poller = threading.Thread(target=poll, args=(ps,))
3. ps.subscribe('foo')

In this case, the health check will be performed and we will still get a race reading from the socket with the poller thread. 
So, my suggestion is to add a new 'cmd_execution_health_check' variable initiated with 'True' to the pubsub class and to set it to False on: 
1. The end of execute_command method, so the health check will be performed only on the first execution), or
2. get_message() function, so the health check will not be performed from the execute_command function at all.
health checks are being done by the get_message() method, so no need to execute it also from the main command execution.  

This change fixes the reported bug.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1415,10 +1415,6 @@ class PubSub:
         before returning. Timeout should be specified as a floating point
         number.
         """
-        if self.cmd_execution_health_check is True:
-            # Health checks will be done within the parse_response method,
-            # cancel health checks from the command_execution method
-            self.cmd_execution_health_check = False
         response = self.parse_response(block=False, timeout=timeout)
         if response:
             return self.handle_message(response, ignore_subscribe_messages)
--- a/redis/client.py
+++ b/redis/client.py
@@ -1215,6 +1215,7 @@ class PubSub:
         self.pending_unsubscribe_channels = set()
         self.patterns = {}
         self.pending_unsubscribe_patterns = set()
+        self.cmd_execution_health_check = True
 
     def close(self):
         self.reset()
@@ -1258,8 +1259,11 @@ class PubSub:
             # were listening to when we were disconnected
             self.connection.register_connect_callback(self.on_connect)
         connection = self.connection
-        kwargs = {'check_health': not self.subscribed}
+        kwargs = {'check_health': self.cmd_execution_health_check}
         self._execute(connection, connection.send_command, *args, **kwargs)
+        if self.cmd_execution_health_check is True:
+            # Run a health check only on the first command execution
+            self.cmd_execution_health_check = False
 
     def _disconnect_raise_connect(self, conn, error):
         """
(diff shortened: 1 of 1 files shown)


chayim:
@barshaul Thanks for contributing a great fix! I ran this overnight and there were zero failures after ~7500 runs. This definitely feels fixed!

bmerry:
> but you may want to see @bmerry comment [here](https://github.com/redis/redis-py/issues/1720#issuecomment-974839627)

FYI my comment was based on the description of the fix - I'd missed that there was a PR implementing the description.

bmerry:
Arguably there is a race condition where a thread might `get_message` immediately after `self.connection` is initialised in `execute_command` but before the command is actually sent. But such a program already has a race condition, since it can't guarantee that `get_message` won't run before `self.connection` is initialised and raise an error. So this seems safe.

It potentially makes the health check less useful though. If the pubsub is used for a bit, then all the subscriptions are removed, then it is used again 10 minutes later, there will be no health check at that point. Maybe that's a reasonable trade-off for correctness.

barshaul:
See #1737 for the closing PR/

PR #1737, opened 2021-11-22
Fixing read race condition during pubsub

### LINKED ISSUE #1720: Race condition in handling health checks for pub-sub

### LINKED ISSUE #1740: Another race condition in health checks and pubsub
**Version**: redis-py 4.0.2, Redis 6.2.6. Also reproduces with #1733 or #1737 applied.

**Platform**: Python 3.8, Ubuntu 20.04

**Description**: This is somewhat similar to #1720, but different. If `get_message` is called between issuing an UNSUBSCRIBE command that removes that last subscription, and getting the unsubscribe message back, and `parse_response` decides it needs to do a health check, then the reply to the health check is mis-interpreted. This is because redis responds differently to PING depending on whether there are active subscriptions or not, and `parse_response` only recognises the one form.

aioredis has the [same bug](https://github.com/aio-libs/aioredis-py/issues/1206), and the [fix](https://github.com/aio-libs/aioredis-py/pull/1207) (recognise both possible responses) would probably work for redis-py too.

Code to reproduce the issue (it's a race condition so might not work every time, but it's been reliable for me). Depends on a redis server on localhost:
```python
#!/usr/bin/env python3

import threading
import time

import redis


def poll(ps, event):
    print(ps.get_message(timeout=5))
    event.wait()
    while True:
        message = ps.get_message(timeout=5)
        if message is not None:
            print(message)
        else:
            break


def main():
    r = redis.Redis.from_url("redis://localhost", health_check_interval=1)
    ps = r.pubsub()
    ps.subscribe("foo")

    event = threading.Event()
    poller = threading.Thread(target=poll, args=(ps, event))
    poller.start()

    time.sleep(2)
    event.set()
    ps.unsubscribe("foo")
    poller.join()


main()
```

It subscribes, then waits 2 seconds (to make the health check interval kick in), then races to call `get_message` and `unsubscribe`.

Output:
```
{'type': 'subscribe', 'pattern': None, 'channel': b'foo', 'data': 1}
{'type': 'unsubscribe', 'pattern': None, 'channel': b'foo', 'data': 0}
{'type': 114, 'pattern': None, 'channel': 101, 'data': 100}
```

The last line is the problem. It's caused by trying to treat a byte string (b"redis-py-health-check") as a list and selecting elements from it (114, 101, 100 are the ASCII codes for 'r', 'e', 'd').

barshaul:
### Pull Request check-list

- [X ] Does `$ tox` pass with this change (including linting)?
- [X ] Is the new or changed code fully tested?

### Description of change

closes #1720 
closes #1740 
closes #1733 

Another implementation to #1720 (first impl: #1733)
In this PR i'm adding an option to call pubsub's method get_message() without subscribing first.
If get_message is called and no channel/pattern is subscribed, the method will return None without trying to read from the connection.
When timeout is passed and no channels are yet subscribed, the get_message() function will wait for the first to arrive - either a subscription has been made or the time has expired.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -1276,6 +1276,7 @@ class PubSub:
         self.shard_hint = shard_hint
         self.ignore_subscribe_messages = ignore_subscribe_messages
         self.connection = None
+        self.subscribed_event = threading.Event()
         # we need to know the encoding options for this connection in order
         # to lookup channel and pattern names for callback handlers.
         self.encoder = encoder
@@ -1315,6 +1316,7 @@ class PubSub:
         self.pending_unsubscribe_channels = set()
         self.patterns = {}
         self.pending_unsubscribe_patterns = set()
+        self.subscribed_event.clear()
 
     def close(self):
         self.reset()
@@ -1340,7 +1342,7 @@ class PubSub:
     @property
     def subscribed(self):
         "Indicates if there are subscriptions to any channels or patterns"
-        return bool(self.channels or self.patterns)
+        return self.subscribed_event.is_set()
 
     def execute_command(self, *args):
         "Execute a publish/subscribe command"
@@ -1443,6 +1445,9 @@ class PubSub:
         # for the reconnection.
         new_patterns = self._normalize_keys(new_patterns)
         self.patterns.update(new_patterns)
+        if not self.subscribed:
+            # Set the subscribed_event flag to True
+            self.subscribed_event.set()
         self.pending_unsubscribe_patterns.difference_update(new_patterns)
         return ret_val
 
@@ -1477,6 +1482,9 @@ class PubSub:
         # for the reconnection.
         new_channels = self._normalize_keys(new_channels)
         self.channels.update(new_channels)
+        if not self.subscribed:
+            # Set the subscribed_event flag to True
+            self.subscribed_event.set()
         self.pending_unsubscribe_channels.difference_update(new_channels)
         return ret_val
 
@@ -1508,29 +1516,24 @@ class PubSub:
         before returning. Timeout should be specified as a floating point
         number.
         """
-        if not self.subscribed and \
-                self.wait_for_subscription(timeout) is False:
-            # The connection isn't subscribed to any channels or patterns, so
-            # no messages are available
-            return None
+        if not self.subscribed:
+            # Wait for subscription
+            deadline = time.time() + timeout
+            if self.subscribed_event.wait(timeout) is True:
+                # The connection was subscribed during the timeout frametime.
+                # The timeout should be adjusted for the time spent waiting
+                # for subscription
+                time_spent = deadline - time.time()
+                timeout = timeout - time_spent
+            else:
+                # The connection isn't subscribed to any channels or patterns,
+                # so no messages are available
+                return None
 
         response = self.parse_response(block=False, timeout=timeout)
         if response:
             return self.handle_message(response, ignore_subscribe_messages)
         return None
-
-    def wait_for_subscription(self, timeout, period=0.25):
-        """
-        Wait until this pubsub connection has been subscribed.
-        Return True if the connection was subscribed during the timeout
-        frametime. Otherwise, return False.
-        """
-        mustend = time.time() + timeout
-        while time.time() < mustend:
-            if self.subscribed:
-                return True
-            time.sleep(period)
-        return False
 
     def ping(self, message=None):
         """
@@ -1580,6 +1583,10 @@ class PubSub:
                 if channel in self.pending_unsubscribe_channels:
                     self.pending_unsubscribe_channels.remove(channel)
                     self.channels.pop(channel, None)
+            if not self.channels and not self.patterns:
+                # There are no subscriptions anymore, set subscribed_event flag
+                # to false
+                self.subscribed_event.clear()
 
         if message_type in self.PUBLISH_MESSAGE_TYPES:
             # if there's a message handler, invoke it
--- a/redis/client.py
+++ b/redis/client.py
@@ -1366,7 +1366,6 @@ class PubSub:
             self.clean_health_check_responses()
         self._execute(connection, connection.send_command, *args, **kwargs)
 
-
     def clean_health_check_responses(self):
         """
         If any health check responses are present, clean them
--- a/redis/client.py
+++ b/redis/client.py
@@ -1480,6 +1480,8 @@ class PubSub:
         if not self.subscribed:
             # Set the subscribed_event flag to True
             self.subscribed_event.set()
+            # Clear the health check counter
+            self.health_check_response_counter = 0
         self.pending_unsubscribe_patterns.difference_update(new_patterns)
         return ret_val
 
@@ -1517,6 +1519,8 @@ class PubSub:
         if not self.subscribed:
             # Set the subscribed_event flag to True
             self.subscribed_event.set()
+            # Clear the health check counter
+            self.health_check_response_counter = 0
         self.pending_unsubscribe_channels.difference_update(new_channels)
         return ret_val
 
--- a/redis/client.py
+++ b/redis/client.py
@@ -1362,7 +1362,21 @@ class PubSub:
             self.connection.register_connect_callback(self.on_connect)
         connection = self.connection
         kwargs = {"check_health": not self.subscribed}
+        if not self.subscribed:
+            self.clean_health_check_responses()
         self._execute(connection, connection.send_command, *args, **kwargs)
+
+
+    def clean_health_check_responses(self):
+        """
+        If any health check responses are present, clean them
+        """
+        conn = self.connection
+        while self._execute(conn, conn.can_read, timeout=0):
+            response = self._execute(conn, conn.read_response)
+            if not self.is_health_check_response(response):
+                raise PubSubError('A non health check response was cleaned by '
+                                  'execute_command: {0}'.format(response))
 
     def _disconnect_raise_connect(self, conn, error):
         """
@@ -1403,17 +1417,21 @@ class PubSub:
             return None
         response = self._execute(conn, conn.read_response)
 
-        if conn.health_check_interval and \
-                response in [
-                self.health_check_response,  # If there was a subscription
-                self.health_check_response_b  # If there wasn't
-                ]:
-            # If there are no subscriptions redis responds to PING command with
-            # a bulk response, instead of a multi-bulk with "pong" and the
-            # response.
+        if self.is_health_check_response(response):
             # ignore the health check message as user might not expect it
             return None
         return response
+
+    def is_health_check_response(self, response):
+        """
+        Check if the response is a health check response.
+        If there are no subscriptions redis responds to PING command with a
+        bulk response, instead of a multi-bulk with "pong" and the response.
+        """
+        return response in [
+                self.health_check_response,  # If there was a subscription
+                self.health_check_response_b  # If there wasn't
+                ]
 
     def check_health(self):
         conn = self.connection
@@ -1529,11 +1547,11 @@ class PubSub:
             # Wait for subscription
             start_time = time.time()
             if self.subscribed_event.wait(timeout) is True:
-                # The connection was subscribed during the timeout frametime.
+                # The connection was subscribed during the timeout time frame.
                 # The timeout should be adjusted based on the time spent
                 # waiting for the subscription
                 time_spent = time.time() - start_time
-                timeout = timeout - time_spent
+                timeout = max(0.0, timeout - time_spent)
             else:
                 # The connection isn't subscribed to any channels or patterns,
                 # so no messages are available
--- a/redis/client.py
+++ b/redis/client.py
@@ -1508,10 +1508,29 @@ class PubSub:
         before returning. Timeout should be specified as a floating point
         number.
         """
+        if not self.subscribed and \
+                self.wait_for_subscription(timeout) is False:
+            # The connection isn't subscribed to any channels or patterns, so
+            # no messages are available
+            return None
+
         response = self.parse_response(block=False, timeout=timeout)
         if response:
             return self.handle_message(response, ignore_subscribe_messages)
         return None
+
+    def wait_for_subscription(self, timeout, period=0.25):
+        """
+        Wait until this pubsub connection has been subscribed.
+        Return True if the connection was subscribed during the timeout
+        frametime. Otherwise, return False.
+        """
+        mustend = time.time() + timeout
+        while time.time() < mustend:
+            if self.subscribed:
+                return True
+            time.sleep(period)
+        return False
 
     def ping(self, message=None):
         """
--- a/redis/client.py
+++ b/redis/client.py
@@ -1370,7 +1370,7 @@ class PubSub:
         ttl = 10
         conn = self.connection
         while self.health_check_response_counter > 0 and ttl > 0:
-            if self._execute(conn, conn.can_read, timeout=1):
+            if self._execute(conn, conn.can_read, timeout=conn.socket_timeout):
                 response = self._execute(conn, conn.read_response)
                 if self.is_health_check_response(response):
                     self.health_check_response_counter -= 1
--- a/redis/client.py
+++ b/redis/client.py
@@ -1282,15 +1282,11 @@ class PubSub:
         self.encoder = encoder
         if self.encoder is None:
             self.encoder = self.connection_pool.get_encoder()
-        self.health_check_response_b = self.encoder.encode(
-            self.HEALTH_CHECK_MESSAGE)
+        self.health_check_response_b = self.encoder.encode(self.HEALTH_CHECK_MESSAGE)
         if self.encoder.decode_responses:
             self.health_check_response = ["pong", self.HEALTH_CHECK_MESSAGE]
         else:
-            self.health_check_response = [
-                b"pong",
-                self.health_check_response_b
-            ]
+            self.health_check_response = [b"pong", self.health_check_response_b]
         self.reset()
 
     def __enter__(self):
@@ -1315,6 +1311,7 @@ class PubSub:
             self.connection_pool.release(self.connection)
             self.connection = None
         self.channels = {}
+        self.health_check_response_counter = 0
         self.pending_unsubscribe_channels = set()
         self.patterns = {}
         self.pending_unsubscribe_patterns = set()
@@ -1370,12 +1367,19 @@ class PubSub:
         """
         If any health check responses are present, clean them
         """
+        ttl = 10
         conn = self.connection
-        while self._execute(conn, conn.can_read, timeout=0):
-            response = self._execute(conn, conn.read_response)
-            if not self.is_health_check_response(response):
-                raise PubSubError('A non health check response was cleaned by '
-                                  'execute_command: {0}'.format(response))
+        while self.health_check_response_counter > 0 and ttl > 0:
+            if self._execute(conn, conn.can_read, timeout=1):
+                response = self._execute(conn, conn.read_response)
+                if self.is_health_check_response(response):
+                    self.health_check_response_counter -= 1
+                else:
+                    raise PubSubError(
+                        "A non health check response was cleaned by "
+                        "execute_command: {0}".format(response)
+                    )
+            ttl -= 1
 
     def _disconnect_raise_connect(self, conn, error):
         """
@@ -1418,6 +1422,7 @@ class PubSub:
 
         if self.is_health_check_response(response):
             # ignore the health check message as user might not expect it
+            self.health_check_response_counter -= 1
             return None
         return response
 
@@ -1428,9 +1433,9 @@ class PubSub:
         bulk response, instead of a multi-bulk with "pong" and the response.
         """
         return response in [
-                self.health_check_response,  # If there was a subscription
-                self.health_check_response_b  # If there wasn't
-                ]
+            self.health_check_response,  # If there was a subscription
+            self.health_check_response_b,  # If there wasn't
+        ]
 
     def check_health(self):
         conn = self.connection
@@ -1442,6 +1447,7 @@ class PubSub:
 
         if conn.health_check_interval and time.time() > conn.next_health_check:
             conn.send_command("PING", self.HEALTH_CHECK_MESSAGE, check_health=False)
+            self.health_check_response_counter += 1
 
     def _normalize_keys(self, data):
         """
--- a/redis/client.py
+++ b/redis/client.py
@@ -1518,12 +1518,12 @@ class PubSub:
         """
         if not self.subscribed:
             # Wait for subscription
-            deadline = time.time() + timeout
+            start_time = time.time()
             if self.subscribed_event.wait(timeout) is True:
                 # The connection was subscribed during the timeout frametime.
-                # The timeout should be adjusted for the time spent waiting
-                # for subscription
-                time_spent = deadline - time.time()
+                # The timeout should be adjusted based on the time spent
+                # waiting for the subscription
+                time_spent = time.time() - start_time
                 timeout = timeout - time_spent
             else:
                 # The connection isn't subscribed to any channels or patterns,
--- a/redis/client.py
+++ b/redis/client.py
@@ -1282,12 +1282,14 @@ class PubSub:
         self.encoder = encoder
         if self.encoder is None:
             self.encoder = self.connection_pool.get_encoder()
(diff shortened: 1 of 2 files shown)


bmerry:
Being able to start polling without making a subscription would be nice - I ran into this limitation recently (although in aioredis). 

An immediate problem I can see with this is that `get_message` can take double the timeout if it first waits just less than `timeout` for a subscription then another `timeout` for a message (plus another 0.25s because of the polling in wait_for_subscription).

I feel like this probably still has a race somewhere, because by the time `wait_for_subscription` returns, the main thread might already have asked to unsubscribe. It may be that works as long as some timing assumptions hold (e.g. round-trip time to the server is less than the health check interval) but breaks if the server suffers high latency. I'll poke at it some more and see if I can produce an explicit example.

bmerry:
Looking at it again, I think the race I was worried about can't happen - provided that there is just one thread calling (un)subscription functions ("main thread") and one thread using `get_message`/`listen` (poller thread). Here's my logic, in case anyone wants to double-check. I consider three possible states:
1. Subscribed: there are subscriptions. `self.subscribed` is true.
2. Semi-subscribed: we've issued UNSUBSCRIBE commands for all subscriptions, but not yet processed the responses. `self.subscribed` is true.
3. Unsubscribed: we've issued UNSUBSCRIBE commands for all subscriptions and processed the responses. `self.subscribed` is false.

When `get_message` passes the wait_for_subscribed` check, we're in either subscribed or semi-subscribed state. The main thread can trigger oscillations between these states, but cannot cause a transition to unsubscribed on its own (the poller thread does that in `handle_message` by processing the unsubscription response). So when `get_message` is reading the socket, we are guaranteed not to be in unsubscribed state. On the other hand, `execute_command` only runs the health check in unsubscribed state, and remains in that state for the duration of the health check.

I think there might still be a bug where `PubSub.check_health` can run in semi-subscribed state, but I'll file that separately if I manage to reproduce it.

redis/client.py
```diff
@@ -1411,11 +1411,30 @@ def get_message(self, ignore_subscribe_messages=False, timeout=0):
         before returning. Timeout should be specified as a floating point
         number.
         """
+        if not self.subscribed and \
+                self.wait_for_subscription(timeout) is False:
```
bmerry, line 1415:
```suggestion
                not self.wait_for_subscription(timeout):
```

redis/client.py
```diff
@@ -1411,11 +1411,30 @@ def get_message(self, ignore_subscribe_messages=False, timeout=0):
...
+            # The connection isn't subscribed to any channels or patterns, so
+            # no messages are available
+            return None
+
         response = self.parse_response(block=False, timeout=timeout)
```
bmerry, line 1420: Need to adjust the timeout for the time spent in `wait_for_subscription`.

redis/client.py
```diff
@@ -1411,11 +1411,30 @@ def get_message(self, ignore_subscribe_messages=False, timeout=0):
...
+        """
+        Wait until this pubsub connection has been subscribed.
+        Return True if the connection was subscribed during the timeout
+        frametime. Otherwise, return False.
+        """
+        mustend = time.time() + timeout
+        while time.time() < mustend:
+            if self.subscribed:
+                return True
+            time.sleep(period)
```
bmerry, line 1435: A `threading.Event` would be cleaner than polling.

redis/client.py
```diff
@@ -1411,11 +1411,30 @@ def get_message(self, ignore_subscribe_messages=False, timeout=0):
...
         return None
 
+    def wait_for_subscription(self, timeout, period=0.25):
+        """
+        Wait until this pubsub connection has been subscribed.
+        Return True if the connection was subscribed during the timeout
+        frametime. Otherwise, return False.
+        """
+        mustend = time.time() + timeout
+        while time.time() < mustend:
```
bmerry, line 1432: `must_end` would be a more readable name (a more standard alternative would be `deadline`).

redis/client.py
barshaul: changed

redis/client.py
barshaul: Changed it to threading.Event and got the wait_for_subscription function

redis/client.py
barshaul: Adjusted on the new commit

redis/client.py
```diff
@@ -1411,6 +1419,20 @@ def get_message(self, ignore_subscribe_messages=False, timeout=0):
...
         number.
         """
+        if not self.subscribed:
+            # Wait for subscription
+            deadline = time.time() + timeout
+            if self.subscribed_event.wait(timeout) is True:
+                # The connection was subscribed during the timeout frametime.
+                # The timeout should be adjusted for the time spent waiting
+                # for subscription
+                time_spent = deadline - time.time()
```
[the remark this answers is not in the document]
barshaul: woops, you're right. i'll change it

barshaul:
Using issue #1740 I found a bug in this fix:
```
#!/usr/bin/env python3

import threading
import time

from redis import Redis


def poll(ps, event=None):
    print(ps.get_message(timeout=5))
    event.wait()
    while True:
        message = ps.get_message(timeout=5)
        if message is not None:
            print(message)
        else:
            break

def main():
    r = Redis.from_url("redis://localhost", health_check_interval=1)
    ps = r.pubsub()
    ps.subscribe("foo")

    event = threading.Event()
    poller = threading.Thread(target=poll, args=(ps, event))
    poller.start()

    time.sleep(2)
    event.set()
    ps.unsubscribe("foo")
    time.sleep(1)
    ps.subscribe("foo")
    poller.join()

while True:
    main()

```

If UNSUBSCRIBE response is received before the PING response received: in this case get_message will poll the unsubscribe response, then it will change the subscribed_event flag to false, and then running get_message again will result in None until a new subscription is made.  However, in this phase, if we call subscribe again, we have the  b"redis-py-health-check" response queued to the socket, so, when we'll run the health check within the execute_command (since now self.subscribed == False), an error will occur: 
redis.exceptions.ConnectionError: Bad response from PING health check
because the response is  b"redis-py-health-check" . 
This issue can be fixed if we'll run the health check from the execute_command method only on the **first** command execution.

I will work on publishing a fix for both.

barshaul:
@bmerry A fix was added for #1740 and to the bug I mentioned in the comment above.

redis/client.py
```diff
@@ -1411,6 +1432,20 @@ def get_message(self, ignore_subscribe_messages=False, timeout=0):
         before returning. Timeout should be specified as a floating point
         number.
         """
+        if not self.subscribed:
+            # Wait for subscription
+            start_time = time.time()
+            if self.subscribed_event.wait(timeout) is True:
```
bmerry, line 1438:
```suggestion
            if self.subscribed_event.wait(timeout):
```

redis/client.py
```diff
@@ -1411,6 +1432,20 @@ def get_message(self, ignore_subscribe_messages=False, timeout=0):
...
+                # The connection was subscribed during the timeout frametime.
```
bmerry, line 1439: What is "frametime"? Do you mean "time frame"?

redis/client.py
```diff
@@ -1411,6 +1432,20 @@ def get_message(self, ignore_subscribe_messages=False, timeout=0):
...
+                # The timeout should be adjusted based on the time spent
+                # waiting for the subscription
+                time_spent = time.time() - start_time
+                timeout = timeout - time_spent
```
bmerry, line 1443: parse_response raises an error when given a negative timeout, which could occur if the event is set just as the timeout expires.
```suggestion
                timeout = max(0.0, timeout - time_spent)
```
Possibly this function should also raise if the initial timeout is negative, so that a negative timeout from the user has consistent behaviour regardless of whether there are subscriptions.

bmerry (commented):
I think the overall approach looks safe now, and I'm unable to crash it with my tests. I've made a suggestion that will fix a new timing-dependent crash.

It is unfortunate that the sequence [subscribe, get_message*, unsubscribe, sleep, subscribe] will no longer benefit from a health check on the later subscribe. In the application where I run into these issues, I have a wrapper class that keeps one permanent PubSub object around and uses it when it wants to wait for some message to be published.

redis/client.py
barshaul: yup. lol. changed

redis/client.py
barshaul: @chayim asked to make statements as explicit as possible

redis/client.py
barshaul: changed

barshaul:
Added a new solution so we could run more than one health check from execute_command.

clean_health_check_responses will be called from exeute_command, before sending the command or initiating a health check, only If (not self.subscribed).  
not self.subscribed can be true in two cases:
1. The first time we subscribe, then we know for sure that the socket is clean and clean_health_check_responses will immediately return
2. In case we subscribe->unsubscribe->subscribe. In this case, just after 'unsubscribed' was executed, there are two options:
    - get_messages() **didn't** execute a health check before pulling the 'unsubscribe' response. it only had the 'unsubscribe' response to pull, then it set is_subscribed to False. If 'subscribe' is being called **before** the 'unsubscribe' response was processed, then is_subscribe will be True, so execute_command will not try to clean the socket. If 'subscribe' is being called **after** the is_subscribed was set to False, execute_command will call clean_health_check_responses and immediately return - since the socket is empty (get_messages() haven't done a health check). Thus execute_command can perform a health check with no problems.
    - get_messages() **did** execute a health check before pulling the 'unsubscribe' response. In this case we have two options for the order of the responses in the socket: 
        1. the health check response is queued before the unsubscribe response. In this case, the health check response will be processed first and the PubSub.parse_response method will ignore it, then it will retrieve the unsubscribe response. 'subscribe' can now execute a health check with no issue. 
        2. the unsubscribe response queued before the health check response. then the get_message() will set is_subscribed event to False and stop pulling messages. so we will have the health check response left unread in the socket. In this stage, if subscribe is executed - self.subscribed is False, and therefore clean_health_check_responses will pull the health_check response and return. Then a health check can be issued by execute_command. 

I couldn't find a scenario in which we'll clean a response that isn't a health check response. However, I added exception throwing in case it did happen, to make debugging easier.
Please verify me and let me know what you think.

barshaul:
> Added a new solution so we could run more than one health check from execute_command.
> 
> clean_health_check_responses will be called from exeute_command, before sending the command or initiating a health check, only If (not self.subscribed). not self.subscribed can be true in two cases:
> 
> 1. The first time we subscribe, then we know for sure that the socket is clean and clean_health_check_responses will immediately return
> 2. In case we subscribe->unsubscribe->subscribe. In this case, just after 'unsubscribed' was executed, there are two options:
>    
>    * get_messages() **didn't** execute a health check before pulling the 'unsubscribe' response. it only had the 'unsubscribe' response to pull, then it set is_subscribed to False. If 'subscribe' is being called **before** the 'unsubscribe' response was processed, then is_subscribe will be True, so execute_command will not try to clean the socket. If 'subscribe' is being called **after** the is_subscribed was set to False, execute_command will call clean_health_check_responses and immediately return - since the socket is empty (get_messages() haven't done a health check). Thus execute_command can perform a health check with no problems.
>    * get_messages() **did** execute a health check before pulling the 'unsubscribe' response. In this case we have two options for the order of the responses in the socket:
>      
>      1. the health check response is queued before the unsubscribe response. In this case, the health check response will be processed first and the PubSub.parse_response method will ignore it, then it will retrieve the unsubscribe response. 'subscribe' can now execute a health check with no issue.
>      2. the unsubscribe response queued before the health check response. then the get_message() will set is_subscribed event to False and stop pulling messages. so we will have the health check response left unread in the socket. In this stage, if subscribe is executed - self.subscribed is False, and therefore clean_health_check_responses will pull the health_check response and return. Then a health check can be issued by execute_command.
> 
> I couldn't find a scenario in which we'll clean a response that isn't a health check response. However, I added exception throwing in case it did happen, to make debugging easier. @bmerry, Please verify me and let me know what you think.

@bmerry, @chayim  I would appreciate a look if you have time.

bmerry:
> @bmerry, @chayim I would appreciate a look if you have time.

I'll take a look now.

redis/client.py
```diff
@@ -1290,8 +1294,21 @@ def execute_command(self, *args):
...
+    def clean_health_check_responses(self):
+        """
+        If any health check responses are present, clean them
+        """
+        conn = self.connection
+        while self._execute(conn, conn.can_read, timeout=0):
+            response = self._execute(conn, conn.read_response)
+            if not self.is_health_check_response(response):
+                raise PubSubError('A non health check response was cleaned by '
+                                  'execute_command: {0}'.format(response))
```
bmerry, line 1310: The idea of this function worries me. It depends on the response arriving in the socket before the function is called. What if the connection is suffering from high latency and the health check response only arrives after this function returns?

What about keeping a counter of outstanding health checks? Increment it when issuing a health check, and decrement it when a response is received (some care might be required to distinguish the two types of health check). Then you will know how many responses need to be cleared out.

barshaul: @bmerry 
I added a health check response counter. I also added a TTL so we won't end up stuck in the cleanup loop if the health check response isn't received for some reason. 
I setup the ttl to 10 seconds (the timeout of each can_read() is 1 second*10 times).
What do you think of this approach?

redis/client.py
```diff
@@ -1358,8 +1359,28 @@ def execute_command(self, *args):
...
         self._execute(connection, connection.send_command, *args, **kwargs)
 
+    def clean_health_check_responses(self):
+        """
+        If any health check responses are present, clean them
+        """
+        ttl = 10
+        conn = self.connection
+        while self.health_check_response_counter > 0 and ttl > 0:
+            if self._execute(conn, conn.can_read, timeout=1):
```
bmerry, line 1373: Should this rather use the `socket_timeout`?

redis/client.py
```diff
@@ -1443,6 +1477,11 @@ def psubscribe(self, *args, **kwargs):
         # for the reconnection.
         new_patterns = self._normalize_keys(new_patterns)
         self.patterns.update(new_patterns)
+        if not self.subscribed:
+            # Set the subscribed_event flag to True
+            self.subscribed_event.set()
+            # Clear the health check counter
+            self.health_check_response_counter = 0
```
bmerry, line 1484: Should this be done at the end of `clean_health_check_response`?

bmerry (commented):
I think the design looks good now. Unfortunately I need to get my own work all sorted out before I go on holiday later this week, so I won't have time to thoroughly test this or to do further reviewing.

redis/client.py
```diff
@@ -1358,8 +1359,28 @@ def execute_command(self, *args):
             self.connection.register_connect_callback(self.on_connect)
         connection = self.connection
         kwargs = {"check_health": not self.subscribed}
+        if not self.subscribed:
```
chayim, line 1362:
```
if self.subscribed is False
```

redis/client.py
chayim: Agreed.

chayim (changes requested):
Using the existing socket timeout throughout?

redis/client.py
barshaul: I don't want to put it there because the following scenario is possible:
1. p.subscribe("foo")
2. health check is performed
3. p.unsubscribe("foo")
4. a health check response still hasn't received
5. p.unsuscribe("foo") 
6. clean_health_check_response is being called by the unsubscribe command, the health check  response hasn't arrived yet and it exists the loop due to ttl runs-out
7. the health check response only now received 
8. p.subscribe() is being called - self.subscribed is still False so a health check will be performed and we should clean the existing health check response before we continue. 
9. If we add 'clean_health_check_response=0' at the end of clean_health_check, we will clean the counter in step 6, so we won't be able to clean the socket from the response on step 8.

redis/client.py
barshaul: changed

chayim (approved):

PR #1765, opened 2021-12-02
Improved RedisCluster's reinitialize_steps and documentation

barshaul:
### Pull Request check-list

### Description of change

1. RedisCluster's reinitialize_steps have been refactored to moved_reinitialize_steps and the documentation has been improved
2. Destructor added to ClusterNode to close connection pool when deleted
3. Added a check to make sure the startup nodes have cluster mode enabled

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -382,7 +382,7 @@ class RedisCluster(RedisClusterCommands):
         cluster_error_retry_attempts=3,
         require_full_coverage=True,
         skip_full_coverage_check=False,
-        moved_reinitialize_steps=10,
+        reinitialize_steps=10,
         read_from_replicas=False,
         url=None,
         retry_on_timeout=False,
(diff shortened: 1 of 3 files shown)


redis/cluster.py
```diff
@@ -727,12 +738,14 @@ def _determine_nodes(self, *args, **kwargs):
...
-        if self.reinitialize_steps == 0:
+        # To reinitialize the cluster on every MOVED error,
+        # set moved_reinitialize_steps to 1.
+        # To avoid reinitializing the cluster on moved errors, set
+        # moved_reinitialize_steps to 0.
+        if self.moved_reinitialize_steps == 0:
             return False
         else:
-            return self.reinitialize_counter % self.reinitialize_steps == 0
+            return self.moved_reinitialize_counter % self.moved_reinitialize_steps == 0
```
chayim, line 748: IMHO we shouldn't rename this at this time.There isn't anything to be gained.

redis/cluster.py
```diff
@@ -1274,6 +1291,12 @@ def create_redis_node(self, host, port, **kwargs):
             r = Redis(host=host, port=port, **kwargs)
         return r
 
+    def _is_cluster_enabled(self, redis_node):
```
chayim, line 1294: Since this is called only once, it belongs in the initialize function. There's no benefit to separating out this check into a function unless there is reuse.

chayim (changes requested):
See changes

redis/cluster.py
barshaul: removed

redis/cluster.py
barshaul: rollbacked

chayim (approved):

PR #1780, opened 2021-12-09
Add packaging to setup_requires, and use >= to play nice to setup.py (fixes #1625)

### LINKED ISSUE #1625: Module installation fails due to missing dependency
https://github.com/redis/redis-py/blob/039488d97ec545b37e903d1b791a88bac8f77973/redis/connection.py#L1
the deprecated distutils was replaced with the packaging module as part of release v4.0.0b1
packaging is not a builtin python module but was not added to setup.py as a dependency which causes applications that require redis-py to fail if packaging isn't already installed on the machine.
the packaging module should probably be added as a dependency in setup.py to resolve this

hartwork:
Fixes unfixed (but closed) issue #1625, more details at https://github.com/redis/redis-py/issues/1625#issuecomment-989340063 . Thanks! :pray: 

### Pull Request check-list

### Description of change

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,9 +26,12 @@ setup(
     author="Redis Inc.",
     author_email="oss@redis.com",
     python_requires=">=3.6",
+    setup_requires=[
+        "packaging>=21.3",
+    ],
     install_requires=[
-        "deprecated==1.2.3",
-        "packaging==21.3",
+        "deprecated>=1.2.3",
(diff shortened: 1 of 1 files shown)


chayim:
Thank you for your contribution @hartwork. Admittedly we don't currently test for a pip install of the package (automatically), though I think we should. It would also help when updates (like this) are missed by me. Much thanks! Merging.

PR #1797, opened 2021-12-16
Socket not closing after server disconnect

pawl:
### Pull Request check-list

### Description of change

This is based on @michael-lazar's recent work on fixing a Celery memory leak caused by py-amqp: https://github.com/celery/py-amqp/pull/374

I recently learned that `socket.shutdown()` can throw an `OSError` when the server caused the disconnection. I also learned that `socket.shutdown` and `socket.close` have different purposes: https://stackoverflow.com/a/598759

So, you still need to run `socket.close` after `socket.shutdown` to allow garbage collection to deallocate the socket.

Currently, redis-py's `Connection.disconnect()` won't continue to close the socket if it encounters an `OSError` on shutdown. This prevents garbage collection from cleaning up the socket and will use more memory than necessary (maybe even causing a leak?).

This PR puts a separate try/except around `socket.close` and `socket.shutdown` to allow the socket to be closed after a failed shutdown.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -717,9 +717,14 @@ class Connection:
         self._parser.on_disconnect()
         if self._sock is None:
             return
-        try:
-            if os.getpid() == self.pid:
+
+        if os.getpid() == self.pid:
+            try:
                 self._sock.shutdown(socket.SHUT_RDWR)
+            except OSError:
+                pass
+
+        try:
             self._sock.close()
         except OSError:
             pass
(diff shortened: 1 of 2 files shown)


chayim:
@pawl Great catch, and a great PR, thank you! I'm particularly appreciative of your adding unit tests for this. Approving for merge, and it'll be in the upcoming release.

auvipy:
which redis release will include this?

PR #1817, opened 2021-12-18
Support for specifying error types with retry

### LINKED ISSUE #1764: Getting ReadOnly error when writing data to AWS elastic cache while the cluster is vertically scaled
I have a python application that writes some data to AWS elastic cache cluster at a regular interval.

Here's an example script that simulates the functionality of my application. It also replicates the error that I have been facing.

```py
from datetime import datetime
import redis
import time
import sys
import random

class Config:
    default_host = "localhost"
    master_host = "xxx.xx.0001.xxxx.cache.amazonaws.com"
    replica_host = "xxx.xx.0001.xxxx.cache.amazonaws.com"
    redis_db = 8
    socket_conn_timeout = 10
    request_delay_sec = 0.1

def get_redis_client():
    return redis.Redis(
        host=Config.master_host,
        db=Config.redis_db,
        socket_connect_timeout=Config.socket_conn_timeout,
    )

def get_random_key_value():
    val = time.time()
    key = "test_key_" + str(random.randint(0, 100))
    return key, val

r = get_redis_client()

r.flushdb()

flag = False

while True:
    try:
        if flag:
            print("beat:", time.time())
        r.set(*get_random_key_value())
        time.sleep(Config.request_delay_sec)
    except redis.RedisError as re:
        print(datetime.now(), "Error:", type(re), re)
        flag = True
        # sys.exit()
    except KeyboardInterrupt:
        print("Stopping loop execution")
        sys.exit()
```

Here are the environment details of my application

 - OS(Ubuntu 18.04)
 - Python(v 3.7.0)
 - Redis-py(v 3.5.3)
 - AWS elastic cache(cluster mode disabled, 1 master node, 1 read replica)

When I scale my AWS elastic cache cluster vertically and the above script is running, I get the following error for few seconds while the cluster scaling-up is in process and then it goes away.

```bash
<class 'redis.exceptions.ReadOnlyError'> You can't write against a read only replica.
```

AWS docs also states that during vertical scaling process some inconsistencies may occur because of data syncing(https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/redis-cluster-vertical-scaling.html)

why this error occurs during the scale up process? how can it be fixed?

I tried the same thing with a golang script and it works perfectly fine.

barshaul:
### Pull Request check-list

- [ X] Does `$ tox` pass with this change (including linting)?
- [ X] Is the new or changed code fully tested?
- [X ] Is there an example added to the examples folder (if applicable)?

### Description of change

The current client implementation supports only retry on TimeoutError.
Added support for passing specific error types to use with the retry mechanism.
The 'retry_on_timeout' argument has not changed for backward compatibility.
Closes #1764


redis/client.py
```diff
@@ -886,8 +887,10 @@ def __init__(
     ):
         """
         Initialize a new Redis client.
-        To specify a retry policy, first set `retry_on_timeout` to `True`
-        then set `retry` to a valid `Retry` object
+        To specify a retry policy for specific errors, first set
```
chayim, line 890: Very happy to see documentation addition for this new flag!

redis/client.py
```diff
@@ -1144,11 +1149,11 @@ def _send_command_parse_response(self, conn, command_name, *args, **options):
...
         """
         Close the connection and raise an exception
-        if retry_on_timeout is not set or the error
-        is not a TimeoutError
+        if retry_on_error is not set or the error
+        is not one of the specified error types
         """
         conn.disconnect()
-        if not (conn.retry_on_timeout and isinstance(error, TimeoutError)):
+        if not (conn.retry_on_error and isinstance(error, tuple(conn.retry_on_error))):
```
chayim, line 1156: What do you think about making this more expressive and positive cased? I'm hoping the below, makes this more clear to the next person who reads this.

```
if (conn.retry_on_error is None and isinstance(error, tuple(conn.retry_on_error))) is False:
    raise error
```

redis/connection.py
```diff
@@ -526,8 +527,10 @@ def __init__(
     ):
         """
         Initialize a new Connection.
-        To specify a retry policy, first set `retry_on_timeout` to `True`
-        then set `retry` to a valid `Retry` object
+        To specify a retry policy for specific errors, first set
```
chayim, line 530: Given this, I think retry_on_error should start as an empty list, rather than None. Rather than change the type (from None), it provides a hint to the future user, at least based on the signature.

chayim (changes requested):
Looks good overall. Given the new retry_on_error having a list value, I really think it's default should be at least, an empty list (see inline).

redis/client.py
barshaul: changed to:
```
        if (
            conn.retry_on_error is None
            or isinstance(error, tuple(conn.retry_on_error)) is False
        ):
            raise error
```

redis/connection.py
barshaul: changed

chayim (approved):

PR #1818, opened 2021-12-20
JSON set_file and set_path support

### LINKED ISSUE #1785: Add helper functions to support loading json from files
Having integrated [RedisJSON](https://redisjson.io) it would be nice if we could extend the json().set() behaviour by some new functions

- [ ] json().set_file() - This would be set the json value stored at the key named _name_ to the contents of the file specified in _obj_, effectively the same behaviour as json().set()

- [ ] json().set_path() - This would be the same as above, except without a _name_ attribute, and pointing to a directory path. All files in that directory would be stored, under the assumption that the filename, minus it's extension would be the key name.

dvora-h:
### Pull Request check-list

### Description of change

Adding the ability to use json.set with a file from disk, or a directory containing multiple json files.

closes #1785

### DIFF
--- a/redis/commands/json/commands.py
+++ b/redis/commands/json/commands.py
@@ -4,6 +4,8 @@ from redis.exceptions import DataError
 
 from .decoders import decode_dict_keys
 from .path import Path
+from json import loads, JSONDecodeError
+import os
 
 
 class JSONCommands:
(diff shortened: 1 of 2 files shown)


redis/commands/json/commands.py
```diff
@@ -213,6 +216,47 @@ def set(self, name, path, obj, nx=False, xx=False, decode_keys=False):
...
+        ``xx`` if set to True, set ``value`` only if it exists.
+        ``decode_keys`` If set to True, the keys of ``obj`` will be decoded
+        with utf-8.
+
+        """
+
+        with open(file_name, "r") as fp:
+            file_content = loads(fp.read())
+
+        return self.set(name, path, file_content, nx, xx, decode_keys)
```
chayim, line 234: Since there are named arguments in json.set, we should pass arguments for this function upstream, accordingly.

```sug
        return self.set(name, path, file_content, nx=nx, xx=xx, decode_keys=decode_keys)
```

redis/commands/json/commands.py
```diff
@@ -213,6 +216,47 @@ def set(self, name, path, obj, nx=False, xx=False, decode_keys=False):
...
+        with utf-8.
+
+        """
+        set_files_result = {}
+        for root, dirs, files in os.walk(root_folder):
+            for file in files:
+                try:
+                    file_name = os.path.join(root, file).rsplit(".")[0]
+                    file_path = os.path.join(root, file)
+                    self.set_file(file_name, json_path, file_path, nx, xx, decode_keys)
```
chayim, line 253: Same here - let's pass named arguments

redis/commands/json/commands.py
```diff
@@ -213,6 +216,47 @@ def set(self, name, path, obj, nx=False, xx=False, decode_keys=False):
...
+        ``xx`` if set to True, set ``value`` only if it exists.
+        ``decode_keys`` If set to True, the keys of ``obj`` will be decoded
+        with utf-8.
+
+        """
+        set_files_result = {}
+        for root, dirs, files in os.walk(root_folder):
+            for file in files:
+                try:
+                    file_name = os.path.join(root, file).rsplit(".")[0]
```
chayim, line 251: We set file_name repeatedly. Let's instead do something like:

```
file_name = os.path.abspath(os.path.join(root, file)
self.set_file(<etc>)
try:
     set_files_result[file_name] = True
except JSONDecodeError:
     set_files_result[file_name] = False
```

chayim (changes requested):
Please see suggestions. It's looking good!

chayim (approved):

PR #1821, opened 2021-12-21
Retry on error exception and timeout fixes

barshaul:
…onError import

### Pull Request check-list

- [X ] Does `$ tox` pass with this change (including linting)?
- [X ] Is the new or changed code fully tested?
- [X ] Is there an example added to the examples folder (if applicable)?

### Description of change

1. Added timeout error to the list of errors to retry.
2. Changed TimeoutError import from redis.exception to RedisTimeoutError so python's TimeoutError won't get override, as it can be also thrown when we try to establish a connection.
3. A ConnectionError import was added from redis.exceptions, since it was mistakenly not imported and the default ConnectionError was being used.
4. Increased sleep time to 0.25sec when ClusterDownError is thrown, so it will let more time for the cluster to get healed.

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -25,9 +25,9 @@ from redis.exceptions import (
     RedisError,
     ResponseError,
     SlotNotCoveredError,
+    TimeoutError,
+    TryAgainError,
 )
-from redis.exceptions import TimeoutError as RedisTimeoutError
-from redis.exceptions import TryAgainError
 from redis.utils import (
     dict_merge,
     list_keys_to_dict,
@@ -378,7 +378,6 @@ class RedisCluster(RedisClusterCommands):
     ERRORS_ALLOW_RETRY = (
         ConnectionError,
         TimeoutError,
-        RedisTimeoutError,
         ClusterDownError,
     )
 
@@ -863,9 +862,10 @@ class RedisCluster(RedisClusterCommands):
             dict<Any, ClusterNode>
         """
         target_nodes_specified = False
-        target_nodes = kwargs.pop("target_nodes", None)
-        if target_nodes is not None and not self._is_nodes_flag(target_nodes):
-            target_nodes = self._parse_target_nodes(target_nodes)
+        target_nodes = None
+        passed_targets = kwargs.pop("target_nodes", None)
+        if passed_targets is not None and not self._is_nodes_flag(passed_targets):
+            target_nodes = self._parse_target_nodes(passed_targets)
             target_nodes_specified = True
         # If an error that allows retrying was thrown, the nodes and slots
         # cache were reinitialized. We will retry executing the command with
@@ -885,7 +885,7 @@ class RedisCluster(RedisClusterCommands):
                 if not target_nodes_specified:
                     # Determine the nodes to execute the command on
                     target_nodes = self._determine_nodes(
-                        *args, **kwargs, nodes_flag=target_nodes
+                        *args, **kwargs, nodes_flag=passed_targets
                     )
                     if not target_nodes:
                         raise RedisClusterException(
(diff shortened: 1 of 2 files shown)


redis/cluster.py
```diff
@@ -17,16 +17,17 @@
...
     MovedError,
     RedisClusterException,
     RedisError,
     ResponseError,
     SlotNotCoveredError,
-    TimeoutError,
-    TryAgainError,
 )
+from redis.exceptions import TimeoutError as RedisTimeoutError
+from redis.exceptions import TryAgainError
```
barshaul, line 30: isort forces to take out the  TryAgainError as well..

redis/cluster.py
```diff
@@ -976,13 +977,6 @@ def _execute_command(self, target_node, *args, **kwargs):
                     # and try again with the new setup
                     self.nodes_manager.initialize()
                     raise
-            except TimeoutError:
-                log.exception("TimeoutError")
-                if connection is not None:
-                    connection.disconnect()
-
-                if ttl < self.RedisClusterRequestTTL / 2:
-                    time.sleep(0.05)
```
barshaul, line 985: We will reinitialize the cluster on timeouts same as for connection errors

redis/cluster.py
chayim: TimeoutError is defined in exceptions.py, and inherited from RedisError.  Either we should use the pure TimeoutError declared there (rather than RedisTimeoutError), or change that error everywhere to be named RedisTimeoutError.

If you want to keep the dedicated (and different) RedisTimeoutError as if it's not a stock error, then we should add a ClusterTimeoutError class and you can capture that. However - I don't think this is the case.

redis/cluster.py
```diff
@@ -953,11 +954,11 @@ def _execute_command(self, target_node, *args, **kwargs):
...
                 return response
 
-            except (RedisClusterException, BusyLoadingError):
-                log.exception("RedisClusterException || BusyLoadingError")
+            except (RedisClusterException, BusyLoadingError) as e:
+                log.exception(type(e))
                 raise
-            except ConnectionError:
-                log.exception("ConnectionError")
+            except (ConnectionError, TimeoutError, RedisTimeoutError) as e:
```
chayim, line 960: Nit: Let's order these by spelling:
```
except (ConnectionError, RedisTimeoutError, TimeoutError) as e:
```

redis/cluster.py
```diff
@@ -1339,9 +1333,9 @@ def initialize(self):
                     raise RedisClusterException(
                         "Cluster mode is not enabled on this node"
                     )
-                cluster_slots = r.execute_command("CLUSTER SLOTS")
+                cluster_slots = str_if_bytes(r.execute_command("CLUSTER SLOTS"))
                 startup_nodes_reachable = True
-            except (ConnectionError, TimeoutError) as e:
+            except (ConnectionError, TimeoutError, RedisTimeoutError) as e:
```
chayim, line 1338: Same ordering comment

chayim (changes requested):
TimeoutError vs RedisTimeoutError - see below.

redis/cluster.py
barshaul: I changed it to catch also socket.timeout (the default TimeoutError) because in one of the tests I did I thought connection.py thrown this error and not the redis.exceptions.TimeoutError. However, I go over again on connection.py code and I can't find a place where it doesn't wrap it with redis's timeoutError. I tried to reproduce it to see if I get again socket.timeout error but I was not able to find it again. 
I reverted it and discard the default timeout error, so we only catch redis's timeoutError in the cluster's code. I tested it again couple more time with live failovers and with short timeout periods and the client handles it properly.
So, as I said - I rolled back the timeout error import etc., please disregard it.

redis/cluster.py
```diff
@@ -1637,7 +1636,7 @@ def __init__(
         cluster_response_callbacks=None,
         startup_nodes=None,
         read_from_replicas=False,
-        cluster_error_retry_attempts=3,
+        cluster_error_retry_attempts=5,
```
chayim, line 1639: Maybe this should be configureable.  It seems likely that a user will want to specify their retries - but with a sane default of course.

barshaul: It is configurable - the user can pass 'cluster_error_retry_attempts' when creating a new RedisCluster instance.
In the tests I ran 3 attempts wasn't enough to handle some of the failover scenarios. obviously it depends on the socket_timeout a user will configure but 5 retries seems sane to me. what do you think?

chayim (approved):

PR #1843, opened 2021-12-30
Add support for LMPOP

dvora-h:
### Pull Request check-list

### Description of change

Add support for LMPOP (redis 7) #1798

### DIFF
--- a/redis/commands/core.py
+++ b/redis/commands/core.py
@@ -1,8 +1,8 @@
 import datetime
 import hashlib
 import time
-from typing import List, Optional
 import warnings
+from typing import List, Optional
 
 from redis.exceptions import ConnectionError, DataError, NoScriptError, RedisError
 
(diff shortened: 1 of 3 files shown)


redis/commands/core.py
```diff
@@ -1926,6 +1926,19 @@ def brpoplpush(self, src, dst, timeout=0):
             timeout = 0
         return self.execute_command("BRPOPLPUSH", src, dst, timeout)
 
+    def lmpop(self, num_keys, keys, *args, count=1):
```
chayim, line 1929: We should split out left|right as an argument named direction.
What's the behaviour meant to be when count is 0 (perhaps we shouldn't have a default count)
Can you add type hints for this - I'd like to start using them.

chayim (approved):
Approved. Will merge as we start pulling in the redis 7 commands.

PR #1845, opened 2022-01-02
Clusters should optionally require full slot coverage

barshaul:
…on, Changed  RedisClient's 'require_full_coverage' default value to false

### Pull Request check-list

### Description of change

In order to avoid the client being a limiting factor for communicating with the server, we should not require full slots coverage by default in the client side.

require_full_coverage parameter: 
            When set to False (default value): the client will not require a
            full coverage of the slots. However, if not all slots are covered,
            and at least one node has 'cluster-require-full-coverage' set to
            'yes,' the server will throw a ClusterDownError for some key-based
            commands. See -
            https://redis.io/topics/cluster-tutorial#redis-cluster-configuration-parameters
            When set to True: all slots must be covered to construct the
            cluster client. If not all slots are covered, RedisClusterException
            will be thrown. 

The 'skip_full_coverage_check' parameter has been removed from RedisClient, and the client no longer checks for 'cluster-require-full-coverage' configuration on the cluster nodes (note - not backward compatible). This check doesn't need to be run since it's an overhead for the client and is already handled by the server's ClusterDown error. Additionally, some managed Redis clusters, such as ElastiCache, don't support the 'config' command, so the client will not work with the default configurations if we keep the node's configuration check. Other clients (e.g. Lettuce, Jedis) doesn't run this check as well.

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -387,8 +387,7 @@ class RedisCluster(RedisClusterCommands):
         port=6379,
         startup_nodes=None,
         cluster_error_retry_attempts=3,
-        require_full_coverage=True,
-        skip_full_coverage_check=False,
+        require_full_coverage=False,
         reinitialize_steps=10,
         read_from_replicas=False,
         url=None,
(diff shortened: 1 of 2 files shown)


redis/cluster.py
```diff
@@ -1450,29 +1419,9 @@ def initialize(self):
             # isn't a full coverage
             raise RedisClusterException(
                 f"All slots are not covered after query all startup_nodes. "
-                f"{len(self.slots_cache)} of {REDIS_CLUSTER_HASH_SLOTS} "
+                f"{len(tmp_slots)} of {REDIS_CLUSTER_HASH_SLOTS} "
```
barshaul, line 1422: small bug fix

chayim: Makes sense - limit based on available as opposed to the entire cluster. This should recover more cleanly.

chayim (approved):

PR #1873, opened 2022-01-11
OCSP Stapling Support

chayim:
### Pull Request check-list

### Description of change


redis/ocsp.py
```diff
@@ -11,8 +11,50 @@
...
+    peer_cert = con.get_peer_certificate().to_cryptography()
+    for c in con.get_peer_cert_chain():
+        if c.to_cryptography().subject == peer_cert.issuer:
+            issuer_cert = c
+            break
+
+    if issuer_cert is None:
+        raise ConnectionError("no matching issuer cert found in certificate chain")
+
+    e = x509.load_pem_x509_certificate(expected)
```
eyalbrami1, line 48: This will fail if user does not provide an expected certificate. do we want to allow users not to provide one? I think so, that's up to a user to decide if he needs it

redis/ocsp.py
```diff
@@ -11,8 +11,50 @@
...
 
+def _check_certificate(certificate):
+    """A wrapper the return the validity of a known ocsp certificate"""
+
+    ocsp_response = ocsp.load_der_ocsp_response(certificate)
+    if ocsp_response.response_status == ocsp.OCSPResponseStatus.UNAUTHORIZED:
+        raise AuthorizationError("you are not authorized to view this ocsp certificate")
+    if ocsp_response.response_status == ocsp.OCSPResponseStatus.SUCCESSFUL:
+        if ocsp_response.certificate_status == ocsp.OCSPCertStatus.REVOKED:
+            return False
```
eyalbrami1, line 22: What if the Cert status is `UNKNOWN`? it is a possible status and I think you should return `False` on that as well

redis/ocsp.py
```diff
@@ -11,8 +11,50 @@
 from redis.exceptions import AuthorizationError, ConnectionError
 
 
+def _check_certificate(certificate):
```
eyalbrami1, line 14: This method does not validate the signature on the stapled response. 
It should verify the signature using the responder's public key (the responder's certificate that includes its private key is included in the stapled OCSP response)

redis/ocsp.py
```diff
@@ -11,8 +11,50 @@
...
+    """A wrapper the return the validity of a known ocsp certificate"""
+
+    ocsp_response = ocsp.load_der_ocsp_response(certificate)
```
eyalbrami1, line 17: The response contains a `thisUpdate` timestamp, the verifier should check that it is valid (i.e. in the past)
the response **may** contain a `nextUpdate` timestamp as well, stating until when is this stapled response valid. if it exists in the response, you should check that the response is still valid

redis/connection.py
```diff
@@ -968,7 +976,33 @@ def _connect(self):
...
+                ocsp_staple_verifier,
+                self.ssl_ocsp_expected_cert,
+            )
+
+            #  need another socket
+            con = OpenSSL.SSL.Connection(self.ssl_ocsp_context, socket.socket())
+            con.request_ocsp()
+            con.connect((self.host, self.port))
+            con.do_handshake()
+            con.shutdown()
```
eyalbrami1, line 1001: isn't this actually closing the connection?

redis/connection.py
```diff
@@ -968,7 +976,33 @@ def _connect(self):
...
             raise RedisError("cryptography is not installed.")
-        elif self.ssl_validate_ocsp is True and CRYPTOGRAPHY_AVAILABLE:
+
+        if self.ssl_ocsp_context is not None and self.ssl_validate_ocsp is not False:
+            raise RedisError(
+                "Either an OCSP context or validation may be specified " "but not both."
+            )
+
+        # validation for the stapled case
+        if self.ssl_ocsp_context:
```
eyalbrami1, line 986: There should be an option for a user to not provide an OpenSSL context (nor an expected certificate as I said in other comments) and have redisPy create this temporary basic OpenSSL context for this OCSP staple verification.
This will make life easy for basic usage that simply asks for some staple verification, something like `request_ocsp_staple_verification=True` and that's it

redis/ocsp.py
chayim: It did in a previous version, restoring. I now check the OCSPResponseStatus and return there - but the next push restores this - excatly here. Good call!

redis/ocsp.py
chayim: _ocsp_response.this_update_ being in the part is fair - the check is ready. If I read [RFC 5019](https://www.rfc-editor.org/rfc/rfc5019#page-7) correctly, a valid responder must produce a response, to allow for caching. This means the cert check will now raise a connection error.  Please correct me if I'm wrong.

redis/connection.py
chayim: Correct. the OCSP validation occurs on a different socket. I was unable to reusse the same socket for the internal ssl implementation and PyOpenSSL. 

In my current thinking - future versions of redis-py (5?) I'd like to see us operate more similarly to node, and have the user supply the context.  But I'd love feedback here.

redis/connection.py
chayim: This will change in a future version - but given our use case, agreed.

redis/ocsp.py
eyalbrami1: nope. `this_update` should simply be in the past, meaning that the current time should not be before `this_update`
even cached responses should be valid between `this_update` and `next_update`, which we should verify that is true (unless `next_update` is not there, then only check that we're after `this_update` timestamp)

chayim: Perhaps I wasn't being clear. In [section 2.2.4](https://www.rfc-editor.org/rfc/rfc5019#section-2.2.4) the RFC states that _Responders MUST always include this value to aid in response caching_, WDYT?

chayim: This appears to be superceded by [RFC 6960#4.2.2.1](https://www.rfc-editor.org/rfc/rfc6960#section-4.2.2.1). But given the language it's unclear. Erring on the side of not requiring a nextUpdate.

redis/connection.py
```diff
@@ -921,6 +924,10 @@ def __init__(
             ssl_ca_path: The path to a directory containing several CA certificates in PEM format. Defaults to None.
             ssl_password: Password for unlocking an encrypted private key. Defaults to None.
 
+            ssl_validate_ocsp: If set, perform a full ocsp validation (i.e not a stapled verification)
+            ssl_ocsp_context: A fully initialized OpenSSL.SSL.Context object to be used in verifying the ssl_ocsp_expected_cert
```
eyalbrami1, line 928: you did not document "ssl_validate_ocsp_stapled"

eyalbrami1 (approved):

redis/connection.py
chayim: Thanks. It''s in our examples. But I've pushed this too now.

whardier:
Is there any specific reason pyopenssl is pinned in setup.py?

PR #1874, opened 2022-01-11
Define incr/decr as aliases of incrby/decrby

salty-horse:
### Pull Request check-list

(I would welcome some help on how to set up the CI tests on my forked repo, if this is necessary (it's already running in this PR). I couldn't figure it out, and the Contributing guide wasn't helpful)

### Description of change

Currently, `incr` and `decr` are implemented as identical wrappers of the functions `incrby` and `decrby`.

Their documentation is also identical to the other functions, with the difference of linking to the original command's Redis reference. `repr` is documented as:
>Increments the value of ``key`` by ``amount``.  If no key exists, the value will be initialized as ``amount``
> For more information check https://redis.io/commands/incr

This is incorrect and misleading, since Redis's documentation of `incr` contains no mention of `amount`.

This change sets up `incr` and `decr` as direct aliases of `incrby` and `decrby` by assignment. This also means they share docstrings, so that both will lead to `incrby` and `decrby`'s original references. I think this makes more sense given the function signatures. (I checked the generated Sphinx documentation and it is identical, other than the Redis URLs).

I also removed the comment in `incrby` that claims it is an alias of `incr`. I think it's the other way around, and I don't think the comment is adding much. Perhaps it's more useful to document the fact that `incr` is an alias of `incrby` for backwards compatibility...?

### DIFF
--- a/redis/commands/core.py
+++ b/redis/commands/core.py
@@ -1159,17 +1159,6 @@ class BasicKeyCommands:
             params.append("REPLACE")
         return self.execute_command("COPY", *params)
 
-    def decr(self, name, amount=1):
-        """
-        Decrements the value of ``key`` by ``amount``.  If no key exists,
-        the value will be initialized as 0 - ``amount``
-
-        For more information check https://redis.io/commands/decr
-        """
-        # An alias for ``decr()``, because it is already implemented
-        # as DECRBY redis command.
-        return self.decrby(name, amount)
-
     def decrby(self, name, amount=1):
         """
         Decrements the value of ``key`` by ``amount``.  If no key exists,
@@ -1178,6 +1167,8 @@ class BasicKeyCommands:
         For more information check https://redis.io/commands/decrby
         """
         return self.execute_command("DECRBY", name, amount)
+
+    decr = decrby
 
     def delete(self, *names):
         """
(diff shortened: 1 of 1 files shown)


dvora-h:
Hi @salty-horse, Thanks for the contribution!
Your suggestion could have been good if decr/incr and decrby/incrby really were the same but in fact decr/incr should always be 1 and has no option to get another amount (It’s a mistaka that it gets), so I suggest just deleting the amount from decr/incr, WDYT?

salty-horse:
I also found the identical API curious. I came across it when changing an `incr` call to `incrby` and finding out `incr` already supports it. But I don't know if it's a mistake - it could have been done intentionally for convenience.

Removing `amount` from `incr`/`decr` is a backwards incompatible change in the API that (from my limited experience) I don't see the benefit for, unless Redis's `INCR x` is more performant than `INCRBY x 1`, or there's another reason to allow redis-py to call it directly.

I don't oppose that change, but since it can affect the existing code of many users, maybe it should be discussed in a wider forum.

dvora-h:
Hi @salty-horse 
You are right, it would be better not to do this so as not to break the API. I just checked the docs and they are not affected by this change, so I accept your PR.

dvora-h (approved):

PR #1879, opened 2022-01-14
Add retries to connections in Sentinel Pools

jdieter:
### Pull Request check-list

### Description of change

Currently, when using sentinel, if a master dies suddenly (i.e. is `kill -9`'d), and the client is configured with retries, the client will attempt to reconnect to the master using the sentinel pool, as expected.

However, if the sentinels have not yet figured out that the master is dead, a `redis.exceptions.ConnectionError` will be returned when trying to re-connect to the old master _even if `ConnectionError` is in the list of retry errors_.

This PR fixes this by making sure that the initial connection itself in the SentinelPool is also retryable.  This has been tested with a script that increments a counter by 1 every second and then running `kill -9` on the Redis master.  As expected, there's a delay while the sentinels figure out what the new master should be, and then the script continues.

I'm not very familiar with the code, so I'm not sure if this is the ideal place for this code.  It is possible that it would make more sense to put it in the connection itself.

While this will need a new test, I'm not sure how to create an automated test that `kill -9`s one of the Redis servers, I'm not sure if this fix needs documentation, and I'm not sure if I should add an example

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -1,13 +1,10 @@
-import copy
 import random
 import weakref
 
-from redis.backoff import NoBackoff
 from redis.client import Redis
 from redis.commands import SentinelCommands
 from redis.connection import Connection, ConnectionPool, SSLConnection
 from redis.exceptions import ConnectionError, ReadOnlyError, ResponseError, TimeoutError
-from redis.retry import Retry
 from redis.utils import str_if_bytes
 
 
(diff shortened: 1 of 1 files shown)


jdieter:
> @jdieter Thank you for contributing this! It would really help bring the sentinel connection in line with the others!
> 
> I can definitely see the argument for adding this to the pool, but I think we probably want this on the SentinelManagedConnection class, instead of the pool itself. This would be consistent with the other retry attempts. WDYT?

Well, it makes the code significantly easier! :)  I've pushed it as a new commitl, but, if you'd prefer, I can merge the two commits into one.

chayim:
@jdieter Awesome. We're going to roll this into redis-py 4.1.1, the upcoming release. Thank you so much, this is perfect!  We squash merge, so all good, all the time.

PR #1894, opened 2022-01-23
Import aioredis

Andrew-Chen-Wang:
This is the initial PR to merge aioredis into redis-py.

What has been done so far:
- [X] Copied files from aioredis to redis-py (I just copied directly, mostly unmodified)
- [ ] De-duplicate code by abstracting away several things (perhaps save for later?)
   - Already in progress though; no need to de-duplicate everything right now.
- [ ] Add type hints to redis-py client and connection? py.typed seems to enable mypy, so perhaps we don't need to add type hints to the sync parts yet (should save for later)
   - Already in progress though; no need to continue.
- [ ] Add aioredis's test files
- [X] Figure out how to add type hints for command mixins

Must dos:
- [ ] Add test suite from aioredis (leaning towards separate tests_async dir; however, locally running test suite when making PRs would be nicer to have async tests run with the sync tests so nothing's missed)
   - The structure would be: instead of `tests/test_bloom.py`, we would have `tests/bloom/test.py` and `tests/bloom/test_async.py`. This also would help to split up large test files like `test_commands.py` 
- [ ] Drop Python 3.6
    - Helps with current annotations that do `arg: int | None`
    - 3.6 already EOL-ed


Andrew-Chen-Wang:
trying to strike the right balance between "holy cow this PR is huge" and "decent product delivery". I think I'll cut the optimization and type annotations for sync redis py. The PR will just be getting aioredis in and we can optimize later.

Issue #1898, opened 2022-01-24
Lock method does not exist in the redis client using cluster mode


jdavidagudelo:
**Version**: 4.1.1

**Platform**: Python 3.10.1 Ubuntu/21.10

**Description**: I am currently using the library https://github.com/Grokzen/redis-py-cluster to access my redis-cluster, I tried to move the code to the latest version of this library in cluster mode, but I found this library does not have a **lock** method when the client is used in cluster mode. Is there any workaround for this, will this library allow locks when using it in cluster mode or are locks not intended to work in cluster mode?

jakebarnwell:
`redis-py-cluster` works fine with locks (noting that redis-py-cluster has its `RedisCluster` class extend redis-py's `Redis` class), so imagine this is mostly an issue of "we just haven't added support yet"?

After some poking, it looks like implementing `RedisCluster.lock()` properly involves doing something like the following:
1. add scripting support (`ScriptCommands`) to `RedisCluster` - needed for lua lock since it uses evalsha etc
2. either add a `.connection_pool.get_encoder()` stub to `RedisCluster` or change `Script` and `Lock` so they can get access to the encoder somehow via `RedisCluster.get_encoder()`... bbecause we have `Redis.connection_pool.get_encoder()` and `RedisCluster.get_encoder(), but there's no such thing as `RedisCluster.connection_pool`

Does this sound right @dvora-h ? Lock support is something I'm interested in as well.

jakebarnwell:
TBH I'm not sure if this is something that someone else is tackling at the moment, but I'm happy to try

chayim:
Correct. We haven't gotten there yet. @jakebarnwell if you'd like to - that would be fantastic!

I've assigned this to you simply because you volunteered. Please feel free to say no, of course :smiley:

jakebarnwell:
Thank you! I've made this PR: https://github.com/redis/redis-py/pull/1937

I'm not sure how to assign somebody to it, but it's there...!

jakebarnwell:
Sorry for the confusion -- #1937 adds scripting support, but does not yet implement the `.lock()` method. https://github.com/redis/redis-py/pull/2013 adds the anticipated `.lock()` method.

PR #1937, opened 2022-02-11
Add cluster support for scripting

jakebarnwell:
### Pull Request check-list

### Description of change

Partly fixes https://github.com/redis/redis-py/issues/1898:

Adds scripting support, but does not yet add `.lock()` to `RedisCluster`. That will come in a following PR.

Much of this change was ~stolen~ adapted from `redis-py-cluster`.

Implements the following commands:
* EVAL: the command is sent to the cluster depending on the keys (i.e., in `EVAL <SCRIPT> num_keys KEY1 ... KEYN ...`); the keys must all be on the same node. If the script requires 0 keys, the command is sent to a random node.
* EVALSHA: same implementation as above
* SCRIPT EXISTS: the command is sent to all primaries. The result is a list of booleans. Each boolean is an AND of "does the script exist on each node". In other words, each boolean is True iff the script exists on all nodes.
* SCRIPT FLUSH: the command is sent to all primaries. The result is a bool AND over all nodes' responses.
* SCRIPT LOAD: the command is sent to all primaries. The result is the SHA1 digest.

I have not (yet?) implemented the `EVAL_RO` commands -- do those need to be implemented? Maybe.

This PR does not implement, nor do I intend to implement:
* SCRIPT KILL
* scripting support with a cluster pipeline

I have added a couple unit tests for the newly-implemented things. ~I have not yet added documentation to the PR, but I will.~


redis/commands/core.py
```diff
@@ -4663,7 +4663,7 @@ def __init__(self, registered_client, script):
         if isinstance(script, str):
             # We need the encoding from the client in order to generate an
             # accurate byte representation of the script
-            encoder = registered_client.connection_pool.get_encoder()
+            encoder = registered_client.get_encoder()
```
jakebarnwell, line 4666: This is technically a breaking change. If any consumers of `redis-py` use Scripts directly, injecting their own `registered_client` that _isn't_ a subclass of `Redis` or `RedisCluster`, this could break for them if they haven't got `client.get_encoder()` defined.

chayim: While I believe it's very unlikely - I'd rather not break the universe. Perhaps we could wrap this with an _if getattr_, check. What do you think @jakebarnwell

tests/test_scripting.py
```diff
@@ -21,24 +21,70 @@
...
 
-    def test_eval(self, r):
+    def test_eval_multiply(self, r):
         r.set("a", 2)
         # 2 * 3 == 6
         assert r.eval(multiply_script, 1, "a", 3) == 6
 
     # @skip_if_server_version_lt("7.0.0") turn on after redis 7 release
+    @pytest.mark.onlynoncluster
     def test_eval_ro(self, unstable_r):
```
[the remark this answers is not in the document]
chayim: I don't think  you should have to deal with this.

redis/commands/core.py
jakebarnwell: That's a good idea! I will try something like that

jakebarnwell:
@chayim can you do another review? :)

chayim (approved):

chayim:
@jakebarnwell This is a beautiful thing. Thank you for the approach, and the changes. I've approved, and we'll merge this into master shortly. It'll be part of the 4.2.0rc1 release.

chayim:
@dvora-h Can you have a look at the conflict here and resolve it for our friend @jakebarnwell prior to the 4.2.0rc1 release?

PR #2009, opened 2022-02-22
Add AsyncFunctionCommands

Andrew-Chen-Wang:
### Pull Request check-list

- [ ] Is the changes added to CHANGES file?

### Description of change

### DIFF
--- a/redis/commands/core.py
+++ b/redis/commands/core.py
@@ -5485,6 +5485,9 @@ class ClusterCommands(CommandsProtocol):
         return self.execute_command("READONLY", **kwargs)
 
 
+AsyncClusterCommands = ClusterCommands
+
+
 class FunctionCommands:
     """
     Redis Function commands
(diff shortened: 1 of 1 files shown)


redis/commands/core.py
```diff
@@ -5592,15 +5601,15 @@ def function_restore(self, payload: str, policy: Optional[str] = "APPEND") -> st
...
+    def function_kill(self) -> Union[Awaitable[str], str]:
         """
         Kill a function that is currently executing.
 
         For more information check https://redis.io/commands/function-kill
         """
         return self.execute_command("FUNCTION KILL")
 
-    def function_stats(self) -> list:
+    def function_stats(self) -> Union[Awaitable[list], list]:
```
chayim, line 5612: @Andrew-Chen-Wang Can you educate me. Why don't we use ```Union[Awaitable[List], List]```

As you know type hints aren't my jam ;)

Andrew-Chen-Wang: They're basically the same. In Python 3.6, you can only do List[element type], but in 3.7+ you can use the built in list like list[element type]. However, in this case, I'm not using the getattr method at all so either List or list works.

Also this was before I found out the other type hints were merged 😅

chayim: HAHAHA Awesome. Thanks again @Andrew-Chen-Wang. I will always appreciate the education here.

PR #2010, opened 2022-02-22
Add async typing support to commands' returns

Andrew-Chen-Wang:
Some typing by @dvora-h didn't include the newly included async response typing.


chayim (approved):
Thanks @Andrew-Chen-Wang as always :). I pushed a small linter fix into your branch. I hope you don't mind!

Andrew-Chen-Wang:
np at all. hopefully i can find a chance to add some typing to the sync side for redis, but no promises 😅 can't wait for those Optionals to go away with future annotations and a pyupgrade

PR #2013, opened 2022-02-22
Implement locks for RedisCluster

jakebarnwell:
### Pull Request check-list

- [x] Is the changes added to CHANGES file?

### Description of change

Adds support for the `lock()` method on `RedisCluster`. Currently, only `Redis` class supports `lock()`. https://github.com/redis/redis-py/pull/1937 added lua scripting support for `RedisCluster` which is a prerequisite for locking support, so with that out of the way, this PR finishes adding support for the `.lock()` method.

The change is relatively simple: we just copy and paste the `Redis.lock()` implementation onto `RedisCluster.lock()`. However, we do change the `Lock` implementation for acquiring the encoder if necessary. Instead of calling `self.redis.connection_pool.get_encoder()`, we now call `self.redis.get_encoder()`. Both of these methods already exist so it _shouldn't_ be breaking (?).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,10 +1,12 @@
-	* Create codeql-analysis.yml (#1988). Thanks @chayim
+  	* Create codeql-analysis.yml (#1988). Thanks @chayim
+    * Add limited support for Lua scripting with RedisCluster
+    * Implement `.lock()` method on RedisCluster
 * 4.1.3 (Feb 8, 2022)
-  * Fix flushdb and flushall (#1926)
-  * Add redis5 and redis4 dockers (#1871)
-  * Change json.clear test multi to be up to date with redisjson (#1922)
-  * Fixing volume for unstable_cluster docker (#1914)
-  * Update changes file with changes since 4.0.0-beta2 (#1915)
+    * Fix flushdb and flushall (#1926)
+    * Add redis5 and redis4 dockers (#1871)
+    * Change json.clear test multi to be up to date with redisjson (#1922)
+    * Fixing volume for unstable_cluster docker (#1914)
+    * Update changes file with changes since 4.0.0-beta2 (#1915)
 * 4.1.2 (Jan 27, 2022)
     * Invalid OCSP certificates should raise ConnectionError on failed validation (#1907)
     * Added retry mechanism on socket timeouts when connecting to the server (#1895)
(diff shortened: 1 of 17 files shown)


redis/lock.py
```diff
@@ -224,7 +224,7 @@ def owned(self):
         # need to always compare bytes to bytes
         # TODO: this can be simplified when the context manager is finished
         if stored_token and not isinstance(stored_token, bytes):
-            encoder = self.redis.connection_pool.get_encoder()
+            encoder = self.redis.get_encoder()
```
jakebarnwell, line 227: I'm beginning to confuse myself a bit -- I think this is a safe change since both redis classes have the `.get_encoder()` method defined on them already? But I'm not positive

CHANGES
```diff
@@ -94,10 +96,10 @@
     * Removing command on initial connections (#1722)
     * Removing hiredis warning when not installed (#1721)
 * 4.0.0 (Nov 15, 2021)
-    * FT.EXPLAINCLI intentionally raising NotImplementedError 
+    * FT.EXPLAINCLI intentionally raising NotImplementedError
```
jakebarnwell, line 99: sorry for these whitespace trims -- my IDE does it 😅

jakebarnwell:
@dvora-h see my [comment](https://github.com/redis/redis-py/issues/1898#issuecomment-1048189465) on https://github.com/redis/redis-py/issues/1898 -- my original PR https://github.com/redis/redis-py/pull/1937 technically didn't solve the issue since the `.lock()` method wasn't implemented in that PR, but this PR does implement the `.lock()` method 😃

CHANGES
chayim: flake8 actually enforces it, but only on python files ;)

redis/lock.py
chayim: I think you're right. Since merging in cluster and asyncio, we actually need to start cleaning and unifying these interfaces. There's a lot of duplication - but the philosophy was first get it in, then clean it up.
@dvora-h concur?

chayim (approved):

jakebarnwell:
I think the tests are flaky 🤔 -- they pass fine on my machine, at least

dvora-h:
Awesome @jakebarnwell, thank you for the PR! I merge it into master.

PR #2048, opened 2022-03-13
Add support to use certificates from string in ssl connection

### LINKED ISSUE #2047: Add support to use certificates from string instead of file
Currently, redis-py support connecting using SSL certificates stored in files by passing the certificate path to the redis connection. This, in turn, is sent to the python/cython SSL module.
Python 3.4 added support for loading certificates from a string in addition to a file. See https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations.
Supporting the parameter `cdata` in redis-py connection will enable to send certificates from a string instead of a file.

All

All

dvora-h:
### Pull Request check-list

### Description of change

closes #2047


tests/test_ssl.py
```diff
@@ -53,7 +53,7 @@ def test_ssl_connection_without_ssl(self, request):
             r.ping()
             assert "Connection closed by server" in str(e)
 
-    def test_validating_self_signed_certificate(self, request):
+    def test_validating_self_signed_file_certificate(self, request):
```
chayim, line 56: Nit: I don't think you need to rename the test. But this PR is great. Approved.

chayim (approved):

PR #2055, opened 2022-03-18
Remove verbose logging from cluster initializers

joni-:
### Pull Request check-list

### Description of change

We are using ClusterPipeline and noticed that our logs are filled with "Creating new instance of ClusterPipeline" entries. I am not sure if these entries provide much value so I thought we could either remove them completely or change the log level to debug.

Another thing that we noticed was that cluster.py is the only module that's doing any kind of logging at all.


chayim:
@joni- Thanks for helping clean this up! This definitely belongs in the 4.2.0 release. Much obliged.

chayim (approved):

Issue #2086, opened 2022-04-05
auth method works wrong for default username


pelid:
**Version**: 4.2.2

**Platform**: Docker image python:3.10

**Description**: Can not connect to Redis with Sentinel and with protection by password.

```python
>>> from redis import Sentinel
>>> 
>>> PASSWORD = 'some password'
>>> sentinel = Sentinel(
...     [('redis', 26379)],
...     sentinel_kwargs={'password': PASSWORD},
...     socket_timeout=0.1
... )
>>> sentinel.sentinel_masters()   # Check sentinel authentication works correctly
True
>>> master = sentinel.master_for('mymaster', socket_timeout=0.1)
>>> master.auth(PASSWORD)
<bound method Redis.execute_command of Redis<SentinelConnectionPool<service=mymaster(master)>>
```

Expects authentication to be done but got `bound method` instead. This behaviour does not correspond to docstring content:

[_redis/commands/core.py:372_](https://github.com/redis/redis-py/blob/143107af05fec7d140a317230becc2b3d353335a/redis/commands/core.py#L372-L378)
```python
def auth(self, password, username=None, **kwargs):
    """
    Authenticates the user. If you do not pass username, Redis will try to
    authenticate for the "default" user. If you do pass username, it will
    authenticate for the given user.
    For more information see https://redis.io/commands/auth
    """
```

Suppose it is a bug:

[_redis/commands/core.py:379_](https://github.com/redis/redis-py/blob/143107af05fec7d140a317230becc2b3d353335a/redis/commands/core.py#L379-L381)
```python
if username:
    return self.execute_command("AUTH", username, password, **kwargs)
return self.execute_command
```

Why return `execute_command` instead of calling it ?

PR #2092, opened 2022-04-07
Fix incorrect return statement in auth (#2086)

kamyabzad:
### Pull Request check-list

### Description of change
Fix a bug in commands.core.ManagementCommands.auth that returned a function handler when provided with a None username

### DIFF
--- a/redis/commands/core.py
+++ b/redis/commands/core.py
@@ -376,9 +376,11 @@ class ManagementCommands(CommandsProtocol):
         authenticate for the given user.
         For more information see https://redis.io/commands/auth
         """
-        if username:
-            return self.execute_command("AUTH", username, password, **kwargs)
-        return self.execute_command
+        pieces = []
+        if username is not None:
+            pieces.append(username)
+        pieces.append(password)
+        return self.execute_command("AUTH", *pieces, **kwargs)
 
     def bgrewriteaof(self, **kwargs):
         """Tell the Redis server to rewrite the AOF file from data in memory.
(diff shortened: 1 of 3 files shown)


redis/commands/core.py
```diff
@@ -369,16 +369,14 @@ class ManagementCommands(CommandsProtocol):
...
         """
         Authenticates the user. If you do not pass username, Redis will try to
         authenticate for the "default" user. If you do pass username, it will
         authenticate for the given user.
         For more information see https://redis.io/commands/auth
         """
-        if username:
-            return self.execute_command("AUTH", username, password, **kwargs)
-        return self.execute_command
+        return self.execute_command("AUTH", username, password, **kwargs)
```
chayim, line 379: Given the [command docs](https://redis.io/commands/auth/) it would appear instead that the following is more accurate. Thoughts?
```
def auth(self, password, username=None, **kwargs):
    parts = []
    if username is not None:
        parts.append(username)
    parts.append(password)
    return self.execute_command("AUTH", *parts, **kwargs)
```

kamyabzad: Thanks for the suggestion, I like it more. But there's a problem, if no password is set for the default user, `AUTH` command will fail with the following error. Won't that be a problem? 
`(error) ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?`

However, `AUTH default ""` returns OK.

kamyabzad:
@chayim Thanks for the suggestion, I like it more. I have updated the code and the tests (also rebased onto upstream).

chayim (approved):

PR #2094, opened 2022-04-07
Minor cleanups in commands/cluster.py

enjoy-binbin:


chayim (approved):
Thanks @enjoy-binbin for helping make the documentation even better!

Issue #2095, opened 2022-04-07
execute_command() behaves strangely in cluster mode


suxb201:
**Version**: redis==4.2.2
**Platform**: MacOS with python==3.10
**Description**: `execute_command` behaves strangely in cluster mode.

run no-key command:
```python
# works fine
rc.cluster_info()
rc.execute_command("CLUSTER INFO")

# RedisClusterException: No way to dispatch this command to Redis Cluster. Missing key.
rc.execute_command("CLUSTER", "INFO")
rc.execute_command("cluster info")
rc.execute_command("cluster", "info")
```

chayim:
@suxb201 can you try using the various functions intended for this purpose, on the cluster interface? I'm asking because underneath execute_command should be used for commands that are unimplemented, and there are function differences associated with the connection.

chayim:
Hi @suxb201 have you tried using the commands directly, for the cluster interface rather than execute command? While perhaps there's improvement for raw commands we can do here (clearly!), at least this should get you over your hump.

suxb201:
We have some scenarios where we use string concatenation commands, so execute_command() is the best choice.

suxb201:
Thanks to @chayim and @dvora-h , the problem has been solved in #2097.

PR #2097, opened 2022-04-08
fix execute_command() determine nodes error when no key command

suxb201:
see #2095


dvora-h (approved):
@suxb201 Thank you for reporting this bug and contribute its fix!

chayim:
@dvora-h note there have been changes since your reviewed fix. I cleared the approved flag as a result.

suxb201:
@chayim just did a git rebase

dvora-h (approved):

PR #2099, opened 2022-04-09
Add Async RedisCluster

utkarshgupta137:
### Pull Request check-list

### Description of change

Add Async Redis Cluster Client


chayim:
@utkarshgupta137 This looks like a great PR in progress. Just checking if you meant for it to be draft.
CC: @dvora-h

utkarshgupta137:
> @utkarshgupta137 This looks like a great PR in progress. Just checking if you meant for it to be draft. CC: @dvora-h

Yeah. I've not modified any documentation or added tests and I've only done basic testing.

utkarshgupta137:
@chayim @dvora-h @Andrew-Chen-Wan @Grokzen

I've got almost all the tests running. I'll see tomorrow why the remaining are failing.

I've divided the PR into 3 commits:
- First one simply duplicates all sync code in async locations. Also copies all mark.onlynoncluster from sync test_commands to asyncio test_commands
- Second just changes all redis client calls in async cluster tests to async/await calls
- Third one is where all the actual changes I've made in async cluster client & test modifications required to get it working.

chayim:
@dvora-h Can you give @utkarshgupta137 a hand with this?  How about integrating the documentation side - adding cluster examples, etc.

@utkarshgupta137 What else can we do to help? The last piece I see is the *test_acl_log* failure, at least from a starting position.

utkarshgupta137:
Looks like the tests are failing on 3.6 due to related issue https://github.com/aio-libs/aioredis-py/issues/1273. Maybe I can add try/except for now?

redis/asyncio/cluster.py
```diff
@@ -0,0 +1,1118 @@
...
+    def __del__(self):
+        try:
+            loop = asyncio.get_event_loop()
+            coro = self.close()
+            if loop.is_running():
+                loop.create_task(coro)
+            else:
+                loop.run_until_complete(coro)
+        except Exception:
+            pass
```
Andrew-Chen-Wang, line 266: I'd rather have it like the client itself which is to get the user to manually close (take a look at the client del code). Thoughts?

Andrew-Chen-Wang (commented):
just an initial read through. This looks great!

redis/commands/parser.py
```diff
@@ -12,7 +12,6 @@ class CommandsParser:
     """
 
     def __init__(self, redis_connection):
-        self.initialized = False
```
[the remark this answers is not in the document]
utkarshgupta137: This is not user anywhere in code.

redis/asyncio/cluster.py
utkarshgupta137: Agreed. It should also solve the CI issues. I'll update the PR.

redis/commands/parser.py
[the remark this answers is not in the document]
Andrew-Chen-Wang: we may need to keep it in for typing purposes to conform to mypy, but i'm not entirely sure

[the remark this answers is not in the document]
utkarshgupta137: I don't see any references to the `initialized` attribute in mypy documentation. If required, I can restore it.

redis/asyncio/cluster.py
```diff
@@ -0,0 +1,1118 @@
...
+        try:
+            if self.redis_connection is not None:
+                loop = asyncio.get_event_loop()
+                coro = self.redis_connection.close(True)
+                if loop.is_running():
+                    loop.create_task(coro)
+                else:
+                    loop.run_until_complete(coro)
+        except Exception:
+            pass
```
utkarshgupta137, line 765: @Andrew-Chen-Wang How should deletion for `ClusterNode` be handled?

redis/commands/parser.py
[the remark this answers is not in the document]
Andrew-Chen-Wang: not a reference in the docs; more so if you do `self.initialized` in a method without initializing it first in the constructor, mypy may raise an exception. I'm not sure tho

redis/asyncio/cluster.py
Andrew-Chen-Wang: correct me if I'm wrong, but on Cluster creation, these ClusterNodes are always with that one Cluster object, as in there is no dynamic changes to these ClusterNodes, right? If so, you can take a look at how I handle connection pools in async Redis's client class's close method that can be replicated in Cluster's close method. 

I have also not take a good look at the code yet, but this is just my initial thought

utkarshgupta137: No, that's the problem. In case of `MOVED` errors or persistent connection errors, the nodes_manager would be reinitialized, which could lead to some Nodes being removed.

Andrew-Chen-Wang:
gotcha. Cluster has an initialize async method. You can make a task that takes a ClusterNode as a param every time node manager creates a NodeCluster. Create a while loop with some small sleep time (because asyncio) that awaits for the task to be canceled. Cancel it when it needs to be removed. Once canceled (it’s an exception in the task itself), you can perform any disconnection on the ClusterNode within that loop.

Something like:

try:
    …
except CanceledError (I think):
    pass
finally:
    # disconnect
________________________________

redis/commands/parser.py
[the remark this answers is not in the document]
utkarshgupta137: This attribute is not used anywhere, not only in this constructor or class but the whole codebase.

utkarshgupta137:
@chayim @Andrew-Chen-Wang Most of the warnings have been resolved. The test_acl_log is still failing on 3.6, I'll check it tomorrow.

Andrew-Chen-Wang:
@chayim when are we dropping Python 3.6? Or start marking it as unneeded for merging.

utkarshgupta137:
test_acl_log is fixed now. It was causing an issue since I had an empty yield statement in async def.

chayim:
@utkarshgupta137 Dropping python 3.6 support isn't a 4.2 thing, but could happen as we start to clean up the repo in 4.3+/5.x.  There's a lot of 3.6 out there - even if I'd love to be 3.7+.

docs/conf.py
```diff
@@ -41,6 +42,10 @@
 autosectionlabel_prefix_document = True
 autosectionlabel_maxdepth = 2
 
+# AutodocTypehints settings.
+always_document_param_types = True
```
chayim, line 46: Awesome. Had no idea this was an option.

redis/commands/cluster.py
```diff
@@ -594,3 +911,326 @@ def readwrite(self, target_nodes=None):
...
+        - nodes flag: ALL_NODES, PRIMARIES, REPLICAS, RANDOM
+        - 'ClusterNode'
+        - 'list(ClusterNodes)'
+        - 'dict(any:clusterNodes)'
+
+    for example:
+        r.cluster_info(target_nodes=RedisCluster.ALL_NODES)
+    """
+
+    def cluster_myid(self, target_node: _TargetNodesT) -> Awaitable:
```
Andrew-Chen-Wang, line 944: code that just runs `self.execute_command` can be abstracted for both sync and async commands to avoid duplication for maintainability purposes.

redis/commands/cluster.py
```diff
@@ -594,3 +911,326 @@ def readwrite(self, target_nodes=None):
...
+        Similar to the CLUSTER DELSLOTS command.
+        The difference is that CLUSTER DELSLOTS takes a list of hash slots to remove
+        from the node, while CLUSTER DELSLOTSRANGE takes a list of slot ranges to remove
+        from the node.
+
+        For more information see https://redis.io/commands/cluster-delslotsrange
+        """
+        return self.execute_command("CLUSTER DELSLOTSRANGE", *slots)
+
+    def cluster_failover(
```
Andrew-Chen-Wang, line 1030: this function should be `async def` and the execute commands should await. This way, if someone tries to await a function that raises an error, a second error won't pop up saying you tried to await an exception (I can't remember if this applies to Python, but that seems to be my logic for previous AsyncCommand mixing).

redis/commands/cluster.py
```diff
@@ -15,8 +43,15 @@
 from .helpers import list_or_args
 from .redismodules import RedisModuleCommands
 
+if TYPE_CHECKING:
+    from redis.asyncio.cluster import ClusterNode
+
+_TargetNodesT = TypeVar(
```
Andrew-Chen-Wang, line 49: Could this theoretically be in the typing module instead? Or the Cluster file rather than the cluster commands file?

Andrew-Chen-Wang (changes requested):

redis/commands/cluster.py
utkarshgupta137: Right now, in the core commands module, we just use the sync commands methods instead of a separate methods with async/await. I just followed the same here.

redis/commands/cluster.py
utkarshgupta137: I was thinking of doing that too. Will update.

redis/commands/cluster.py
utkarshgupta137: Sure, makes sense.

utkarshgupta137:
Benchmarks agains aredis & aioredis-cluster: https://docs.google.com/spreadsheets/d/1jra5fz1k7trjD8vHSwjPaWhGqni5mfFbQqP1hpbNz7o/edit?usp=sharing

TLDR: `aredis` is the fastest in most cases, but `aioredis-cluster` is faster if running many commands with `asyncio.gather` . Also `aredis` maybe faster without `uvloop` if number of operations is less.

utkarshgupta137:
The optimization changes along with https://github.com/redis/redis-py/pull/2118 has brought the performance in line or even better in many cases with aredis & aioredis-cluser. aredis cluster is still faster if all commands are execute sequentially. See: https://docs.google.com/spreadsheets/d/1jra5fz1k7trjD8vHSwjPaWhGqni5mfFbQqP1hpbNz7o/edit#gid=640869726

redis/asyncio/parser.py
```diff
@@ -0,0 +1,104 @@
...
+    # https://github.com/redis/redis/pull/8324
+    async def get_keys(
+        self, redis_conn: "ClusterNode", *args
+    ) -> Optional[Union[List[str], List[bytes]]]:
+        if len(args) < 2:
+            # The command has no keys in it
+            return None
+
+        try:
+            command = self.commands[args[0]]
```
chayim, line 52: WDYT about catching IndexError here too, and responding to the user

redis/asyncio/parser.py
```diff
@@ -0,0 +1,104 @@
...
+        return args[command["first_key_pos"] : last_key_pos + 1 : command["step_count"]]
+
+    async def _get_moveable_keys(
+        self, redis_conn: "ClusterNode", *args
+    ) -> Optional[List[str]]:
+        pieces = []
+        cmd_name = args[0]
+        # The command name should be splitted into separate arguments,
+        # e.g. 'MEMORY USAGE' will be splitted into ['MEMORY', 'USAGE']
+        pieces = pieces + cmd_name.split()
```
chayim, line 91:
```
pieces += cmd_name.split() + list(args[1:])
```

chayim (changes requested):
@utkarshgupta137 This is really, really close. I added tiny comments - that to be honest, I'm okay with eliminating. Also, it works nicely :).

I think we can get this in a *4.3.0* release, immediately - as this necessitates a minor version bump.  WDYT?

redis/asyncio/parser.py
utkarshgupta137: There won't be an IndexError here due to check at L47

utkarshgupta137:
@chayim Can you hold off on merging this? I think making it thread-safe is trivial.

chayim:
Of course - but beware of scope creep.

utkarshgupta137:
@chayim According to https://docs.python.org/3/library/asyncio-dev.html#asyncio-multithreading, `asyncio` is not meant to be used with threading anyway, so I've removed the threading related checks.

Grokzen:
@utkarshgupta137 I am no longer part of working with redis clients here or in my own repo any longer. I have handed the future of cluster related things to this repo so there won't be any need for me to review or be a part of any cluster related development :)

The major good part of this PR is that you get asyncio into here, the major pain going forward tho will be that it seems you have to clone and duplicate the normal code and asyncio code =( I am sure maintaining that is going to be a much bigger job after this get in.

Is the plan to require any future modifications and contributions to make both normal and asyncio code compatible with any change? or will it be more on a "lets hope it will happen" basis?

utkarshgupta137:
@Grokzen I will carry over all the changes/optimisations I've made to async to the sync client soon. For the long term, there is a discussion in https://github.com/redis/redis-py/issues/2119 regarding moving to sansio approach. It is IMO a much better approach for supporting both sync/async & I can port the cluster code to it if we go with it.

chayim (approved):

chayim:
@utkarshgupta137 I just merged master in. We're *very*, *very* close!

karwinliu:
Does this support pipeline by any chance?

utkarshgupta137:
> Does this support pipeline by any chance?

Not at the moment. I just wanted to add basic support first and see if there are any issues. I will try to add pipeline support as well as a few other missing features like locks, pubsub, monitor etc.

karwinliu:
> 

Thanks @utkarshgupta137. What you have done on the async support is really helpful! Our project relies heavily on pipelines, so I wonder if there's any rough timeline you have in mind for these supports.

utkarshgupta137:
> > 
> 
> Thanks @utkarshgupta137. What you have done on the async support is really helpful! Our project relies heavily on pipelines, so I wonder if there's any rough timeline you have in mind for these supports, or if there's any workaround for the time-being.

I've been considering switching to pipelines as well. But between me writing the code, tests, review, & release, it may take 2-3 weeks. In the meantime, I think there are a couple of other packages with async cluster like aredis which may've support for it, but I'm not sure.

akx:
Oof, there's a lot of copy-paste in this PR... My PyCharm's Locate Duplicates is just lighting up :(

Issue #2103, opened 2022-04-13
Connection class disconnects on BaseException


kristjanvalur:
**Version**: 4.2.2

**Platform**: Windows 11

**Description**:  The methods `send_packed_command` and `read_response` on the `asyncio.connection.Connection` class
have exception handlers which catch `BaseException` and call `self.disconnect()` before re-raising.

This is rather unfortunate, because it makes it impossible to use an outer `Timeout` around redis methods.  For example, the following pattern fails:

```python
async def get_next_message_or_None():
        async with async_timeout.timeout(0.9) as self.timeout_manager:
                # blocking method to return messages
                async for message in self.pubsub.listen():
                        return message
```

This is because internally, the timeout will raise a `CancelledError` in the task.  This is a base exception and isn't converted into a TimeoutError until higher in the call stack.

Generally, BaseExceptions should not be caught.   In this case, it would be prudent to change this to catch `Exception` instead.

ikonst:
Strangely in #2499 I found the opposite to be true.

Similarly to asyncio's `CancelledError`, gevent's timeouts manifest as a base exception raised from a socket operation (an "await" in asyncio). With existing code, if a cancellation occurs while reading from the socket, the response stays in the socket's buffer and ends up being picked up as the response for the next parsed command, and at that point the redis connection is perpetually broken. Is this not an issue for asyncio code?

kristjanvalur:
No, it is the application which decides, in response to a Timeout, to either 

- retry the reading of the reponse or
- discard the connection.

The Timout should leave the connection in the state that it was, so that the application can decide how to respond to the timout.  Typically, the application can do something else, such as updating a progress bar, and then resume reading the response. 
Or, it can fail and give up and close the connection.

ikonst:
As discussed on another thread,

> it can fail and give up and close the connection

this is true for `PubSub` which has a `connection` property, but not for `Redis` when using a connection pool, since it doesn't know which connection is in unstable state.

kristjanvalur:
The connection pool is optional.   It is then, `ConnectionPool` which needs to make this decision.  It is the ConnectionPool which must not place a connection with an incompletely processed command into its pool.

Chronial:
[...]

PR #2104, opened 2022-04-13
Catch `Exception` and not `BaseException` in the `Connection`

kristjanvalur:
### Pull Request check-list

### Description of change

Fixes Issue #2103:
Raising a BaseException inside a `Connection`'s send or receive oprations causes a `disconnect`().  BaseExceptions are commonly used for _out of band_ messages in Python and normally not be expecially acted upon, only be handled at the top-level 

See also:  #2356 which focuses on a different aspcect of error handling.

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -502,8 +502,6 @@ class HiredisParser(BaseParser):
             # data was read from the socket and added to the buffer.
             # return True to indicate that data was read.
             return True
-        except asyncio.CancelledError:
-            raise
         except (socket.timeout, asyncio.TimeoutError):
             if raise_on_timeout:
                 raise TimeoutError("Timeout reading from socket") from None
@@ -721,7 +719,7 @@ class Connection:
                 lambda: self._connect(), lambda error: self.disconnect()
             )
         except asyncio.CancelledError:
-            raise
+            raise  # in 3.7 and earlier, this is an Exception, not BaseException
         except (socket.timeout, asyncio.TimeoutError):
             raise TimeoutError("Timeout connecting to server")
         except OSError as e:
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -916,7 +916,7 @@ class Connection:
             raise ConnectionError(
                 f"Error {err_no} while writing to socket. {errmsg}."
             ) from e
-        except BaseException:
+        except Exception:
             await self.disconnect()
             raise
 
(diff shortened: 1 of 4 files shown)


WisdomPill (commented):
LGTM!

Can you add some tests for it?

chayim:
@kristjanvalur Why did you catch Exception rather that BaseException? Was there an inheritance issue you saw?

WisdomPill:
@chayim have a look here #2103

kristjanvalur:
I can add tests.  As I explained in the defect, `BaseException` is usually not caught except in top level handlers, unless it is to implement something like `async_timeout.Timeout()`  .  `BaseException` was added to Python precisely to be a sort of _meta_ exception that regular exception handling doesn't touch, the `Exception` being the most general class usually touched in code that isn't trying to be too clever.

kristjanvalur:
In fact, I don't see why there is a `except Exception: self.disconnect(); raise` at all. Doesn't seem like something the Connection should do if it gets an _unexpected_ exception.  IMHO, this clause should be removed alltogether.

kristjanvalur:
Hm, now it is mysteriously failing.  it asserts on the "assert  pubsub.connection.is_connected"
I´m running tests both on windows and in a linux dev container using
`pytest tests/test_async/test_pubsub.py`.   I don't get the test failures above.

kristjanvalur:
Well, it seems I cracked the unittest problem I was having.  Turned out that only in Python 3.8 did asyncio.CancelledError get promoted to BaseException.
Fixed up the test to skip older python versions and added Mock-based tests for exception propagation for both async and blocking redis.

kristjanvalur:
any responses?  @WisdomPill ?

kristjanvalur:
I'll clean up the commit history a bit...

kristjanvalur:
yet another fluke test failure.

kristjanvalur:
I added the Connection.disconnect(nowait) optional arg.  When `true`, the disconnect() call will not attempt to wait for the flushing of the write buffer during close().  This is an async operation, and when calling "disconnect" as a response to an error, the async close can cause other errors to be raised.  Notably, if running within a "timeout" context, a `CancelledError` may get thrown.

In the case of an error, there should be no need to wait for a flush of the write buffer.

akx:
I bumped into this too – trying to e.g. ctrl+c out of py.test requires a couple of tries because `KeyboardInterrupt` (a `BaseException`) is caught where it really must not be.

dvora-h (approved):

kristjanvalur:
woo-hoo!

dvora-h:
Thanks for this! And sorry this (and your other PR's) waited so long...

PR #2108, opened 2022-04-14
Add typings

soukyomi:
### Pull Request check-list
  - **NOTE:** The only failed check didn't fail because of changes of this PR.

### Description of change
Add typings for all components (functions and classes) of the library.

### Typing progress
- [ ] `__init__.py`
- [ ] `backoff.py`
- [ ] `client.py`
- [ ] `cluster.py`
- [ ] `compat.py`
- [ ] `connection.py`
  - **WIP:** [■■■■----------------] 22.38% completed (15/67)
- [ ] `crc.py`
- [ ] `exceptions.py`
- [ ] `lock.py`
- [ ] `ocsp.py`
- [ ] `retry.py`
- [ ] `sentinel.py`
- [ ] `typing.py`
- [ ] `utils.py`
- [ ] `redis.commands` module
- [ ] `redis.asyncio` module

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -5,12 +5,23 @@ import os
 import socket
 import sys
 import threading
-import weakref
 from itertools import chain
 from queue import Empty, Full, LifoQueue, Queue
 from time import time
-from typing import Any, Dict, List, Optional, Set, Type
+from typing import (
+    Any,
+    Dict,
+    List,
+    Mapping,
+    Optional,
+    Protocol,
+    Set,
+    Tuple,
+    Type,
+    Union,
+)
 from urllib.parse import parse_qs, unquote, urlparse
+from weakref import WeakMethod
 
 from packaging.version import Version
 
@@ -84,6 +95,11 @@ MODULE_EXPORTS_DATA_TYPES_ERROR = (
     "exports one or more module-side data "
     "types, can't unload"
 )
+
+
+class ConnectCallbackProtocol(Protocol):
+    def __call__(self, connection: "Connection") -> None:
+        ...
 
 
 class Encoder:
@@ -510,30 +526,60 @@ else:
 class Connection:
     "Manages TCP communication to and from a Redis server"
 
+    __slots__ = (
+        "pid",
+        "host",
+        "port",
+        "db",
+        "username",
+        "client_name",
+        "password",
+        "socket_timeout",
+        "socket_connect_timeout",
+        "socket_keepalive",
+        "socket_keepalive_options",
+        "socket_type",
+        "retry_on_timeout",
+        "retry_on_error",
+        "retry",
+        "health_check_interval",
+        "next_health_check",
+        "redis_connect_func",
+        "encoder",
+        "_sock",
+        "_socket_read_size",
+        "_connect_callbacks",
+        "_buffer_cutoff",
+        "_parser",
+    )
+
     def __init__(
         self,
-        host="localhost",
-        port=6379,
-        db=0,
-        password=None,
-        socket_timeout=None,
-        socket_connect_timeout=None,
-        socket_keepalive=False,
-        socket_keepalive_options=None,
-        socket_type=0,
-        retry_on_timeout=False,
-        retry_on_error=[],
-        encoding="utf-8",
-        encoding_errors="strict",
-        decode_responses=False,
-        parser_class=DefaultParser,
-        socket_read_size=65536,
-        health_check_interval=0,
-        client_name=None,
-        username=None,
-        retry=None,
-        redis_connect_func=None,
-    ):
+        *,
+        host: str = "localhost",
+        port: Union[str, int] = 6379,
+        db: Union[str, int] = 0,
+        password: Optional[str] = None,
+        socket_timeout: Optional[float] = None,
+        socket_connect_timeout: Optional[float] = None,
+        socket_keepalive: bool = False,
+        socket_keepalive_options: Optional[
+            Mapping[int, Union[int, bytes]]
+        ] = None,
+        socket_type: int = 0,
+        retry_on_timeout: bool = False,
+        retry_on_error: List[Type[Exception]] = [],
+        encoding: str = "utf-8",
+        encoding_errors: str = "strict",
+        decode_responses: bool = False,
+        parser_class: Type[BaseParser] = DefaultParser,
+        socket_read_size: int = 65536,
+        health_check_interval: float = 0,
+        client_name: Optional[str] = None,
+        username: Optional[str] = None,
+        retry: Optional[Retry] = None,
+        redis_connect_func: Optional[ConnectCallbackProtocol] = None,
+    ) -> None:
         """
         Initialize a new Connection.
         To specify a retry policy for specific errors, first set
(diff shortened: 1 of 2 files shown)


redis/connection.py
```diff
@@ -1371,7 +1381,13 @@ def _checkpid(self):
...
                 self._fork_lock.release()
 
-    def get_connection(self, command_name, *keys, **options):
+    def get_connection(
+        self,
+        command_name: str,
+        # TODO: Type the next two parameters.
+        *keys,
+        **options
+    ) -> Connection:
```
soukyomi, line 1390: I couldn't figure out what types the x and y parameters take. If anyone knows please tell me.

soukyomi, line 1390: I couldn't figure out what types the `*keys` and `**options` parameters take. If anyone knows please tell me.

utkarshgupta137: https://google.github.io/pytype/faq.html#how-do-i-annotate-args-and-kwargs

utkarshgupta137:
I would recommend using https://github.com/Instagram/MonkeyType or https://github.com/google/pytype to get started. MonkeyType should help in adding types everywhere & pytype should find more types after monkeytyping is done.

redis/connection.py
soukyomi: > https://google.github.io/pytype/faq.html#how-do-i-annotate-args-and-kwargs

I know how to type `*args` and `**kwargs`, the question was what do these two parameters take (only integers? Strings? etc.)

soukyomi:
> I would recommend using https://github.com/Instagram/MonkeyType or https://github.com/google/pytype to get started. MonkeyType should help in adding types everywhere & pytype should find more types after monkeytyping is done.

Neither of these two libraries helped me much at first so I'm having to manually type (not that it's really a problem).

soukyomi:
I will start some rewriting on library's code because some functions/methods have some redundancies. In Python, we should all be consenting adults when programming and we shouldn't parse every parameter passed into a function. We should just consider that the user is passing the correct type when using a function/method (especially when dealing with internals, where the user should not have access). 

I'm going to spend a few days working on this and then release a fairly large commit, after which I need someone to review my code, thanks.

redis/connection.py
chayim: The right thing to do is set a minimum of:
*keys: List
**options: Dict

Those are accurate typings. It doesn't help the sub options of the various items - but it's realistic.

utkarshgupta137: @chayim For `*args` & `**kwargs`, the correct way to type them is to specify the inner type, not List or Dict (that is assumed). See the link I shared.

redis/connection.py
```diff
@@ -1390,7 +1524,7 @@ def get_connection(self, command_name, *keys, **options):
             # closed. either way, reconnect and verify everything is good.
             try:
                 if connection.can_read():
-                    raise ConnectionError("Connection has data")
+                    raise ConnectionEr`ror("Connection has data")
```
Andrew-Chen-Wang, line 1527:
```suggestion
                    raise ConnectionError("Connection has data")
```

redis/connection.py
```diff
@@ -1527,7 +1670,7 @@ def reset(self):
 
         # Keep a list of actual connection instances so that we can
         # disconnect them later.
-        self._connections = []
+        self._connections: List[Connection] = []
```
Andrew-Chen-Wang, line 1673: if you're using `__future__` annotations, then you don't need to use `List[Connection]`; built-in list is ok: `list[Connection]` which removes an import from typing

redis/connection.py
```diff
@@ -1143,15 +1241,20 @@ def _error_message(self, exception):
 FALSE_STRINGS = ("0", "F", "FALSE", "N", "NO")
 
 
-def to_bool(value):
+def to_bool(value: Optional[str]) -> Optional[bool]:
     if value is None or value == "":
         return None
-    if isinstance(value, str) and value.upper() in FALSE_STRINGS:
+    if value.upper() in FALSE_STRINGS:
```
Andrew-Chen-Wang, line 1247: Please do not reimplement functionality or make a potentially breaking change when it comes to adding typing support; we don't know what developers are doing, and not everyone likes typing and thus won't follow the rules.

Andrew-Chen-Wang (changes requested):
thanks for this!

redis/connection.py
soukyomi: Hi, thanks for the feedback. I removed this code snippet as Mypy was complaining about redundant code (`value` will always be string in this code snippet, so there's no reason to check if it's an instance of it).

> we don't know what developers are doing

Thanks for the sincerity, but that part of the sentence really worries me. If you guys don't know what's being put in the code itself, it can be inferred that bad or even malicious code can be inserted into the library (of course, I'm exaggerating a lot, but I think you get the point).

In Python, it is inferred that we are [consenting adults](https://mail.python.org/pipermail/tutor/2003-October/025932.html), this means that the types passed are considered to be the correct ones, and if you are not sure about this, it is safer to create a CI by running a type checker on every commit/pull request (and require every pull request to be properly typed).

If you want, I can create a workflow that runs a type checker in this pull request. That way you guys will be sure that the types are correct.

Andrew-Chen-Wang: haha I'm sure the types seem fine based on which commands use this util. We know what the function is for. What we don't know is whether a user has imported the function for themselves.

redis/connection.py
soukyomi: That's ok, I'm just worried about the code conflicting until then. Anything I fix, anyway. I'll first do a commit using `typing.List` and then do another one by fixing it.

redis/connection.py
soukyomi: > What we don't know is whether a user has imported the function for themselves.

I'm not sure I understand this part of the answer. By "user" do you mean the person who will be using the library? If yes, this answer makes no sense. It's the user's responsibility to pass the correct type to the parameter, and this function isn't even in the documentation so there's no chance of anyone using it (that's why I considered it as an internal).

redis/connection.py
WisdomPill: I am afraid that `list[Connection]` is just a python3.10 thing

WisdomPill:
My two cents.

I think this PR will be too long and big, maybe you could split it for every different module?

It is just not healthy to make big PRs, every time somebody merges something you will need to rebase and probably solve merge conflicts this is going to touch almost every file.

If we split this I would be more than happy in stepping in and type some module as well. I already have in my todo list to make `lock` more consistent

soukyomi:
> First and foremost, thanks for all the great work!
> 
> My two cents.
> 
> I think this PR will be too long and big, maybe you could split it for every different module?
> 
> It is just not healthy to make big PRs, every time somebody merges something you will need to rebase and probably solve merge conflicts this is going to touch almost every file.
> 
> If we split this I would be more than happy in stepping in and type some module as well. I already have in my todo list to make `lock` more consistent

Hi, thanks for the reply! It is a pleasure to be able to help with the project.

While I was developing this PR, I also thought the same thing, but I was afraid to do it. I fully agree that large PR are unhealthy for both the person developing it and the people reviewing it. I'll try to split Pull Request into others, thanks again!

redis/connection.py
Andrew-Chen-Wang: it is also available via future.annotations

WisdomPill: that's interesting... thanks for letting me know

WisdomPill:
@soukyomi would you like to divide the work? maybe we can create different issues for every module and assign them?

soukyomi:
> @soukyomi would you like to divide the work? maybe we can create different issues for every module and assign them?

Hi! I would like to share the work, but I should only deal with it next week.

WisdomPill:
@soukyomi let's create separate issues for every module? 

@chayim should we take this conversation to a discussion thread?

PR #2114, opened 2022-04-19
Support password with special characters

bale836:
When the password contains "#, @, : ?, etc" which is used by urlparse(), the redis cannot to connect, fix this scenario

### Pull Request check-list

### Description of change


WisdomPill (commented):
Technically you should unquote every part used for the connection, like username.

Furthermore there are other types of connections, I would recommend to use a url in the meantime like [here](https://github.com/redis/redis-py/blob/b721eedd35d52835a96722c95b89ac60660f5211/redis/connection.py#L1274)

chayim:
@dvora-h can you please have a look at this? I agree with @WisdomPill, we should be running unquote on parts.  Perhaps not username however, as usernames must be  /a-zA-Z0-9/

PR #2118, opened 2022-04-21
cluster: add C speedups for key_slot

utkarshgupta137:
### Pull Request check-list

### Description of change

Cherry-picked: https://github.com/NoneGG/aredis/pull/57 & https://github.com/NoneGG/aredis/pull/59

C version is approx. 4x faster than the python version.


utkarshgupta137:
@chayim @dvora-h

utkarshgupta137:
@chayim @dvora-h Thoughts?

utkarshgupta137:
@chayim @dvora-h Any updates on this? Will you consider merging this in?
If the maintainability of C code is an issue, I can also write this in Rust.
Ideally, we should move the whole Node determination part of the Cluster client into C/Rust.

chayim:
I think the right answer is using the rust bridge - agreed. However, I'm worried about general maintainability in the client - this is my only concern.

As long as users without compilers (let's call them base users) can have what they need to work via a stock pip install, I see this being sane. But - rust over C, since this would at least be in line with what other modules (i.e cryptography) has done.

@dvora-h comments?

utkarshgupta137:
I have not worked with rust in python yet, so it may take time to get started. I'll see when I've time.

karwinliu:
> With the C version, the normal requests are ~5% faster & pipeline requests are ~15% faster.

Bumped into this PR while going over the open ones with some other bug fixes, and this is really awesome stuff as we are actually suffering from some performance issues and `pack_commands` indeed stood out from our profiling. Will keep a close look

akx:
Sounds like these speedups should be shipped as a separate opt-in module like `hiredis` is a separate opt-in parser for the protocol... (Even better, maybe this should be part of `hiredis`?)

chayim:
> Sounds like these speedups should be shipped as a separate opt-in module like `hiredis` is a separate opt-in parser for the protocol... (Even better, maybe this should be part of `hiredis`?)

I think these belong in redis-py rather than hiredis. However, as mentioned, they really should be in rust. setuptools and core changes recently have led to embracing rust for these things - and it's something I fully support.

Is anyone interested in helping @dvora-h associate these in rust?

chayim:
The upside, and downside of the size of this community is the number of PRs, and the speed of change :)

I only see this one opened, as authored by you. Is there another one I'm missing?

utkarshgupta137:
> The upside, and downside of the size of this community is the number of PRs, and the speed of change :)
> 
> I only see this one opened, as authored by you. Is there another one I'm missing? 

My PRs got merged in July itself, but there has been no stable release since then. I'm unable to use this library in production due to a critical bug in 4.3.x async cluster. I would first like to start using it in production, before making more contributions.

chayim:
True. 4.4.x is so big that we're taking our time with milestones. Much like a couple of other big releases. FYI I'd like to have a 4.4 release shortly.

PR #2155, opened 2022-04-30
Fix incorrect return annotation in asyncio.lock

JelleZijlstra:
### Pull Request check-list

### Description of change

NoReturn should be used only when the function never returns. In this case, the awaitable returns None if releasing the lock succeeds, so `Awaitable[None]` is right.

Noticed this while reviewing python/typeshed#7676


chayim:
Thank you @JelleZijlstra this is great! We'd love to get everything, properly typed, inserting a py.typed when ready. Keep 'em coming!

PR #2198, opened 2022-05-24
Add `query_params` to FT.PROFILE

dvora-h:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/commands/search/commands.py
+++ b/redis/commands/search/commands.py
@@ -1,6 +1,6 @@
 import itertools
 import time
-from typing import Dict, Union
+from typing import Dict, Optional, Union
 
 from redis.client import Pipeline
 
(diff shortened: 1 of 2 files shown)


redis/commands/search/commands.py
```diff
@@ -506,7 +506,7 @@ def _get_aggregate_result(self, raw, query, has_cursor):
 
         return AggregateResult(rows, cursor, schema)
 
-    def profile(self, query, limited=False):
+    def profile(self, query, query_params=None, limited=False):
```
chayim, line 509: This now changes the API order - which makes it breaking.  Can we instead put query_params after limited?

chayim: We should probably also change the documentation to help people understand that it's a query + query_params.

redis/commands/search/commands.py
```diff
@@ -506,16 +508,17 @@ def _get_aggregate_result(self, raw, query, has_cursor):
 
         return AggregateResult(rows, cursor, schema)
 
-    def profile(self, query, limited=False):
+    def profile(self, query, limited=False, query_params=None):
```
chayim, line 511: Since we're already here can we add type hints to this function

PR #2201, opened 2022-05-26
set default response_callbacks to redis.asyncio.cluster.ClusterNode

rapidia:
### Pull Request check-list

### Description of change

`asyncio.cluster.ClusterNode` needs default response_callbacks, not `None`.

When I initialize `RedisCluster` with `startup_nodes`(instead `host=` and `port=`), `response_callbacks` of ClusterNode will be `None`, unless I specify `response_callbacks`.
It could be assigned when create ClusterNode, but having default value looks more reasonable.

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -725,7 +725,7 @@ class ClusterNode:
         server_type: Optional[str] = None,
         max_connections: int = 2 ** 31,
         connection_class: Type[Connection] = Connection,
-        response_callbacks: Dict = None,
+        response_callbacks: Dict = RedisCluster.RESPONSE_CALLBACKS,
         **connection_kwargs,
     ) -> None:
         if host == "localhost":
--- a/tests/test_asyncio/test_cluster.py
+++ b/tests/test_asyncio/test_cluster.py
@@ -2230,3 +2230,20 @@ class TestNodesManager:
                 async with RedisCluster(startup_nodes=[node_1, node_2]) as rc:
                     assert rc.get_node(host=default_host, port=7001) is not None
                     assert rc.get_node(host=default_host, port=7002) is not None
+
+
+@pytest.mark.onlycluster
+class TestClusterNode:
+    """
+    Tests for the ClusterNode class
+    """
+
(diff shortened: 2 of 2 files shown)


utkarshgupta137:
+1. Would you be able to add a test as well?

rapidia:
@utkarshgupta137
I don't know how to test this case, properly. because, this is a just assignment of a value, not a logic.

basically, If `response_callbacks` is None, `execute_command` function will be [brake](https://github.com/redis/redis-py/blob/9167a0e10d5acd9d76edd887300c75cfeb07a7c3/redis/asyncio/cluster.py#L824) with `TypeError: 'NoneType' object is not subscriptable`.

in the test case of `test_startup_nodes`, you are [mocking execute_command](https://github.com/redis/redis-py/blob/9167a0e10d5acd9d76edd887300c75cfeb07a7c3/tests/test_asyncio/test_cluster.py#L91-L104), so you couldn't find this problem.

I can check response_callbacks of ClusterNode is not None. but still, you can set response_callbacks to None at runtime. I don't think it's a meaningful test case.

to prevent response_callbacks to None, there could be lots of ways.
I could make getter/setter for response_callbacks. but, I think it's getting out of purpose of this PR.

utkarshgupta137:
Even if response_callbacks is a dictionary, then it would fail at https://github.com/redis/redis-py/blob/master/redis/asyncio/cluster.py#L972. So you need to add a test along the lines of:
```
node = ClusterNode(host="localhost", port=16379)
rc = RedisCluster(startup_nodes=[node])
assert await rc.set("A", 1)
assert await rc.get("A") == 1
```

utkarshgupta137:
The test looks good. But its scope is not limited to just this argument. It would be better to merge it into [test_startup_nodes](https://github.com/redis/redis-py/blob/master/tests/test_asyncio/test_cluster.py#L228).

utkarshgupta137:
Awesome! @chayim @dvora-h This is a somewhat major bug, it should be released on priority.

rapidia:
@utkarshgupta137 thank you for your advice 👍🏼

dvora-h (approved):
@rapidia Great catch, and a great PR, thank you!

PR #2213, opened 2022-06-03
Fix timezone handling for datetime to unixtime conversions

joekohlsdorf:
### Pull Request check-list

### Description of change

datetime objects are supported to set expire, these can have timezones.
mktime was used to convert these to unixtime. mktime in Python however is not timezone aware, it expects the input to be UTC and redis-py did not convert the datetime timestamps to UTC before calling mktime.

This can lead to:
1) When the datetime timestamp is within DST, mktime fails with "OverflowError: mktime argument out of range" because UTC doesn't have DST. This depends on libc versions.
2) Setting incorrect expire times because the input datetime object has a timezone but is passed to mktime without converting to UTC first.

Since Python 3.3 datetime has a timestamp function which converts to UTC before calling mktime internally.
I changed all direct uses of mktime to use this instead.

### How to reproduce the issues

DST:
```
import datetime
import time
from dateutil.tz import tzfile

dt = datetime.datetime(2000, 1, 1, 0, 0, 0, 0, tzinfo=tzfile('/usr/share/zoneinfo/America/Santiago'))
print(dt.timestamp())
# this will fail with OverflowError because the timezone has DST flag set to 1 at this date
print(time.mktime(dt.timetuple()))
```

No DST, conversion works but results in incorrect Unix timestamp due to non-timezone aware conversion:
```
import datetime
import time
from dateutil.tz import tzfile

dt = datetime.datetime(2000, 6, 1, 0, 0, 0, 0, tzinfo=tzfile('/usr/share/zoneinfo/America/Santiago'))
print(dt.timestamp())
# conversion works because DST flag set to 0 but unixtime is wrong
print(time.mktime(dt.timetuple()))
```


redis/commands/core.py
```diff
@@ -1762,14 +1761,12 @@ def getex(
         if exat is not None:
             pieces.append("EXAT")
             if isinstance(exat, datetime.datetime):
-                s = int(exat.microsecond / 1000000)
```
joekohlsdorf, line 1765: This can be removed because the maxvalue for `microsecond` is 999999 resulting in `s` being 0 for any valid value.

redis/commands/core.py
```diff
@@ -1762,14 +1761,12 @@ def getex(
...
-                exat = int(time.mktime(exat.timetuple())) + s
+                exat = int(exat.timestamp())
             pieces.append(exat)
         if pxat is not None:
             pieces.append("PXAT")
             if isinstance(pxat, datetime.datetime):
-                ms = int(pxat.microsecond / 1000)
```
joekohlsdorf, line 1771: This can be removed because `timestamp()` returns a fractional result and multiplying with 1000 will therefore already consider ms.

dvora-h:
@joekohlsdorf Sorry it took so long to respond. The PR looks good! I just merged master in to check if it works with the latest changes and we can approve & merge it

dvora-h (approved):

Issue #2119, opened 2022-04-24
[Discussion] Revisit How We Handle Async and Sync


chayim:
As of now, redis-py has support for async, and sync usage. The goal is to continue having these within the same library.

However, things seem 'off', or at least unwieldy when handling both cases.  While the APIs need to stay the same, this Issue exists to discuss in full - how we can improve this.  Let's start this conversation here.

Relevant conversation in another PR: https://github.com/redis/redis-py/pull/2096#discussion_r847382190

CC: @Andrew-Chen-Wang, @dvora-h, @WisdomPill 
Also please feel free to add any others :+1:

Andrew-Chen-Wang:
I guess first obligatory question is what are your current pain points?

One thing that pains me was the size of commands/core.py. Potential solutions:

1. A pyi file to store annotations and docs
2. Use `__call__` to execute methods that only simply return `execute_command`
3. Put mixins in their own dedicated file

I don't think there's a way for us to combine async and sync methods of core classes like Redis and AsyncRedis since I/O in Python is split by how functions are implemented (eg utilization of async/await keywords). It's just the troubles of maintaining a sync and async API in Python -- or at least as far as I can tell given Elasticsearch-py, Django, and psycopg. 

cc @seandstewart

seandstewart:
Hey folks - 

This is precisely why I opened https://github.com/redis/redis-py/issues/2039.

We need to move our protocol implementation for serialization and deserialization away from the IO boundary so that we can reduce the amount of doubled implementation but also so that we can properly type our IO implementations without violating the Liskov substitution principle.

seandstewart:
This will also make testing protocols far simpler!

Andrew-Chen-Wang:
I'm +1 on the SansIO approach for the aforementioned reasons.

I think there's a steep learning curve to it though (short docs, good talk; difficult to gain that particular mindset). It just reminds me of the complicated structure, at least to me, of aioredis 1.3.1. I also think once PubSub is implemented SansIO style (maybe take inspiration from wsproto? I'm also just reading up on the `websockets` package's move towards SansIO in addition to how their integrations worked with packages that did both HTTP and ws), it would get the go-ahead and the current implementation would become legacy.

WisdomPill:
First time I saw this SansIO and it looks promising and clean.

@seandstewart thanks for bringing this up.

One question, how will this make typing behave?
(I just started reading about SansIO, so this might be a stupid question)

[...]

PR #2217, opened 2022-06-05
async_cluster: fix max_connections/ssl & improve args

utkarshgupta137:
### Pull Request check-list

### Description of change

- set proper connection_class if ssl = True
- pass max_connections/connection_class to ClusterNode
- pass parser_class to Connection instead of changing it in on_connect
- only pass redis_connect_func if read_from_replicas = True
- add connection_error_retry_attempts parameter
- skip is_connected check in acquire_connection as it is already checked in send_packed_command

BREAKING:
- RedisCluster args except host & port are kw-only now
- RedisCluster will no longer accept unknown arguments
- RedisCluster will no longer accept url as an argument. Use RedisCluster.from_url
- RedisCluster.require_full_coverage defaults to True
- ClusterNode args except host, port, & server_type are kw-only now

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -17,7 +17,13 @@ from typing import (
 )
 
 from redis.asyncio.client import ResponseCallbackT
-from redis.asyncio.connection import Connection, DefaultParser, Encoder, parse_url
+from redis.asyncio.connection import (
+    Connection,
+    DefaultParser,
+    Encoder,
+    SSLConnection,
+    parse_url,
+)
 from redis.asyncio.parser import CommandsParser
 from redis.client import EMPTY_RESPONSE, NEVER_DECODE, AbstractRedis
 from redis.cluster import (
@@ -42,6 +48,7 @@ from redis.exceptions import (
     ConnectionError,
     DataError,
     MasterDownError,
+    MaxConnectionsError,
     MovedError,
     RedisClusterException,
     ResponseError,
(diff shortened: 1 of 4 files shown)


redis/asyncio/cluster.py
```diff
@@ -234,87 +208,132 @@ def from_url(cls, url: str, **kwargs: Any) -> "RedisCluster":
     def __init__(
         self,
         host: Optional[str] = None,
-        port: int = 6379,
+        port: Union[str, int] = 6379,
+        *,
```
dvora-h, line 212: @utkarshgupta137 Why do you think it's better to force send arguments as kw only?

chayim: This starts us down a path split between the two connection objects.
Having that conversation is reasonable, but IMHO not in the 4.x series. It should be part of a general architectural change (i.e maybe sansio and friends).

chayim: Realistically, I think the issue is that we don't have the same arguments for all (supported) connection types where possible.  I'm okay with adding arguments to sync these, as long as we don't change ordering.

Basically - no backwards compatibility issues.

utkarshgupta137:
@chayim @dvora-h Any updates on this?

utkarshgupta137:
@chayim @dvora-h Would dropping the kw-only required be enough to get this merged for now? I'm still strongly in favour of it, but this PR contains other important fixes which should be merged while we continue to have that discussion.

redis/asyncio/cluster.py
chayim: I think the two clients actually should probably inherit from a common base, then work it through there...

chayim:
I think this is fine. @dvora-h want to do another run through to make sure there isn't an argument that's **inadvertently removed (added is fine)**. I think this belongs in 4.4.x

redis/asyncio/cluster.py
utkarshgupta137: This is actually how it works in the [rust client](https://docs.rs/redis/latest/redis/struct.Client.html#method.get_connection). Although, in their case, the base class is the client from which you get sync or async connections. I think that is a better experience from a user perspective, but it would be a breaking change and probably needs to be discussed in #2119.

utkarshgupta137:
Ok, in that case, I'll just rebase this PR.

dvora-h:
@utkarshgupta137 @chayim  I think that except the kwargs issue - this can be merged

dvora-h (approved):

utkarshgupta137:
@chayim @dvora-h It has been 3 months since this PR was raised & 2 months since it was merged. This is a major bug fix which should've been released instantly. Can we please create a minor release with this and other async cluster related bug fixes?

PR #2232, opened 2022-06-15
Correct retention_msecs value

earthgecko:
### Description of change

The `retention_msecs` takes a millisecond value not a seconds value, having checked and tested redis/commands/timeseries/commands.py is not adjusting seconds to milliseconds it is passing it through as is, therefore the statement in this doc is incorrect as it creates a time series with a retention period of 5 milliseconds not 5 seconds as stated.


dvora-h (approved):

dvora-h:
@earthgecko Thank you for helping make our documentation better!

PR #2236, opened 2022-06-16
Uppercase commands in CommandsParser.get_keys

falk-h:
### Pull Request check-list

I haven't been able to get `tox` to pass locally. I'm unsure what the problem is, but it seems like it's related to `uvloop`. I'm sure it's just a problem with my dev environment as everything passes in CI.

### Description of change

I ran into issues when migrating from the blocking `RedisCluster` to `asyncio.RedisCluster` where `asyncio.CommandsParser.get_keys` would raise errors about commands not existing, even though everything worked with the blocking client. Passing the commands to `asyncio.RedisCluster.execute_command` as uppercase strings fixed the problem.

This change makes sure that commands passed to `asyncio.CommandsParser.get_keys` are always uppercased when used with the `self.commands` dict. This behavior is consistent with the blocking client. I thought that the commands maybe should be uppercased before being passed to `get_keys` but I haven't been able to find any code that looks like it does that. For reference, I was using commands from a proprietary Redis module.

### DIFF
--- a/redis/asyncio/parser.py
+++ b/redis/asyncio/parser.py
@@ -55,14 +55,14 @@ class CommandsParser:
             # try to split the command name and to take only the main command
             # e.g. 'memory' for 'memory usage'
             args = args[0].split() + list(args[1:])
-            cmd_name = args[0]
+            cmd_name = args[0].upper()
             if cmd_name not in self.commands:
                 # We'll try to reinitialize the commands cache, if the engine
                 # version has changed, the commands may not be current
                 await self.initialize()
                 if cmd_name not in self.commands:
                     raise RedisError(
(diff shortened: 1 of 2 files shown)


redis/asyncio/parser.py
```diff
@@ -50,19 +50,19 @@ async def get_keys(self, *args: Any) -> Optional[Tuple[str, ...]]:
             return None
 
         try:
-            command = self.commands[args[0]]
+            command = self.commands[args[0].upper()]
```
utkarshgupta137, line 53: Since most users won't use raw commands, we should only keep this in the except block. Please remove this change.

dvora-h (approved):
@falk-h Nice catch, thank you!

PR #2244, opened 2022-06-22
Added dynamic_startup_nodes configuration to RedisCluster.

barshaul:
### Pull Request check-list

### Description of change

Added dynamic_startup_nodes configuration to RedisCluster.
By default, when reinitializing the cluster topology, the nodesManager will use only the initial passed startup nodes to refresh the cluster topology. If dynamic_startup_nodes is set to true, the nodesManager will set the startup nodes to all of the discovered nodes and they will be used to determine the cluster nodes-slots mapping in the next topology refresh.

When true, the initial passed startup nodes will be removed if their endpoints aren't listed in the CLUSTER SLOTS output. For example, if dynamic DNS endpoints are used for startup nodes, but CLUSTER SLOTS lists specific IP addresses, the passed startup nodes will be overridden with the specific nodes' IP. 
Therefore, it is recommended to keep dynamic_startup_nodes set to false with DNS endpoints.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -11,7 +11,7 @@
     * Fix broken connection writer lock-up for asyncio (#2065)
     * Fix auth bug when provided with no username (#2086)
     * Fix missing ClusterPipeline._lock (#2189)
-
+    * Added dynaminc_startup_nodes configuration to RedisCluster
 * 4.1.3 (Feb 8, 2022)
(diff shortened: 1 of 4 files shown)


dvora-h (approved):

barshaul:
> @barshaul tests fails after merge master in.

taking a look.

PR #2252, opened 2022-06-27
Fix `XAUTOCLAIM` to return the full response

### LINKED ISSUE #2226: xautoclaim discards first reply part (id of last scanned stream item)
**Version**: redis-py 4.3.3, redis 7.0.0

**Platform**: python 3.9, ubuntu 20.04

**Description**: I tried to implement the iterative XAUTOCLAIM as described in [Redis docs](https://redis.io/commands/xautoclaim/):

> The command returns the claimed entries as an array. It also returns a stream ID intended for cursor-like use as the <start> argument for its subsequent call. When there are no remaining PEL entries, the command returns the special 0-0 ID to signal completion.
> 
> ...
> 
> Return
>Array reply, specifically:
>An array with three elements:
> 1. A stream ID to be used as the <start> argument for the next call to XAUTOCLAIM.
> 2. An array containing all the successfully claimed messages in the same format as [XRANGE](https://redis.io/commands/xrange).
> 3. An array containing message IDs that no longer exist in the stream, and were deleted from the PEL in which they were found.

Unfortunately, I found out that redis-py drops first element of array completely and returns just the second part (claimed messages), as can be seen in [parse_xautoclaim](https://github.com/redis/redis-py/blob/master/redis/client.py#L310). So it is not possible to figure out the start_id for the next call.

What is the rationale behind that decision? It looks like there are cases when id of last scanned stream item is quite useful.

dvora-h:
### Pull Request check-list

### Description of change

Fixes #2226

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -310,7 +310,8 @@ def parse_xclaim(response, **options):
 def parse_xautoclaim(response, **options):
     if options.get("parse_justid", False):
         return response[1]
-    return parse_stream_list(response[1])
+    response[1] = parse_stream_list(response[1])
+    return response
 
 
 def parse_xinfo_stream(response, **options):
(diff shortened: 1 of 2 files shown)


chayim (approved):
Super tiny + clean. Nice!

PR #2261, opened 2022-07-05
CredentialsProvider class added to support password rotation

barshaul:
### Pull Request check-list

### Description of change

Instead of providing a simple pair of username+password, the users can create a CredentialsProvider object with their own credentials supplier function, and have redis-py call it whenever a new connection is created.
By doing so, users will be able to fetch the current password and rotate credentials without having to create a new client.

See related feature request in Lettuce: https://github.com/lettuce-io/lettuce-core/issues/1774

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -938,6 +938,7 @@ class Redis(AbstractRedis, RedisModuleCommands, CoreCommands, SentinelCommands):
         username=None,
         retry=None,
         redis_connect_func=None,
+        credentials_provider=None,
     ):
         """
         Initialize a new Redis client.
@@ -985,6 +986,7 @@ class Redis(AbstractRedis, RedisModuleCommands, CoreCommands, SentinelCommands):
                 "health_check_interval": health_check_interval,
                 "client_name": client_name,
                 "redis_connect_func": redis_connect_func,
+                "credentials_provider": credentials_provider,
             }
             # based on input, setup appropriate connection args
             if unix_socket_path is not None:
--- a/redis/client.py
+++ b/redis/client.py
@@ -13,6 +13,7 @@ from redis.commands import (
     list_or_args,
 )
 from redis.connection import ConnectionPool, SSLConnection, UnixDomainSocketConnection
+from redis.credentials import CredentialProvider
 from redis.exceptions import (
     ConnectionError,
     ExecAbortError,
@@ -938,7 +939,7 @@ class Redis(AbstractRedis, RedisModuleCommands, CoreCommands, SentinelCommands):
         username=None,
         retry=None,
         redis_connect_func=None,
-        credentials_provider=None,
+        credential_provider: CredentialProvider = None,
     ):
         """
         Initialize a new Redis client.
@@ -986,7 +987,7 @@ class Redis(AbstractRedis, RedisModuleCommands, CoreCommands, SentinelCommands):
                 "health_check_interval": health_check_interval,
                 "client_name": client_name,
                 "redis_connect_func": redis_connect_func,
-                "credentials_provider": credentials_provider,
+                "credential_provider": credential_provider,
             }
             # based on input, setup appropriate connection args
             if unix_socket_path is not None:
--- a/redis/client.py
+++ b/redis/client.py
@@ -5,6 +5,7 @@ import threading
 import time
 import warnings
 from itertools import chain
+from typing import Optional
 
 from redis.commands import (
     CoreCommands,
@@ -939,7 +940,7 @@ class Redis(AbstractRedis, RedisModuleCommands, CoreCommands, SentinelCommands):
         username=None,
         retry=None,
         redis_connect_func=None,
-        credential_provider: CredentialProvider = None,
+        credential_provider: Optional[CredentialProvider] = None,
     ):
         """
         Initialize a new Redis client.
--- a/redis/client.py
+++ b/redis/client.py
@@ -865,7 +865,7 @@ class Redis(AbstractRedis, RedisModuleCommands, CoreCommands, SentinelCommands):
 
             redis://[[username]:[password]]@localhost:6379/0
             rediss://[[username]:[password]]@localhost:6379/0
-            unix://[[username]:[password]]@/path/to/socket.sock?db=0
+            unix://[username@]/path/to/socket.sock?db=0[&password=password]
 
         Three URL schemes are supported:
 
@@ -1262,12 +1262,17 @@ class Redis(AbstractRedis, RedisModuleCommands, CoreCommands, SentinelCommands):
         try:
             if NEVER_DECODE in options:
                 response = connection.read_response(disable_decoding=True)
+                options.pop(NEVER_DECODE)
             else:
                 response = connection.read_response()
         except ResponseError:
             if EMPTY_RESPONSE in options:
                 return options[EMPTY_RESPONSE]
             raise
+
+        if EMPTY_RESPONSE in options:
+            options.pop(EMPTY_RESPONSE)
+
         if command_name in self.response_callbacks:
             return self.response_callbacks[command_name](response, **options)
         return response
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -389,7 +389,7 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands):
 
             redis://[[username]:[password]]@localhost:6379/0
             rediss://[[username]:[password]]@localhost:6379/0
-            unix://[[username]:[password]]@/path/to/socket.sock?db=0
+            unix://[username@]/path/to/socket.sock?db=0[&password=password]
 
         Three URL schemes are supported:
 
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -121,6 +121,7 @@ REDIS_ALLOWED_KEYS = (
     "connection_class",
     "connection_pool",
     "client_name",
+    "credentials_provider",
     "db",
     "decode_responses",
     "encoding",
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -121,7 +121,7 @@ REDIS_ALLOWED_KEYS = (
     "connection_class",
     "connection_pool",
     "client_name",
-    "credentials_provider",
+    "credential_provider",
     "db",
     "decode_responses",
     "encoding",
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -8,6 +8,7 @@ import weakref
 from itertools import chain
 from queue import Empty, Full, LifoQueue
 from time import time
+from typing import Optional
 from urllib.parse import parse_qs, unquote, urlparse
 
 from redis.backoff import NoBackoff
@@ -526,8 +527,6 @@ class Connection:
             )
 
         self.credential_provider = credential_provider
-        self.username = username
-        self.password = password
         if username or password:
             # Keep backward compatibility by creating a static credential provider
             # for the passed username and password
@@ -563,6 +562,38 @@ class Connection:
         self.set_parser(parser_class)
         self._connect_callbacks = []
         self._buffer_cutoff = 6000
+
+    @property
+    def password(self) -> Optional[str]:
+        if self.credential_provider is not None:
+            return self.credential_provider.password
+        else:
+            return None
+
+    @password.setter
+    def password(self, value: Optional[str]):
+        if value is None:
+            # Delete the credential provider
+            self.credential_provider = None
+            return
+        if self.credential_provider is not None:
+            self.credential_provider.password = value
+        else:
+            self.credential_provider = StaticCredentialProvider(password=value)
+
+    @property
+    def username(self) -> Optional[str]:
+        if self.credential_provider is not None:
+            return self.credential_provider.username
+        else:
+            return None
+
+    @username.setter
+    def username(self, value: Optional[str]):
+        if self.credential_provider is not None:
+            self.credential_provider.username = value
+        else:
+            self.credential_provider = StaticCredentialProvider(username=value)
 
     def __repr__(self):
         repr_args = ",".join([f"{k}={v}" for k, v in self.repr_pieces()])
@@ -1087,8 +1118,6 @@ class UnixDomainSocketConnection(Connection):
                 "2. 'credential_provider'"
             )
         self.credential_provider = credential_provider
-        self.username = username
-        self.password = password
         if username or password:
             # Keep backward compatibility by creating a static credential provider
             # for the passed username and password
@@ -1120,6 +1149,38 @@ class UnixDomainSocketConnection(Connection):
         self.set_parser(parser_class)
         self._connect_callbacks = []
         self._buffer_cutoff = 6000
+
+    @property
+    def password(self) -> Optional[str]:
+        if self.credential_provider is not None:
+            return self.credential_provider.password
+        else:
+            return None
+
+    @password.setter
+    def password(self, value: Optional[str]):
+        if value is None:
+            # Delete the credential provider
+            self.credential_provider = None
+            return
+        if self.credential_provider is not None:
+            self.credential_provider.password = value
+        else:
+            self.credential_provider = StaticCredentialProvider(password=value)
+
+    @property
+    def username(self) -> Optional[str]:
+        if self.credential_provider is not None:
+            return self.credential_provider.username
+        else:
+            return None
+
+    @username.setter
+    def username(self, value: Optional[str]):
+        if self.credential_provider is not None:
+            self.credential_provider.username = value
+        else:
+            self.credential_provider = StaticCredentialProvider(username=value)
 
     def repr_pieces(self):
         pieces = [("path", self.path), ("db", self.db)]
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -8,6 +8,7 @@ import weakref
 from itertools import chain
 from queue import Empty, Full, LifoQueue
 from time import time
+from typing import Optional
 from urllib.parse import parse_qs, unquote, urlparse
 
 from redis.backoff import NoBackoff
@@ -503,7 +504,7 @@ class Connection:
         username=None,
         retry=None,
         redis_connect_func=None,
-        credential_provider: CredentialProvider = None,
+        credential_provider: Optional[CredentialProvider] = None,
     ):
         """
         Initialize a new Connection.
@@ -1062,7 +1063,7 @@ class UnixDomainSocketConnection(Connection):
         client_name=None,
         retry=None,
         redis_connect_func=None,
-        credential_provider: CredentialProvider = None,
+        credential_provider: Optional[CredentialProvider] = None,
     ):
         """
         Initialize a new UnixDomainSocketConnection.
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -517,9 +517,20 @@ class Connection:
         self.port = int(port)
         self.db = db
         self.client_name = client_name
+        if (username or password) and credential_provider is not None:
+            raise DataError(
+                "'username' and 'password' cannot be passed along with 'credential_provider'. "
+                "Please provide only one of the following arguments: \n"
+                "1. 'password' and (optional) 'username'\n"
+                "2. 'credential_provider'"
+            )
+
         self.credential_provider = credential_provider
-        if (username or password) and self.credential_provider is None:
-            # username and password backward compatibility
+        self.username = username
+        self.password = password
+        if username or password:
+            # Keep backward compatibility by creating a static credential provider
+            # for the passed username and password
             self.credential_provider = StaticCredentialProvider(username, password)
         self.socket_timeout = socket_timeout
         self.socket_connect_timeout = socket_connect_timeout or socket_timeout
@@ -695,7 +706,7 @@ class Connection:
                 # https://github.com/andymccurdy/redis-py/issues/1274
                 self.send_command(
                     "AUTH",
-                    self.credential_provider.password,
+                    auth_args[-1],
                     check_health=False,
                 )
                 auth_response = self.read_response()
@@ -1068,9 +1079,19 @@ class UnixDomainSocketConnection(Connection):
         self.path = path
         self.db = db
         self.client_name = client_name
+        if (username or password) and credential_provider is not None:
+            raise DataError(
+                "'username' and 'password' cannot be passed along with 'credential_provider'. "
+                "Please provide only one of the following arguments: \n"
+                "1. 'password' and (optional) 'username'\n"
+                "2. 'credential_provider'"
+            )
         self.credential_provider = credential_provider
-        if (username or password) and self.credential_provider is None:
-            # username and password backward compatibility
+        self.username = username
+        self.password = password
+        if username or password:
+            # Keep backward compatibility by creating a static credential provider
+            # for the passed username and password
             self.credential_provider = StaticCredentialProvider(username, password)
         self.socket_timeout = socket_timeout
         self.retry_on_timeout = retry_on_timeout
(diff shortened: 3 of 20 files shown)


dvora-h (commented):
Except these tiny comments - it looks great!

barshaul:
Added type hints + moved CredentialsProvider to a separate file

redis/credentials.py
```diff
@@ -0,0 +1,41 @@
+class CredentialsProvider:
+    def __init__(
+        self,
+        username: str = "",
+        password: str = "",
+        supplier: callable = None,
```
chayim, line 6: I think you want supplier to be Optional[Callable]

redis/connection.py
```diff
@@ -538,9 +540,10 @@ def __init__(
         self.host = host
         self.port = int(port)
         self.db = db
-        self.username = username
         self.client_name = client_name
-        self.password = password
+        self.credentials_provider = credentials_provider
+        if not self.credentials_provider and (username or password):
```
chayim, line 545: Can we order change? Positive ifs are easier to read than negatives - to be kind to our community.

redis/credentials.py
```diff
@@ -0,0 +1,41 @@
...
+    def get_credentials(self):
+        if self.supplier:
+            self.username, self.password = self.supplier(*self.args, **self.kwargs)
+        if self.username:
+            auth_args = (self.username, self.password or "")
+        else:
+            auth_args = (self.password,)
+        return auth_args
+
+    def get_password(self, call_supplier: bool = True):
```
chayim, line 33: What do you think about making username and password properties - and instead having get_password and get_username be helper functions that if call_supplier is not called, return the property

redis/credentials.py
```diff
@@ -0,0 +1,41 @@
...
+        :param args: arguments to pass to the supplier function
+        :param kwargs: keyword arguments to pass to the supplier function
+        """
+        self.username = username
+        self.password = password
+        self.supplier = supplier
+        self.args = args
+        self.kwargs = kwargs
+
+    def get_credentials(self):
```
chayim, line 24: Same. Perhaps a property

redis/credentials.py
```diff
@@ -0,0 +1,41 @@
...
+        if self.supplier:
+            self.username, self.password = self.supplier(*self.args, **self.kwargs)
+        if self.username:
+            auth_args = (self.username, self.password or "")
```
chayim, line 28: Why not always auth_args = self.username, self.password

redis/credentials.py
```diff
@@ -0,0 +1,41 @@
...
+        else:
+            auth_args = (self.password,)
```
chayim, line 30: I think we should always pass user and password. The implementation of the callable should handle the details internally.

barshaul: Changed

redis/credentials.py
barshaul: Changed

redis/credentials.py
barshaul: Changed to property

redis/credentials.py
barshaul: Changed to property

barshaul:
@chayim @dvora-h - Finished round, ready for review :). Tests failed here from some internal error, in my fork it seems that tests are passing:
https://github.com/barshaul/redis-py/pull/16#partial-pull-merging

redis/client.py
```diff
@@ -938,6 +938,7 @@ def __init__(
         username=None,
         retry=None,
         redis_connect_func=None,
+        credentials_provider=None,
```
chayim, line 941: Can this be credential_provider - there aren't many.
Also, a type hint please on all new items

redis/client.py
```diff
@@ -979,6 +980,7 @@ def __init__(
                 "health_check_interval": health_check_interval,
                 "client_name": client_name,
                 "redis_connect_func": redis_connect_func,
+                "credentials_provider": credentials_provider,
```
chayim, line 983: Same

redis/cluster.py
```diff
@@ -121,6 +121,7 @@ def parse_cluster_shards(resp, **options):
     "connection_class",
     "connection_pool",
     "client_name",
+    "credentials_provider",
```
chayim, line 124: Again

redis/connection.py
chayim, line 545: I don't believe this is the correct logic for all use cases.

What are your thoughts on the enterprise pattern where a user connects to a service with an initial set (confirmation of connection basically) and then needs to change authentication methods?

redis/connection.py
```diff
@@ -699,12 +702,9 @@ def on_connect(self):
...
         self._parser.on_connect(self)
 
-        # if username and/or password are set, authenticate
-        if self.username or self.password:
-            if self.username:
-                auth_args = (self.username, self.password or "")
-            else:
-                auth_args = (self.password,)
+        # if credentials provider is set, authenticate
+        if self.credentials_provider:
```
chayim, line 706: Better explicit than implicit [PEP20](https://peps.python.org/pep-0020/)
```
if self.credentials_provider is not None
``

File: redis/connection.py [hunk 11]
```diff
@@ -1085,9 +1090,10 @@ def __init__(
         self.pid = os.getpid()
         self.path = path
         self.db = db
-        self.username = username
         self.client_name = client_name
-        self.password = password
+        self.credentials_provider = credentials_provider
```
chayim, line 1094: Same question regarding the use case previously mentioned.

redis/credentials.py
```diff
@@ -0,0 +1,50 @@
+from typing import Callable, Optional
+
+
+class CredentialsProvider:
+    def __init__(
+        self,
+        username: str = "",
```
chayim, line 7: In theory the default can be None, so you may want a Union here. Both for username and password

redis/credentials.py
```diff
@@ -0,0 +1,50 @@
...
+        password: str = "",
+        supplier: Optional[Callable] = None,
+        *args,
+        **kwargs,
+    ):
+        """
+        Initialize a new Credentials Provider.
+        :param supplier: a supplier function that returns the username and password.
+                         def supplier(arg1, arg2, ...) -> (username, password)
+                         For examples see examples/connection_examples.ipynb
```
chayim, line 17:
```
                         See examples/connection_examples.ipynb
```

redis/credentials.py
```diff
@@ -0,0 +1,50 @@
...
+        :param args: arguments to pass to the supplier function
+        :param kwargs: keyword arguments to pass to the supplier function
+        """
+        self._username = "" if username is None else username
```
chayim, line 21: If this is meant to be internal, and private ideally it would be self.__username__.

Same for __passsword__

redis/credentials.py
```diff
@@ -0,0 +1,50 @@
...
+        self._password = "" if password is None else password
+        self.supplier = supplier
+        self.args = args
```
chayim, line 24: Args and kwargs apply to a function. Ideally at a class level these items would have specified names that aren't this generic.

redis/credentials.py
```diff
@@ -0,0 +1,50 @@
...
+        self.kwargs = kwargs
+
+    def get_credentials(self):
+        if self.supplier:
```
chayim, line 28: [Explicit vs Implicit ](https://peps.python.org/pep-0020/)
```
if self.supplier is not None
```

redis/credentials.py
```diff
@@ -0,0 +1,50 @@
...
+            self.username, self.password = self.supplier(*self.args, **self.kwargs)
+        return self._username, self._password
+
+    @property
+    def password(self):
+        if self.supplier and not self._password:
```
chayim, line 34: Explicit vs implicit

chayim:
@barshaul Thanks as always. I left a tonne of comments - and will give it another once over. As part of this, I suggest a MockCredentialProvider or similar. It would be nice to provide an example of a standard provider, perhaps implemented by a single (useless) function. This is how @dvora-h initially looked at this PR in fact.

redis/connection.py
barshaul: This check was added only for backward compatibility with username and password.
If the credentials need to be change I think it should be implemented into the credentials supplier function. 
e.g. 
```
def supplier(init_user, init_pass):
   if (init_user or init_pass) and not (self.username or self.password):
        self.username = init_user
        self.password = init_pass
        return self.username, self.password
    return callSupplier()
```

Another option is to add call_supplier kwargs to the CredentialProvider then create a supplier function that changes this filed after the first use:
```
def supplier_use_init_set():
    call_supplier =  self.kwargs.get("call_supplier", True)
    if call_supplier:
        return callSupplier()
    # use the init set only for the first time
    self.kwargs.update({"call_supplier": True})
    return self.username, self.password
    
cred_provider = CredentialProvider(username="init_user", password="init_pass", call_supplier=False, supplier=supplier_use_init_set)
```
Either way I think this logic should be part of the CredentialProvider, and passing Redis with username and password should only be left for backward compatibility.

barshaul: I can add an example for one of those to the connection examples doc.
WDYT?

redis/credentials.py
barshaul: Not sure I understand you here, _username/ __username is the python way of saying a member is private, isn't it?
also self.username is a property so I can’t change the member to self.username too

redis/connection.py
barshaul: I can also do:
```
        self.credential_provider = credential_provider
        if username or password:
            if self.credential_provider is not None:
                self.credential_provider.username = username
                self.credential_provider.password = password
            else:
                self.credential_provider = CredentialProvider(username, password)
```
but it won't be specific for the mentioned use case

barshaul:
@chayim Finished round. I left some of the reviews open with questions

redis/credentials.py
barshaul: changed to __password

redis/connection.py
```diff
@@ -548,9 +550,11 @@ def __init__(
         self.host = host
         self.port = int(port)
         self.db = db
-        self.username = username
         self.client_name = client_name
-        self.password = password
```
akx, line 553: Removing these attributes could be a breaking change for some apps.

redis/connection.py
```diff
@@ -548,9 +550,11 @@ def __init__(
...
+        self.credential_provider = credential_provider
+        if (username or password) and self.credential_provider is None:
+            # username and password backward compatibility
+            self.credential_provider = StaticCredentialProvider(username, password)
```
akx, line 557: I think it'd be wise to raise an error if both username+password _and_ a provider are passed in, seeing as they smell mutually exclusive.

redis/connection.py
```diff
@@ -726,7 +727,11 @@ def on_connect(self):
                 # server seems to be < 6.0.0 which expects a single password
                 # arg. retry auth with just the password.
                 # https://github.com/andymccurdy/redis-py/issues/1274
-                self.send_command("AUTH", self.password, check_health=False)
+                self.send_command(
+                    "AUTH",
+                    self.credential_provider.password,
```
akx, line 732: Shouldn't this use a password from `auth_args` instead of a static attribute?

redis/connection.py
```diff
@@ -1095,9 +1101,11 @@ def __init__(
         self.pid = os.getpid()
         self.path = path
         self.db = db
-        self.username = username
         self.client_name = client_name
-        self.password = password
+        self.credential_provider = credential_provider
+        if (username or password) and self.credential_provider is None:
+            # username and password backward compatibility
+            self.credential_provider = StaticCredentialProvider(username, password)
```
akx, line 1108: Same comments about backward compat and mutual exclusivity here.

redis/credentials.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Initialize a new Credentials Provider.
+        :param supplier: a supplier function that returns the username and password.
+                         def supplier(self, arg1, arg2, ...) -> (username, password)
+                         See examples/connection_examples.ipynb
+        :param supplier_args: arguments to pass to the supplier function
+        :param supplier_kwargs: keyword arguments to pass to the supplier function
+        """
+        self.__username = "" if username is None else username
+        self.__password = "" if password is None else password
```
akx, line 22: Maybe just a single underscore? I don't think it's a good idea to use Python's private attribute mangling here.

redis/credentials.py
```diff
@@ -0,0 +1,77 @@
+from typing import Callable, Optional, Union
+
+
+class CredentialProvider:
```
akx, line 4: I'm not sure why this needs to be a class at all – couldn't a provider just be a function `get_credentials(*, username, password)`...

akx (changes requested):

redis/credentials.py
barshaul: @chayim you asked for double underscore, should I revert it?

redis/credentials.py
barshaul: Different credential providers will need different arguments to be passed, so we can save within the credential provider object the required args/kwargs.

redis/connection.py
barshaul: added

redis/connection.py
barshaul: added back

redis/connection.py
barshaul: added

redis/credentials.py
akx: You're right: a credential provider could indeed just be a `get_credentials()` function with no knowledge of _any_ arguments. 😉 

```python
def get_very_specific_credential_provider(some_auth_url):
    def provider():
        # ... do some API call..?
        return ("foo", "bar")
    return provider

c = Client(
    credential_provider=get_very_specific_credential_provider("https://secret..."),
)
```

barshaul: Yup, that would do the job. :)
@chayim WDYT?

dvora-h:
@barshaul See tests failure

barshaul:
> @barshaul See tests failure

@dvora-h fixed, not sure why the tests are still queued

redis/credentials.py
chayim: Sorry - these should be single, per side as opposed to dundered, or prefix dunder

redis/credentials.py
chayim: There's no reason that during the instantiation of all Redis connection objects a function can't be passed in, as opposed to the class being discussed. However, I think the current class implementation is more readable for **the community that uses this**. I'm partial for erring towards readability, and the clear separation it provides.

dvora-h (approved):

redis/credentials.py
barshaul: @chayim - self._username_ ?

barshaul: or `self._username` as we started with

chayim: single per side == self._username_

barshaul: changed

redis/credentials.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Initialize a new Credentials Provider.
+        :param supplier: a supplier function that returns the username and password.
+                         def supplier(self, arg1, arg2, ...) -> (username, password)
+                         See examples/connection_examples.ipynb
+        :param supplier_args: arguments to pass to the supplier function
+        :param supplier_kwargs: keyword arguments to pass to the supplier function
+        """
+        self._username_ = "" if username is None else username
+        self._password_ = "" if password is None else password
```
akx, line 22: Shouldn't these be `self._username` and `self._password` if you want to signal these to be "private"? I haven't ever seen `_foo_` as a pattern for that in all of the Python code I've read.

redis/credentials.py
akx: @chayim @barshaul Well, this class looks like it's become quite complicated and hard-to-follow.

It has an username/password that may not be used at all, args and kwargs that may not be used at all.

If you want a class, then why not just 

```
class CredentialProvider:
    def get_credentials(self) -> Union[Tuple[str], Tuple[str, str]]:
        raise NotImplementedError("get_credentials must be implemented")


class UsernamePasswordCredentialProvider(CredentialProvider):
    def __init__(self, username: Optional[str]=None, password: Optional[str]=None):
        self.username = username
        self.password = password

    def get_credentials(self):
        if self.username:
            return (self.username, self._password)
        return (self.password,)
```

and leave the rest (e.g. implementing a specific provider with a specific `get_credentials`) to the users?

As an aside – should `get_credentials` maybe also support being async? :)

redis/connection.py
```diff
@@ -514,9 +516,22 @@ def __init__(
...
+                "2. 'credential_provider'"
+            )
+
+        self.credential_provider = credential_provider
+        self.username = username
         self.password = password
+        if username or password:
+            # Keep backward compatibility by creating a static credential provider
+            # for the passed username and password
+            self.credential_provider = StaticCredentialProvider(username, password)
```
akx, line 534: This could break some backward compatibility in that setting `username` and `password` on an existing `Connection` won't affect the credential provider created here.

redis/credentials.py
barshaul: I’ll let @chayim answer, he’s the python expert

redis/connection.py
barshaul: Changed username&password to properties to address this issue

akx: I don't think that's a good idea - now you may end up inadvertently setting username and password on a provider that doesn't do anything with the information. Instead, I'd just do

```python
credentials = (self.credential_provider or StaticCredentialProvider(self.username, self.password)).get_credentials()
```
where you're getting the credentials...

redis/connection.py
```diff
@@ -675,12 +686,14 @@ def on_connect(self):
...
-                auth_args = (self.username, self.password or "")
-            else:
-                auth_args = (self.password,)
+        # if credential provider or username and/or password are set, authenticate
+        if self.credential_provider or (self.username or self.password):
+            cred_provider = (
+                self.credential_provider
+                if self.credential_provider
+                else StaticCredentialProvider(self.username, self.password)
+            )
```
akx, line 695: Simply put as
```suggestion
            cred_provider = self.credential_provider or StaticCredentialProvider(self.username, self.password)
```

redis/connection.py
```diff
@@ -692,7 +705,11 @@ def on_connect(self):
                 # server seems to be < 6.0.0 which expects a single password
                 # arg. retry auth with just the password.
                 # https://github.com/andymccurdy/redis-py/issues/1274
-                self.send_command("AUTH", self.password, check_health=False)
+                self.send_command(
+                    "AUTH",
+                    auth_args[-1],
+                    check_health=False,
+                )
```
akx, line 712: Did this really need to be split into multiple lines? `self.password` and `auth_args[-1]` are the same length, so an autoformatter shouldn't complain.

redis/credentials.py
akx: With the changes now done to `Connection`, I really do believe that the implementation I suggest above would be the better fit, and the current `CredentialProvider` is _way_ too generic. I sincerely ask you to revise it.

tests/test_commands.py
```diff
@@ -95,7 +96,9 @@ def teardown():
                 # error when switching to the db 9 because we're not authenticated yet
                 # setting the password on the connection itself triggers the
                 # authentication in the connection's `on_connect` method
-                r.connection.password = temp_pass
+                r.connection.credential_provider = StaticCredentialProvider(
+                    password=temp_pass
+                )
```
akx, line 101: This... shouldn't be required since the static credential provider is now lazily created?

tests/test_credentials.py
```diff
@@ -0,0 +1,235 @@
...
+        assert str_if_bytes(conn.read_response()) == "PONG"
+        assert conn.username == username
+        assert conn.password == password
+        init_acl_user(r, request, new_username, new_password)
+        conn.password = new_password
+        conn.username = new_username
+        conn.send_command("PING")
+        assert str_if_bytes(conn.read_response()) == "PONG"
+        conn.username = None
+        conn.password = None
```
akx, line 205: What do these lines test?

tests/test_credentials.py
```diff
@@ -0,0 +1,235 @@
...
+        password = "password"
+        creds_provider = CredentialProvider(password=password)
+        init_required_pass(r, request, password)
+        assert r.auth(creds_provider.password) is True
+
+        r2 = _get_client(
+            redis.Redis, request, flushdb=False, credential_provider=creds_provider
+        )
+
+        assert r2.ping()
```
akx, line 87: Should this not be 
```suggestion
        assert r2.ping() is True
```
to match the rest of the tests?

tests/test_credentials.py
```diff
@@ -0,0 +1,235 @@
...
+
+    @skip_if_redis_enterprise()
+    def test_credential_provider_without_supplier_acl_user_and_pass(self, r, request):
+        # test for other users
+        username = "username"
+        password = "password"
+
+        init_acl_user(r, request, username, password)
+        creds_provider = CredentialProvider(username, password)
```
akx, line 96: Isn't this the same as `StaticCredentialProvider` now..? IOW, the current design is quite [TMTOWDI](https://en.wikipedia.org/wiki/There%27s_more_than_one_way_to_do_it), which is against [the Zen of Python](https://peps.python.org/pep-0020/#the-zen-of-python)'s 
> There should be one-- and preferably only one --obvious way to do it.

tenet.

akx (changes requested):
As said in the inline comments, please reconsider the CredentialsProvider base class a bit..?

redis/credentials.py
chayim: TBC these aren't the changes that concern me. But, yes - I suggested wrapped underscores (a mistake!. Based on the standard, they should be single sided as @akx suggested.

chayim:
@dvora-h once @barshaul addresses @akx suggestions, I think an end-to-end (re)review of this is worthwhile. It can't make rc3 but could make rc4, potentially.

redis/credentials.py
barshaul: changed

redis/connection.py
```diff
@@ -502,6 +503,7 @@ def __init__(
         username=None,
         retry=None,
         redis_connect_func=None,
+        credential_provider: CredentialProvider = None,
```
akx, line 506: Considering the default is `None`, I believe this should be
```suggestion
        credential_provider: Optional[CredentialProvider] = None,
```

redis/connection.py
```diff
@@ -1050,6 +1062,7 @@ def __init__(
         client_name=None,
         retry=None,
         redis_connect_func=None,
+        credential_provider: CredentialProvider = None,
```
akx, line 1065: Same here: considering the default is `None`, I believe this should be
```suggestion
        credential_provider: Optional[CredentialProvider] = None,
```

barshaul: changed

barshaul:
@chayim @dvora-h @akx  - finished round

redis/credentials.py
chayim: ```get_credentials``` should definitely support an async all, to keep this aligned.  

On the class inputs side - I don't think it's unreasonable to pass in an optional ```CredentialProvider``` on a per redis-connection type. I see classes as more useful given state needs, depending on the provider. Maybe instead, ```CredentialProvider``` shouldn't require any specified arguments, instead be a descendent of an ABC that accepts ```**kwargs```. This could address the same need, and reduce the complexity, but yes - be perhaps generic enough (though perhaps something shared).

@akx WDYT needlessly generic? Having a way to do this for systems that are completely different makes it problematic.

akx: @chayim The "needlessly generic" comment was outdated by https://github.com/redis/redis-py/pull/2261/commits/abe613773b070754f54e7e797d3a7af022673700, which removed the "username + password + supplier + args + kwargs" mode.

dvora-h:
@barshaul I think that after adding async support we are good to go.

dvora-h:
LGTM now!

Issue #2089, opened 2022-04-06
(asyncio) PubSub does not automatically reconnect


kristjanvalur:
**Version**: 4.2.2

**Platform**: python 3.8 on windows

**Description**: Running a pub-sub task in a loop, requires me to manually call pubsub.connection.connect()

So:

I'm new to python redis, having previously used hiredis and written clients for UnrealEngine using it.

So, I was pleasantly surprised at how the *client* typically handles reconnections for you, in case there are disconnects.
However, this does not happen with the PubSub client.

Please note that I am using the `redis.asyncio` module here.

There is a retry mechanism in PubSub._execute, which callse PubSub._disconnect_raise_connect in case of failure.
However, that function only calls `connect()` in case of a `TimeoutError.`

If there is a disconnect, for example, because of the connection to the server disappearing for a few seconds, or other,
the connection stays in a disconnected state.

In my tests, I killed and restarted the server and the `PubSub.connection` stayed in a `is_connected==False` state, with any read operations resulting in a `ConnectionError("Connection closed by server.")`.

Previously the loop looked somethign like:

```python
async def loop(self):
    await self.perform_any_subscriptions()
    while True:
        try:
            await self.loop_step()
        except redis.ConnectionError:
            pass

async def loop_step(self):
    async for message in self.pubsub.listen():
        await self.dispatch_message(message)
        break
```

If the connection becomes disconnected, the `listen()` will simply continually raise  `ConnectionError("Connection closed by server.")`.

What I had to do, as a workaround, was to do something like this, for my listening loop:

```python
async def loop(self):
    while True:
        await self.pubsub.connection.call_with_retry(
            self.loop_step,
            lambda e: if isinstance(e, redis.ConnectionError) and self.pubsub.connection and not self.pubsub.connection.is_connected: await self.pubsubconnection.connect()
            )
```
(use the call_with_retry as a handy catch-all system)

I'm not sure if this is expected, or if it is an error, but from the documentation I'm reading, it seems like PubSub _ought_ to be re-usable across disconnects, automatically reconnecting any previous subscriptions, etc.

### UPDATE:
  After PR #2148, the async `PubSub` class now has a `connect()` method, which can be used to remedy
this.  The top loop becomes:

```python
async def loop(self):
    await self.perform_any_subscriptions()
    while True:
        try:
            await self.pubsub.connect()
            await self.loop_step()
        except redis.ConnectionError:
            pass
```

[...]

PR #2281, opened 2022-07-17
automatically reconnect pubsub when reading messages in blocking mode

kristjanvalur:
### Pull Request check-list

### Description of change

Issue #2089 demonstrates how a pub-sub connection doesn't automatically reconnect.  Further investigation showed that this was due to the fact that the `connection` only performed a just-in-time connect in the `can_read()` method, which is only invoked if the operation to read a message has a _timeout_.  In particular, using `listen()` to get the message, instead of the `get_message()` with a timeout, the `can_read()` method isn't called, and so a `connect()` call is not made.

This issue is present in both non-async and async versions of the library as demonstrated with the unit-test code.

The PR fixes this by adding a just-in-time connect to the `read_response` call of both the `Connection` object.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -815,8 +815,6 @@ class Connection:
 
     def read_response(self, disable_decoding=False):
         """Read the response from a previously sent command"""
-        if not self._sock:
-            self.connect()
         try:
             hosterr = f"{self.host}:{self.port}"
         except AttributeError:
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -815,6 +815,8 @@ class Connection:
 
     def read_response(self, disable_decoding=False):
         """Read the response from a previously sent command"""
+        if not self._sock:
+            self.connect()
         try:
             hosterr = f"{self.host}:{self.port}"
         except AttributeError:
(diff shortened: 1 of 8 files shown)


redis/connection.py
```diff
@@ -815,6 +815,8 @@ def can_read(self, timeout=0):
 
     def read_response(self, disable_decoding=False):
         """Read the response from a previously sent command"""
+        if not self._sock:
+            self.connect()
```
utkarshgupta137, line 819: Shouldn't this be added to the pub/sub code instead of adding this check for every redis request?

kristjanvalur: Possibly.  read_response() is after all used in other places than the pub-sub.  I'll see if I can do this differently.

kristjanvalur: Thanks for the suggestion.  Now the can_read/read_response interaction, along with an extra `connect()`, is wrapped in a lambda passed to `PubSub._execute()`

dvora-h (approved):

dvora-h:
@kristjanvalur Thanks, great PR!

dvora-h:
@kristjanvalur Can you resolve the conflicts?

agronholm:
Why is the async reconnection method different from the sync one? Is either one documented anywhere?

kristjanvalur:
Not sure what you mean.  Both code paths now contain an extra `conn.connect()` in the PubSub read method.

PR #2295, opened 2022-07-20
Simplify async timeouts and  allowing `timeout=None` in `PubSub.get_message()` to wait forever

kristjanvalur:
### Pull Request check-list

### Description of change

async `PubSub` no longer does a separate `can_read` for every message read.  Instead async timeout mechanisms are used at the highest level to time out operations.

The use of `can_read` is a holdover from the non-async times, when timeout had to happen at the lowest level, at the actual blocking socket call.  With asyncio, timeouts happen higher up in the stack.  This allows for simplifications, because there is no longer any need to have timeout code lower down in the stack.

`can_read()` is still kept around, but its only use now is assertions in the `ConnectionPool`.  This could be simplified further, since having `can_read()` around requires additional code paths for read, but not consumed, data.

Obsolete code pertaining to blocking exceptions was also removed.

Additionally, `get_message()` now accepts `timeout=None` argument to wait indefinitely for a message.

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextlib
 import copy
 import enum
 import errno
@@ -54,6 +55,19 @@ from redis.utils import HIREDIS_AVAILABLE, str_if_bytes
 hiredis = None
 if HIREDIS_AVAILABLE:
     import hiredis
+
+if sys.version_info[:2] >= (3, 10):
+    nullcontext = contextlib.nullcontext()
+else:
+
+    class NullContext:
+        async def __aenter__(self):
+            pass
+
+        async def __aexit__(self, *args):
+            pass
+
+    nullcontext = NullContext()
 
 NONBLOCKING_EXCEPTION_ERROR_NUMBERS = {
     BlockingIOError: errno.EWOULDBLOCK,
@@ -922,19 +936,23 @@ class Connection:
                 f"Error while reading from {self.host}:{self.port}: {e.args}"
             )
 
-    async def read_response(self, disable_decoding: bool = False):
+    async def read_response(
+        self,
+        disable_decoding: bool = False,
+        timeout: Optional[float] = None,
+    ):
         """Read the response from a previously sent command"""
+        read_timeout = timeout if timeout is not None else self.socket_timeout
         try:
-            if self.socket_timeout:
-                async with async_timeout.timeout(self.socket_timeout):
-                    response = await self._parser.read_response(
-                        disable_decoding=disable_decoding
-                    )
-            else:
+            async with async_timeout.timeout(read_timeout):
                 response = await self._parser.read_response(
                     disable_decoding=disable_decoding
                 )
         except asyncio.TimeoutError:
+            if timeout is not None:
+                # user requested timeout, return None
+                return None
+            # it was a self.socket_timeout error.
             await self.disconnect(nowait=True)
             raise TimeoutError(f"Timeout reading from {self.host}:{self.port}")
         except OSError as e:
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -2,7 +2,6 @@ import asyncio
 import contextlib
 import copy
 import enum
-import errno
 import inspect
 import io
 import os
@@ -68,16 +67,6 @@ else:
             pass
 
     nullcontext = NullContext()
-
-NONBLOCKING_EXCEPTION_ERROR_NUMBERS = {
-    BlockingIOError: errno.EWOULDBLOCK,
-    ssl.SSLWantReadError: 2,
-    ssl.SSLWantWriteError: 2,
-    ssl.SSLError: 2,
-}
-
-NONBLOCKING_EXCEPTIONS = tuple(NONBLOCKING_EXCEPTION_ERROR_NUMBERS.keys())
-
 
 SYM_STAR = b"*"
 SYM_DOLLAR = b"$"
@@ -243,11 +232,9 @@ class SocketBuffer:
         self,
         stream_reader: asyncio.StreamReader,
         socket_read_size: int,
-        socket_timeout: Optional[float],
     ):
         self._stream: Optional[asyncio.StreamReader] = stream_reader
         self.socket_read_size = socket_read_size
-        self.socket_timeout = socket_timeout
         self._buffer: Optional[io.BytesIO] = io.BytesIO()
         # number of bytes written to the buffer from the socket
         self.bytes_written = 0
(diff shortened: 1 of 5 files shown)


kristjanvalur:
I'm submitting this for review, even though there are one or two weird sporatic test failures that don't seem to have anything to do with this.  (Note the added code for killing containers, seems to have fixed test stability when container ports were in-use when starting tox run, probably due to rest-runner being re-used.)

redis/asyncio/connection.py
```diff
@@ -922,55 +881,27 @@ async def can_read(self, timeout: float = 0):
...
+        if not self.is_connected:
+            await self.connect()
         try:
-            if self.socket_timeout:
-                async with async_timeout.timeout(self.socket_timeout):
-                    response = await self._parser.read_response(
-                        disable_decoding=disable_decoding
-                    )
-            else:
+            async with lock_ctxt, async_timeout.timeout(read_timeout):
```
utkarshgupta137, line 896: Wouldn't an if/else on with_lock directly be more performant instead of a null context? Same for async_timeout. Even with 0 timeout, async_timeout adds significant latency.

kristjanvalur: That is a matter of opinion.  pub-sub is primarily IO bound.  An application spends its time *waiting* for a message.  Once a message arrives, two extra calls to __aexit__() make no difference in the large scheme of things.

The `not with_lock` is a special case, used only in cluster mode.
the async_timeout is usually needed, and always for timeout=0.  It could be skipped for `timeout==None` but this blocking case is, again, an uncommon special case.

There is an argument for reducing code duplication and simplifying program flow to ensure correctness, even if the cost is an extra __aexit__ call in special circumstances.    The alternative is to have *four* differenc control paths, for `with_lock == True/False` and `timeout is None or not None`.  In my opinion, that is not very pythonic.

utkarshgupta137: Normally I would always argue against code duplication, but the foremost feature of redis is its speed. If we've to make small ergonomic sacrifices so that the code is faster, then I think it should be considered.
Also, async calls add significantly more overhead than sync calls, which is why I even bothered commenting. Out of the two, we should at least have an if/else for the timeout block, as it adds a lot of latency.
And if you are keen on removing the if/else for _without_lock function, can you at least directly call this new function instead of adding a level of indirection in cluster?

kristjanvalur: I'll oblige with the changes you suggest, of course, because getting fixes into async redis-py has been a very arduous process indeed.

But as a python developer for 20 years, allow me to suggest that the changes will not amount to any measurable difference.  Remember, this is Python we're talking about.  Redis may be fast, but all that is dwarfed by the fact that we are running inside Python.  We already have layers and layers of buffers, abstractions, and classes, even lower than this function of which you speak. An extra null function call, in some minority special case, just does not matter.

My suggestion would be, instead, to simplify the code as much as possible removing interdependencies and special cases, so that in further steps, we can remove those intermediate buffering layers.  The async code in redis-py suffers too much from being verbatim copy of the blocking code, with the odd `async` keyword thrown in.

kristjanvalur: I didn't add the direct call to the without_lock function in cluster since I didn't want the changes to spread too far, nor change the api of the connection.py too be backwards incompatible.  But removing the special _without_lock function seems very reasonable.

(why the without-lock is required in the cluster client is also not clear to me, not documented anywhere, maybe you can explain why there is no need for locking in the cluster client?  I've never used cluster)

kristjanvalur: Btw, maybe you are confused about `async_timeout.timeout()`.  If `timeout==None`, then this is just a null context manager, and the result is a *blocking* operator, with negligible overhead added.   If `timeout==0` then this amounts to *polling* the connection.  the timeout context manager doesn't add any IO latency as such, any *latency* it adds comes from the cost of a context manager and the cost of calling Python functions.

Btw, I once advocated that the Python interpreter accept `None` as a null context manager, effectively omitting it, making the special `contextlib.nullcontext()` unnecessary, but it was rejected by the core devs:
https://bugs.python.org/issue18677

utkarshgupta137: Async Cluster implementation is not using the connection pool from asyncio client implementation, which uses locks for providing multi-process & multi-threading support. Async Cluster doesn't support multi-process yet & multi-threading isn't supported with asyncio anyway: https://docs.python.org/3/library/asyncio-dev.html#asyncio-multithreading.

It also uses deques instead of lists/sets, so adding/removing connections from the pool is already atomic. These differences improved speed by 9-10% AFAIR. The deque change should be adopted for standalone async clients as well.

We can keep this function for backwards compatibility, but can you please change the functions in cluster client?

utkarshgupta137: async_timeout doesn't return nullcontext if timeout is none: https://github.com/aio-libs/async-timeout/blob/master/async_timeout/__init__.py#L119

kristjanvalur: right not an actual nullcontext, that was just a turn of phrase, but it is still a no-op.  if `timeout is None`, it does nothing.

Anyway, I´ve made the fixes you suggested, permuting the context manager code.  I also changed the calls to the _without_lock from the Cluster code and tests.

Which actually got me thinking, locking (async) really shouldn't be done anyway.  It should be up to the client to perform locking if there is more than one `Task` reading from a connection at the same time, which is not necessarily a good architecture anyway.  The connection pool should perform any necessary locking to be async-safe, since it is designed to be used by many clients.

How about removing the lock alltogether, making sure that an underlying connection pool is safe to use by multiple Tasks?  I'll look the code over, and see if it makes sense.

kristjanvalur: Yes, the `with_lock` is just wrapped around the parser `read_response`.  it doesn't even cover any ConnectionPool code.  Unless I'm mistaken, we can eliminate the lock completely.
Also, the use of the lock in the ConnectionPool can be removed.  Locks are only needed in async code if there are any (potentially blocking) async calls going on.  The connection management can be made completely lock-free if desired.

I can wrap more such changes into this PR, or make a a separate one for removing asyncio locking.  What do you think?

kristjanvalur: The connectionpool can be simplified.  It is already not thread-safe.  (since async IO and threading doesn't mix.  And having it Fork-safe is also dubious.

kristjanvalur: I also suggest that the fork-protection be removed from asyncio connection pools.  When forking using asyncio, all bets are off.  Threading and forking are fundamentally incompatible with asyncio.  I´ll make a separate PR for that.

utkarshgupta137: I would say a separate PR would probably be better, since it may need other changes and/or improved tests.

I agree, threading support should neither be supported nor be expected.
Regarding removing multiprocessing support, I'm definitely in favour of it. But @chayim @dvora-h would've to take a call for release plan.

chayim: I prefer tight, surgical PRs. It allows us to both review code, comment effectively - and frankly, helps us not miss things. This includes surgical tests.

Feature parity (where possible) belongs in all of our usage models:
1. Sync
2. Async
3. Sync Cluster
4. Async Cluster

I'm generally against removing features that people are using. However, I'm pro pre-announced deprecation (see the Python announcement in the README).

.github/workflows/integration.yaml
```diff
@@ -56,6 +56,7 @@ jobs:
          run: |
            pip install -U setuptools wheel
            pip install -r dev_requirements.txt
+           [ $(docker ps -aq | wc -l) -ne 0 ] && docker rm -f $(docker ps -aq)
```
chayim, line 59: This shouldn't be necessary in GH Actions. What did you run into here?

redis/asyncio/connection.py
```diff
@@ -56,15 +56,18 @@
 if HIREDIS_AVAILABLE:
     import hiredis
 
-NONBLOCKING_EXCEPTION_ERROR_NUMBERS = {
-    BlockingIOError: errno.EWOULDBLOCK,
-    ssl.SSLWantReadError: 2,
-    ssl.SSLWantWriteError: 2,
-    ssl.SSLError: 2,
-}
+if sys.version_info[:2] >= (3, 10):
```
chayim, line 59: nullcontext is available as of 3.7, but the [async context manager](https://docs.python.org/3/glossary.html#term-asynchronous-context-manager) is available since 3.10

redis/asyncio/connection.py
```diff
@@ -471,42 +451,21 @@ async def can_read(self, timeout: float):
...
-            # original exception.
-            allowed = NONBLOCKING_EXCEPTION_ERROR_NUMBERS.get(ex.__class__, -1)
-            if not raise_on_timeout and ex.errno == allowed:
+                    return await self.read_from_socket()
+            except asyncio.TimeoutError:
                 return False
-            raise ConnectionError(f"Error while reading from socket: {ex.args}")
+        return True
+
+    async def read_from_socket(self):
```
chayim, line 461: @dvora-h can you dig into this function?

kristjanvalur:
A separate PR, #2308, proposes to remove this read lock entirely, and would simplify this PR further.

redis/asyncio/connection.py
kristjanvalur: This has been removed.

.github/workflows/integration.yaml
kristjanvalur: It is one of the weird rare failure cases with gh.  I agree that it shouldn't be needed, but what I ran into was that the redis container couldn't start because the port was in use.  I have no idea if this fixed it, but I didn't see it again.

kristjanvalur: removed.  tests pass.

dvora-h:
@kristjanvalur Can you merge master in so I can see if tests pass and merge this PR?

utkarshgupta137 (approved):

chayim:
@kristjanvalur I think a better title for this may be your comment in the CHANGES file - which completely changes how I think about this PR. IMHO this becomes a feature. For 4.4.0-rc3. WDYT?

kristjanvalur:
Well, the `timeout=None` was actually just something I threw in there as a bonus.  My primary concern with this PR was to remove the cruft in the `async` code that exists because it is a direc copy of the `non-async` code.  Timeouts in blocking code need to be handled at the IO level, while in `async` they are generally free, and implemented via the event loop.

I had actually been thinking about removing this "feature" and adding it separately, post-hoc, to make the PR more "clean".

But whatever you think is best, I'm happy with.  I think it is important to straighten out and simplify the async connection code so that we can more easily work on simplifications closer to the metal, for example, remove double buffering, etc.

Would you like me to change the title?

chayim:
Given that we'd like to get this into 4.4.0 rc2... I'm voting for take as is. GitHub really needs a "100%" reaction.  I think this is great :)

chayim (approved):

dvora-h:
@kristjanvalur last one...

kristjanvalur:
Ok, this one was a bit trickier, lots of related commits that went in... lets see how it does in CI

kristjanvalur:
Thanks.  That was a bunch of related PRs that all sort of complemented each other.  I hope with this in place to be able to do further work, particularly on the hiredis parser, to reduce overhead.

PR #2316, opened 2022-08-02
Fix async SEARCH pipeline

### LINKED ISSUE #2279: RediSearch: search command doesn't support asyncio Pipeline
The RediSearch search command returns an instance of the `Result` class except when the Redis client is a `Pipeline` because `Pipeline` returns itself instead of a result when you execute a command. There's code that checks for this in both the `SearchCommands` and `AsyncSearchCommands` classes:
https://github.com/redis/redis-py/blob/4b0543d567aef36ac467ce495d831a24575d8d5b/redis/commands/search/commands.py#L414
https://github.com/redis/redis-py/blob/4b0543d567aef36ac467ce495d831a24575d8d5b/redis/commands/search/commands.py#L883
However, this check doesn't work if the `Pipeline` is from the `redis.asyncio.client` module. The following modification should fix the issue:
```python
from redis.client import Pipeline
from redis.asyncio.client import Pipeline as AsyncPipeline
...
        if isinstance(res, Pipeline) or isinstance(res, AsyncPipeline):
            return res
...
```
I'm not sure if it makes sense to check for both `Pipeline` types or if the `SearchCommands` class should check for just `Pipeline` and the `AsyncSearchCommands` class should check for just `AsyncPipeline`. Let me know and I can make a PR. Or feel free to make the changes yourself if that's easier. Thanks!

dvora-h:
### Pull Request check-list

### Description of change

Fixes #2279


redis/commands/search/__init__.py
```diff
@@ -167,5 +167,5 @@ class Pipeline(SearchCommands, redis.client.Pipeline):
     """Pipeline for the module."""
 
 
-class AsyncPipeline(AsyncSearchCommands, AsyncioPipeline):
+class AsyncPipeline(AsyncSearchCommands, AsyncioPipeline, Pipeline):
```
chayim, line 170: ahh

chayim (approved):

PR #2322, opened 2022-08-04
Search test - Ignore order of the items in the response

dvora-h:
### Pull Request check-list

### Description of change

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -965,7 +965,7 @@ def test_aggregations_groupby(client):
 
     res = client.ft().aggregate(req).rows[0]
     assert res[1] == "redis"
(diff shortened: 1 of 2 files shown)


chayim (approved):

tests/test_search.py
```diff
@@ -965,7 +965,7 @@ def test_aggregations_groupby(client):
 
     res = client.ft().aggregate(req).rows[0]
     assert res[1] == "redis"
-    assert res[3] == ["RediSearch", "RedisAI", "RedisJson"]
+    assert set(res[3]) == {"RediSearch", "RedisAI", "RedisJson"}
```
chayim, line 968: So much smarter than what I was thinking. But, I'm not convinced it always works - in older python versions.

Instead can we do an assert on each item in res[3]?

PR #2343, opened 2022-08-19
Be more strict about url scheme parsing

vanschelven:
### Pull Request check-list

### Description of change

The error message implied that urls had to start with `scheme://`.
However, if the double slash was left out, the url parsed just fine
and the part that was ostensibly intended to be the hostname ended
up as part of the path, whereas the default (localhost) would be
used for the hostname. This commit makes the check as strict as the
error message implies by including a check for the double slash.


vanschelven:
I think the pipeline troubles are independent of this fix (see https://github.com/redis/redis-py/issues/2344)

dvora-h (approved):

PR #2345, opened 2022-08-19
Do not implicitly default to localhost when using from_url

vanschelven:
### Pull Request check-list

### Description of change

Do not implicitly default to `localhost` when using from_url. 

This avoids much confusion when `from_url` is passed a url without an explicit hostname because of some error in the passed url. Passed urls could be incorrect for any number of reasons, e.g. when they are incorrectly constructed using environment variables in some CI/CD process. Whatever the reason, simply trying to connect to localhost (and getting a connection error, or worse: no error at all) will not be helpful in finding the root cause of the problem.

A counterpoint to this PR could be: strictly speaking urls without explicit hosts are valid, and a case could be made that relying on that behavior is perfectly valid.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -1219,6 +1219,9 @@ def parse_url(url):
         if url.scheme == "rediss":
             kwargs["connection_class"] = SSLConnection
 
+    if url.scheme != "unix" and "host" not in kwargs:
+        raise ValueError("Redis URL must specify an explicit hostname")
+
     return kwargs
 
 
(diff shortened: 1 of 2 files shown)


vanschelven:
Converted this to a draft... as it stands it's obviously unworkable, because for `unix:///` schemes the requirement that a host should be provided cannot be reasonably made. And perhaps there are other issues, I didn't check yet given the redness of the tests.

chayim:
This would represent a breaking change to the client - and is the way this (and other) clients have worked for years.  This isn't a small change to make - and I don't see the benefit, as opposed to the danger.

PR #2356, opened 2022-08-24
Add `nowait` flag to `asyncio.Connection.disconnect()`

kristjanvalur:
### Pull Request check-list

### Description of change

When the `asyncio.Connection` object encounters an exception, it often performs a `self.disconnect()`.  This change
adds a `nowait` flag to this method, so that in the error case, `disconnect()` does not attempt to await a `wait_closed()` on the output stream.  Awaiting a buffer method at this point can cause other errors to occur.
Notably an `asyncio.CancelledError` could happen, in case the entire operation is enclosed in a `async_timeout.timeout()` context manager.

At any rate, if the connection is in an error state, there is no need to wait for the write buffer flush.

This PR is separate from a related pr, #2104, both about making exception handling and cancelling (via Timeout) possible.

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -847,7 +847,7 @@ class Connection:
                     if os.getpid() == self.pid:
                         self._writer.close()  # type: ignore[union-attr]
                         # wait for close to finish, except when handling errors and
-                        # forcecully disconnecting.
+                        # forcefully disconnecting.
                         if not nowait:
                             await self._writer.wait_closed()  # type: ignore[union-attr]
                 except OSError:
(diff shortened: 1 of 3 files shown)


kristjanvalur:
No idea why redismod, `test_alias_basic` now fails.  These tests sure are flakey sometimes.

dvora-h (approved):

redis/asyncio/connection.py
```diff
@@ -849,8 +849,9 @@ async def disconnect(self) -> None:
                 try:
                     if os.getpid() == self.pid:
                         self._writer.close()  # type: ignore[union-attr]
-                        # py3.6 doesn't have this method
-                        if hasattr(self._writer, "wait_closed"):
+                        # wait for close to finish, except when handling errors and
+                        # forcecully disconnecting.
```
akx, line 853:
```suggestion
                        # forcefully disconnecting.
```

dvora-h:
@kristjanvalur Can you resolve the conflicts with master?

PR #2359, opened 2022-08-25
Add support for string responses in ping

dvora-h:
Change `ping` response callback so that it can also return any string and not only true if pong and false otherwise.

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -100,6 +100,12 @@ def parse_object(response, infotype):
     return response
 
 
+def parse_ping(response):
+    if str_if_bytes(response) == "PONG":
+        return True
+    return response
+
+
 def parse_info(response):
     """Parse the result of Redis's INFO command into a Python dict"""
     info = {}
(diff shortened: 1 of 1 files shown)


levincarmel:
@chayim @dvora-h 
Hi, I talked to Oran about the new PING behavior, and this is what he said:

_"the documentation may be lacking, but `PING` always has been a status check command.
it can return `PONG` in normal circumstance, and will return other response like `LOADING`, `NOAUTH`, `BUSY` and now also `MASTERDOWN`.
I think the client (or our cluster) should handle it much the same as it handles the `LOADING` error.
Both mean that you may be able to get `INFO` or `CONFIG`, but not interact with data.
AFAIK all errors throw an exception, the LOADING one can be set to be ignored, and i think that should be done for `MASTERDOWN` too, but not for other errors."_

So we want redis-py to handle `MASTERDOWN` the same way it handles `LOADING`.

PR #2360, opened 2022-08-25
Dev/no can read

kristjanvalur:
### Pull Request check-list

### Description of change

The `asyncio.Connection.can_read()` API is unnecessary in async-io, since polling a connection is best done with timeout.
Keeping this api adds complexity in the lower layers of the asyncio.Connection since it requires reading data and then "unreading" them for subsequent `read` operations.

Further, this code is almost exclusively used by assertion macros, to make sure that there is no data present on a  connection pulled from a connection pool.  For this use case, it is fine that the read data is discarded.

The only remaining use case was in the `PubSub` implementation, but that can be done differently.

Renaming the method to `can_read_destructive()` and simplifying it, getting rid of the "keep any read data for next read" logic, paves the way for further simplifications of the asyncio connection code, and other pull requests in the same ares (getting rid of locking, buffering, simplifying timeouts, etc)

The PR contains elements from Pull Requests #2356 and #2104 as necessary for it to work properly

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -24,6 +24,8 @@ from typing import (
     cast,
 )
 
+import async_timeout
+
 from redis.asyncio.connection import (
     Connection,
     ConnectionPool,
@@ -755,12 +757,16 @@ class PubSub:
         await self.check_health()
 
         async def try_read():
+            if not conn.is_connected:
+                await conn.connect()
             if not block:
-                if not await conn.can_read(timeout=timeout):
+                try:
+                    async with async_timeout.timeout(timeout):
+                        return await conn.read_response()
+                except asyncio.TimeoutError:
                     return None
             else:
-                await conn.connect()
-            return await conn.read_response()
+                return await conn.read_response()
 
         response = await self._execute(conn, try_read)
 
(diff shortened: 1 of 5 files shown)


utkarshgupta137 (approved):

dvora-h:
@kristjanvalur Can you resolve the conflicts with master?

kristjanvalur:
No problem, it is best that I resolve these myself anyway.

dvora-h (approved):

Chronial:
This PR broke async `PubSub.get_message()`:
* It changes the interface in an unexpected way: Before, the `timeout` parameter was a timeout for waiting for a new message. Now it is a timeout for the whole receive and parse operation, i.e. it will interrupt message receiving/parsing if it takes too long.
* If that happens (receive interrupted), the currently processing message gets dropped and the connection corrupted. In the past, this would have been detected and the connection closed. But that check was removed in #2104. Now this PR just leaves the Connection of the `PubSub` in a corrupted state and following operations on that connection will fail in unexpected ways.

The timeout parameter should only affect the wait operation and never cause messages to be dropped.

kristjanvalur:
Indeed.  See issue #2513

Issue #2354, opened 2022-08-24
Can not exit the redis-cluster client after redis is down


suxb201:
**Version** and **Platform**:
- Darwin Kernel Version 21.6.0
- Python 3.10.4
- redis-py 4.4.0rc1

**Description**: 
After redis is down, you can not exit the redis-cluster client
```
16:17:46.524 error occurred when writing data: ERROR sending "cluster slots" command to redis server 127.0.0.1:30008. error: cannot pickle '_thread.lock' object
Exception ignored in: <function ClusterNode.__del__ at 0x7f79d9e03640>
Traceback (most recent call last):
  File "/opt/miniconda3/envs/py310/lib/python3.10/site-packages/redis/cluster.py", line 1191, in __del__
    if self.redis_connection is not None:
AttributeError: 'ClusterNode' object has no attribute 'redis_connection'
Exception ignored in: <function ClusterNode.__del__ at 0x7f79d9e03640>
Traceback (most recent call last):
  File "/opt/miniconda3/envs/py310/lib/python3.10/site-packages/redis/cluster.py", line 1191, in __del__
    if self.redis_connection is not None:
AttributeError: 'ClusterNode' object has no attribute 'redis_connection'
Exception ignored in: <function ClusterNode.__del__ at 0x7f79d9e03640>
Traceback (most recent call last):
  File "/opt/miniconda3/envs/py310/lib/python3.10/site-packages/redis/cluster.py", line 1191, in __del__
    if self.redis_connection is not None:
AttributeError: 'ClusterNode' object has no attribute 'redis_connection'
```

dvora-h:
Fixed in #2377

Issue #2297, opened 2022-07-21
Redis cluster connection fails when slave goes offline


theblazehen:
**Version**: What redis-py and what redis version is the issue happening on?
redis-py: 4.3.4
redis: keydb 6.3.1

**Platform**: Python 3.8.10 on Ubuntu 20.04

**Description**: 
I have a 6 node cluster, 3 masters and 3 slaves.
I initialize the client with
```
rc = redis.cluster.RedisCluster(
    host="keydb-staging-1.lxd",
    port=6379,
    decode_responses=True,
    retry=redis.retry.Retry(redis.backoff.ConstantBackoff(1), 20),
)
```

When I shut down one master then everything keeps running fine:
```
>>> rc.ping()
True
```

However, when I shut down one slave (All masters are online) I get the following when attempting to run a ping:
```
>>> rc.ping()
<class 'redis.exceptions.ConnectionError'>
Traceback (most recent call last):
  File "/home/ubuntu/venv/deepalert/lib/python3.8/site-packages/redis/connection.py", line 1393, in get_connection
    if connection.can_read():
  File "/home/ubuntu/venv/deepalert/lib/python3.8/site-packages/redis/connection.py", line 809, in can_read
    return self._parser.can_read(timeout)
  File "/home/ubuntu/venv/deepalert/lib/python3.8/site-packages/redis/connection.py", line 411, in can_read
    return self.read_from_socket(timeout=timeout, raise_on_timeout=False)
  File "/home/ubuntu/venv/deepalert/lib/python3.8/site-packages/redis/connection.py", line 423, in read_from_socket
    raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
redis.exceptions.ConnectionError: Connection closed by server.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ubuntu/venv/deepalert/lib/python3.8/site-packages/redis/connection.py", line 611, in connect
    sock = self.retry.call_with_retry(
  File "/home/ubuntu/venv/deepalert/lib/python3.8/site-packages/redis/retry.py", line 46, in call_with_retry
    return do()
  File "/home/ubuntu/venv/deepalert/lib/python3.8/site-packages/redis/connection.py", line 612, in <lambda>
    lambda: self._connect(), lambda error: self.disconnect(error)
  File "/home/ubuntu/venv/deepalert/lib/python3.8/site-packages/redis/connection.py", line 677, in _connect
    raise err
[...]

PR #2377, opened 2022-09-10
Failover handling improvements for RedisCluster and Async RedisCluster

barshaul:
### Pull Request check-list

### Failover handling improvements for RedisCluster and Async RedisCluster
1. **Handle cluster connection errors in the retry object** -Removed handling of timeouts/connection errors within the cluster loop to reduce retries duplication with Retry objects and enable backoff strategies. Each cluster node will share the cluster's Retry object for retry and backoff policy. 
2. **Fixed "cannot pickle '_thread.lock' object" bug**- When we use NodesManager's initialize function to refresh the cluster topology after a failover, if the chosen startup node was the failed node, we received the above exception when trying to copy the connection kwargs (traceback at the end), and this error was raised to the user since it was not expected. It caused the initialize loop to get stuck on the failed node and not continue. Removed the copy of kwargs as it no longer needed since we call str_if_bytes on the conncetion's responses. We now throw exceptions in the initialize() loop only if there are no more startup nodes available. Closing #2354 #2297 
3. **Removing the failed target node from the seed nodes when connection/timeout errors occur:** the failed target node will return to the startup nodes if discovered on the cluster slots, but we'll skip it in this initialize iteration.
4.  **Backwards comparability for connection_error_retry_attempts** - connection_error_retry_attempts can still be used to pass the number of retries to execute on connection/timeout error. If connection_error_retry_attempts = 0 and retry object isn't being passed, no retries will be done on connecton/timeout errors for the cluster.

```
Traceback (most recent call last):   File "/home/ubuntu/redis-py/redis/cluster.py", line 1424, in initialize

    copy_kwargs = copy.deepcopy(kwargs)

  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy

    y = copier(x, memo)

...
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy

    y = copier(x, memo)

  File "/usr/lib/python3.10/copy.py", line 231, in _deepcopy_dict

    y[deepcopy(key, memo)] = deepcopy(value, memo)

  File "/usr/lib/python3.10/copy.py", line 161, in deepcopy

    rv = reductor(4)

TypeError: cannot pickle '_thread.lock' object
```

### DIFF
--- a/redis/__init__.py
+++ b/redis/__init__.py
@@ -1,6 +1,6 @@
 import sys
 
-from redis.backoff import get_default_backoff
+from redis.backoff import default_backoff
 from redis.client import Redis, StrictRedis
 from redis.cluster import RedisCluster
 from redis.connection import (
@@ -65,7 +65,7 @@ __all__ = [
     "ConnectionPool",
     "DataError",
     "from_url",
-    "get_default_backoff",
+    "default_backoff",
     "InvalidResponse",
     "PubSubError",
     "ReadOnlyError",
--- a/redis/__init__.py
+++ b/redis/__init__.py
@@ -1,5 +1,6 @@
 import sys
 
+from redis.backoff import get_default_backoff
 from redis.client import Redis, StrictRedis
 from redis.cluster import RedisCluster
 from redis.connection import (
@@ -64,6 +65,7 @@ __all__ = [
     "ConnectionPool",
     "DataError",
     "from_url",
+    "get_default_backoff",
     "InvalidResponse",
     "PubSubError",
     "ReadOnlyError",
--- a/redis/asyncio/__init__.py
+++ b/redis/asyncio/__init__.py
@@ -15,6 +15,7 @@ from redis.asyncio.sentinel import (
     SentinelManagedSSLConnection,
 )
 from redis.asyncio.utils import from_url
+from redis.backoff import get_default_backoff
 from redis.exceptions import (
     AuthenticationError,
     AuthenticationWrongNumberOfArgsError,
@@ -43,6 +44,7 @@ __all__ = [
     "ConnectionPool",
     "DataError",
     "from_url",
+    "get_default_backoff",
     "InvalidResponse",
     "PubSubError",
     "ReadOnlyError",
--- a/redis/asyncio/__init__.py
+++ b/redis/asyncio/__init__.py
@@ -15,7 +15,7 @@ from redis.asyncio.sentinel import (
     SentinelManagedSSLConnection,
 )
 from redis.asyncio.utils import from_url
-from redis.backoff import get_default_backoff
+from redis.backoff import default_backoff
 from redis.exceptions import (
     AuthenticationError,
     AuthenticationWrongNumberOfArgsError,
@@ -44,7 +44,7 @@ __all__ = [
     "ConnectionPool",
     "DataError",
     "from_url",
-    "get_default_backoff",
+    "default_backoff",
     "InvalidResponse",
     "PubSubError",
     "ReadOnlyError",
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -26,7 +26,7 @@ from redis.asyncio.connection import (
 )
 from redis.asyncio.parser import CommandsParser
 from redis.asyncio.retry import Retry
-from redis.backoff import get_default_backoff
+from redis.backoff import default_backoff
 from redis.client import EMPTY_RESPONSE, NEVER_DECODE, AbstractRedis
 from redis.cluster import (
     PIPELINE_BLOCKED_COMMANDS,
@@ -138,7 +138,7 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
         | Number of times to retry before reinitializing when :class:`~.TimeoutError`
           or :class:`~.ConnectionError` are encountered.
           The default backoff strategy will be set if Retry object is not passed (see
-          get_default_backoff in backoff.py). To change it, pass a custom Retry object
+          default_backoff in backoff.py). To change it, pass a custom Retry object
           using the "retry" keyword.
     :param max_connections:
         | Maximum number of connections per node. If there are no free connections & the
@@ -309,7 +309,7 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
         if retry or retry_on_error or connection_error_retry_attempts > 0:
             # Set a retry object for all cluster nodes
             self.retry = retry or Retry(
-                get_default_backoff(), connection_error_retry_attempts
+                default_backoff(), connection_error_retry_attempts
             )
             if retry_on_error is None:
                 # Default errors for retrying
@@ -642,7 +642,9 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
             target_nodes_specified = True
             retry_attempts = 0
 
-        while True:
+        # Add one for the first execution
+        execute_attempts = 1 + retry_attempts
+        for _ in range(execute_attempts):
             if self._initialize:
                 await self.initialize()
             try:
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -110,10 +110,10 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
     :param startup_nodes:
         | :class:`~.ClusterNode` to used as a startup node
     :param require_full_coverage:
-        | When set to ``False``: the client will not require a full coverage of the
-          slots. However, if not all slots are covered, and at least one node has
-          ``cluster-require-full-coverage`` set to ``yes``, the server will throw a
-          :class:`~.ClusterDownError` for some key-based commands.
+        | When set to ``False``: the client will not require a full coverage of
+          the slots. However, if not all slots are covered, and at least one node
+          has ``cluster-require-full-coverage`` set to ``yes``, the server will throw
+          a :class:`~.ClusterDownError` for some key-based commands.
         | When set to ``True``: all slots must be covered to construct the cluster
           client. If not all slots are covered, :class:`~.RedisClusterException` will be
           thrown.
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -707,7 +707,7 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
                 return await target_node.execute_command(*args, **kwargs)
             except (BusyLoadingError, MaxConnectionsError):
                 raise
-            except (ConnectionError, TimeoutError) as e:
+            except (ConnectionError, TimeoutError):
                 # Connection retries are being handled in the node's
                 # Retry object.
                 # Remove the failed node from the startup nodes before we try
@@ -1163,8 +1163,8 @@ class NodesManager:
 
         if not startup_nodes_reachable:
             raise RedisClusterException(
-                f"Redis Cluster cannot be connected. Please provide at least "
-                f"one reachable node"
+                "Redis Cluster cannot be connected. Please provide at least "
+                "one reachable node"
             ) from exception
 
         # Check if the slots are not fully covered
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -25,6 +25,8 @@ from redis.asyncio.connection import (
     parse_url,
 )
 from redis.asyncio.parser import CommandsParser
+from redis.asyncio.retry import Retry
+from redis.backoff import get_default_backoff
 from redis.client import EMPTY_RESPONSE, NEVER_DECODE, AbstractRedis
 from redis.cluster import (
     PIPELINE_BLOCKED_COMMANDS,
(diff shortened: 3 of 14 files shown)


redis/asyncio/cluster.py
```diff
@@ -232,6 +237,8 @@ def __init__(
         socket_keepalive: bool = False,
         socket_keepalive_options: Optional[Mapping[int, Union[int, bytes]]] = None,
         socket_timeout: Optional[float] = None,
+        retry: Optional["Retry"] = None,
```
chayim, line 240: I'd argue if retry_on_error is added and retry is added - you effectively have duplication. WDYT?

redis/asyncio/connection.py
```diff
@@ -650,7 +650,7 @@ def __init__(
             retry_on_error.append(socket.timeout)
             retry_on_error.append(asyncio.TimeoutError)
         self.retry_on_error = retry_on_error
-        if retry_on_error:
+        if retry or retry_on_error:
```
chayim, line 653: Seems duplicative. First there's a check for retry or retry_on_error, but then there's an explicit retry check. WDYT about reworking this block?

redis/backoff.py
```diff
@@ -1,6 +1,9 @@
 import random
 from abc import ABC, abstractmethod
 
+DEFAULT_CAP = 0.512
```
chayim, line 4: Let's comment on why these numbers matter, and help the community contribute!

redis/asyncio/cluster.py
```diff
@@ -691,25 +713,18 @@ async def _execute_command(
...
+            except (BusyLoadingError, MaxConnectionsError):
+                raise
+            except (ConnectionError, TimeoutError):
+                # Connection retries are being handled in the node's
+                # Retry object.
+                # Remove the failed node from the startup nodes before we try
+                # to reinitialize the cluster
+                self.nodes_manager.startup_nodes.pop(target_node.name, None)
+                # Hard force of reinitialize of the node/slots setup
+                # and try again with the new setup
```
chayim, line 725: Good idea!

redis/asyncio/cluster.py
```diff
@@ -614,9 +640,9 @@ async def execute_command(self, *args: EncodableT, **kwargs: Any) -> Any:
         if passed_targets and not self._is_node_flag(passed_targets):
             target_nodes = self._parse_target_nodes(passed_targets)
             target_nodes_specified = True
-            retry_attempts = 1
+            retry_attempts = 0
 
-        for _ in range(retry_attempts):
+        while True:
```
chayim, line 645: Can we hit a case where this fails to exit and just sits here? Counter impl?

redis/__init__.py
```diff
@@ -1,5 +1,6 @@
 import sys
 
+from redis.backoff import get_default_backoff
```
chayim, line 3: Can we change get_default_backoff to default_backoff? It's return implies a get

redis/asyncio/__init__.py
```diff
@@ -15,6 +15,7 @@
     SentinelManagedSSLConnection,
 )
 from redis.asyncio.utils import from_url
+from redis.backoff import get_default_backoff
```
chayim, line 18: Same idea. get_default_backoff to default_backoff? IMHO it's implied.

redis/cluster.py
```diff
@@ -425,27 +426,30 @@ class initializer. In the case of conflicting arguments, querystring
...
-        reinitialize_steps=10,
-        read_from_replicas=False,
-        dynamic_startup_nodes=True,
-        url=None,
+        host: Optional[str] = None,
+        port: int = 6379,
+        startup_nodes: Optional[List["ClusterNode"]] = None,
+        cluster_error_retry_attempts: int = 3,
+        connection_error_retry_attempts: int = 3,
+        retry: Optional["Retry"] = None,
```
chayim, line 434: What do you think about collapsing retry and retry_on_error?

redis/cluster.py
```diff
@@ -455,22 +459,29 @@ def __init__(
...
+             Number of times to retry before raising an error when
+             :class:`~.TimeoutError` or :class:`~.ConnectionError` or
+             :class:`~.ClusterDownError` are encountered
+        :param connection_error_retry_attempts:
+            Number of times to retry before reinitializing when :class:`~.TimeoutError`
+            or :class:`~.ConnectionError` are encountered.
+            The default backoff strategy will be set if Retry object is not passed (see
+            get_default_backoff in backoff.py). To change it, pass a custom Retry object
+            using the "retry" keyword.
         :reinitialize_steps: 'int'
```
chayim, line 485: I think 10 is too many - but yes I agree that 1/3 is too small. What in your experience leads you to 10?

chayim (changes requested):
@barshaul  I really like this PR - a lot. I don't think it'll make it into 4.4.0-rc2, but should (with the other one) make it into rc3.  Please take a look at the suggestions.

redis/asyncio/cluster.py
```diff
@@ -298,6 +306,17 @@ def __init__(
             # Call our on_connect function to configure READONLY mode
             kwargs["redis_connect_func"] = self.on_connect
 
+        if retry or retry_on_error or connection_error_retry_attempts > 0:
```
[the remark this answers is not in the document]
barshaul: So in the async cluster (and only in the async code) we have this additional argument: 'connection_error_retry_attempts', which is mutually exclusive with 'retry'. 
Our guide line is to keep backward compatibility, and since the default value of [connection_error_retry_attempts is set to 5](https://github.com/redis/redis-py/blob/6b3e0b491c4a348ca2e7a332e6f3b23a5da3c461/redis/asyncio/cluster.py#L217), we need to try to maintain this default retry behavior. 
There are couple of options I thought of to maintain the same behavior, but none of them isn't great:
The first is to leave the default value as is and raise exceptions when mutually exclusive arguments passed together:
```
        if retry and retry_on_error:
            raise DataError("'retry' cannot be passed along with "
                            "'retry_on_error'. "
                            "To control these variables pass a "
                            "Retry object set with the required retries "
                            "and/or errors.")
        if retry and retry._retries != connection_error_retry_attempts:
            raise DataError("Retry attempts set in the passed 'retry' object do not match "
                            "'connection_error_retry_attempts'."
                            "'connection_error_retry_attempts' should be set to None or to "
                            "the number of retries set in the Retry object.")
```
However, as connection_error_retry_attempts == 5 by default, if someone will create a new async cluster instance and pass a retry object it will require him to set connection_error_retry_attempts to None/the same retries set in the retry object. e.g.:
```
r = RedisCluster(host, port, retry=Retry(backoff(), 4), connection_error_retry_attempts=None)
```
which obviously doesn't look good.
To solve it, we can change the connection_error_retry_attempts default value to None/0 by default, but then users that were used to get 5 retries will now have none. Do you think that it's something we can do if we'll announce about this change in the release notes?
Another option is to allow passing retry object with different retry attempts than in connection_error_retry_attempts and ignore the later when retry is passed. We can add a documentation of that, or even log a warning stating that we ignored connection_error_retry_attempts, but I don't think it would be readable at all for the user. 

WDYT? do you have a better option?

redis/asyncio/__init__.py
barshaul: changed

redis/__init__.py
barshaul: changed

redis/asyncio/cluster.py
barshaul: changed

redis/cluster.py
barshaul: There's a latency tradeoff between getting a moved error and rotating the request to another node, which will result in 2 call rounds, and between reinitializing the cluster on a moved error which is an expensive operation that will also require another call to the redirected node. As a new cluster map can take some time to form (e.g. when adding more shards to the cluster, when migrating slots from one node to another), we can receive lots of moved errors during this time. By setting the reinitiailize steps to 10, we will not have to call the expensive reinitialize call on every moved error, giving the cluster more time to reach the final slot map before querying for it. The slot map will be changed at once in other scenarios, such as when the primary fails and a failover is done, so reinitializing the cluster map after the first Moved error should suffice, and if we did indeed query for the cluster slot map after the first move, we could potentially save 9 more moved errors.
So, I don't think we have a best value that fits all cases. Some other clients reinitialize the topology for each moved error and some continue to change only the specific slot owner when moved occurred.
We can reduce it to 5 (or less) if you feel 10 is too much, either way some cases will benefit from either one.

redis/backoff.py
barshaul: Added

redis/cluster.py
barshaul: We can remove retry_on_error all together, as the Retry object can be set with specific errors (supported_errors) and it defaults to ConnectionError and TimeoutError. The only issue with removing the retry_on_error is backward compatibility.

redis/asyncio/cluster.py
```diff
@@ -318,8 +337,7 @@ def __init__(
         self.read_from_replicas = read_from_replicas
         self.reinitialize_steps = reinitialize_steps
         self.cluster_error_retry_attempts = cluster_error_retry_attempts
-        self.connection_error_retry_attempts = connection_error_retry_attempts
-
+        self.retry = retry
```
utkarshgupta137, line 340: Please update `__slots__` as well.

barshaul: Reverting these deletion to maintain backward compatibility

utkarshgupta137: I don't think this line is required. It will overwrite the default implementation on line 312.

redis/asyncio/cluster.py
```diff
@@ -477,6 +497,14 @@ def get_connection_kwargs(self) -> Dict[str, Optional[Any]]:
         """Get the kwargs passed to :class:`~redis.asyncio.connection.Connection`."""
         return self.connection_kwargs
 
+    def get_retry(self) -> Optional["Retry"]:
+        return self.retry
+
+    def set_retry(self, retry: "Retry") -> None:
+        self.retry = retry
+        for node in self.get_nodes():
+            node.connection_kwargs.update({"retry": retry})
```
utkarshgupta137, line 506: Shouldn't we also update the retry attribute for existing connections?

barshaul: good catch, added

redis/cluster.py
chayim: I'd rather do this removal in redis-py <next major version> So.. yuck, but let's leave it.

redis/cluster.py
chayim: I don't think there's a right answer here, without data. Given that they can be overriden - that's part of the answer. I'd rather allow for a lower retry threshold (5), while allowing users to override. Reset to 5 seems reasonable - but again, I have no measured, scientific reason to dictate any number

redis/asyncio/cluster.py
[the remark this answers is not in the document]
chayim: Retry consistency (with the other interfaces as well) is generally what we should be aiming for. I think the entire retry piece is off - across all our interfaces.  In general, we should have a Retry object, which during initialization reqiures a number of retries - across the board. This will break everything, but IMHO is a worthy breakage (in the future).

For this release and these blocks...  I think that a variant of the if/else check is reasonable. Raising an exception when mutually exclusive arguments are passed in is fine - so long as doctstings and contributed examples indicate that this is the desired behaviour.

redis/cluster.py
barshaul: changed to 5.

barshaul:
@chayim @dvora-h finished round

chayim:
@dvora-h let's do one more review to make sure we're good to go.

dvora-h:
I think this is ready

dvora-h (approved):

redis/asyncio/cluster.py
```diff
@@ -298,6 +306,17 @@ def __init__(
             # Call our on_connect function to configure READONLY mode
             kwargs["redis_connect_func"] = self.on_connect
 
+        if retry or retry_on_error or connection_error_retry_attempts > 0:
+            # Set a retry object for all cluster nodes
+            self.retry = retry or Retry(
+                default_backoff(), connection_error_retry_attempts
+            )
+            if retry_on_error is None:
```
utkarshgupta137, line 314: If an empty list is passed, it will not get the default value. Better to do: `if not retry_on_error`

redis/asyncio/cluster.py
utkarshgupta137: Not resolved.

redis/asyncio/cluster.py
chayim: @utkarshgupta137 is correct. This only checks the None value, but an empty list can be passed, given the type hint.

latelan:
I met the same problem with ```TypeError: cannot pickle '_thread.lock' object```,  Is this PR missed in 4.4.0-rc3?

chayim:
> I met the same problem with `TypeError: cannot pickle '_thread.lock' object`, Is this PR missed in 4.4.0-rc3?

This PR does not exist in 4.4.0-rc3. We're looking at 4.4.0-rc4 for this...

redis/asyncio/cluster.py
```diff
@@ -319,7 +339,6 @@ def __init__(
         self.reinitialize_steps = reinitialize_steps
         self.cluster_error_retry_attempts = cluster_error_retry_attempts
         self.connection_error_retry_attempts = connection_error_retry_attempts
-
```
utkarshgupta137, line 322: Nitpick: Don't delete this line.

dvora-h: I didn't delete it, just moved it to line 309

chayim (approved):

PR #2386, opened 2022-09-19
Supply chain risk reduction: remove dependency on library named `deprecated`

akx:
No need for an external library just for 5 annotations.

### Pull Request check-list

### Description of change

This PR removes the dependency on the `@deprecated` library in favor of a simple 8-line function.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -22,6 +22,7 @@
     * Fix RedisCluster to immediately raise AuthenticationError without a retry
     * ClusterPipeline Doesn't Handle ConnectionError for Dead Hosts (#2225)
     * Remove compatibility code for old versions of Hiredis, drop Packaging dependency
+    * The `deprecated` library is no longer a dependency
 
 * 4.1.3 (Feb 8, 2022)
     * Fix flushdb and flushall (#1926)
--- a/redis/commands/bf/commands.py
+++ b/redis/commands/bf/commands.py
@@ -1,8 +1,6 @@
-from deprecated import deprecated
-
 from redis.client import NEVER_DECODE
 from redis.exceptions import ModuleError
-from redis.utils import HIREDIS_AVAILABLE
+from redis.utils import HIREDIS_AVAILABLE, deprecated_function
 
 BF_RESERVE = "BF.RESERVE"
 BF_ADD = "BF.ADD"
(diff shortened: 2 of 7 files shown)


chayim:
While this removes the dependency, I don't really see this as necessary. It's less expressive than a decorator - and I'd like to keep that interface. If you'd like to introduce a decorator that provides the same namespace, that's a separate conversation. 

As a project, it would appear that [deprecated](https://github.com/tantale/deprecated) has stalled. But, the need to decorate functions, _potentially_ classes, and **definitely** arguments exists.

akx:
@chayim Thanks for the comments.

My reasoning is that pulling in two dependencies (`deprecated` and its dependency `wrapt`) just to basically format a string and call `warnings.warn()` in 5 places is unwarranted. I also feel there's a nonzero supply chain risk in it for such a heavily used project such as `redis-py`. (One could also argue that it's more cost-effective bandwidth-wise, and better for the environment to not have the world necessarily [download those extra dependencies some million times per day](https://pepy.tech/project/redis)...)

As for the expressiveness, if you're thinking about e.g. introspecting functions to see whether they're `@deprecated`, I don't think that really matters..?

For instance, Django happily just calls `warnings.warn` without a wrapper wherever they need to warn about a deprecation, á la https://github.com/django/django/blob/10bb21e71e3668f770493e2af0d0e0271830f7a3/django/db/models/expressions.py#L1551-L1561 ...

akx:
@chayim Have you had a chance to look at my comments above?

chayim:
@akx I'm more than happy to reduce any (and frankly all) dependencies. I'd like to hard reduce them in general, and keep things tidy. However, I'd like to maintain deprecation as a decorated interface, rather than an in-function call.

The differences may be minor, but I believe it improves readability, which is something we generally need to improve.

I'm VERY pro removing the dependency. However, that merge would be contingent on a _decorator_ providing the same functionality. Any chance you're interested in modifying your PR accordingly (pretty please!)?

akx:
@chayim Sure, I added a `deprecated_function` decorator that does what `deprecated` did for functions. When single parameters may be deprecated in the future, the function would call `warn_deprecated` directly.

chayim:
@dvora-h when you're cool with this... let's get it into the next milestone (this week!)?.  Unfortunately @akx I can't review this due to timing.  But, love that we're going down this path...

dvora-h:
@akx LGTM!
Can you just add your decorator to the deprecated commands in SEARCH so the CI will pass? I added deprecation to them after you opened this PR...

akx:
@dvora-h Sure! Rebased & fixed search.

dvora-h (approved):

PR #2392, opened 2022-09-20
Add support for new TDIGEST features and changes

dvora-h:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [ ] Does `$ tox` pass with this change (including linting)?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?
- [ ] Was the change added to CHANGES file?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

_Please provide a description of the change here._

PR #2394, opened 2022-09-21
Reinstate codecov in CI

akx:
### Pull Request check-list

### Description of change

This PR reinstates (and upgrades) the `codecov` CI action that was removed in #2392 (321c57faff6f9fbc5cc6cc21533dac2090f6d360).


dvora-h (commented):
Why not just uncomment lines 84-88?

akx:
@dvora-h 

> Why not just uncomment lines 84-88?

Because the commented lines above it were/are on the wrong indent level, and it was hard to see the correct indentation without moving the block.

Also, the execution order of the "show results" action and this don't really matter.

This also changes the action version to v3, and removes the token line, so the diff is probably about the same size regardless.

dvora-h:
The execution order really doesn't matter, but:
1. We prefer to include the name.
2. I don't understand why we don't need the token?

akx:
> The execution order really doesn't matter, but:
> 
> 1. We prefer to include the name.

Well, sure, I can put it back. 🤷 

> 2. I don't understand why we don't need the token?

Because this is a public/OSS repository, and public repositories don't need tokens on Codecov.

dvora-h (approved):

PR #2404, opened 2022-09-29
Add to_string method for GeoValue

AnneYang720:
### Pull Request check-list

### Description of change

Add to_string method for GeoValue. The lack of to_string will cause NotImplementedError when building query string from Nodes.

### DIFF
--- a/redis/commands/search/querystring.py
+++ b/redis/commands/search/querystring.py
@@ -133,12 +133,7 @@ class GeoValue(Value):
         self.unit = unit
 
     def to_string(self):
-        return "[{0} {1} {2} {3}]".format(
-            self.lon,
-            self.lat,
-            self.radius,
-            self.unit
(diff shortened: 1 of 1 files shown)


redis/commands/search/querystring.py
```diff
@@ -132,6 +132,14 @@ def __init__(self, lon, lat, radius, unit="km"):
         self.radius = radius
         self.unit = unit
 
+    def to_string(self):
+        return "[{0} {1} {2} {3}]".format(
+            self.lon,
+            self.lat,
+            self.radius,
+            self.unit
+        )
```
akx, line 141: This would be more succinct as an f-string:
```suggestion
        return f"[{self.lon} {self.lat} {self.radius} {self.unit}]"
```

dvora-h (approved):

PR #2416, opened 2022-10-13
Fix issue with `pack_commands` returning an empty byte sequence

jmcbailey:
### Pull Request check-list

### Description of change

Under certain circumstances `pack_commands()` adds an empty byte sequence (`b""`) to the returned list, which when connecting to a Redis server over SSL could cause an EOF error:

```
redis.exceptions.ConnectionError: Error while reading from socket: (8, 'EOF occurred in violation of protocol (_ssl.c:2483)')
```
In our case we are running an async Falcon application using the `uvicorn` worker type; we ran into the problem when using a Redis pipeline to perform an `MSET` followed by one or more `EXPIRE` commands. The values in the `MSET` are images, generally around 5 to 6kb in size. From what I can see it was certain images within a very specific size range, where the length 
 of the value was just under 6000, that caused the problem (6000 being the `buffer_cutoff` value used in `pack_commands`).

(Note that in some/many cases the `b""` doesn't seem to cause a problem; I wrote a simple script to reproduce the error, executing the code in the standard `asyncio` event loop, but couldn't reproduce it that way. It was only by creating a simple Falcon/uvicorn app and executing it within that that I could reproduce it; I guess it's to do with the fact that uvicorn uses `uvloop` instead of the asyncio event loop, and there must be some difference in the underlying SSL transport.)

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -922,7 +922,8 @@ class Connection:
                     or chunklen > buffer_cutoff
                     or isinstance(chunk, memoryview)
                 ):
-                    output.append(SYM_EMPTY.join(pieces))
+                    if pieces:
+                        output.append(SYM_EMPTY.join(pieces))
                     buffer_length = 0
                     pieces = []
 
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -930,7 +930,8 @@ class Connection:
                     or chunklen > buffer_cutoff
                     or isinstance(chunk, memoryview)
                 ):
-                    output.append(SYM_EMPTY.join(pieces))
+                    if pieces:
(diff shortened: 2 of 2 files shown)


chayim:
@jmcbailey Can you think of a test that can help us reproduce this?

jmcbailey:
> @jmcbailey Can you think of a test that can help us reproduce this?

@chayim I can certainly add a unit test for the method(s) to cover this scenario. Adding a test to reproduce the SSL error would be trickier...

chayim:
@jmcbailey Does this only **occur** with SSL? I'm happy to forgo the SSL side if they can be isolated.

@dvora-h can you help @jmcbailey out here?

dvora-h (approved):

PR #2418, opened 2022-10-15
Remove the superflous SocketBuffer from asyncio PythonParser

kristjanvalur:
### Pull Request check-list

### Description of change

asyncio `PyhonParser` was using a `SocketBuffer` class to store incoming data.  This is probably a carry-over from the 
non-async version, but in asyncio, it is not needed.  We can read directly from the `_stream` which is a `StreamReader` with
an internal buffer of 64KB. 

This change removes redundant code and speeds up python parsing by relying on already-provided buffering and `readline()` functionality.

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -207,12 +207,7 @@ class BaseParser:
 
 
 class SocketBuffer:
-    """Async-friendly re-impl of redis-py's SocketBuffer.
-
-    TODO: We're currently passing through two buffers,
-        the asyncio.StreamReader and this. I imagine we can reduce the layers here
-        while maintaining compliance with prior art.
-    """
+    """Async-friendly re-impl of redis-py's SocketBuffer."""
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


kristjanvalur:
In addition, the `socket_read_size` is now  more or less unused.  it is used by `HiredisParser` as the batch size to move between the input stream and the hiredis parser.  I'd like to remove it from the Parsers, use an arbitrary large value in the HiredisParser, and instead use it when instantiating the stream, using `open_connection(limit=socket_read_size)` or similar which will directly affect read buffering.

kristjanvalur:
@dvora-h , is this something you'd consided including in the next release?  It sort of forms the conclusion to a lot of my asyncio related simplifications in the library.

dvora-h (approved):

dvora-h:
@kristjanvalur Thanks for this!
Can you please remove the timeseries tests changes from this PR?
Because:
1. It is not related here, we prefer to do such things in separate PR's
2. Following changes in the module, this is happening there now and there is no need for it... (we have already changed the test accordingly)

kristjanvalur:
Sure, I can remove that.  Please understand, I find myself often adding unrelated fixes to unittests when doing these pull requests, because the unittests are sometimes very flaky.  Of course they're not part of the PR as such, but I _do_ like my PRs to pass the CI pipeline regardless.

dvora-h (approved):

kristjanvalur:
Regarding the socket_read_size, I'll make a separate PR for that once this one is merged, so that you can evaluate that change separately.

dvora-h:
Thanks @kristjanvalur for all your hard work!

Issue #2420, opened 2022-10-19
AsyncIO Cluster Mode Lock doesn't work


KMilhan:
**Version**: 4.3.x - 4.4.0rc2

**Platform**: Python 3.10 on Ubuntu 20.04

**Description**: AsyncIO Cluster mode doesn't seem to support Lock

Reproducible code is this

```python
async def test_aiolock():
    from redis.asyncio.lock import Lock as ALock

    r: redis.asyncio.RedisCluster = aredis()  # Async Redis cluster connection
    counter = 0
    async with ALock(
        redis=r, name="test-lock", timeout=None, sleep=0.01, blocking=True
    ):
        counter += 1
    assert counter == 1
```
Log is as follow

```log
============================= test session starts ==============================
collecting ... collected 1 item

test_cache.py::test_aiolock FAILED                                       [100%]
tests/test_cache.py:30 (test_aiolock)
async def test_aiolock():
        from redis.asyncio.lock import Lock as ALock
    
        r: redis.asyncio.RedisCluster = aredis()
        counter = 0
>       async with ALock(
            redis=r, name="test-lock", timeout=None, sleep=0.01, blocking=True
        ):

test_cache.py:36: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../venv/lib/python3.10/site-packages/redis/asyncio/lock.py:147: in __init__
    self.register_scripts()
../../venv/lib/python3.10/site-packages/redis/asyncio/lock.py:153: in register_scripts
    cls.lua_release = client.register_script(cls.LUA_RELEASE_SCRIPT)
../../venv/lib/python3.10/site-packages/redis/commands/core.py:5204: in register_script
    return AsyncScript(self, script)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <redis.commands.core.AsyncScript object at 0x7f6d92d03d30>
registered_client = <redis.asyncio.cluster.RedisCluster object at 0x7f6d933e6fb0>
script = "\n        local token = redis.call('get', KEYS[1])\n        if not token or token ~= ARGV[1] then\n            return 0\n        end\n        redis.call('del', KEYS[1])\n        return 1\n    "

[...]

PR #2446, opened 2022-11-06
Enable AsyncIO cluster mode lock

KMilhan:
### Pull Request check-list

- [x] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?  -> ***Couldn't configure my fork's Action -> CI here passed***

### Description of change

Enable AsyncIO cluster mode lock. AsyncIO Cluster mode has not been providing `Lock`, due to the fact that the `core.py` trying to find the `connection_pool` attribute where it's not provided in the AsyncIO cluster client. This change will try to find `get_encoder` in the client if `connection_pool` is unavailable.

Also, this change copies the` lock` method of AsyncIO client (please let me know if this is not appropriate), providing the same API to the users whether they are using non-cluster or cluster.

An example is not added as there is no currently available cluster-mode-only example for asyncio Redis.

Closes: #2420

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -23,6 +23,7 @@
     * ClusterPipeline Doesn't Handle ConnectionError for Dead Hosts (#2225)
     * Remove compatibility code for old versions of Hiredis, drop Packaging dependency
     * The `deprecated` library is no longer a dependency
+    * Enable Lock for asyncio cluster mode
 
 * 4.1.3 (Feb 8, 2022)
     * Fix flushdb and flushall (#1926)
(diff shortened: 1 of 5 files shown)


KMilhan:
Maybe we can simplify by using `.get_encoder()` only in Async[Cluster]?
Below is from redis/asyncio/client.py
```python
    def get_encoder(self):
        """Get the connection pool's encoder"""
        return self.connection_pool.get_encoder()
```

KMilhan:
I also have no clue why CodeDev reported such a big drop in the coverage, as I actually expanded the test set by including cluster mode for asyncio lock (probably this shouldn't change the coverage)

utkarshgupta137 (approved):

dvora-h (approved):
LGTM!

PR #2452, opened 2022-11-11
Add OpenTelemetry example with Uptrace backend

vmihailenco:
This PR adds some docs that explain how start monitoring redis-py performance using [OpenTelemetry](https://opentelemetry.io/) and [Uptrace](https://github.com/uptrace/uptrace). Here is how the [doc](https://vmihailenco-redis-py.readthedocs.io/en/latest/opentelemetry.html) page will look like.

I've briefly discussed this on Slack with @chayim and he encouraged me to open a PR to start a discussion. So here I am :)


chayim:
@vmihailenco I think we should probably move these things from the docker folder (to start) into docs/examples. Perhaps even start a subsection on monitoring. WDYT?

docs/opentelemetry.rst
```diff
@@ -0,0 +1,102 @@
...
+OpenTelemetry instrumentation
+-----------------------------
+
+Instrumentations are plugins for popular frameworks and libraries that use OpenTelemetry API to record important operations, for example, HTTP requests, DB queries, logs, errors, and more.
+
+To install OpenTelemetry `instrumentation <https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/redis/redis.html>`_ for redis-py:
+
+.. code-block:: shell
+
+   pip install opentelemetry-instrumentation-redis==0.35b0
```
chayim, line 20: IMHO if we tie this to a version, it's one more thing to maintain.

docs/opentelemetry.rst
```diff
@@ -0,0 +1,102 @@
...
+   from opentelemetry.instrumentation.redis import RedisInstrumentor
+
+   RedisInstrumentor().instrument()
+
+Once the code is patched, you can use redis-py as usually:
+
+.. code-block:: python
+
+   # Sync client
+   client = redis.StrictRedis(host="localhost", port=6379)
```
chayim, line 35: StrictRedis is now redis - so redis.Redis?

docs/opentelemetry.rst
```diff
@@ -0,0 +1,102 @@
...
+   client.get("my-key")
+
+   # Async client
+   client = redis.asyncio.Redis(host="localhost", port=6379)
```
chayim, line 39: Actually, also defaults are localhost/6379

docs/opentelemetry.rst
vmihailenco: Agreed, it is a copy-paste mistake

docs/opentelemetry.rst
vmihailenco: Done

docs/opentelemetry.rst
vmihailenco: Yeah, but my thinking is that with explicit values users can more easily copy-paste the example and change credentials. WDYT?

vmihailenco:
>I think we should probably move these things from the docker folder (to start) into docs/examples. Perhaps even start a subsection on monitoring. WDYT?

A subsection is probably too much to start with, but I will move everything to docs/examples :+1:

vmihailenco:
>Perhaps even start a subsection on monitoring

I think I misread subsection as subfolder. I guess you mean a section in examples.rst? I will play with it.

vmihailenco:
@chayim PTAL. All should be fixed, but I was not able to update `examples.rst` in a meaningful manner. Perhaps you could show me an example?

vmihailenco:
The CI failure seems unrelated...

vmihailenco:
@chayim I've added the [Jupyter notebook](https://vmihailenco-redis-py.readthedocs.io/en/latest/examples/opentelemetry_api_examples.html) we've discussed and updated the title. PTAL

dvora-h (approved):

vmihailenco:
@dvora-h thanks!

PR #2463, opened 2022-11-20
Added a replacement for the default cluster node in the event of failure.

barshaul:
### Pull Request check-list

### Description of change

For non-key based commands, we choose the default cluster node to execute the command on (for example, PING, CLUSTER NODES, CONFIG, etc..).

When the default node fails, we'll replace it with a new node before retrying the command.
Improves failover handling. 

- Separated AUTH cluster/non-cluster tests

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1501,7 +1501,7 @@ class ClusterPipeline(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterComm
                         raise result
 
             if is_default_node:
-                self.replace_default_node()
+                client.replace_default_node()
 
         return [cmd.result for cmd in stack]
 
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -11,7 +11,6 @@ from typing import (
     List,
     Mapping,
     Optional,
-    Tuple,
     Type,
     TypeVar,
     Union,
@@ -517,14 +516,9 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
 
     async def _determine_nodes(
         self, command: str, *args: Any, node_flag: Optional[str] = None
-    ) -> Tuple[List["ClusterNode"], bool]:
-        """Determine which nodes should be executed the command on
-
-        Returns:
-            tuple[list[Type[ClusterNode]], bool]:
-                A tuple containing a list of target nodes and a bool indicating
-                if the return node was chosen because it is the default node
-        """
+    ) -> List["ClusterNode"]:
+        # Determine which nodes should be executed the command on.
+        # Returns a list of target nodes.
         if not node_flag:
             # get the nodes group for this command if it was predefined
             node_flag = self.command_flags.get(command)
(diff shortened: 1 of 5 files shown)


dvora-h (approved):

redis/asyncio/cluster.py
```diff
@@ -516,35 +517,44 @@ def set_response_callback(self, command: str, callback: ResponseCallbackT) -> No
...
         self, command: str, *args: Any, node_flag: Optional[str] = None
-    ) -> List["ClusterNode"]:
+    ) -> Tuple[List["ClusterNode"], bool]:
+        """Determine which nodes should be executed the command on
+
+        Returns:
+            tuple[list[Type[ClusterNode]], bool]:
+                A tuple containing a list of target nodes and a bool indicating
+                if the return node was chosen because it is the default node
+        """
```
utkarshgupta137, line 527: This is an internal function, so we should avoid adding docs. We can keep this in the comments if required.

redis/asyncio/cluster.py
```diff
@@ -516,35 +517,44 @@ def set_response_callback(self, command: str, callback: ResponseCallbackT) -> No
 
     async def _determine_nodes(
         self, command: str, *args: Any, node_flag: Optional[str] = None
-    ) -> List["ClusterNode"]:
+    ) -> Tuple[List["ClusterNode"], bool]:
```
utkarshgupta137, line 520: Since `is_default_node` value will only be checked in the rare case of cluster failover or scaling/re-sharding, might it be better to compare the target_node with self.nodes_manager.default_node instead of changing the return signature? Even though this is an internal function, users/packages might still be overwriting it to add/modify functionality.

barshaul: Changed

redis/asyncio/cluster.py
barshaul: Changed

redis/asyncio/cluster.py
```diff
@@ -664,6 +670,11 @@ async def execute_command(self, *args: EncodableT, **kwargs: Any) -> Any:
                         raise RedisClusterException(
                             f"No targets were found to execute {args} command on"
                         )
+                    if (
+                        len(target_nodes) == 1
+                        and target_nodes[0] == self.get_default_node()
+                    ):
+                        is_default_node = True
```
utkarshgupta137, line 677: This check could be moved inside the `if self._initialize:` block.

redis/asyncio/cluster.py
```diff
@@ -1450,8 +1462,9 @@ async def _execute(
                     )
             if len(target_nodes) > 1:
                 raise RedisClusterException(f"Too many targets for command {cmd.args}")
-
             node = target_nodes[0]
+            if node == client.get_default_node():
+                is_default_node = True
```
utkarshgupta137, line 1467: This is inside a for loop, so you might want an or operation here.

barshaul: Hey, please see the changes in the new commit.

redis/asyncio/cluster.py
barshaul: changed

utkarshgupta137 (approved):

Issue #2407, opened 2022-10-05
4.4.0rc2: read() called while another coroutine is already waiting for incoming data


JonasKs:
**Version**: [4.4.0rc2](https://github.com/redis/redis-py/releases/tag/v4.4.0rc2)

**Platform**: MacOS and Windows. Python 3.10.6.

**Description**: We've noticed that we are no longer able to `enqueue_job()` with [`arq`](https://github.com/samuelcolvin/arq). All of this **works** in `4.4.0rc1`. 

```
Traceback (most recent call last):
  File "<redacted>", line <redacted>, in <redacted>
    await redis.enqueue_job(
  File "<redacted>", line 137, in enqueue_job
    if any(await asyncio.gather(pipe.exists(job_key), pipe.exists(result_key_prefix + job_id))):
  File "<redacted>/lib/python3.10/site-packages/redis/asyncio/client.py", line 1161, in immediate_execute_command
    return await conn.retry.call_with_retry(
  File "<redacted>lib/python3.10/site-packages/redis/asyncio/retry.py", line 59, in call_with_retry
    return await do()
  File "<redacted>lib/python3.10/site-packages/redis/asyncio/client.py", line 463, in _send_command_parse_response
    return await self.parse_response(conn, command_name, **options)
  File "<redacted>lib/python3.10/site-packages/redis/asyncio/client.py", line 1293, in parse_response
    result = await super().parse_response(connection, command_name, **options)
  File "/<redacted>/lib/python3.10/site-packages/redis/asyncio/client.py", line 505, in parse_response
    response = await connection.read_response()
  File "<redacted>/lib/python3.10/site-packages/redis/asyncio/connection.py", line 886, in read_response
    response = await self._parser.read_response(
  File "<redacted>/lib/python3.10/site-packages/redis/asyncio/connection.py", line 468, in read_response
    await self.read_from_socket()
  File "<redacted>/lib/python3.10/site-packages/redis/asyncio/connection.py", line 451, in read_from_socket
    buffer = await self._stream.read(self._read_size)
  File "<redacted>/lib/python3.10/asyncio/streams.py", line 669, in read
    await self._wait_for_data('read')
  File "<redacted>lib/python3.10/asyncio/streams.py", line 488, in _wait_for_data
    raise RuntimeError(
RuntimeError: read() called while another coroutine is already waiting for incoming data
```

I can try to provide a reproducible example later if you'd like, but that would require `arq` as a dependency.

[...]

PR #2471, opened 2022-11-24
Async clusters: Support creating locks inside async functions

utkarshgupta137:
### Pull Request check-list

### Description of change

This allows creating the client at the module level or before the event loop has been initialized.
Ref: https://github.com/encode/starlette/issues/1315#issuecomment-1096256231

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -356,11 +356,13 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
         )
 
         self._initialize = True
-        self._lock = asyncio.Lock()
+        self._lock: Optional[asyncio.Lock] = None
 
     async def initialize(self) -> "RedisCluster":
         """Get all nodes from startup nodes & creates connections if not initialized."""
         if self._initialize:
+            if not self._lock:
+                self._lock = asyncio.Lock()
             async with self._lock:
                 if self._initialize:
                     try:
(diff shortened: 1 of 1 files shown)


utkarshgupta137:
@chayim @dvora-h This is a minor fix which should be included in 4.4.0. If it is still far away, I can raise a PR for the 4.3 branch as well.

chayim:
@utkarshgupta137 We're still waiting on #2407 and given your fix.. I'm pretty good with this getting into 4.4 as that merge. Let's give a bit longer (end of week maybe) for [your PR in arq](https://github.com/samuelcolvin/arq/pull/366), and then we can merge it down + decide.

Does that work for you? I'd really like to **not** have multiple releases :D

utkarshgupta137:
> @utkarshgupta137 We're still waiting on #2407 and given your fix.. I'm pretty good with this getting into 4.4 as that merge. Let's give a bit longer (end of week maybe) for [your PR in arq](https://github.com/samuelcolvin/arq/pull/366), and then we can merge it down + decide.
> 
> Does that work for you? I'd really like to **not** have multiple releases :D

Sure, it would work for me. But I don't think it is reasonable to wait for a third-party package to fix what is essentially a misuse (undefined behaviour) on their end. For all we know, there could be many other packages out there misusing some functionality of this package...
Also, there are other breaking changes in this release, so ideally, packages using redis as a dependency should have a version cap.

chayim:
@utkarshgupta137 I randomly head out to the internet and contribute version caps for reasons like this. Part of a general goal of not breaking the community means I was happy to delay in this case, especially with the backport to 4.3.5.

Thanks again for the fix in arq - this goes into 4.4x.

chayim (approved):

Issue #2474, opened 2022-11-29
Community Question: True/False Encoding


chayim:
Currently, the redis-py encoder does not allow a user to store true booleans (True/False) within redis. Near as I can determine, this is due to the history of the library, as opposed to a limitation itself.

As part of *starting* to clean up parsers and encoders, I've removed this limitation in my personal branch. This means that a user, would be able to:

```
r = redis.Redis()
r.set('a', True)
r.set('b', False)
```

This change means that a ```redis.exceptions.DataError``` would not be raised, and the string representation of True/False would be stored. It would appear that the [RESP2](https://redis.io/docs/reference/protocol-spec/) protocol does not support booleans. As a result, for RESP2 backed Redis, we cannot return a True boolean without parsing the string (IMHO expensive, we should not). In [RESP3](https://github.com/antirez/RESP3/blob/master/spec.md) there is true boolean support, and this is a non-issue, hence these changes, as we think about them.

Given the above, what does the community think? Does **fixing** the encoding so that a user can set a True/False (which yes, is encoded) constitute a breaking change? Is anyone *opposed* to this going in to RESP2 protocol support? I would argue that this provides users with functionality that they can choose to use, but I'd like to hear from the community, so that it's either not a breaking change, or an intentionally breaking change.

judahrand:
I think this does constitute a potentially breaking change imo. Even with a flag somewhere to enable to behaviour it doesn't seem that valuable as the user can do this parsing themselves in any case. If you need to set the flag or do the parsing I can't see it being very different. Support for RESP3, however, would be much much more valuable. I've been hoping for this for a long time to simplify use cases which use Server-assisted Client Side Caching.

PR #2476, opened 2022-11-29
Migrating parsers to a single namespace

chayim:
As part of beginning support for RESP3, I think it makes sense to clean up our parser layout. This is a forklift, not a facelift, meaning the methods themelves haven't been changed, inasmuch organized into a single place. 

However, a byproduct of this is a small improvement to sync cluster key retrieval (6% it seems, based on tessing), due to changing a type and split, into a single line. This is part of the thinking that started the conversation in #2474


redis/parsers/commands.py
```diff
@@ -1,6 +1,11 @@
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Union
```
utkarshgupta137, line 1: This file & class were always misnamed. It should ideally have "cluster" somewhere in there.

chayim:
abandoned, thanks to 5.x

PR #2477, opened 2022-11-29
Fix redis version check in ACL tests

uglide:
Tests with sanitize-payload assertions require Redis 6.2 and fail on Redis 6.0.

### Pull Request check-list

### Description of change

### DIFF
--- a/tests/test_asyncio/test_commands.py
+++ b/tests/test_asyncio/test_commands.py
@@ -19,6 +19,7 @@ from tests.conftest import (
 )
 
 REDIS_6_VERSION = "5.9.0"
+REDIS_6_2_VERSION = "6.2.0"
 
 
 @pytest_asyncio.fixture()
@@ -114,7 +115,7 @@ class TestRedisCommands:
         password = await r.acl_genpass()
         assert isinstance(password, str)
 
-    @skip_if_server_version_lt(REDIS_6_VERSION)
+    @skip_if_server_version_lt(REDIS_6_2_VERSION)
     @skip_if_server_version_gte("7.0.0")
     async def test_acl_getuser_setuser(self, r_teardown):
         username = "redis-py-user"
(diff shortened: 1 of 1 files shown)


tests/test_asyncio/test_commands.py
```diff
@@ -114,7 +115,7 @@ async def test_acl_genpass(self, r: redis.Redis):
         password = await r.acl_genpass()
         assert isinstance(password, str)
 
-    @skip_if_server_version_lt(REDIS_6_VERSION)
+    @skip_if_server_version_lt(REDIS_6_2_VERSION)
```
chayim, line 118: IMHO, maybe we should change REDIS_6_VERSION instead, to reflect a real release (6.0.0), and update tests associated here? I believe this was originally a placeholder we inherited when merging the projects.

It looks like [REDIS_6_VERSION](https://sourcegraph.com/github.com/redis/redis-py/-/blob/tests/test_asyncio/test_commands.py?L21:1&subtree=true#tab=references) is only used by these tests, meaning updating it, and the tests herein should go together given the helps docs (example: [ACL SETUSER](https://redis.io/commands/acl-setuser/)

WDYT?

uglide:
@chayim It looks like there are no docs for this except comments in redis.conf sample file https://github.com/redis/redis/commit/ca1c182567add4092e9cb6ea829e9c5193e8fd55

chayim:
> It looks like there are no docs for this except comments in redis.conf sample file https://github.com/redis/redis/commit/ca1c182567add4092e9cb6ea829e9c5193e8fd55

In that case it looks like we should walk back the tests. This isn't a client feature. IMHO we should remove the acl flag, since that's on the redis side. But, we should still update the tests to reflect the right parent version.

WDYT?

uglide:
@chayim I agree that this PR is insufficient to address the root cause problem, which is redis-py should be tested against Redis 5.x, 6.x, 6.2 and 7.0

PR #2478, opened 2022-11-29
Adding connection step to bloom filter examples

### LINKED ISSUE #2415: Error in documentation regarding initialisation of RedisBloom and TopK
[Link to documentation](https://redis-py.readthedocs.io/en/stable/redismodules.html)

**Version**: redis-py version: 4.3.4 & redis-stack running on Docker (v6.2.7)

**Platform**: Python 3.10.6 on macOS
**Description**: Creating a bloom filter or TopK client according to the documentation throws an AttributeError

Sample:
```python
>>> import redis
>>> filter = redis.bf().create("bloom", 0.01, 1000)
AttributeError: module 'redis' has no attribute 'bf'
```

Solution:

```python
>>> from redis import Redis
>>> r = Redis()
>>> bf = r.bf()
>>> bf.create("bloom", 0.01, 1000)
True
```

I am facing similar issue with TopK filters as well.

I can raise a PR to solve this issue as well.

nermiller:
Fixes #2415, adds a connect step to bloom commands

### DIFF
--- a/docs/redismodules.rst
+++ b/docs/redismodules.rst
@@ -12,28 +12,28 @@ These are the commands for interacting with the `RedisBloom module <https://redi
 **Create and add to a bloom filter**
 
 .. code-block:: python
-
+   
+    import redis 
     r = redis.Redis()
-    import redis
-    filter = redis.bf().create("bloom", 0.01, 1000)
-    filter.add("bloom", "foo")
+    r.bf().create("bloom", 0.01, 1000) 
+    r.bf().add("bloom", "foo")
 
 **Create and add to a cuckoo filter**
 
 .. code-block:: python
 
+    import redis 
     r = redis.Redis()
-    import redis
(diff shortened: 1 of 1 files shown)


savynorem (changes requested):
@nermiller sorry I didn't articulate this well - the connection will need to be after the import, then the connection name is what needs to be used - also, the return from create is a boolean, it can't actually be used to alter the data structure, so they'll need to be more like
`
import redis
r = redis.Redis()
r.bf().create("bloom", .01, 1000)
r.bf().add("bloom", foo)
`

The return values are pretty much only used as a success / fail / did something change type of measure, they don't return a way to access the object.

nermiller:
@savynorem please review again. Thanks!

savynorem (approved):
👍  looks great to me!

chayim (approved):

PR #2494, opened 2022-12-08
Allow EVAL_RO and EVALSHA_RO to be routed to read replica

### LINKED ISSUE #2490: EVAL_RO, EVALSHA_RO doesn't get routed to replicas
Thanks for wanting to report an issue you've found in redis-py. Please delete this text and fill in the template below.  
It is of course not always possible to reduce your code to a small test case, but it's highly appreciated to have as much data as possible. Thank you!

redis-py 4.4.0, redis 7.0

python 3.11.0 on Mac

Commands like `EVAL_RO` and `EVALSHA_RO` are always directed to the master, even though they can be routed to replicas as they are read-only commands

I suspect the cause to be READ_COMMANDS missing few read only commands
 https://github.com/redis/redis-py/blob/6219574b042a6596b150ca8248441198f01f8c87/redis/commands/cluster.py#L49-L93

Can i make a PR to add those commands or is there any other rationale that prevents adding EVAL_RO and EVALSHA_RO into READ_COMMANDS?

zakaf:
### Pull Request check-list

### Description of change
- fixes #2490 by adding EVAL_RO and EVALSHA_RO to READ_COMMANDS
- minor typo fix (Lue -> Lua)
- `test_reading_from_replicas_in_round_robin` asserts multiple `GET` commands getting routed to master and replica
  - two `READONLY` calls would only happen with 3 or more `GET` commands only if there are two nodes for the slot

In addition, i've tested locally using the below method
1. running MONITOR command on both the master and the replica node
2. RUN EVAL_RO and EVALSHA_RO multiple times
3. see the output of MONITOR to confirm that commands were routed in a round-robin manner


zakaf:
@chayim i've created a PR to address #2494. If you have any concern with the PR, please let me know

dvora-h (approved):

PR #2504, opened 2022-12-12
Raising NotImplementedError for certain CLUSTER commands

### LINKED ISSUE #1968: Add support for missing command CLUSTER BUMPEPOCH

### LINKED ISSUE #1969: Add support for missing command CLUSTER FLUSHSLOTS

dvora-h:
### Pull Request check-list

### Description of change

closes #1968 
closes #1969

### DIFF
--- a/redis/commands/cluster.py
+++ b/redis/commands/cluster.py
@@ -644,6 +644,16 @@ class ClusterManagementCommands(ManagementCommands):
         """
         return self.execute_command("CLUSTER LINKS", target_nodes=target_node)
 
+    def cluster_flushslots(self, target_nodes: Optional["TargetNodesT"] = None) -> None:
+        raise NotImplementedError(
+            "CLUSTER FLUSHSLOTS is intentionally not implemented in the client."
+        )
+
+    def cluster_bumpepoch(self, target_nodes: Optional["TargetNodesT"] = None) -> None:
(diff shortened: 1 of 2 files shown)


dvora-h:
@shacharPash @vladvildanov

chayim (approved):

redis/commands/cluster.py
```diff
@@ -644,6 +644,16 @@ def cluster_links(self, target_node: "TargetNodesT") -> ResponseT:
         """
         return self.execute_command("CLUSTER LINKS", target_nodes=target_node)
 
+    def cluster_flushslots(self, target_nodes: Optional["TargetNodesT"] = None) -> None:
+        raise NotImplementedError(
+            "CLUSTER FLUSHSLOTS is intentionally not implemented in the client."
```
vladvildanov, line 649: Would it be better to move this functionality into client rather than specify it for each command? So it will raise an exception whenever any not implemented command will be called

shacharPash (commented):
I didn't quite understand why these commands need to be written just for them to send an error that says that on purpose they are not implemented, wouldn't that just confuse the user?

redis/commands/cluster.py
chayim: Our standard is to do this here... in python at least, we share this way.

Issue #2499, opened 2022-12-10
BaseException at I/O corrupts Connection


ikonst:
tl;dr repeat of https://github.com/redis/redis-py/issues/360

My codebase uses gevent (w/[monkey-patching](https://www.gevent.org/api/gevent.monkey.html)) for concurrent workers, and schedules a [gevent.Timeout](https://www.gevent.org/api/gevent.timeout.html) to force a deadline on the workers. Since a timeout causes an exception to be raised from an arbitrary "yield point", there's risk of corrupting shared state if code is not written to be exception-safe.

The code talks to redis over a client that's shared between the workers. Socket I/O is a "yield point". Sending a command successfully, but then failing to read the entire response off the socket, gets `Connection` into an undefined state: a subsequent command would try to parse a response intended for a previous command, and in many cases would fail. Here's what a typical error would look like:
```
  File "site-packages/redis/client.py", line 3401, in parse_response
    result = Redis.parse_response(
  File "site-packages/redis/client.py", line 774, in parse_response
    return self.response_callbacks[command_name](response, **options)
  File "site-packages/redis/client.py", line 528, in <lambda>
    'SET': lambda r: r and nativestr(r) == 'OK',
  File "site-packages/redis/_compat.py", line 135, in nativestr
    return x if isinstance(x, str) else x.decode('utf-8', 'replace')
AttributeError: 'int' object has no attribute 'decode'
```
As you can see, we're trying to parse an integer response to a previous command (e.g. `":12345\r\n"`) as a string response to a `SET` command.

The `except Exception` block in `redis.connection.Connection.read_response` is intended to handle all sorts of parsing errors by disconnecting the connection:
https://github.com/redis/redis-py/blob/6219574b042a6596b150ca8248441198f01f8c87/redis/connection.py#L819-L821
but perhaps it could be changed to `except:` since e.g. `gevent.Timeout` is intentionally a `BaseException` to suggest that it should not be suppressed (and we are indeed not _suppressing_ it).

ikonst:
Digging through history:
- this was already reported in #1128
- then fixed in #1129 (cc @Chronial), released in 3.2.0
- then reverted in #2104 (cc @kristjanvalur), released in 4.4.0rc2

ikonst:
P.S. I can think of a few alternatives to `self.disconnect()` where the onus on recovery is on the next caller rather than in the midst of `BaseException` handling.

[...]

PR #2505, opened 2022-12-12
Add regression test for #360

ikonst:
### Pull Request check-list

### Description of change

Adding a regression test for #360. Since it currently fails due to #2499, marking the `BaseException` parameter with `xfail`: a fix for #2499 will have to update that.

### DIFF
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -112,3 +112,37 @@ async def test_connect_timeout_error_without_retry():
         await conn.connect()
     assert conn._connect.call_count == 1
     assert str(e.value) == "Timeout connecting to server"
+
+
+@pytest.mark.parametrize(
+    "exc_type",
+    [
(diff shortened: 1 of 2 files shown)


kristjanvalur:
Good stuff.  Let me see if I can resolve this in an elegant way.

kristjanvalur:
Ok, taking another look at this, this does not in fact seem to be a regression test for the issue.  I would expect exactly the behaviour that this test demonstrates.  That if you are interrupted in reading, then when you read again, you get the previous response.

The problem does not manifest at this level.  it manifests when issuing a command from the `Redis` class and the "response" part of it is interrupted.  Re-issuing a command on that connection will return an invalid response.   The correct behaviour, for `Redis` is to close the connection.

I'll see if I can produce a different regression test for the problem.

ikonst:
> That if you are interrupted in reading, then when you read again, you get the previous response.

Fair enough.

PR #2506, opened 2022-12-12
Revert #2104, provide way to disable disconnects in `read_response()`

### LINKED ISSUE #2103: Connection class disconnects on BaseException

kristjanvalur:
### Pull Request check-list

### Description of change

PR #2104, a fix to #2103 caused a regression for issue #1128.  This has been dicussed in issue #2499.

This PR provides 

- A regression test for issue #1128
- A revert of pr #2104, along with comments in the code
- A different way to fix #2103 with an optional argument to `Connection.read_response()`

A lot of code uses the Command/Response pattern and if an exception occurs during that, the conneciton is in an
indeterminate state:

 - `Connection.send_command()` has no way to continue sending a command if there is a benign interrupt, e.g. a Timeout.  Ensuring that the connection is closed in this case is prudent, because some code does capture BaseExceptions, e.g. `gevent` style code.
 - Code often pairs `send_command()` and `read_response()` and even if the error happens in `read_response()` (which is in principle resumable), application code has no way of knowing and so closing the connection is necessary.
 
 To support `PubSub`, and other cases where an application may want to properly handle Timeouts and other errors while reading reponses, a new argument is added to `read_response()`, which disables the automatic `disconnect()` of the connection when an exception is occurred.  It is then the responsibility of the caller to take appropriate action.


tests/test_asyncio/test_commands.py
```diff
@@ -2999,6 +3001,36 @@ async def test_module_list(self, r: redis.Redis):
...
+        await ready.wait()
+        await asyncio.sleep(0.01)
+        # the task is now sleeping, lets send it an exception
+        task.cancel()
+        try:
+            async with async_timeout.timeout(0.1):
+                assert await task == "done"
+        except asyncio.TimeoutError:
+            task.cancel()
+            await task
```
ikonst, line 3032: Is this a 2nd phase of the same test?

tests/test_commands.py
```diff
@@ -4712,6 +4716,34 @@ def test_psync(self, r):
...
+                # blocking pop
+                with patch.object(
+                    socket.socket, "recv_into", side_effect=CancelledError
+                ) as mock_recv:
+                    r.brpop(["nonexist"])
+            except CancelledError:
+                print("canc")
+                pass  # we got some BaseException.
+            # if all is well, we can continue.
+            r.set("status", "down")  # should not hang
```
ikonst, line 4739: Does this also exercise the "command expects int but got back str" scenario?

tests/test_commands.py
```diff
@@ -4712,6 +4716,34 @@ def test_psync(self, r):
...
+            return "done"
```
ikonst, line 4740:
```suggestion
```

tests/test_commands.py
```diff
@@ -4712,6 +4716,34 @@ def test_psync(self, r):
...
+    @pytest.mark.onlynoncluster
+    def test_interrupted_command(self, r: redis.Redis):
+        """
+        Regression test for issue #1128:  An Un-handled BaseException
+        will leave the socket with un-read response to a previous
+        command.
+        """
+        print("start")
+
+        def helper():
```
ikonst, line 4728:
```suggestion
        def helper() -> None:
```
IIUC there's no way to get a retval from a thread, though you can pass a mutable object and use that to communicate

Chronial:
Note that this is not enough to fix everything broken by #2104. At least the [other callsites](https://github.com/search?q=repo%3Aredis%2Fredis-py+send_command&type=code) of `send_command` and `read_response` are also affected.

ikonst:
Also, might be good to remove the `except Exception:` unless there's a distinct reason for it.

tests/test_commands.py
kristjanvalur: No.  It excercises the originally reported scenario:  A blocking wait is initiated, and even though a new command is sent on the channel, the server is still in blocking wait mode.
It is harder to interrupt a half-read response, although not impossible by using mocking.  I can look into that.

tests/test_commands.py
kristjanvalur: Also, there is no type hinting in redis-py, not going to start adding that to unit tests :)

kristjanvalur:
> Also, might be good to remove the `except Exception:` unless there's a specific reason for it.

Which one is that?

tests/test_commands.py
ikonst: Yeah, I mostly just meant for you to remove the retval. No need to type-annotate, though I do notice that some stuff is type-annotated.

tests/test_commands.py
ikonst: That scenario doesn't require half-read. It's an interrupt between send_command and parse_response:
```
send_command("EXISTS foo")
<interrupt>
send_command("GET foo")
parse_response() --> expected str, got int (from previous command)
```

tests/test_commands.py
```diff
@@ -4723,18 +4723,14 @@ def test_interrupted_command(self, r: redis.Redis):
         will leave the socket with un-read response to a previous
         command.
         """
-        print("start")
-
+        
         def helper():
-            try:
+            with pytest.raises(BaseException):
```
ikonst, line 4728: Note that unless you propagate exceptions from the thread to the test, we'd be unaware of this pytest.raises assertion failing.

Chronial:
Since you seem really committed to go down this route:

You now disconnect on the outside for any exception. Are you sure this is correct? This will at least disconnect twice sometimes. And are there really no exceptions that should go through without disconnect?

Since you are against rolling back #2104, you should denote the API change in the functions affected, because they now have a rather surprising API. `send_command` needs a comment explaining that if it raises, the connection is now in an undefined state and the caller needs to disconnect it. Same for `read_response` and `parse_response`.

If you succeed with #2510, that comment could change on the async response functions to explaining that if the Exception was a CancellationException, the connection is not undefined, but still holds the previous response. If it is another Exception like KeyboardInterrupt, the connection will still be broken.

ikonst:
Do we agree that under all scenarios, interrupted `send_command`:
- leaves the `socket` in a **defined state**
- leaves the `redis.Connection` in an **undefined state** (per RESP protocol)
?

So expecting `socket` to handle this (e.g. by disconnecting) would be wrong, since `socket` will be making assumptions about the protocol being implemented. However, should the responsibility fall unto `redis.Connection` or unto all its callers?

Arguments in favor of making `redis.Connection.send_command` responsible:
- Less code repetition
- Lower chance of forgetting to do it at the call sites
- Scope: `redis.Connection` _has_ knowledge of RESP, and it's expected that it'd make decisions based on that knowledge

tobymao:
Would it be easier to just roll back #2104?

ikonst:
@tobymao no, because it was done for a reason

tests/test_asyncio/test_commands.py
```diff
@@ -2999,6 +3001,32 @@ async def test_module_list(self, r: redis.Redis):
...
+        """
+        ready = asyncio.Event()
+
+        async def helper():
+            with pytest.raises(asyncio.CancelledError):
+                # blocking pop
+                ready.set()
+                await r.brpop(["nonexist"])
+            # if all is well, we can continue.  The following should not hang.
+            await r.set("status", "down")
```
ikonst, line 3019: Still tbd to validate that not only you don't "hang", but also that the next read response matches the command.

Chronial:
> Maybe a side-note, but when interruption can happen at any point, as it is with KeyboardInterrupt, I think no approach could protect against an undefined state, so maybe there's no use thinking about it.

The old state before #2104 seemed to manage just fine.

ikonst:
> Before https://github.com/redis/redis-py/pull/2104 this seemed to work quite well.

You can't guarantee that `KeyboardInterrupt` doesn't happen twice, once when you `recv` and another time before you run `disconnect()`. If something happened, it's by chance, because most of the time you had to Ctrl-C was during lengthy I/O calls.

Chronial:
In production, where this matters nobody is pressing Ctrl-C. And if they are, they should know not to press it twice if they want sensible shutdown. In the important scenario, like the one [you explained before](https://github.com/redis/redis-py/issues/2499#issuecomment-1348746819), the exception is raised by a framework. Obviously it should only raise a single one.

tests/test_commands.py
kristjanvalur: I'll add a half-read test case.

tests/test_commands.py
kristjanvalur: You're right, there is the odd type hinting, but It is not consistently applied.  I guess people add it to help their dev tools with code completion.

kristjanvalur:
> > > Also, might be good to remove the except Exception: unless there's a specific reason for it.
> 
Right, this needs to be BaseException (naked `except` clauses are deprectated).

Ok, let's give this PR another whirl.

kristjanvalur:
> You now disconnect on the outside for any exception. Are you sure this is correct? This will at least disconnect twice sometimes. And are there really no exceptions that should go through without disconnect?
Disconnecting twice is safe.
I think we have established that there should be no exceptions going through without a disconnect.  That was previously the case with _all_ exceptions while reading a response.  Now BaseExceptions don't disconnect, in the `Connection` but they do cause a disconnect in the scope of a `command/response` pair.
> 
> Since you are against rolling back #2104, you should denote the API change in the functions affected, because they now have a rather surprising API. `send_command` needs a comment explaining that if it raises, the connection is now in an undefined state and the caller needs to disconnect it. Same for `read_response` and `parse_response`.

This is incorrect.  It it is only in the case of BaseExceptions that it is kept open, exceptions that are not handled or caught anywhere and typically result in an exit.  (In async code, the TimeoutError _does_ result in a `disconnect()`.
 However, since there is no way to restart a partially completed send, I see no harm in closing it.  Particularly since it may help the code jive with gevent.  I'll make a PR to fix the `send_response.`
 
> If you succeed with #2510, that comment could change on the async response functions to explaining that if the Exception was a CancellationException, the connection is not undefined, but still holds the previous response. If it is another Exception like KeyboardInterrupt, the connection will still be broken.

Fair point.  Yes, a timeout on read_response really should mean that the application can retry, and that should be documented, I guess, if the opposite is already undocumented.   Otherwise, in all APIs that I am familiar with, a `timeout` on an operation which provides a "timeout" parameter, means that the operation can be retried.

Chronial:
>>You now disconnect on the outside for any exception. Are you sure this is correct? This will at least disconnect twice sometimes. And are there really no exceptions that should go through without disconnect?
Disconnecting twice is safe.
>
>I think we have established that there should be no exceptions going through without a disconnect. That was previously the case with all exceptions while reading a response. Now BaseExceptions don't disconnect, in the Connection but they do cause a 
disconnect in the scope of a command/response pair.

Yes: "That was previously the case with all exceptions **while reading a response**". You extended the scope of the try-catch. Now, for example, any exception  [raised by a response callback](https://github.com/redis/redis-py/blob/74c251a19d211b3dcf555adc4a68bdc6f097f763/redis/client.py#L1284) will also disconnect the connection. That was not previously the case.

kristjanvalur:
> Yes: "That was previously the case with all exceptions **while reading a response**". You extended the scope of the try-catch. Now, for example, any exception [raised by a response callback](https://github.com/redis/redis-py/blob/74c251a19d211b3dcf555adc4a68bdc6f097f763/redis/client.py#L1284) will also disconnect the connection. That was not previously the case.

Okay, I think it is simplest that I restore the original behaviour of connection.parse_response(), since there simply seem to be too many cases that rely on it.  Even though none of them was caught by the test suite.  I'll modify this PR so that the BaseException behaviour becomes optional.

redis/asyncio/connection.py
```diff
@@ -763,7 +763,11 @@ async def send_packed_command(
             raise ConnectionError(
                 f"Error {err_no} while writing to socket. {errmsg}."
             ) from e
-        except Exception:
+        except BaseException:
+            # The send_packed_command api does not support re-trying a partially
+            # sent message, so there is no point in keeping the connection open.
```
ikonst, line 768: Consider: The lack of support is really RESP's (the wire protocol).

redis/asyncio/connection.py
```diff
@@ -763,7 +763,11 @@ async def send_packed_command(
...
+            # An unknown number of bytes has been sent and the connection is therefore
```
ikonst, line 769: Even if you knew how many bytes were sent (I think you can know), the problem is you can't "take them back".

redis/asyncio/connection.py
```diff
@@ -787,6 +791,7 @@ async def read_response(
         self,
         disable_decoding: bool = False,
         timeout: Optional[float] = None,
+        disconnect_on_error: bool = True,
```
ikonst, line 794: This function starts having a lot of position arguments that are of primitive types. Consider making additional ones kw-only (we can't change existing ones without making a breaking change):
```suggestion
        *,
        disconnect_on_error: bool = True,
```

redis/asyncio/connection.py
kristjanvalur: Actually no.  You can keep the buffer of data (the message) around in a send buffer.  the actual socket.send() message lets you know how much of this has been successfully delivered to underlying layers.  So a re-triable API could, for example, accept a `None` as the data, and then continue sending from where it left off, if it has a buffer in place.  Consider this api:
```python
def send_packed_command(message: Optional[str]):
   """ Send the packed command, or, if None, a previous call must not have completed and the function will
        will resume sending that data over the wire.
  """
  ...
```

redis/asyncio/connection.py
kristjanvalur: It's not about "taking them back", its about "trying again to send the rest".  This could (theoretically) be done in case of a timeout, and is the way that many `socket.sendall()` implementations work.  (the underlying `socket.send()` sends a potential subset of the data and needs to be retried.)

send_packed_api() could be architected in a similar way, but maintain an internal buffer of unsent data so that it can be re-tried.

redis/asyncio/connection.py
kristjanvalur: Good suggestion.  I'll add that to the `timeout` too, since that's one that I added and it is used only as a kwarg.

redis/asyncio/connection.py
ikonst: Yes, on the Python side (e.g. in `Connection`) we could choose to implement a write buffer and "continue where we left off". As [I mentioned](https://github.com/redis/redis-py/issues/2499#issuecomment-1346737191) before, the problem with a write buffer would be the potential of flushing out a command that's stale from an app code's perspective.

What I'd rather see, is being able to, instead of flushing a command that can be stale, to send a short control sequence that would cancel a command that's half-written into the socket, turning it into a no-op. This is the feature that RESP protocol doesn't support.

So:
1. Write buffer: Not an inherent limitation in Python, we simply don't want to do it because it's bad UX.
2. Cancel a half-written command: Inherent limitation of RESP.

Hence the "problem" is RESP, not `send_packed_command`.

redis/asyncio/connection.py
ikonst: But of course it's not reasonable to flush a command after an arbitrary delay (could be milliseconds, could be days). The "workaround" supported by RESP is not reasonable for what Redis is.

kristjanvalur: I'm not sure I understand what you mean.   All I'm saying is that sometimes your `send()` can time out, for various reasons, buffers, server responsiveness, or whatever.   It's technically possible to engineer things such that you can at least be sure that you've delivered all of your payload down to the OS to deliver over the wire.    What happens after that is anyone's guess.  Typically, then, you'd listen for the response, and time out in the listening phase if nothing happens.

This is in no way dissimilar to http, and doesn't have anything to do with RESP as such.  

Anyway, I thought I'd comment on that, a _different_ API, allowing one to retry a partially completed `send()` operation, could mean that it would make sense to keep the connection open.  But the current API has no semantics that woudl allow a "retry".

redis/asyncio/connection.py
kristjanvalur: I see what you are saying, but I'm not even thinking about the "cancel" of the command.  That's something I've never seen and in my mind completely unnecessary.  For that, you can simply disconnect.   I'm primarily thinking about Timeouts, where the transmitting of the command over the wire times out, and you want to maybe do some housekeeping and then continue trying to send the rest of the command.   

In the context of this PR, do you suggest that I change this comment somehow?  Is it confusing?  I always find that code comments help future contributors.  A similar comment would have probably caused me to do my original PR differently :)

redis/asyncio/connection.py
kristjanvalur: also, see the above discussion :)

redis/asyncio/connection.py
ikonst: Yes, I'm discussing how we phrase the comment :) I think I understand now.

**My perspective:**
I called `send` and caught an exception. Next time I call `send`, it might be a different command, at a different time, on a different thread. I wouldn't want `send_packed_command` to flush my previous command at an arbitrary time in the future. I would settle for flushing a short "cancel sequence" instead, but RESP doesn't support that, so instead I'd like it to disconnect.

From that perspective, it's a RESP limitation.

**Your perspective:**
I called `send` and caught an exception. Now I want to retry immediately, but `send_packed_command` didn't tell me _where_ it stopped nor allows me to resume half-way, so I can't and it's a limitation of `send_packed_command`.

From that perspective, it's a `send_packed_command` limitation.

kristjanvalur: Yes, I understand.  Let me try to formulate the comment better.

redis/asyncio/connection.py
```diff
@@ -764,10 +764,10 @@ async def send_packed_command(
...
             ) from e
         except BaseException:
-            # The send_packed_command api does not support re-trying a partially
-            # sent message, so there is no point in keeping the connection open.
-            # An unknown number of bytes has been sent and the connection is therefore
-            # unusable.
+            # BaseExceptions can be raised when a send*() is not finished, e.g. due
+            # to a timeout.  Ideally, a caller could then re-try, to continue
+            # sending partially-sent data. However, the send_packed_command() API
+            # does not it so there is no point in keeping the connection open.
```
ikonst, line 770: 💯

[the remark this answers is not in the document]
kristjanvalur: yes, sorry, was a bit fast, should be fixed now;.

redis/asyncio/connection.py
```diff
@@ -764,10 +764,10 @@ async def send_packed_command(
                 f"Error {err_no} while writing to socket. {errmsg}."
             ) from e
         except BaseException:
-            # BaseExceptions can be raised when a send*() is not finished, e.g. due
+            # BaseExceptions can be raised when a send() is not finished, e.g. due
```
ikonst, line 767: technically `sendall`, no? :)

kristjanvalur: Yes, but I'm referring to the sort of classic POSIX socket API.    I guess I could just say "socket send" to keep it ambiguous enough and avoid confusion.

kristjanvalur: there, I added an update.  maybe it is better?

dvora-h:
@kristjanvalur Thanks for this PR! Can you please resolve the conflicts here?

kristjanvalur:
There are errors due to a recent merge, I'll fix that...

chayim:
@kristjanvalur I see the green passes. Ready for a review?

kristjanvalur:
Yes, this has been ready for a long time.  The non-green are the spurious failures that the test suite has.

dvora-h:
@chayim Can you review this?

kristjanvalur:
I'm investigating the test failure.

kristjanvalur:
This is new behaviour with the 3.11 timeout context manager.  It doesn't work if there has been a previous cancel on the task.
https://github.com/python/cpython/issues/102780

kristjanvalur:
Oh dear.  Now it is hanging because it is no longer possible to cancel an operation, due to this code:
```python
        return await asyncio.shield(
            self._try_send_command_parse_response(conn, *args, **options)
        )
```
in client.py.
The send command and parse is shielded, so cancelling the command won't cancel it.  This leaves the `__single_conn_lock` held.
I wonder why this was added?  It breaks things.

kristjanvalur:
This happened in pr #2666

ikonst:
This issue will your white whale. I'm sorry :) but thanks for the efforts.

kristjanvalur:
Closed this now that #2695 has been merged (a re-creation of this PR)

PR #2510, opened 2022-12-13
Make PythonParser resumable

kristjanvalur:
### Pull Request check-list

### Description of change

See issue #2513

Recent changes in redis.asyncio changed the way timeouts are handled, by catching timeouts on a higher level and not leave a `Connection` object in a failed state.  Timeouts are now received directly when doing IO, instead of running a preliminary `can_read()` test with timeout.  This resulted in significant simplification of the logic and consolidated different timeout handling.

Unfortunately, the `PythonParser`, unlike the `HiredisParser` was not able to deal with IO being interrupted as it didn't maintain internal state.  A TimeoutException caught while parsing a response would cause data to be lost.  The `HiredisParser`, however, cleanly separates parsing and IO, and maintains its own internal parse buffer which is fed by IO separately.

This PR fixes `PythonParser` to keep a buffer of incomplete responses, so that if IO is interrupted, the parser's `read_response()` method can be called again, if it is interrupted during IO.

See also PR #2512 which suggests a slightly different (and probably simpler and better) approach, more in line with how `HiredisParser` works.

See issue #2499 where this problem was pointed out.

### DIFF
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -162,6 +162,7 @@ class InterruptingReader:
         return result
 
 
+@pytest.mark.onlynoncluster
 async def test_connection_parse_response_resume(r: redis.Redis):
     """
     This test verifies that the Connection parser,
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -7,6 +7,7 @@ import pytest
 
 import redis
 from redis.asyncio.connection import (
+    BaseParser,
     Connection,
     PythonParser,
     UnixDomainSocketConnection,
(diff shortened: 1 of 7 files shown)


kristjanvalur:
A separate approach is also possible, similar to HiRedis, where a single `fill()` method places network data into the buffer and the `_read()` and `_readline()` methods work only with previously read data.

tests/test_asyncio/test_connection.py
```diff
@@ -112,3 +117,79 @@ async def test_connect_timeout_error_without_retry():
         await conn.connect()
     assert conn._connect.call_count == 1
     assert str(e.value) == "Timeout connecting to server"
+
+
+class FakeStream:
```
chayim, line 122: maybe Mock instead of Fake? But.. yes!

chayim:
@kristjanvalur I like this approach better, it definitely aligns more with hiredis. @dvora-h can you possibly test both and profile them? I can't tell from reading if one seems more efficient than the other.

kristjanvalur:
> @kristjanvalur I like this approach better, it definitely aligns more with hiredis. @dvora-h can you possibly test both and profile them? I can't tell from reading if one seems more efficient than the other.

This one is slightly more involved.
The other is more like HiRedis, and maybe easier to reason about.  The other has the drawback that when the buffer needs refilling, the parsing needs to be restarted.  This happens a lot for large payloads.

I'm the author of an experimental library, asynkit. https://github.com/kristjanvalur/py-asynkit  I've been thinking about adding a feature which would allow _suspendable_ and _resumable_ functions like this, using the `async` and `await` keywords, but without actual IO, to be able to maintain state of a callstack that is paused somehow.  There is no code yet, just an Idea in my head, but that would make for a very clean approach when implemented.

tests/test_asyncio/test_connection.py
kristjanvalur: Sounds fine, I can move these to a more generic place.  A MockSocket and MockStream

chayim:
@dvora-h I'm voting on this one..... Let's add to 4.4.x

chayim (approved):

PR #2512, opened 2022-12-14
Make PythonParser resumable (alternative)

kristjanvalur:
### Pull Request check-list

### Description of change

See issue #2513

Recent changes in redis.asyncio changed the way timeouts are handled, by catching timeouts on a higher level and not leave a `Connection` object in a failed state.  Timeouts are now received directly when doing IO, instead of running a preliminary `can_read()` test with timeout.  This resulted in significant simplification of the logic and consolidated different timeout handling.

Unfortunately, the `PythonParser`, unlike the `HiredisParser` was not able to deal with IO being interrupted as it didn't maintain internal state.  A TimeoutException caught while parsing a response would cause data to be lost.  The `HiredisParser`, however, cleanly separates parsing and IO, and maintains its own internal parse buffer which is fed by IO separately.

This PR fixes `PythonParser` to keep a buffer of incomplete responses, so that if IO is interrupted, the parser's `read_response()` method can be called again, if it is interrupted during IO.

**Note:** This is an alternative change to pr #2510

See issue #2499 where this problem was pointed out.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,4 @@
+    * Make PythonParser resumable in case of error (#2512)
     * Add `timeout=None` in `SentinelConnectionManager.read_response`
     * Documentation fix: password protected socket connection (#2374)
     * Allow `timeout=None` in `PubSub.get_message()` to wait forever
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -208,18 +208,13 @@ class BaseParser:
 class PythonParser(BaseParser):
     """Plain Python parsing class"""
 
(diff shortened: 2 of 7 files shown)


dvora-h:
@kristjanvalur Do you have any idea why the tests are canceling? It only happens in these 2 PR's...

kristjanvalur:
> @kristjanvalur Do you have any idea why the tests are canceling? It only happens in these 2 PR's...

No.  I'll investigate.

kristjanvalur:
current failure is with sync hiredis parser in 3.11 cluster mode, code which is not changed by this.

kristjanvalur:
Closed, since #2510 was merged (an alternative implementation)

PR #2516, opened 2022-12-15
Close a Connection whenever an exception is raised for send_command()

kristjanvalur:
The api does not allow for a "resume", e.g. on a timeout, because an unknown number of bytes has been sent and an internal send state is not maintained.  Therefore, there is no point in keeping the connection open.

### Pull Request check-list

### Description of change

The discussion at #2560 pointed out that there is no way an open Connecition after a failed `send_command()` call can be useful.  This PR therefore closes the `Connection` on _every_ exception in that case, thus partially reverts a change made in #2104


kristjanvalur:
Closing this in favor of #2506

PR #2518, opened 2022-12-18
BlockingConnectionPool simplification: use counter instead of connection list

Fogapod:
### Pull Request check-list

I do not have docker on my machine and i couldn't find a way to make it work with podman. I hope repo CI can run test and catch bugs.
There are a lot of linting errors/warnings about docstring formatting in cluster.py file which I haven't touched.
Action doesn't start for my branch. I assume it only works for PRs

### Description of change

Existing code juggling full queue of `None`s and using extra list to store connections looked weird so I simplified it to use a simple counter to track how many connections were allocated.
Also added slots to pools.

I tested my change with the following code using different `max_connections` values:
```py
import asyncio

from redis import asyncio as redis


async def main() -> None:
    pool = redis.BlockingConnectionPool(max_connections=50)
    async with redis.Redis(connection_pool=pool) as rd:
        await asyncio.gather(*[rd.set(str(i), i) for i in range(200_000)])


asyncio.run(main())
```
It also performs a few % better than existing solution, probably because of slots.

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1237,6 +1237,19 @@ class ConnectionPool:
     ``connection_class``.
     """
 
+    __slots__ = (
+        "connection_class",
+        "connection_kwargs",
+        "max_connections",
+        "_fork_lock",
+        "_lock",
+        "_created_connections",
+        "_available_connections",
+        "_in_use_connections",
+        "encoder_class",
+        "pid",
+    )
+
     @classmethod
     def from_url(cls: Type[_CP], url: str, **kwargs) -> _CP:
         """
@@ -1519,18 +1532,23 @@ class BlockingConnectionPool(ConnectionPool):
         >>> pool = BlockingConnectionPool(timeout=5)
     """
 
+    __slots__ = (
+        "queue_class",
+        "timeout",
+        "pool",
+    )
+
     def __init__(
         self,
         max_connections: int = 50,
         timeout: Optional[int] = 20,
         connection_class: Type[Connection] = Connection,
-        queue_class: Type[asyncio.Queue] = asyncio.LifoQueue,
+        queue_class: Type[asyncio.Queue] = asyncio.Queue,
         **connection_kwargs,
     ):
 
         self.queue_class = queue_class
         self.timeout = timeout
-        self._connections: List[Connection]
         super().__init__(
             connection_class=connection_class,
             max_connections=max_connections,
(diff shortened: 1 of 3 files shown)


redis/asyncio/connection.py
```diff
@@ -1646,7 +1652,10 @@ async def disconnect(self, inuse_connections: bool = True):
         self._checkpid()
         async with self._lock:
             resp = await asyncio.gather(
-                *(connection.disconnect() for connection in self._connections),
```
kristjanvalur, line 1649: This disconnects only those connections _in_ the pool, not _all_ connections as before.

redis/asyncio/connection.py
```diff
@@ -1519,36 +1532,34 @@ class BlockingConnectionPool(ConnectionPool):
...
+        "timeout",
+        "pool",
+    )
+
     def __init__(
         self,
         max_connections: int = 50,
         timeout: Optional[int] = 20,
         connection_class: Type[Connection] = Connection,
-        queue_class: Type[asyncio.Queue] = asyncio.LifoQueue,
```
kristjanvalur, line 1527: The LifoQueue ensures that _hot_ connections are the ones re-used first, which is probably beneficial.

Fogapod: There was no difference in my simple tests, maybe i did something wrong but Lifo had 0 impact

kristjanvalur: Well, I´m not sure there will be a difference, or that this was the original intent of the LifoQueue.  But LIFO is often used in caching, to maintain _hot_ information.  There is a degenerate case which can happen with Fifo, which is that if you pull always the oldest connections, you may end up with all the connections having timed out, or been disconnected due to idleness or other such things.   But maybe that doesn't matter at all.   Maybe the only intention of the LifoQueue was to keep those `None` objects at the end.

Fogapod:
I added `_in_use_connections` set similar to non-blocking pool to not leak in-use connections. I am unsure about what it means for pool to "own" a connection so I assumed that it means different pool created it. This is why I call `self._in_use_connections.remove(connection)` only if pool owns connection

Fogapod:
Test is failing because of `__slots__`. Unsure how to fix mock yet: `AttributeError: 'ConnectionPool' object attribute 'get_connection' is read-only`

kristjanvalur:
Btw, this is probably obsoleted with the merging of #2911

Fogapod:
It is

Issue #2349, opened 2022-08-21
AttributeError: 'NoneType' object has no attribute 'feed' (possibly race condition on async disconnect)


bentheiii:
**Version**: 4.3.4
**Platform**: python 3.8.13 on azure

**Description**: unfortunately I have not been able to replicate this error locally, I'm guessing it's a bizarre race condition (possibly an async resurgence of #732)

```
AttributeError: 'NoneType' object has no attribute 'feed'
  File "***", line **
    await pipeline.execute()
  File "redis/asyncio/client.py", line 1352, in execute
    return await conn.retry.call_with_retry(
  File "redis/asyncio/retry.py", line 59, in call_with_retry
    return await do()
  File "redis/asyncio/client.py", line 1266, in _execute_pipeline
    await self.parse_response(connection, args[0], **options)
  File "redis/asyncio/client.py", line 1291, in parse_response
    result = await super().parse_response(connection, command_name, **options)
  File "redis/asyncio/client.py", line 505, in parse_response
    response = await connection.read_response()
  File "redis/asyncio/connection.py", line 931, in read_response
    response = await self._parser.read_response(
  File "redis/asyncio/connection.py", line 526, in read_response
    await self.read_from_socket()
  File "redis/asyncio/connection.py", line 491, in read_from_socket
    self._reader.feed(buffer)
```

emtrane:
I'm able to reproduce the same very consistently with a script like this.
```
import asyncio
import wrapt
from redis.asyncio.cluster import RedisCluster

async def main():
  host = "localhost"
  port = 6379
  MAX_HASH_SLOTS = 16384

  def _execute_command_wrapper():
      async def _wrapped(f, instance, args, kwargs):
          if len(args) == 1:
              if args[0] == "INFO":
                  return {"cluster_enabled": True}
              elif args[0] == "CLUSTER SLOTS":
                  return [[0, MAX_HASH_SLOTS - 1, [host, port]]]

          return await f(*args, **kwargs)

      return _wrapped

  wrapt.wrap_function_wrapper(
      "redis.asyncio.cluster", "ClusterNode.execute_command", _execute_command_wrapper()
  )

  rc = RedisCluster(
    host=host, 
    port=port
  )

  async def t():
    for i in range(1000):
        async with rc.pipeline() as p:
            p.sadd('foo', 'bar')
            p.expire('foo', 3)
            p.incr('bar')
            p.expire('bar', 3)
            results = await p.execute()
            print(results)

[...]

PR #2557, opened 2023-01-20
Fix issue 2349: Let async HiredisParser finish parsing after a Connection.disconnect()

kristjanvalur:
### Pull Request check-list

### Description of change

This change addresses an issue raised in issue #2349.

It ensures that an async `Connection.read_response()` call can succeed even if a simultaneous `Connection.disconnect()`
call is issued.  Socket disconnects are _graceful_ in that one closes the send direction and can then wait for any data
in transit to arrive, until an EOF is delivered from the remote side.  In `asyncio` parlance this means that to gracefully close a socket, you close the `send` _stream_ and can then read from the `receive` stream until you reach an End Of File.

This change makes sure that the internal `BaseParser` state is maintained so that it can continue to request and parse chunks for as long as it can.   Previously, the HiredisParser would get its internal state cleared immediately, and this would cause the read, in progress, to fail with an unexpected error.  Now an ongoing `read_response()` will continue until it succeeds or encounters an error such as EOF.

A new `read_response()` issued **after** a `disconnect()` call has been issued (but before it completes) will immediately fail, howerver.  This is to maintain previous semantics, even though we **could** simply allow it to go through and fail later when it attemts to read from a closed connection.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,4 @@
+    * Add test and fix async HiredisParser when reading during a disconnect() (#2349)
     * Simplify synchronous SocketBuffer state management
     * Fix string cleanse in Redis Graph
     * Make PythonParser resumable in case of error (#2510)
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -350,13 +350,14 @@ class PythonParser(BaseParser):
 class HiredisParser(BaseParser):
     """Parser class for connections using Hiredis"""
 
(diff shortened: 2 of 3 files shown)


kristjanvalur:
random test failure is random.

dvora-h (approved):

dvora-h:
@kristjanvalur Thanks for this PR! Can you please resolve the conflicts so I can merge it?

dvora-h (approved):

Issue #2540, opened 2023-01-04
RuntimeError: readuntil() called while another coroutine is already waiting for incoming data


Tiendil:
**Version**: `redis==4.4.0`, Redis: `redis:6.2-alpine`  (docker)

**Platform**: Python 3.9.16, docker `python:3.9`

**Description**: Redis client raises RuntimeError while doing multiple concurrency operations.

Reproduced on `4.4.0`, ok on `4.3.5`

Example code:

```
import asyncio
import uuid

from redis.asyncio import Redis


async def main():

    client = Redis(
        host="redis",
        port=6379,
        db=0,
        username=None,
        password=None,
        retry_on_timeout=True,
        single_connection_client=True,
        client_name='test_client',
    )

    # At my PC stable reproduces with n=10, but let set n=100 for better guarantees
    n = 100

    async def problem():
        await client.delete(uuid.uuid4().hex)
        await client.delete(uuid.uuid4().hex)  # <- problem is at this line

    operations = [problem() for i in range(n)]

    await asyncio.gather(*operations)

    print('success!')
```

Stacktrace:

```
Traceback (most recent call last):
  File "/app/x.py", line 36, in <module>
    asyncio.get_event_loop().run_until_complete(main())
  File "/usr/local/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/app/x.py", line 33, in main
    await asyncio.gather(*operations)
  File "/app/x.py", line 29, in problem
    await client.delete(uuid.uuid4().hex)  # <- problem is at this line
  File "/tmp/venv/lib/python3.9/site-packages/redis/asyncio/client.py", line 505, in execute_command
    return await conn.retry.call_with_retry(
  File "/tmp/venv/lib/python3.9/site-packages/redis/asyncio/retry.py", line 59, in call_with_retry
    return await do()
  File "/tmp/venv/lib/python3.9/site-packages/redis/asyncio/client.py", line 481, in _send_command_parse_response
    return await self.parse_response(conn, command_name, **options)
  File "/tmp/venv/lib/python3.9/site-packages/redis/asyncio/client.py", line 524, in parse_response
    response = await connection.read_response()
  File "/tmp/venv/lib/python3.9/site-packages/redis/asyncio/connection.py", line 800, in read_response
    response = await self._parser.read_response(
  File "/tmp/venv/lib/python3.9/site-packages/redis/asyncio/connection.py", line 245, in read_response
    raw = await self._readline()
[...]

PR #2568, opened 2023-01-28
Fix issue 2540: Synchronise concurrent command calls to single-client to single-client mode

Vivanov98:
### Pull Request check-list

### Description of change

Fixing issue [#2540](https://github.com/redis/redis-py/issues/2540) (example script in issue tested and bug no longer occurs)

The underlying cause is that the connection in single-client mode isn't protected by a lock, resulting in multiple coroutines reading/writing to a connection stream when multiple commands are called in parallel (forexample with asyncio.gather)

Additionally, redundant connections can be made if the first time the client is used, it is used in parallel (forexample with asyncio.gather).

A simple lock mechanism has been added to the `execute_command` function to ensure that single-client mode is always executing commands in sequence. This lock is also used each time the `initialize` method is called, ensuring that multiple connections aren't spun up if the first time the client is called happens to be with a parallel mechanism.

Worth mentioning that the use cases for using parallel calling mechanisms in single-client mode are probably limited, though this still seems useful to have in terms of consistency.

### DIFF
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -6,6 +6,7 @@ from unittest.mock import patch
 import pytest
 
 import redis
+from redis.asyncio import Redis
 from redis.asyncio.connection import (
     BaseParser,
     Connection,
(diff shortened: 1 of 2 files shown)


tests/test_asyncio/test_connection.py
```diff
@@ -41,6 +42,50 @@ async def test_invalid_response(create_redis):
...
+
+    async def get_conn(_):
+        # Validate only one client is created in single-client mode when
+        # concurrent requests are made
+        nonlocal init_call_count
+        await asyncio.sleep(0.01)
+        init_call_count += 1
+        return mock_conn
+
+    with mock.patch.object(r.connection_pool, "get_connection", get_conn):
```
Vivanov98, line 81: The only deterministic test to validate the locking does what we want that I could think of required patching the connection and call_with_retry objects. This test is consistent, and would fail without the client changes.

dvora-h (approved):

PR #2570, opened 2023-02-03
Use hiredis::pack_command to serialized the commands.

prokazov:
@chayim  @dvora-h
### Pull Request check-list

- [x ] Does `$ tox` pass with this change (including linting)?
- [ x] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ x] Is the new or changed code fully tested?
- [ ?] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [n/a ] Is there an example added to the examples folder (if applicable)?
- [ x] Was the change added to CHANGES file?

    Historically, hiredis-py (Python extension module written in C and utilizing hiredis library) has been used only for deserialization of the Redis command's replies. With the new hiredis-py 2.2.1 release it's possible to do the opposite operation - serialization of the Python tuples, representing commands in the redis-py into a RESP-encoded bytes buffer. This PR:
1) Extracts the pack command functionality form the Connection class and delegates it to an external object.
2) PythonRespSerializer class that implements the extracted functionality.
3) HiredisRespSerializer class that uses hiredis-py to serialize a redis-py command.


redis/connection.py
```diff
@@ -607,6 +684,14 @@ def __del__(self):
         except Exception:
             pass
 
+    def _construct_command_packer(self, packer):
+        if packer is not None:
+            return packer
+        elif HIREDIS_PACK_AVAILABLE:
+            return HiredisRespSerializer()
+        else:
+            return PythonRespSerializer(self._buffer_cutoff, self.encoder.encode)
```
spicy-sauce, line 693: seems that elif and else are not needed:
```
if packer is not None:
    return packer

if HIREDIS_PACK_AVAILABLE:
    return HiredisRespSerializer()

return PythonRespSerializer(self._buffer_cutoff, self.encoder.encode)
```

spicy-sauce (changes requested):

chayim:
@prokazov given that hiredis isn't a dependency, but is an extra (i.e. ```pip install redis[hiredis]``` we can't enforce a hiredis version on existing users.  This means that the hiredis check now becomes more complicated.

We need to wrap the pack_command call, with a check to ensure that hiredis is >= 2.2.1, specifically due to this change, otherwise we'll break the community.

Similarly @dvora-h this becomes 4.5.0 as discussed.

zalmane:
@chayim Originally the check was checking for a version. It was later changed to check for the existence of the pack_command exposed in the hiredis. So effectively this validates if the hiredis-py contains this new method so it is backwards compatible.

redis/connection.py
prokazov: Technically - yes, we can get rid of `else` but , my understanding of general sentiment regarding `elif ` in Python is: it's cleaner way to show that conditions intended to be mutually exclusive.

dvora-h (approved):
LGTM!

PR #2575, opened 2023-02-06
if an index error occurs while fetching a node, return master node

zakaf:
### Pull Request check-list

### Description of change

When an AWS Elasticache shard is added to an existing cluster, below exception occurs in redis-py.
```
  ...
  File "/usr/lib/python3.8-pyston2.3/site-packages/redis/cluster.py", line 876, in _determine_nodes
    node = self.nodes_manager.get_node_from_slot(
  File "/usr/lib/python3.8-pyston2.3/site-packages/redis/cluster.py", line 1382, in get_node_from_slot
    return self.slots_cache[slot][node_idx]
IndexError: list index out of range
```

The above exception seems to occur, when slots_cache changes dynamically, while get_node_from_slot is being executed.
In order to avoid the above exception, return the primary node, when an IndexError occurs

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -1379,7 +1379,10 @@ class NodesManager:
             # randomly choose one of the replicas
             node_idx = random.randint(1, len(self.slots_cache[slot]) - 1)
 
-        return self.slots_cache[slot][node_idx]
+        try:
+            return self.slots_cache[slot][node_idx]
+        except IndexError:
+            return self.slots_cache[slot][0]
 
     def get_nodes_by_server_type(self, server_type):
         """


zach-iee:
@dvora-h @chayim if you guys have spare time, could you take a look at this PR?

zakaf:
@dvora-h @chayim if you guys have spare time, could you take a look at this PR?

gordo-veda:
This is still a problem... the code around here should be hardened such that we throw a proper error to handle resharding, or in this case, trying another node seems harmless (since the node will point you elsewhere if it doesn't have your lot).

Is it safe to assume the zeroth node is the primary? (maybe not) but finding any working node that can point you to the right slot is i likely good enough.

PR #2596, opened 2023-02-23
Speeding up the protocol parsing

chayim:
This PR makes small changes to the protocol parsers. Specifically, this reduces the number of branches, and composite branches (if then if) that happen along the way. 

As part of this change, fewer cast operations now happen, and the response comparison is (when necessary) only on the first byte of the response. This in turn makes the comparison cost lower (memory), and faster, due to the lack of repeated casting.

This is true of both async and sync.

There are no user facing changes.

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -267,9 +267,6 @@ class PythonParser(BaseParser):
         response: Any
         byte, response = raw[:1], raw[1:]
 
-        if byte not in (b"-", b"+", b":", b"$", b"*"):
-            raise InvalidResponse(f"Protocol Error: {raw!r}")
-
         # server returned an error
         if byte == b"-":
             response = response.decode("utf-8", errors="replace")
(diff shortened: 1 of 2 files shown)


redis/asyncio/connection.py
```diff
@@ -291,19 +288,21 @@ async def _read_response(
...
-                return None
+        elif byte == b"*" and response == b"-1":
+            return None
+        elif byte == b"*" and response != b"-1":
             response = [
-                (await self._read_response(disable_decoding)) for _ in range(length)
+                (await self._read_response(disable_decoding))
+                for _ in range(int(response))  # noqa
             ]
+        elif byte not in (b"-", b"+", b":", b"$", b"*"):
```
Ernest0x, line 303: Maybe this line could be just `elif byte != b"+":`. All other comparisons are known to be true here because none of the previous ifs/elifs matched.

redis/connection.py
```diff
@@ -381,20 +378,21 @@ def _read_response(self, disable_decoding=False):
...
-                return None
+        elif byte == b"*" and response == b"-1":
+            return None
+        elif byte == b"*" and response != b"-1":
             response = [
                 self._read_response(disable_decoding=disable_decoding)
-                for i in range(length)
+                for i in range(int(response))
             ]
+        elif byte not in (b"-", b"+", b":", b"$", b"*"):
```
Ernest0x, line 393: Maybe this line could be just `elif byte != b"+":`. All other comparisons are known to be true here because none of the previous ifs/elifs matched.

chayim: @Ernest0x please take another look.. I've actually made it simpler.. and everything passes currently. But - I would love more eyes.

chayim:
@dvora-h any idea why this hiredis test is failing? Have you seen this before? It shouldn't in any way be connected to these changes.

redis/connection.py
Ernest0x: @chayim looks good. A minor thing could be that `int(response)` which is done in several places. Perhaps a variable could be set for that.

chayim: The problem is that doing that adds overhead and slows the parsing down (a bit, sure).  This is one of those cases where the speed improvement dictates something that is less "PEP 1"... But then again, it follows the [zen](https://peps.python.org/pep-0020/)

dvora-h (approved):

Issue #2593, opened 2023-02-19
Asyncio TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'


DanielK990:
**Version**: 

redis 5.0.7
pip list:
channels           4.0.0
channels-redis     4.0.0
redis              4.5.1

**Platform**: Windows 10 WSL2, Python 3.8.10, redis server runs on a Raspberry Pi

**Description**: 

I am using Django channels with Redis.

I am using a channel to send a message from 

View ---> AsyncConsumer (background worker)

def myview(request):
    async_to_sync(channel_layer.send)(        # <--- here the error occurs
                "start",
                {
                    "type": "start",
                    },
                )

Another message is send from this AsyncConsumer --> AsyncWebsocketConsumer

class MyAsyncConsumer(AsyncConsumer):
    async def start(self, event):
        logger = logging.getLogger('django')
        logger.info(event)
        await self.channel_layer.send(
                event['channel'],
                {
                    "type": "finish",
                },
            )

Finally in the WebsocketConsumer, sends a Websocket message to the Websocket Client

class MyWebsocketConsumer(AsyncWebsocketConsumer):

    async def finish(self, data):
        logger = logging.getLogger('django')
        logger.info("Reply from Worker")
        message = {
            "type" : "finish",
        }
        await self.send(json.JSONEncoder().encode(message))

When sending the message from View ---> AsyncConsumer, there is the following error:

Internal Server Error: /
Traceback (most recent call last):
  File "/mnt/d/Documents/python/myapp/myapp/.venv/lib/python3.8/site-packages/asgiref/sync.py", line 486, in thread_handler
    raise exc_info[1]
  File "/mnt/d/Documents/python/myapp/myapp/.venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 42, in inner
    response = await get_response(request)
  File "/mnt/d/Documents/python/myapp/myapp/.venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 253, in _get_response_async
    response = await wrapped_callback(
  File "/mnt/d/Documents/python/myapp/myapp/.venv/lib/python3.8/site-packages/asgiref/sync.py", line 448, in __call__
    ret = await asyncio.wait_for(future, timeout=None)
  File "/usr/lib/python3.8/asyncio/tasks.py", line 455, in wait_for
    return await fut
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
[...]

PR #2607, opened 2023-03-09
Fix handling of `CancelledError` by async client

jajce:
### Pull Request check-list

### Description of change

This PR is supposed to fix the issue #2593 . 
The problem occurs when invoking `cancel()` on a task which was executing any redis command. Invoking `cancel()` resulted in interruption of StreamReader `readline()` method by the CancelledError. When it happened it resulted in not reading the response from StreamReader after command execution. It caused all future responses to be mixed up between commands.

For example if we canceled a `ping()` command and executed `exists('foo')` command afterwards StreamReader reads previously not read `b"+PONG\r\n"`  instead of `b":0\r\n"` which results in error when trying to parse raw data and cast it to the correct type. (This example is implemented as a working test)

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -517,6 +517,12 @@ class Redis(
                 ),
                 lambda error: self._disconnect_raise(conn, error),
             )
+        except asyncio.CancelledError:
+            # If we're cancelled and cancel happened after we sent the command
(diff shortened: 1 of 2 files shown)


dvora-h:
#2593 already fixed in #2695

PR #2611, opened 2023-03-13
update json().arrindex() default values

### LINKED ISSUE #2481: arrindex returns -1 when finding the last value in the array (and when start and stop are not specified)
Version: redis-py 4.1.4, redis 6.2.7

Platform: Python 3.10.5 on MacOS 12.6.1

Description: When using arrindex, the behavior for finding the last element of the array, and finding no element in the array is the same when default start and stop are unspecified.

For example, I have an array ['test', 'test2' and 'test3']

When executing

```
redis_client.json().arrindex(f'type:key','arr_name','test3')

redis_client.json().arrindex(f'type:key','arr_name','test4')

```
The response is the same [-1].

It works fine when I search

```
redis_client.json().arrindex(f'type:key','arr_name','test3',start=0, stop=1000000)
```

davemcphee:
### Pull Request check-list

### Description of change

Refactor default `start` / `stop` values in the `json().arrindex()` function. Previously, if `stop` was _unset_, the stop value would default to -1. If the value being searched was the last element in an array, and `stop` was set to -1, the server would reply with -1 instead of the element's index.

eg.: given a json object myarr: `{"l": ["a", "b", "c", "d"]}`:

```
localhost:36379> JSON.ARRINDEX myarr $.l '"d"'            # no start or stop set
1) (integer) 3
localhost:36379> JSON.ARRINDEX myarr $.l '"d"' 0 10000    # stop set to some huge value
1) (integer) 3
localhost:36379> JSON.ARRINDEX myarr $.l '"d"' 0 -1       # stop set to -1
1) (integer) -1
localhost:36379> JSON.ARRINDEX myarr $.l '"d"' 0          # start set to 0, but stop omitted
1) (integer) 3
```

By only appending `start` and `stop` values to the `JSON.ARRINDEX` command if they're set, the python client more closely mimics redis' behaviour. 

Fixes #2481

### DIFF
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -166,6 +166,8 @@ def test_arrindex(client):
     client.json().set("arr", Path.root_path(), [0, 1, 2, 3, 4])
     assert 1 == client.json().arrindex("arr", Path.root_path(), 1)
     assert -1 == client.json().arrindex("arr", Path.root_path(), 1, 2)
+    assert 4 == client.json().arrindex("arr", Path.root_path(), 4)
+    # assert -1 == client.json().arrindex("arr", Path.root_path(), 4, start=0, stop=-1)
 
 
 @pytest.mark.redismod
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -167,7 +167,7 @@ def test_arrindex(client):
     assert 1 == client.json().arrindex("arr", Path.root_path(), 1)
     assert -1 == client.json().arrindex("arr", Path.root_path(), 1, 2)
     assert 4 == client.json().arrindex("arr", Path.root_path(), 4)
-    # assert -1 == client.json().arrindex("arr", Path.root_path(), 4, start=0, stop=-1)
+    assert -1 == client.json().arrindex("arr", Path.root_path(), 4, start=0, stop=-1)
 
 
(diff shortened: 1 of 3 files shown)


tests/test_json.py
```diff
@@ -166,6 +166,8 @@ def test_arrindex(client):
     client.json().set("arr", Path.root_path(), [0, 1, 2, 3, 4])
     assert 1 == client.json().arrindex("arr", Path.root_path(), 1)
     assert -1 == client.json().arrindex("arr", Path.root_path(), 1, 2)
+    assert 4 == client.json().arrindex("arr", Path.root_path(), 4)
+    # assert -1 == client.json().arrindex("arr", Path.root_path(), 4, start=0, stop=-1)
```
dvora-h, line 170: Why is this line commented out?

dvora-h (commented):
Thanks for the fix!

tests/test_json.py
davemcphee: Glad you caught this - check was failing due to a logic bug in `json().arrindex()`. Fixed and uncommented test.

davemcphee:
Fixed the bool checks - forgot that 0 is falsey. Added a few more unit tests.

PR #2641, opened 2023-03-22
AsyncIO Race Condition Fix

### LINKED ISSUE #2579: [Bug] Aioredis cancellation of blocking operations stucks next operations
**Version**: redis-py 4.4.2, redis 7.0.8-1

**Platform**: Python 3.9.16 on Arch Linux

**Description**: Aioredis cancellation of blocking operations, like blpop, does not work correctly. Operation seems cancelled, but any next operation, like rpush will stuck, key does not matter. If there is timeout - stuck will last as long as time is out, otherwise forever.

**Example**
```
import asyncio
import time
from typing import Awaitable, List, TypeVar, cast

from redis import asyncio as aioredis

_T = TypeVar('_T')


async def _task_1(redis: aioredis.Redis) -> str:
    async with redis as r:
        await cast(Awaitable[List[str]], r.blpop('key_1', timeout=10))

    return 'task_1_completed'


async def _task_2() -> str:
    await asyncio.sleep(1)

    return 'task_2_completed'


async def _first_completed(*tasks: Awaitable[_T]) -> _T:
    done, pending = await asyncio.wait(
        [asyncio.ensure_future(task) for task in tasks], return_when=asyncio.FIRST_COMPLETED
    )

    for task in pending:
        task.cancel()

    await asyncio.wait(pending)

    print(done)
    print(pending)

    return done.pop().result()


async def _redis_rpush(redis: aioredis.Redis) -> None:
    async with redis as r:
        await cast(Awaitable[int], r.rpush('key_2', ''))


async def _main() -> None:
    redis = aioredis.from_url('redis://localhost/3', encoding='utf-8', decode_responses=True)

    print(await _first_completed(_task_1(redis), _task_2()))

    rpush_start_time = time.time()
    await _redis_rpush(redis)
    rpush_end_time = time.time()

    print(f"rpush seconds: {rpush_end_time - rpush_start_time}")

    await cast(Awaitable[bool], redis.flushdb())


if __name__ == '__main__':
    asyncio.run(_main())

```
**Output**
```
{<Task finished name='Task-3' coro=<_task_2() done, defined at /home/merunes/vhome/proj/work/robohood-roboapi/robohood_roboapi/main.py:17> result='task_2_completed'>}
{<Task cancelled name='Task-2' coro=<_task_1() done, defined at /home/merunes/vhome/proj/work/robohood-roboapi/robohood_roboapi/main.py:10>>}
task_2_completed
rpush seconds: 9.014980554580688
```

rpush seconds are weird, it seems like internally blpop wasn't cancelled and rpush is waiting it to complete.

**PS**:
Maybe connected with:

https://github.com/redis/redis-py/issues/2028
https://github.com/redis/redis-py/issues/2551
https://github.com/redis/redis-py/issues/2523

### LINKED ISSUE #2624: Off by 1 - Canceling async Redis command leaves connection open, in unsafe state for future commands
**Version**: 4.5.1

**Platform**:  Python 3.8 / Ubuntu (but really, any platform will likely suffer the same issue)

**Description**:  Canceling async Redis command leaves connection open, in unsafe state for future commands

The issue here is really the same as #2579, except that it generalizes it to all commands (as opposed to just blocking commands).

If async Redis request is canceled at the right time, after the command was sent but before the response was received and parsed, the connection is left in an unsafe state for future commands. The following redis operation on the same connection will send the command, and then promptly continue to read the response from the previous, canceled command. From that point on, the connection will remain in this weird, off-by-1 state. 

Here's a script reproducing the problem:
```python
import asyncio
import sys


async def main():
    myhost, mypassword = sys.argv[1:]
    async with Redis(host=myhost, password=mypassword, ssl=True, single_connection_client=True) as r:

        await r.set('foo', 'foo')
        await r.set('bar', 'bar')

        t = asyncio.create_task(r.get('foo'))
        await asyncio.sleep(0.001)
        t.cancel()
        try:
            await t
            print('try again, we did not cancel the task in time')
            print('managed to cancel the task, connection is left open with unread response')

        print('bar:', await r.get('bar'))
        print('ping:', await r.ping())
        print('foo:', await r.get('foo'))


if __name__ == '__main__':
    asyncio.run(main())
```

Running this against a server in the cloud (such that requests typically take > 5ms to complete) results in this:
```
$ python redis_cancel.py hostname password
managed to cancel the task, connection is left open with unread response
bar: b'foo'
ping: False
foo: b'PONG'
```

I believe the solution is simple. [This method](https://github.com/redis/redis-py/blob/1b2f408259405d412d7530291902f9e0c8bd34b3/redis/asyncio/client.py#L483) needs to be modified to disconnect connection when current request is canceled. For example, do this:
```python
    async def _send_command_parse_response(self, conn, command_name, *args, **options):
        """
        Send a command and parse the response
        """
        try:
            return await self.parse_response(conn, command_name, **options)
            await conn.disconnect()
            raise
```

chayim:
closes #2624 
closes #2579


drago-balto:
@chayim I think this only fixed the pipeline, while the race condition still exists in the `Redis` class:
https://github.com/redis/redis-py/blob/66a4d6b2a493dd3a20cc299ab5fef3c14baad965/redis/asyncio/client.py#L514

I can still reproduce #2624 using the latest redis-py `4.5.3`.

Keep in mind that you need slow enough Redis connection to reproduce this, otherwise the cancellation does not happen simply because the entire RESP exchange with Redis happens before there is a chance to cancel the task.

Also, are you sure that shielding from cancellation is the right approach? For example, this means that there would be no way to cancel a blocking BLPOP.

so1n:
> @chayim I think this only fixed the pipeline, while the race condition still exists in the `Redis` class:
> 
> https://github.com/redis/redis-py/blob/66a4d6b2a493dd3a20cc299ab5fef3c14baad965/redis/asyncio/client.py#L514
> 
> I can still reproduce #2624 using the latest redis-py `4.5.3`.
> 
> Keep in mind that you need slow enough Redis connection to reproduce this, otherwise the cancellation does not happen simply because the entire RESP exchange with Redis happens before there is a chance to cancel the task.
> 
> Also, are you sure that shielding from cancellation is the right approach? For example, this means that there would be no way to cancel a blocking BLPOP.

I found the same problem. Just install the Redis service locally and run the following code and it will definitely reproduce:
```Python
import asyncio


async def main():
    async with Redis(single_connection_client=True) as r:

        await r.set('foo', 'foo')
        await r.set('bar', 'bar')

        t = asyncio.create_task(r.get('foo'))
        await asyncio.sleep(0)  # <--- must 0 
        t.cancel()
        try:
            await t
            print('try again, we did not cancel the task in time')
        except asyncio.CancelledError as e:
            print('managed to cancel the task, connection is left open with unread response')

        print('bar:', await r.get('bar'))
        print('ping:', await r.ping())
        print('foo:', await r.get('foo'))


if __name__ == '__main__':
    asyncio.run(main())
```

Issue #2633, opened 2023-03-21
New `asyncio.exceptions.CancelledError` from 4.5.2


rra:
**Version**: 4.5.2
**Platform**: Python 3.11.2 on Debian unstable
**Description**:

Starting with 4.5.2, I'm seeing a new exception when attempting to write to Redis using the asyncio layer. The problem seems to be related to FastAPI: if I create a FastAPI app with any Starlette custom middlware, and then try to set a key in Redis using a separate connection pool created with `redis.asyncio.from_url`, the set request fails immediately with an uncaught `asyncio.exceptions.CancelledError` and the following backtrace:

```
.tox/py/lib/python3.11/site-packages/gafaelfawr/storage/base.py:140: in store
    await self._redis.set(key, encrypted_data, ex=lifetime)
.tox/py/lib/python3.11/site-packages/redis/asyncio/client.py:509: in execute_command
    conn = self.connection or await pool.get_connection(command_name, **options)
.tox/py/lib/python3.11/site-packages/redis/asyncio/connection.py:1408: in get_connection
    if await connection.can_read_destructive():
.tox/py/lib/python3.11/site-packages/redis/asyncio/connection.py:817: in can_read_destructive
    return await self._parser.can_read_destructive()
.tox/py/lib/python3.11/site-packages/redis/asyncio/connection.py:250: in can_read_destructive
    return await self._stream.read(1)
/usr/lib/python3.11/asyncio/streams.py:689: in read
    await self._wait_for_data('read')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <StreamReader transport=<_SelectorSocketTransport fd=15 read=polling write=<idle, bufsize=0>>>
func_name = 'read'

    async def _wait_for_data(self, func_name):
        """Wait until feed_data() or feed_eof() is called.
    
        If stream was paused, automatically resume it.
        """
        # StreamReader uses a future to link the protocol feed_data() method
        # to a read coroutine. Running two read coroutines at the same time
        # would have an unexpected behaviour. It would not possible to know
        # which coroutine would get the next data.
        if self._waiter is not None:
            raise RuntimeError(
                f'{func_name}() called while another coroutine is '
                f'already waiting for incoming data')
    
        assert not self._eof, '_wait_for_data after EOF'
    
        # Waiting for data while paused will make deadlock, so prevent it.
[...]

PR #2659, opened 2023-03-23
fix: do not use asyncio's timeout lib before 3.11.2

bellini666:
There's an issue in asyncio's timeout lib before 3.11.3 that causes async calls to raise `CancelledError`.

This is a cpython issue that was fixed in this commit [1] and cherry-picked to previous versions, meaning 3.11.3 will work correctly.

Check [2] for more info.

[1] https://github.com/python/cpython/commit/04adf2df395ded81922c71360a5d66b597471e49
[2] https://github.com/redis/redis-py/issues/2633

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -25,7 +25,9 @@ from typing import (
 )
 from urllib.parse import ParseResult, parse_qs, unquote, urlparse
 
-if sys.version_info.major >= 3 and sys.version_info.minor >= 11:
+# the functionality is available in 3.11.x but has a major issue before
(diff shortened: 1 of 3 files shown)


sileht (approved):

chayim:
@bellini666 Great hunt + find. IMHO we can release this with 4.5.x only.
@dvora-h for the merge circuit.

chayim (approved):

bellini666:
@dvora-h @chayim just to know, do I need to create a PR to cherry-pick this to 4.5.x or my work here is done? =P

Issue #2665, opened 2023-03-25
Canceling async Redis command leaves connection open, in unsafe state for future commands


drago-balto:
**Version**: 4.5.3

**Platform**: Python 3.8 on Ubuntu  / Generic

**Description**: Canceling async Redis command leaves connection open, in unsafe state for future commands

This is a reincarnation of #2624, which was closed with an incomplete fix and a possibly unreliable test case. This is the same issue that recently got a lot of attention due to ChatGPT outage, and that remains only partially fixed. The cancellation shielding introduced in #2641 addressed only the cancellation of Redis pipeline operation, but non-pipelined ops are still vulnerable.

This time I am attaching a script that reproduces the issue reliably without relying on an external, slow Redis server. This is achieved by inserting a small TCP socket proxy between the Redis client and local Redis server, with the proxy introducing a 0.1 second delay when sending data in either direction. 

Running this script with a Redis server running locally on port 6379 produces the following output:
```
$ python redis_cancel.py 
managed to cancel the task, connection is left open with unread response
bar: b'foo'
ping: False
foo: b'PONG'
```

```python
import asyncio

from redis.asyncio import Redis


async def pipe(reader: asyncio.StreamReader, writer: asyncio.StreamWriter, delay: float, name=''):
    while data := await reader.read(1000):
        # print(name, 'received:', data)
        await asyncio.sleep(delay)
        writer.write(data)
        await writer.drain()


class DelayProxy:

    def __init__(self, addr, redis_addr, delay: float):
        self.addr = addr
        self.redis_addr = redis_addr
        self.delay = delay

    async def start(self):
        server = await asyncio.start_server(self.handle, *self.addr)
        asyncio.create_task(server.serve_forever())

    async def handle(self, reader, writer):
        # establish connection to redis
        redis_reader, redis_writer = await asyncio.open_connection(*self.redis_addr)
        pipe1 = asyncio.create_task(pipe(reader, redis_writer, self.delay, 'to redis:'))
        pipe2 = asyncio.create_task(pipe(redis_reader, writer, self.delay, 'from redis:'))
        await asyncio.gather(pipe1, pipe2)


async def main():

    # create a tcp socket proxy that relays data to Redis and back, inserting 0.1 seconds of delay
    dp = DelayProxy(addr=('localhost', 6380), redis_addr=('localhost', 6379), delay=0.1)
    await dp.start()

    # note that we connect to proxy, rather than to Redis directly
[...]

PR #2666, opened 2023-03-26
Fixing cancelled async futures

chayim:
The following pull request contains changes to Async cancelling.  Specifically, all cases where a command is sent to redis, via execute_command (and friends), and now wrapped with a shield. This builds on the helpful proxy provided in #2665.

This change covers the following usage patterns.

* Async
* Async pipeline
* Async pubsub
* Async sentinel
* Async cluster
* Async cluster pipelines

Reproducible tests currently exist for async, async pipeline, and async cluster. But note - the async cluster pipeline seems at best incorrect.

Feedback and changes are very much welcomed. Bonus points for assistance with testing,


tests/test_asyncio/test_cwe_404.py
```diff
@@ -0,0 +1,126 @@
...
+import pytest
+
+from redis.asyncio import Redis
+from redis.asyncio.cluster import RedisCluster
+
+
+async def pipe(
+    reader: asyncio.StreamReader, writer: asyncio.StreamWriter, delay: float, name=""
+):
+    while data := await reader.read(1000):
```
jamestiotio, line 13: The walrus operator was only introduced in Python 3.8 (see [PEP 572](https://peps.python.org/pep-0572/)).

Since this library supports Python 3.7 as well, this line should be changed to:

```suggestion
    while True:
        data = await reader.read(1000)
        if not data:
            break
```

chayim: Thank you :D

.github/workflows/integration.yaml
```diff
@@ -51,6 +51,7 @@ jobs:
      timeout-minutes: 30
      strategy:
        max-parallel: 15
+       fail-fast: false
```
sileht, line 54: Looks unrelated.

chayim: This is so that we can see all the tests at once. It's not related - it's just to allow faster iteration.

redis/asyncio/client.py
```diff
@@ -511,12 +511,17 @@ async def execute_command(self, *args, **options):
         if self.single_connection_client:
             await self._single_conn_lock.acquire()
         try:
-            return await conn.retry.call_with_retry(
-                lambda: self._send_command_parse_response(
-                    conn, command_name, *args, **options
-                ),
-                lambda error: self._disconnect_raise(conn, error),
+            return await asyncio.shield(
```
fps7806, line 514: The try...finally should be inside the asyncio.shield if you want to prevent disconnects.

redis/asyncio/client.py
```diff
@@ -759,7 +764,12 @@ async def _disconnect_raise_connect(self, conn, error):
         if retry_on_timeout is not set or the error
         is not a TimeoutError. Otherwise, try to reconnect
         """
-        await conn.disconnect()
+        try:
+            await asyncio.shield(conn.disconnect())
```
fps7806, line 768: I think this leads to race conditions since it doesn't guarantee that it will be disconnected.

sileht (commented):
Even in theory, this change should fix the issue. asyncio.CancelledError may never reaches this redispy code due to this bug: https://github.com/aio-libs/async-timeout/issues/229.

chayim:
@fps7806  @sileht I saw that when doing some research on this a couple of days back. That's what lead down some of the paths.

IMHO the right thing to do on the security side - is this series of changes. Yes, this can lead to the side affect of a true connection issue, if the connection has been cancelled. I think, looking at that we can safely say that the use of the client should handle that, as opposed to the client.

We're going to have a look at some try/catch work here. If try/catch within the shield triggers properly - I think we're good for release. 

Targeting tomorrow.

redis/asyncio/client.py
```diff
@@ -1191,13 +1222,13 @@ async def immediate_execute_command(self, *args, **options):
...
-            lambda: self._send_command_parse_response(
-                conn, command_name, *args, **options
-            ),
-            lambda error: self._disconnect_reset_raise(conn, error),
-        )
+        try:
+            return await asyncio.shield(
+                self._try_send_command_parse_response(conn, *args, **options)
+            )
+        except asyncio.CancelledError:
```
fps7806, line 1229: This one is not necessary since try_response already has the disconnect

redis/asyncio/client.py
```diff
@@ -1386,17 +1430,10 @@ async def execute(self, raise_on_error: bool = True):
...
-                )
+                self._try_execute(conn, execute, stack, raise_on_error)
             )
-        except asyncio.CancelledError:
-            # not supposed to be possible, yet here we are
-            await conn.disconnect(nowait=True)
-            raise
-        finally:
-            await self.reset()
+        except RuntimeError:
```
fps7806, line 1435: What happens if we Canceled (so this block raises CanceledError) and then get a RuntimeError inside the shield (unhandled by this try...except since we already exited this block)

chayim: So the issue was that I couldn't successfully get that to descend in 100% of cases. The Cancelled isn't triggered - which looks very similar to the aiohttp bug - it should be.

But - thanks to testing in multiple python versions, I can also see that there are instances where the RuntimeError doesn't triggers, so a finally wrapper is also needed. Fascinated - I'll probably update a couple of existing bugs in the ecosystem.

dvora-h (approved):

risicle:
:heart: for the backports, folks.

kristjanvalur:
This change makes it impossible to cancel a command and retry, because the cancellation never happens and the single connection lock is held indefinetly.   So, the unit-tests I devised for pr #2506 no longer work.

Is this really intentional?  Not to cancel operation at all?   Is there a tl/dr somewhere on why this is desirable?

kristjanvalur:
Anyone reading the code will be clueless.  "Wait, it is creating a new `Task` for each operation?  That adds latency. and then shielding it?  Whatever for?"  Even the unit tests have no comments on them on what specifically they are trying to fix.

kristjanvalur:
This testing code now breaks.  Is there a workaround?
```python
        ready = asyncio.Event()

        async def helper():
            with pytest.raises(asyncio.CancelledError):
                # blocking pop
                ready.set()
                await r.brpop(["nonexist"])
  
          # if all is well, we can continue.  The following should not hang.
    # DEADLOCK, here:
            await r.set("status", "down")

        task = asyncio.create_task(helper())
        await ready.wait()
        await asyncio.sleep(0.01)
        # the task is now sleeping, lets send it an exception
        task.cancel()  
```
The task was interrupted with a CancelledError.  but the inner task (shileded()) was not interrupted and continues to wait, hanging on to the `_single_connection_lock` and the second attempt at `r.set()` is now deadlocked.

Basically, task interruption has been nerfed...  Is there a workaround?

agronholm:
For what reason does this shield the sub-task running `_try_send_command_parse_response()`? If I cancel a task that is performing a blocking command on Redis, the shielded subtask crashes with `redis.exceptions.ConnectionError: Connection closed by server.` when the client is closed.

In other words: **Always** manage your tasks! Having unmanaged tasks lying around is bad practice.

kristjanvalur:
I believe this PR was a mistake.  It is an attempt to fix an issuce caused by a regression that I caused last year, but without understanding the problem.  I have had a fix for that regression lying around for four months without being accepted.  Please see 
the current pr #2695 which undoes this change and correctly fixes the underlying problem.

PR #2672, opened 2023-03-29
Version 4.3.7

### LINKED ISSUE #2665: Canceling async Redis command leaves connection open, in unsafe state for future commands

dvora-h:
### Pull Request check-list

### Description of change

Closes #2665

### DIFF
--- a/tests/test_asyncio/test_cwe_404.py
+++ b/tests/test_asyncio/test_cwe_404.py
@@ -1,0 +1,146 @@
+import asyncio
+import sys
+
+import pytest
+
+from redis.asyncio import Redis
+from redis.asyncio.cluster import RedisCluster
+
+
+async def pipe(
+    reader: asyncio.StreamReader, writer: asyncio.StreamWriter, delay: float, name=""
+):
+    while True:
+        data = await reader.read(1000)
+        if not data:
+            break
+        await asyncio.sleep(delay)
+        writer.write(data)
+        await writer.drain()
+
+
+class DelayProxy:
+    def __init__(self, addr, redis_addr, delay: float):
+        self.addr = addr
+        self.redis_addr = redis_addr
+        self.delay = delay
+
+    async def start(self):
+        self.server = await asyncio.start_server(self.handle, *self.addr)
+        self.ROUTINE = asyncio.create_task(self.server.serve_forever())
+
+    async def handle(self, reader, writer):
+        # establish connection to redis
+        redis_reader, redis_writer = await asyncio.open_connection(*self.redis_addr)
+        pipe1 = asyncio.create_task(pipe(reader, redis_writer, self.delay, "to redis:"))
+        pipe2 = asyncio.create_task(
+            pipe(redis_reader, writer, self.delay, "from redis:")
+        )
+        await asyncio.gather(pipe1, pipe2)
+
+    async def stop(self):
+        # clean up enough so that we can reuse the looper
+        self.ROUTINE.cancel()
+        loop = self.server.get_loop()
+        await loop.shutdown_asyncgens()
+
+
+@pytest.mark.onlynoncluster
+@pytest.mark.parametrize("delay", argvalues=[0.05, 0.5, 1, 2])
+async def test_standalone(delay):
+
+    # create a tcp socket proxy that relays data to Redis and back,
+    # inserting 0.1 seconds of delay
+    dp = DelayProxy(
+        addr=("localhost", 5380), redis_addr=("localhost", 6379), delay=delay * 2
+    )
+    await dp.start()
+
+    for b in [True, False]:
+        # note that we connect to proxy, rather than to Redis directly
+        async with Redis(host="localhost", port=5380, single_connection_client=b) as r:
+
+            await r.set("foo", "foo")
+            await r.set("bar", "bar")
+
+            t = asyncio.create_task(r.get("foo"))
+            await asyncio.sleep(delay)
+            t.cancel()
+            try:
+                await t
+                sys.stderr.write("try again, we did not cancel the task in time\n")
+            except asyncio.CancelledError:
+                sys.stderr.write(
+                    "canceled task, connection is left open with unread response\n"
+                )
+
+            assert await r.get("bar") == b"bar"
+            assert await r.ping()
+            assert await r.get("foo") == b"foo"
+
+    await dp.stop()
+
+
+@pytest.mark.onlynoncluster
+@pytest.mark.parametrize("delay", argvalues=[0.05, 0.5, 1, 2])
+async def test_standalone_pipeline(delay):
+    dp = DelayProxy(
+        addr=("localhost", 5380), redis_addr=("localhost", 6379), delay=delay * 2
+    )
+    await dp.start()
+    async with Redis(host="localhost", port=5380) as r:
+        await r.set("foo", "foo")
+        await r.set("bar", "bar")
+
+        pipe = r.pipeline()
+
+        pipe2 = r.pipeline()
+        pipe2.get("bar")
+        pipe2.ping()
+        pipe2.get("foo")
+
+        t = asyncio.create_task(pipe.get("foo").execute())
+        await asyncio.sleep(delay)
+        t.cancel()
+
+        pipe.get("bar")
+        pipe.ping()
+        pipe.get("foo")
+        pipe.reset()
+
+        assert await pipe.execute() is None
+
+        # validating that the pipeline can be used as it could previously
+        pipe.get("bar")
+        pipe.ping()
+        pipe.get("foo")
+        assert await pipe.execute() == [b"bar", True, b"foo"]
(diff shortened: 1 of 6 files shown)


tests/test_asyncio/test_cwe_404.py
```diff
@@ -0,0 +1,148 @@
...
+
+class DelayProxy:
+    def __init__(self, addr, redis_addr, delay: float):
+        self.addr = addr
+        self.redis_addr = redis_addr
+        self.delay = delay
+
+    async def start(self):
+        self.server = await asyncio.start_server(self.handle, *self.addr)
+        self.ROUTINE = asyncio.create_task(self.server.serve_forever())
```
jamestiotio, line 30:
```suggestion
        self.ROUTINE = asyncio.ensure_future(self.server.serve_forever())
```

tests/test_asyncio/test_cwe_404.py
```diff
@@ -0,0 +1,148 @@
...
+
+    async def handle(self, reader, writer):
+        # establish connection to redis
+        redis_reader, redis_writer = await asyncio.open_connection(*self.redis_addr)
+        pipe1 = asyncio.create_task(pipe(reader, redis_writer, self.delay, "to redis:"))
```
jamestiotio, line 35:
```suggestion
        pipe1 = asyncio.ensure_future(pipe(reader, redis_writer, self.delay, "to redis:"))
```

tests/test_asyncio/test_cwe_404.py
```diff
@@ -0,0 +1,148 @@
...
+        pipe2 = asyncio.create_task(
```
jamestiotio, line 36:
```suggestion
        pipe2 = asyncio.ensure_future(
```

tests/test_asyncio/test_cwe_404.py
```diff
@@ -0,0 +1,148 @@
...
+    await dp.start()
+
+    for b in [True, False]:
+        # note that we connect to proxy, rather than to Redis directly
+        async with Redis(host="localhost", port=5380, single_connection_client=b) as r:
+
+            await r.set("foo", "foo")
+            await r.set("bar", "bar")
+
+            t = asyncio.create_task(r.get("foo"))
```
jamestiotio, line 66:
```suggestion
            t = asyncio.ensure_future(r.get("foo"))
```

tests/test_asyncio/test_cwe_404.py
```diff
@@ -0,0 +1,148 @@
...
+        await r.set("bar", "bar")
+
+        pipe = r.pipeline()
+
+        pipe2 = r.pipeline()
+        pipe2.get("bar")
+        pipe2.ping()
+        pipe2.get("foo")
+
+        t = asyncio.create_task(pipe.get("foo").execute())
```
jamestiotio, line 103:
```suggestion
        t = asyncio.ensure_future(pipe.get("foo").execute())
```

tests/test_asyncio/test_cwe_404.py
```diff
@@ -0,0 +1,148 @@
...
+
+    dp = DelayProxy(addr=("localhost", 5381), redis_addr=("localhost", 6372), delay=0.1)
+    await dp.start()
+
+    r = RedisCluster.from_url("redis://localhost:5381")
+    await r.initialize()
+    await r.set("foo", "foo")
+    await r.set("bar", "bar")
+
+    t = asyncio.create_task(r.get("foo"))
```
jamestiotio, line 136:
```suggestion
    t = asyncio.ensure_future(r.get("foo"))
```

jamestiotio (changes requested):
For Python 3.6, these calls to `asyncio.create_task` should be modified to `asyncio.ensure_future` instead.

EDIT: Seems like `asyncio.CancelledError` existed in [Python 3.6.0's code](https://github.com/python/cpython/blob/v3.6.0/Lib/asyncio/base_futures.py#L9) after all, it was just undocumented (earliest documented page of [asyncio exceptions](https://docs.python.org/3.7/library/asyncio-exceptions.html) was for Python 3.7). Ignore my previous comment about `asyncio.CancelledError`.

auvipy:
will be looking forward to this release

chayim:
Using the proxy changes from @drago-balto, combined with the current 4.3 branch yields a sideaffect in this particular case. Given the output errors on RuntimeError - we can now be positive, that the fix is truly caught.

However, the byproduct of that is this given case. If you cancel an executing pipeline (as [in our test](https://github.com/redis/redis-py/pull/2672/files#diff-a90c3a19dad7803c9726358f223a3b7b8fb23ccb1d573d580d6640f04fdd3f27R103) then the pipeline object cannot be reused.

Technically, it's always been possible (and works) to reuse the pipeline object, rather than reinitialize it:

```
r = Redis()
pipe = r.pipeline()
pipe.set("foo", "bar")
pipe.execute()

pipe.set("bar", "blee")
pipe.get("foo")
pipe.execute()
````

However, in 4.3 this won't be possible *should* the async be canceled. IMHO this is reasonable, it's a security fix we're backporting to a dead branch. Meaning if you end up in this situation, you can easily catch RuntimeError and move along.

Thoughts community?

chayim:
@auvipy I haven't seen feedback from anyone.. I let it bake for two weeks, to make sure there's real interest here. Can you confirm - is there - and does this fix work happily for you?  Anyone else, please feel free to weigh in.

This version is old enough, that I want to double-check here.

auvipy (commented):
at celery we use mostly the sync versions codes so far

auvipy:
but we are py 3.7+ and pypy 3.8+

chayim:
Right. Since celery is py 3.7+ this PR isn't needed. That's already covered by existing releases.

@dvora-h Let's let this bake. If no one *needs* the change, specifically due to python 3.6, we abandon.

auvipy:
but we need to calm down pyup

chayim:
> but we need to calm down pyup

That sounds like a misconfiguration of pyup and IMHO not a reason to release this. I'd rather not extend a potential issue.

Celery releases clearly support Python 3.7+, especially since 3.6 has been end of lifed.  I went back through the history of the repository and Pypi, and haven't seen an active version support Python 3.6.

This has sat long enough, I think we can close. Thank you everyone who helps us get to this point!

PR #2680, opened 2023-04-01
Fix redis waring https://github.com/aio-libs/aioredis-py/issues/1137

kozzztik:
Asyncio pubsub impementation raises pytest warnings in my project, that looks like:
```
    File "c:\Users\kozzz\PycharmProjects\tulius\tulius\websockets\user_session.py", line 97, in _pubsub_task
      await self.pubsub.run(
  RuntimeError: coroutine ignored GeneratorExit

    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
```

The reason is that PubSub.run() function suppress GeneratorExit exception. Python garbage collector when deleting couroutines first throws GeneratorExit to them to be sure, all context managers will be correctly closed and expects that it will be raised to show, that all finalization flow is completed. And it adds warnging if it wasn't raised.

I don't found a way to catch it in tests of redis-py, and event don't see this warning there for some reason, so provide this fix without tests change.


redis/asyncio/client.py
```diff
@@ -1043,7 +1043,7 @@ async def run(
                 await self.get_message(
                     ignore_subscribe_messages=True, timeout=poll_timeout
                 )
-            except asyncio.CancelledError:
+            except (asyncio.CancelledError, GeneratorExit):
```
chayim, line 1046: I've never actually reproduced this @kozzztik . Can you (or community @kristjanvalur maybe) weigh in here?

I don't want to catch a random exception I have yet to reproduce.

kristjanvalur: The thinking is correct. GeneratorExit generally should not be handled, any more than CancelledError. However I'd prefer to change this so that the next 'except' clause catches Exception, rather than BaseException. BaseException should almost never be caught. CancelledError still needs its own reraise handler for 3.7 compatibility.

redis/asyncio/client.py
```diff
@@ -1043,7 +1043,7 @@ async def run(
...
                 raise
             except BaseException as e:
```
kristjanvalur, line 1048:
```suggestion
            except Exception as e:
```

redis/asyncio/client.py
kristjanvalur, line 1046:
```suggestion
            except asyncio.CancelledError:
```

kristjanvalur (commented):
This is a cleaner way to do it.  BaseExceptions should generally not be caught.  The need to catch CancelledError and re-raise it is because it was an `Exception` in 3.7 and promoted to a `BaseExcetpion` only in 3.8.  So, for 3.7 we must make sure it isn't handled with the more general `Exception`.

Issue #2693, opened 2023-04-04
Usage of `hiredis.Reader` in asyncio leads to cyclic reference, causing a memory leak


oranav:
**Version**: redis-py 4.5.4, Redis version irrelevant

**Platform**: macOS 13.2.1

**Description**: There's an inherent memory leak in `redis.asyncio.connection.HiredisParser`. This is actually `hiredis-py`'s fault, but I suggest that we implement stuff a bit different to workaround this issue until <https://github.com/redis/hiredis-py/pull/163> is merged and a new version of `hiredis-py` is released.

This is a result of #2557.

Inside `HiredisParser.on_connect`, an `hiredis.Reader` object is constructed with `replyError` set to `self.parse_error`:

<https://github.com/redis/redis-py/blob/e1017fd77afd2f56dca90f986fc82e398e518a26/redis/asyncio/connection.py#L368-L376>

At this point, we have the `HiredisParser` object pointing to the `hiredis.Reader` object (through `_reader`).

`hiredis.Reader.__init__` reads `replyError` and sets it within `self.replyErrorClass`:

<https://github.com/redis/hiredis-py/blob/8adb1b3cb38b82cdc73fa2d72879712da1f74e70/src/reader.c#L288>

Now we have a cyclic reference:

`HiredisParser._reader` references to `hiredis.Reader`;
`hiredis.Reader.replyErrorClass` references to `HiredisParser.parse_error` (which is a bound method, so its references to `self`).

However, `hiredis.Reader` doesn't support garbage collection!

<https://github.com/redis/hiredis-py/blob/8adb1b3cb38b82cdc73fa2d72879712da1f74e70/src/reader.c#L47>

It doesn't declare [`Py_TPFLAGS_HAVE_GC`](https://docs.python.org/3/c-api/typeobj.html#Py_TPFLAGS_HAVE_GC).

You can also see that it doesn't implement [`tp_traverse`](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse):
<https://github.com/redis/hiredis-py/blob/8adb1b3cb38b82cdc73fa2d72879712da1f74e70/src/reader.c#L49>

So these objects are never garbage collected, leading to a memory leak.

The non-asyncio variant doesn't suffer from this problem, as its `HiredisParser.on_disconnect` deletes the reference to `hiredis.Reader`, breaking the cycle. #2557 removed this from the async variant, which has introduced the problem.

[...]

PR #2694, opened 2023-04-04
Fix memory leak caused by hiredis in asyncio case

oranav:
### Pull Request check-list

### Description of change

Change `BaseParser.parse_error` to be a class method, so that no reference is being held from `hiredis.Reader` to the `HiredisParser` instance. This is a workaround until a proper fix in [hiredis-py ](https://github.com/redis/hiredis-py)is implemented (<https://github.com/redis/hiredis-py/pull/163>).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,4 @@
+    * asyncio: Fix memory leak caused by hiredis (#2693)
     * Allow data to drain from async PythonParser when reading during a disconnect()
     * Use asyncio.timeout() instead of async_timeout.timeout() for python >= 3.11 (#2602)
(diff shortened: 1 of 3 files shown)


oranav:
Actually this can't be merged now as it causes a regression of #2557 (which has led to this bug)...

oranav:
@chayim Please note that after https://github.com/redis/hiredis-py/pull/163 is merged, there won't be a memory leak anymore.
That being said, I think it's still worthwhile to merge this PR, as `parse_error` is really a class method (semantically speaking). It doesn't belong to the instance.

dvora-h:
Fixes #2693

PR #2701, opened 2023-04-10
support report libname and libver to Redis

binfeng-xin:
Reference: https://github.com/redis/redis/pull/11758
Resolves: https://github.com/redis/redis-py/issues/2682

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -55,7 +55,7 @@ from redis.exceptions import (
     TimeoutError,
 )
 from redis.typing import EncodableT, EncodedT
-from redis.utils import HIREDIS_AVAILABLE, str_if_bytes
+from redis.utils import HIREDIS_AVAILABLE, get_lib_version, str_if_bytes
 
 hiredis = None
 if HIREDIS_AVAILABLE:
@@ -718,6 +718,14 @@ class Connection:
             await self.send_command("CLIENT", "SETNAME", self.client_name)
             if str_if_bytes(await self.read_response()) != "OK":
                 raise ConnectionError("Error setting client name")
+
+        try:
+            await self.send_command("CLIENT", "SETINFO", "LIB-NAME", "redis-py")
+            await self.read_response()
+            await self.send_command("CLIENT", "SETINFO", "LIB-VERSION", get_lib_version())
+            await self.read_response()
+        except ResponseError:
+            pass
 
         # if a database is specified, switch to it
         if self.db:
--- a/redis/commands/core.py
+++ b/redis/commands/core.py
@@ -706,6 +706,14 @@ class ManagementCommands(CommandsProtocol):
         """
         return self.execute_command("CLIENT SETNAME", name, **kwargs)
 
+    def client_setinfo(self, attr: str, value: str, **kwargs) -> ResponseT:
+        """
+        Sets the current connection libname or version
+
+        For mor information see https://redis.io/commands/client-setinfo
+        """
+        return self.execute_command("CLIENT SETINFO", attr, value, **kwargs)
+
     def client_unblock(
         self, client_id: int, error: bool = False, **kwargs
     ) -> ResponseT:
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -38,6 +38,7 @@ from redis.utils import (
     CRYPTOGRAPHY_AVAILABLE,
     HIREDIS_AVAILABLE,
     HIREDIS_PACK_AVAILABLE,
+    get_lib_version,
     str_if_bytes,
 )
 
@@ -769,6 +770,14 @@ class AbstractConnection:
             self.send_command("CLIENT", "SETNAME", self.client_name)
             if str_if_bytes(self.read_response()) != "OK":
                 raise ConnectionError("Error setting client name")
+
+        try:
+            self.send_command("CLIENT", "SETINFO", "LIB-NAME", "redis-py")
+            self.read_response()
+            self.send_command("CLIENT", "SETINFO", "LIB-VERSION", get_lib_version())
+            self.read_response()
+        except ResponseError:
+            pass
 
         # if a database is specified, switch to it
         if self.db:
--- a/redis/utils.py
+++ b/redis/utils.py
@@ -1,3 +1,4 @@
+import sys
 from contextlib import contextmanager
 from functools import wraps
 from typing import Any, Dict, Mapping, Union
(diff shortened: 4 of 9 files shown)


tests/test_asyncio/test_cluster.py
```diff
@@ -870,6 +870,14 @@ async def test_client_setname(self, r: RedisCluster) -> None:
         client_name = await r.client_getname(target_nodes=node)
         assert client_name == "redis_py_test"
 
+    async def test_client_setinfo(self, r: RedisCluster) -> None:
```
dvora-h, line 873:
```suggestion
    @skip_if_server_version_lt("7.2.0")
    async def test_client_setinfo(self, r: RedisCluster) -> None:
```

dvora-h (changes requested):

tests/test_asyncio/test_cluster.py
binfeng-xin: Thank you for your reminder.

redis/asyncio/client.py
```diff
@@ -171,6 +171,7 @@ def __init__(
         single_connection_client: bool = False,
         health_check_interval: int = 0,
         client_name: Optional[str] = None,
+        set_lib: bool = True,
```
chayim, line 174: Any specific reason you made this optional on connect? I could see instead, sending it as part of the connect statement - and if the redis version doesn't support it, grabbing this in a try. WDYT?

binfeng-xin: Good idea! 
Now it's as part of the connect statement, also you can set it using the client_setinfo command.

redis/asyncio/connection.py
```diff
@@ -718,7 +720,14 @@ async def on_connect(self) -> None:
...
             if str_if_bytes(await self.read_response()) != "OK":
                 raise ConnectionError("Error setting client name")
-
+        if self.is_connected is False:
+            try:
+                await self.send_command("CLIENT", "SETINFO", "LIB-NAME", "redis-py")
+                await self.read_response()
+                await self.send_command("CLIENT", "SETINFO", "LIB-VERSION", get_lib())
+                await self.read_response()
+            except Exception:
```
dvora-h, line 729: I think it is better to catch more specific exception, maybe ResponseError, WDYT?

redis/asyncio/connection.py
```diff
@@ -504,6 +505,7 @@ def __init__(
         socket_read_size: int = 65536,
         health_check_interval: float = 0,
         client_name: Optional[str] = None,
+        set_lib: bool = False,
```
dvora-h, line 508: Why do you need it? I can't see where you use it.

redis/commands/core.py
```diff
@@ -706,6 +706,14 @@ def client_setname(self, name: str, **kwargs) -> ResponseT:
         """
         return self.execute_command("CLIENT SETNAME", name, **kwargs)
 
+    def client_setinfo(self, attr: str, value: str, **kwargs) -> ResponseT:
+        """
+        Sets the current connection libname
```
dvora-h, line 711: or version

redis/connection.py
```diff
@@ -739,6 +740,10 @@ def _error_message(self, exception):
     def on_connect(self):
         "Initialize the connection, authenticate and select a database"
         self._parser.on_connect(self)
+        if ssl_available:
+            authened = False
```
dvora-h, line 744: Why do you need this flag?

redis/utils.py
```diff
@@ -110,3 +116,11 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+def get_lib():
```
dvora-h, line 121: I think get_lib_version is more correct name because it is more informative WDYT?

redis/asyncio/connection.py
```diff
@@ -718,7 +720,14 @@ async def on_connect(self) -> None:
             await self.send_command("CLIENT", "SETNAME", self.client_name)
             if str_if_bytes(await self.read_response()) != "OK":
                 raise ConnectionError("Error setting client name")
-
+        if self.is_connected is False:
```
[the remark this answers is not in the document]
binfeng-xin: It's for the test_cwe_404.py in test_asyncio.In the test_standalone_pipeline test case, an error will be raised if the statement is not added when single_connection_client is true.

You can try running this test case.

```python
@pytest.mark.onlynoncluster
@pytest.mark.parametrize("delay", argvalues=[0.05, 0.5, 1, 2])
async def test_standalone_pipeline(delay):
    dp = DelayProxy(
        addr=("localhost", 5380), redis_addr=("localhost", 6379), delay=delay * 2
    )
    await dp.start()
    for b in [True, False]:
        async with Redis(host="localhost", port=5380, single_connection_client=b) as r:
            await r.set("foo", "foo")
            await r.set("bar", "bar")

            pipe = r.pipeline()

            pipe2 = r.pipeline()
            pipe2.get("bar")
            pipe2.ping()
            pipe2.get("foo")

            t = asyncio.create_task(pipe.get("foo").execute())
            await asyncio.sleep(delay)
            t.cancel()

            pipe.get("bar")
            pipe.ping()
            pipe.get("foo")
            pipe.reset()

            assert await pipe.execute() is None

            # validating that the pipeline can be used as it could previously
            pipe.get("bar")
            pipe.ping()
            pipe.get("foo")
            assert await pipe.execute() == [b"bar", True, b"foo"]
            assert await pipe2.execute() == [b"bar", True, b"foo"]

    await dp.stop()

redis/connection.py
binfeng-xin: It's for the test_ssl.py in test. In the test_ssl_connection_without_ssl test case, stack overflow occur if authentication fails and the command is sent.

```python
    def test_ssl_connection_without_ssl(self, request):
        ssl_url = request.config.option.redis_ssl_url
        p = urlparse(ssl_url)[1].split(":")
        r = redis.Redis(host=p[0], port=p[1], ssl=False)

        with pytest.raises(ConnectionError) as e:
            r.ping()
            assert "Connection closed by server" in str(e)

dvora-h: You don't need it after you catch ResponseError

redis/asyncio/connection.py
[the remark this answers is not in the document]
dvora-h: Can you add test that check the on_connect() set library name & version without calling client_setinfo() ? I think it won't work with this if condition...

chayim:
Supplanted by #2857. Closing.

PR #2702, opened 2023-04-10
Updated AWS Elasticache IAM Connection Example

NickG123:
### Pull Request check-list

### Description of change

As far as I can tell, the existing example for connecting to a Redis instance with ElastiCache IAM credential provider does not work, and has never been tested.  In particular, it relies on a function of the boto3 elasticache client `generate_iam_auth_token` which does not exist (the only reference I can find to it anywhere is in this repo)

This code was developed with reference from the [AWS Docs](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/auth-iam.html) in order to develop a working solution


barshaul (approved):
Since the release of the relevant boto3 IAM function is being delayed, this appears to be a good option until then. Thanks!

PR #2706, opened 2023-04-13
Add RedisCluster.remap_host_port, Update tests for CWE 404

kristjanvalur:
### Pull Request check-list

This is a spin-off from PR #2695.  It only contains updates/fixes to the new unittest `test_cwe_404.py` and disables a test which doesn't work currently.

### Description of change

- Respect configuration settings
- Add an event to know when DelayProxy has finished sending a chunk to the server
- Default with 0 delay, but turn it on at critical parts (performance)
- Add missing `await` and fix incorrect tests
- Mark still-failing "Pipeline" tests
- Add code to the `RedisCluster` allowing host port remapping so that proxies can be used in front of a cluster.
- Update the "cluster" tests to use a set of DelayProxies, for each cluster node.

This change does not document the "host_port_remap" feature.  It if is considered a worthwhile addition to redis-py on its own, I can put it into a separate PR and add a test and documentation.

### DIFF
--- a/tests/test_asyncio/test_cwe_404.py
+++ b/tests/test_asyncio/test_cwe_404.py
@@ -120,9 +120,10 @@ async def test_standalone_pipeline(delay, redis_addr):
             pipe.get("bar")
             pipe.ping()
             pipe.get("foo")
-            pipe.reset()
+            await pipe.reset()
 
-            assert await pipe.execute() is None
+            # check that the pipeline is empty after reset
+            assert await pipe.execute() == []
 
             # validating that the pipeline can be used as it could previously
             pipe.get("bar")
--- a/tests/test_asyncio/test_cwe_404.py
+++ b/tests/test_asyncio/test_cwe_404.py
@@ -1,10 +1,23 @@
 import asyncio
 import sys
+import urllib.parse
 
 import pytest
 
 from redis.asyncio import Redis
 from redis.asyncio.cluster import RedisCluster
+from redis.asyncio.connection import async_timeout
+
+
+@pytest.fixture
+def redis_addr(request):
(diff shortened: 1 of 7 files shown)


chayim:
@dvora-h mind having a look? Let's see if we can get this into 4.5.5 with the upcoming release.

kristjanvalur:
Just to iterate, this is really just #2695 but without the code change to async connections.  i.e. just fixing up the unit tests.  We could merge this as a precursor to rebasing that one, now that we have already merged #2726.  We would then end up with three somewhat orthogonal and self-contained prs.

tests/test_asyncio/test_cwe_404.py
```diff
@@ -1,147 +1,261 @@
...
-):
-    while True:
-        data = await reader.read(1000)
-        if not data:
-            break
-        await asyncio.sleep(delay)
-        writer.write(data)
-        await writer.drain()
+@pytest.fixture
+def redis_addr(request):
```
dvora-h, line 13: @kristjanvalur WDYT about moving this fixture to conftest.py?
Now we have the exact same code 3 times (here, in [test_cluster.py](https://github.com/redis/redis-py/blob/6d32503a58d841a41556cebcf188add324c6ce86/tests/test_cluster.py#LL128C1-L129C1), and in [test_asyncio/test_cluster.py](https://github.com/redis/redis-py/blob/6d32503a58d841a41556cebcf188add324c6ce86/tests/test_asyncio/test_cluster.py#L105))

kristjanvalur: there is already a "master_host" fixture, I'll use that instead.

dvora-h (approved):

PR #2718, opened 2023-04-21
Fix incorrect usage of once flag in async Sentinel

felipou:
### Pull Request check-list

### Description of change

In the execute_command of the async Sentinel, the once flag was being
used incorrectly, with its meaning inverted (which could also be checked by comparing
with the non-async version), so I just inverted the if and else bodies.

This isn't being caught by the tests currently because the tests of commands
that use this flag do not check their results/effects (for example the "test_ckquorum"
test), but the code is covered. I'm not sure what could be done here, as I'm not yet an expert
with the Sentinel feature, and also I'm not used with the tests in this repo. If this is something that is really
wanted, could you provide some pointers as to how you think we should test this?

In the checklist above, I've left the CI tests unmarked because I do not know how to enable them in my fork, I don't have any experience with Github Actions (I tried, but it seems I need to configure a self-hosted runner? Maybe I'm not looking at the right places).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,4 @@
+    * Fix incorrect usage of once flag in async Sentinel
     * asyncio: Fix memory leak caused by hiredis (#2693)
     * Allow data to drain from async PythonParser when reading during a disconnect()
     * Use asyncio.timeout() instead of async_timeout.timeout() for python >= 3.11 (#2602)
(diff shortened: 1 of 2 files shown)


chayim (approved):

chayim (approved):

chayim:
Thank you @felipou. The resulting CI failures don't appear related to this issue.
@dvora-h IMHO good for merging

PR #2719, opened 2023-04-21
Fix potential race condition during disconnection

Anthchirp:
### Pull Request check-list

### Description of change

When the `disconnect()` function is called twice in parallel it is possible that one thread deletes the `self._sock` reference, while the other thread will attempt to call `.close()` on it, leading to an `AttributeError`.

This situation can routinely be encountered by closing the connection in a `PubSubWorkerThread` error handler in a blocking thread (ie. with `sleep_time==None`), and then calling `.close()` on the `PubSub` object. The main thread will then run into the `disconnect()` function, and the listener thread is woken up by the closure and will race into the `disconnect()` function, too.

This can be fixed easily by copying the object reference before doing the `None`-check, [similar to what we do in the `redis.client.close()` function](https://github.com/redis/redis-py/blob/7fc4c76c778163c21d396f99dcc710d99942895f/redis/client.py#L1218-L1228).

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -778,20 +778,22 @@ class AbstractConnection:
     def disconnect(self, *args):
         "Disconnects from the Redis server"
         self._parser.on_disconnect()
-        if self._sock is None:
+
+        conn_sock = self._sock
+        self._sock = None
+        if conn_sock is None:
             return
 
         if os.getpid() == self.pid:
             try:
-                self._sock.shutdown(socket.SHUT_RDWR)
+                conn_sock.shutdown(socket.SHUT_RDWR)
             except OSError:
                 pass
 
         try:
-            self._sock.close()
+            conn_sock.close()
         except OSError:
             pass
-        self._sock = None
 
     def _send_ping(self):
         """Send PING, expect PONG in return"""


chayim:
Can you share the race condition where this happens, or potentially happens? I don't really see an improvement here - but could easily be missing something.

Anthchirp:
The setup is essentially:
```python
class Example:
    def __init__(self):
        def stop_on_errors(ex: BaseException, pubsub, thread):
            if not thread._running.is_set():
                # Ignore errors during connection closure
                return
            self._thread_exception = ex
            thread.stop()

        self._redis_pubsub: redis.client.PubSub = ...
        self._thread = self._redis_pubsub.run_in_thread(
            sleep_time=None, daemon=True, exception_handler=stop_on_errors
        )

        ... # Do some stuff

        # Shutdown sequence
        # First, tell the listener thread to exit the main loop
        self._thread.stop()
        # The thread is doing blocked reads, so we wake it up by closing the
        # underlying connection
        self._redis_pubsub.close()  # one half of the race condition goes here
        # Wait for thread to stop
        self._thread.join()
```
The race condition is between the main tread running `.close()` on the PubSub object (required to wake up the redis thread), and the redis thread running `.close()` on the PubSub object ([due to the listener thread control flow](https://github.com/redis/redis-py/blob/7fc4c76c778163c21d396f99dcc710d99942895f/redis/client.py#L1793)).
One `.close()` call succeeds and terminates, but if the other thread has gone past the `if self._sock is None:` check already it will result in:
image

dvora-h (approved):

PR #2721, opened 2023-04-23
Add support for PubSub with RESP3 parser

dvora-h:
### Pull Request check-list

### Description of change

#2648

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -57,7 +57,7 @@ from redis.exceptions import (
     WatchError,
 )
 from redis.typing import ChannelT, EncodableT, KeyT
-from redis.utils import safe_str, str_if_bytes
+from redis.utils import HIREDIS_AVAILABLE, _set_info_logger, safe_str, str_if_bytes
 
 PubSubHandler = Callable[[Dict[str, str]], Awaitable[None]]
 _KeyT = TypeVar("_KeyT", bound=KeyT)
@@ -656,6 +656,7 @@ class PubSub:
         shard_hint: Optional[str] = None,
         ignore_subscribe_messages: bool = False,
         encoder=None,
+        push_handler_func: Optional[Callable] = None,
     ):
         self.connection_pool = connection_pool
         self.shard_hint = shard_hint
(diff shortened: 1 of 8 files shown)


redis/client.py
```diff
@@ -1429,6 +1429,7 @@ def __init__(
         shard_hint=None,
         ignore_subscribe_messages=False,
         encoder=None,
+        push_handler=None,
```
chayim, line 1432: ```push_handler_func```

and the type hint that goes with it

redis/utils.py
```diff
@@ -117,3 +118,12 @@ def wrapper(*args, **kwargs):
...
     return decorator
+
+
+logger = logging.getLogger("push_response")
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler()
+handler.setLevel(logging.INFO)
+logger.addHandler(handler)
+
+INFO_LOGGER = logger
```
chayim, line 129: INFO_LOGGER Should instead be _info_logger, so that we don't get caught in the same issue as the parsers rename

redis/utils.py
```diff
@@ -117,3 +118,12 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+logger = logging.getLogger("push_response")
```
chayim, line 123: instead, can this be a function? So the first time you'd call it, meaning the logger is triggered. Something like

def setup_logger(logevel=logging.INFO, handler=[logging.StreamHandler()]:
# do things here

This way, a user can also use our log handler and instantly override, just override their own handler if they want.

We need to only export the **name** of the logger. Then after the first time getLogger is called by python.. it's generally available

chayim (changes requested):
Small changes - the logger approach

redis/asyncio/client.py
```diff
@@ -656,6 +656,7 @@ def __init__(
         shard_hint: Optional[str] = None,
         ignore_subscribe_messages: bool = False,
         encoder=None,
+        push_handler_func: Optional[Callable] = None,
```
chayim, line 659: really happy to see the same option name :heart:

chayim (approved):

PR #2726, opened 2023-04-27
add "address_remap" feature to RedisCluster

kristjanvalur:
### Pull Request check-list

### Description of change

A new parameter, `address_remap` is added to both both `RedisCluster` and `asyncio.RedisCluster`.  This allows remapping the addresses that the cluster nodes advertise about themselves, to something which is accessible to a client.  For example, the nodes might sit behind a NAT, or a reverse proxy, that the nodes themselves don't know about.  The `address_remap` will ensure that the `RedisCluster` object will connect to the _remapped_ addresses when connecting to the cluster.

This PR could be considered a precursor to  #2695 and #2706  since both of these provide this functionality in order to facilitate unit-tests, albeit only for the `asyncio.RedisCluster`.  Accepting this PR would allow those other PRs to be simplified.

### Reference

This feature is inspired by the `host_port_remap` feature described here: https://redis-py-cluster.readthedocs.io/en/stable/client.html
but we use a more flexible _callback_ based mechanism rather than a dictionary of rules.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,4 @@
+    * Add `address_remap` parameter to `RedisCluster`
     * Fix incorrect usage of once flag in async Sentinel
     * asyncio: Fix memory leak caused by hiredis (#2693)
     * Allow data to drain from async PythonParser when reading during a disconnect()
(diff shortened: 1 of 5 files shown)


chayim:
Hey @kristjanvalur Just wanted to let you know that the failing test appears to be a red herring. Not sure if you're ready for review and whatnot, but I did want to let you know ;)

Issue #2730, opened 2023-04-30
Commands don't implement `CommandsProtocol.connection_pool`


Avasam:
**Version**:  redis-py 4.5.4

**Platform**: Any

**Description**:

The `CommandsProtocol` protocol at `redis/typing.py` has a `connection_pool` member. However, all commands that subclasses the protocol seem to be missing `connection_pool` in their implementation. This causes issues with type-checking tools expecting the commands to respect the Protocol.

Avasam:
I could make a PR to fix it, but the problem is that I don't know what was the **intention**. Is the protocol wrong for wanting `connection_pool` ? Or are the commands wrong for not implementing it? Is the variable set dynamically later down the line? If so, where, and should `connection_pool` be potentially `None`?

petyaslavova:
Closing this issue as a duplicate of #3574 (I leave the newer one, because it is easier to find it). I'm planning to work on a fix soon.

PR #2738, opened 2023-05-03
Add py.typed in accordance with PEP-561

zmievsa:
### Pull Request check-list

### Description of change

Add py.typed in accordance with PEP-561. Otherwise type checkers might ignore type hints from this package (particularly old versions of pyright)


chayim:
@Ovsyanka83 thanks kindly! But.. the library isn't fully typed at this time. Want to help out by contributing type hints?

Ovsyanka83:
Would love to. I think, the PEP allows us to specify that the typing is partial

chayim:
Amazing, didn't know about the partial bit. Any idea how typeshed handles that? Those types are incorrect - but if they won't absorb (even after the 6 month cycle) it's... perhaps less valuable. WDYT?

Ovsyanka83:
I will try to leave my full reply and research into that next week or, in the worst case, by the end of the month.

akx:
@chayim We added types to Babel in the last release, see https://github.com/python-babel/babel/pull/975 and https://github.com/python/typeshed/pull/9832 for relevant discussion.

chayim:
Thanks @akx any idea how partially typed works? typeshed is wrong in all sorts of places for redis-py... The library is only partially typed.. hence the Q

akx:
@chayim I'd _assume_ that anything not explicitly typed will remain any-typed, just as if it hadn't been typed at all. But you'd probably not want to mix and match typeshed types with the native ones.

chayim:
I have the same assumption.. hence my worry.  Maybe we should get this into 5 instead - along with more typings.

chayim (approved):

JelleZijlstra:
> typeshed is wrong in all sorts of places for redis-py

@chayim If you have concrete examples of this, please let the typeshed project know and we'll try to fix it.

Great to see redis-py adding py.typed though!

Avasam:
(not a typeshed maintainer, just a contributor), FWIW, https://github.com/redis/redis-py/issues/2730 has been blocking me from fixing a handful of redis typing issues. And furthermore enabling more tools and checks in typeshed that ensures proper typing.

Concerning incomplete typing in both the base library (here) and typeshed, refer to my comment there: https://github.com/python/typeshed/issues/10592#issuecomment-1706972996 I fully support type definitions moving fully into Redis itself. But if it can't be done quickly or completely, Typeshed should be able to provide explicitly partial stubs and gradually remove modules as they are type-completed in Redis.

Issue #2749, opened 2023-05-07
After Upgrading to 4.5.4 we're getting `Task was destroyed but it is pending!`


rohanerw:
**Version**: What redis-py and what redis version is the issue happening on? 
**redis-py version** - 4.5.4
**python version** -  3.10.10

**Description**: Description of your issue, stack traces from errors and code that reproduces the issue
We recently bumped `redis-py` to its latest version 4.5.4. After that on sentry we started to see these exceptions `Task was destroyed but it is pending!`

There's nothing much in stack trace just this text.

```
Task was destroyed but it is pending!
task: <Task pending name='Task-634' coro=<Connection.disconnect() done, defined at /envs/django-proj/bin/python3.10/site-packages/redis/asyncio/connection.py:727> wait_for=<Future pending cb=[Task.task_wakeup()]>>
```

This is happening with the tasks that uses redis in between, to me it appears that there's some leak here.
Somehow I'm not able to reproduce this locally otherwise I could have provided steps to reproduce.

Any help is appreciated and big thanks for building this cool library.

chayim:
@dvora-h As you're getting a 4.5.5 ready, I wonder if some of @kristjanvalur recent contributions address this!

kristjanvalur:
It looks like a task created specifically to disconnect.  From what I can tell, this happens only in cluster code.  Are you using a redis cluster and `asyncio.RedisCluster`?

rohanerw:
@kristjanvalur Yes we're using redis cluster via sentinel.

 And no we're not using `asyncio.RedisCluster`

kristjanvalur:
> And no we're not using `asyncio.RedisCluster`

Sorry, let me rephrase the question:  What mechanism provided by `redis.asyncio` are you using to communicate with your cluster?  Which `redis` classes do you use?

rohanerw:
> What mechanism provided by redis.asyncio are you using to communicate with your cluster?

We're not using anything related to `redis.asyncio` to communicate with cluster.

> Which redis classes do you use?

`redis.sentinel.Sentinel`

Here's a simple example.

```
def get_redis(read_only: bool)
  from redis.sentinel import Sentinel
  sentinel = Sentinel([(URL, PORT)])
  if read_only:
      return sentinel.slave_for(...)
  return sentinel.master_for(...)

```

@kristjanvalur I hope it clarifies my previous comment.

kristjanvalur:
[...]

PR #2755, opened 2023-05-08
Issue #2749: Remove unnecessary __del__ handlers

kristjanvalur:
There normally should be no logic attached to del.  Cleanly disconnecting network resources is not needed at that time.

It is possible that this code was added in the past to address some problem, but there are no comments to suggest what they may have been.  As this is only invoked during garbage collection, no one cares anymore whether the socket is cleanly
and correctly closed.  Invoking new background at such time is not good practice.

### Pull Request check-list

### Description of change

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,4 @@
+    * Fix #2749, remove unnecessary __del__ logic to close connections.
     * Fix #2754, adding a missing argument to SentinelManagedConnection
     * Fix `xadd` command to accept non-negative `maxlen` including 0
     * Revert #2104, #2673, add `disconnect_on_error` option to `read_response()` (issues #2506, #2624)
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -181,12 +181,6 @@ class BaseParser:
         self._read_size = socket_read_size
         self._connected = False
 
-    def __del__(self):
-        try:
-            self.on_disconnect()
-        except Exception:
-            pass
-
     @classmethod
     def parse_error(cls, response: str) -> ResponseError:
         """Parse an error response"""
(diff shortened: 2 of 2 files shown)


kristjanvalur:
Unit test failures seem completely unrelated.  This project really needs a round of fixing sporadically failing unittests.

kristjanvalur:
__del__() is only ever called if an object is being garbage collected.  So, no one is interested in it anymore.  There is no logic running which requires that the connection be cleanly disconnected.  It will be sufficient to just also allow the underlying socket to also be garbage collected.

In BaseParser, the __del__ is used to call on_disconnect.  But by this time, no one cares (all it does is to set a state flag in the parser)

In the Connection, all it does is cause an underlying socket to be cleanly disconnected.  But by this time, all data has been flushed anyway.  And doing things like running the event loop from a __del__ method is not nice at all.

__del__() is very rarely used.  It is mostly used for:

1. Resource warnings.  It is a good place to detect, e.g. that a socket was not closed, and that someone should have actually used a context manager or a `close()` method somewhere.
2. To resurrect objects, e.g. when doing object caching.  but this behaviour is implementation dependend.]

If you read the documentation for __del__ in the python docs, it should make clear that it is not something you want to use for program control.   You really cannot use it for RAII like c++ destructors, for instance.

kristjanvalur:
After fixing the subscription iterator, it is finally possible to get deterministic behaviour for subscription finalizers.  Previously they could get bot GeneratorExit _and_ a CancelledError.  it made it impossible to do any async logic in there, like wait for condition variables etc.

dvora-h (approved):

kristjanvalur:
🎆

PR #2778, opened 2023-05-25
Introduce OutOfMemoryError exception for Redis write command rejections due to OOM errors

bmacphee:
- handle returned "OOM" error code string by raising explicit exception class

### Pull Request check-list

### Description of change

I ran across this when using `noeviction` as the eviction policy. I thought it would be a bit more convenient to have the exception for it, so I followed the pattern of the other error handling.  I can address the other items in the check-list if it is something that would be valuable.


bmacphee:
On the CI workflow action, I see all sorts of "RuntimeError: Event loop is closed" errors, for things unrelated to this change.  Not sure what is going on there, so I will leave the box unticked.

chayim (approved):
LGTM. Thanks a bunch!

bmacphee:
I did some quick manual testing of both the `redis` and `redis.asyncio` packages to double check the `try/except` works when connected with a real server that has hit the memory limit (rather than one that is forced to emit an error as in the tests).  Is that sufficient to check the third box?  

Any other docs/examples or further changes required?  It looks like the exceptions are documented automatically from the class docstring.

chayim:
@bmacphee I think this makes sense as is - docs generated from docstring is the current (emphasis on current) path.

PR #2785, opened 2023-05-29
Removing compat

chayim:
compat was used to handle type loading - but those types are provided by typing. We need to run these through the python 3.7 tests, because it may fail entirely there - so this is a draft PR.


chayim:
Not worth it. Abandoning

PR #2816, opened 2023-06-28
Save a reference to created async tasks, to avoid tasks potentially disappearing

anio:
> Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done.
>
> [python docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)

### Pull Request check-list

### Description of change

Please review the link: https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/


redis/asyncio/cluster.py
```diff
@@ -1117,13 +1117,13 @@ def set_nodes(
         if remove_old:
             for name in list(old.keys()):
                 if name not in new:
-                    asyncio.create_task(old.pop(name).disconnect())
+                    task = asyncio.create_task(old.pop(name).disconnect())
```
dvora-h, line 1120:
```suggestion
                    task = asyncio.create_task(old.pop(name).disconnect())  # noqa
```

redis/asyncio/cluster.py
```diff
@@ -1117,13 +1117,13 @@ def set_nodes(
...
 
         for name, node in new.items():
             if name in old:
                 if old[name] is node:
                     continue
-                asyncio.create_task(old[name].disconnect())
+                task = asyncio.create_task(old[name].disconnect())
```
dvora-h, line 1126:
```suggestion
                task = asyncio.create_task(old[name].disconnect())  # noqa
```

redis/asyncio/cluster.py
dvora-h: Skip linter check

redis/asyncio/cluster.py
dvora-h: same

dvora-h (changes requested):

dvora-h (approved):

jakob-keller:
@anio: Thanks for bringing the issue up!

I am not convinced that the fix is appropriate, though. As the [source you provide](https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/) mentions:

> The solution recommended in the docs is to keep a reference to the task for as long as you need it to live.

The way I understand it is, that we need to save the task reference beyond the scope of the local / loop variable. As it stands, the reference will be lost almost immediately, rendering the fix ineffective.

PR #2855, opened 2023-07-18
Create dt_list.py

sav-norem:
as part of the effort to get tabbed examples in client languages on the redis.io website, this file contains the examples for the list datatype in python with the appropriate short codes to be pulled in

### DIFF
--- a/doctests/dt_list.py
+++ b/doctests/dt_list.py
@@ -1,0 +1,303 @@
+# EXAMPLE: list_tutorial
+# HIDE_START
+import redis
+
+r = redis.Redis(decode_responses=True)
+r.delete("bikes:repairs")
+r.delete("bikes:finished")
+# HIDE_END
+
+# STEP_START queue
+res1 = r.lpush("bikes:repairs", "bike:1")
+print(res1)  # >>> 1
+
+res2 = r.lpush("bikes:repairs", "bike:2")
+print(res2)  # >>> 2
+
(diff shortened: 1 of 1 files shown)


sav-norem:
@uglide and lists!

doctests/dt_list.py
```diff
@@ -0,0 +1,324 @@
+# EXAMPLE: list_tutorial
+# HIDE_START
+import redis
+
+r = redis.Redis(decode_responses=True)
+r.delete("bikes:repairs")
```
uglide, line 6: Please wrap those two lines in the REMOVE_START/END block to hide this cleanup code from the users.

doctests/dt_list.py
```diff
@@ -0,0 +1,324 @@
...
+r.delete("bikes:finished")
+# HIDE_END
+
+# STEP_START queue
```
uglide, line 10: See my comment here https://github.com/redis/redis-doc/pull/2488/files#r1269382572

uglide (changes requested):
Great work!
We can make this example more concise by explaining List basics around one pattern Producer/Consumer which is extensively covered in the current version of the doc here https://redis.io/docs/data-types/lists/

sav-norem:
@uglide I believe those changes would be better addressed with an issue for the docs team to rework the structure of that content

sav-norem:

PR #2857, opened 2023-07-19
Add support for `CLIENT SETINFO`

### LINKED ISSUE #2682: Add support for new redis command CLIENT SETINFO
7.2.0 adds support for the new redis command CLIENT SETINFO. We need to add support, documented [here](https://redis.io/commands/CLIENT-SETINFO).

### LINKED ISSUE #2856: Clients should identify themselves on connect
[CLIENT SETINFO](https://redis.io/commands/client-setinfo/) is a new fire-and-forget command that is available as part of redis 7.2. Clients should send the lowercased version of the client name (i.e redis-py, jedis, go-redis) as part of the initial connect, after authenticating to the server.

dvora-h:
### Pull Request check-list

### Description of change

closes #2856 
closes #2682

### DIFF
--- a/tests/test_lock.py
+++ b/tests/test_lock.py
@@ -101,11 +101,12 @@ class TestLock:
         assert lock1.acquire(blocking=False)
         bt = 0.4
         sleep = 0.05
+        fudge_factor = 0.05
         lock2 = self.get_lock(r, "foo", sleep=sleep, blocking_timeout=bt)
         start = time.monotonic()
         assert not lock2.acquire()
         # The elapsed duration should be less than the total blocking_timeout
-        assert bt > (time.monotonic() - start) > bt - sleep
+        assert (bt + fudge_factor) > (time.monotonic() - start) > bt - sleep
         lock1.release()
 
     def test_context_manager(self, r):


redis/asyncio/connection.py
```diff
@@ -347,6 +347,15 @@ async def on_connect(self) -> None:
             if str_if_bytes(await self.read_response()) != "OK":
                 raise ConnectionError("Error setting client name")
 
+        try:
+            # set the library name and version
+            await self.send_command("CLIENT", "SETINFO", "LIB-NAME", "redis-py")
+            await self.read_response()
```
chayim, line 353: Any reason to not pipeline this?,,, and just not read the response? From the docs:
`Client libraries are expected to pipeline `

redis/connection.py
```diff
@@ -360,6 +361,15 @@ def on_connect(self):
             if str_if_bytes(self.read_response()) != "OK":
                 raise ConnectionError("Error setting client name")
 
+        try:
+            # set the library name and version
+            self.send_command("CLIENT", "SETINFO", "LIB-NAME", "redis-py")
```
chayim, line 366: same re pipeline

redis/utils.py
```diff
@@ -27,6 +28,11 @@
 except ImportError:
     CRYPTOGRAPHY_AVAILABLE = False
 
+if sys.version_info >= (3, 8):
```
chayim, line 31: There's a VERSION available in __init__ for you

tests/test_asyncio/test_commands.py
```diff
@@ -355,6 +355,14 @@ async def test_client_setname(self, r: redis.Redis):
             r, await r.client_getname(), "redis_py_test", b"redis_py_test"
         )
 
+    @skip_if_server_version_lt("7.2.0")
+    async def test_client_setinfo(self, r: redis.Redis):
+        assert await r.client_setinfo("lib-name", "test")
+        assert await r.client_setinfo("lib-ver", "123")
```
chayim, line 361: can we:
- set with the master version
- set with fake versions, to test?

tests/test_commands.py
```diff
@@ -554,6 +554,14 @@ def test_client_setname(self, r):
         assert r.client_setname("redis_py_test")
         assert_resp_response(r, r.client_getname(), "redis_py_test", b"redis_py_test")
 
+    @skip_if_server_version_lt("7.2.0")
+    def test_client_setinfo(self, r: redis.Redis):
+        assert r.client_setinfo("lib-name", "test")
```
chayim, line 559: I'd love to use the library name redis-py as a const, even here

chayim:
@kristjanvalur I'm digging through what I think is the initial issue (#2349) and subsequent PR (#2557), and the recent fix in #2859.. and to be honest, a bit flummoxed on why the test _tests/test_asyncio/test_connection.py::test_connection_disconect_race[AsyncRESP2Parser]_ fails, and then cascades.

It ends with a timeout issue - but there's something missing in my understanding. Can I talk you into having a look :pray: ? I will continue to do so.

chayim:
So in theory it doesn't, which is why I'm flumoxxed. On connection we now call redis' identification commands (this is good!) and allow users to not do so should they choose.

But, this test fails because we hit the 30s timeout - which was the purpose of the test. Further down the failure deck, everything fails due to what appears to be data in the async retrieval side.

On the bright side the stack traces haven't been useful. Using [this CI](https://github.com/redis/redis-py/actions/runs/5711572716/job/15473434838?pr=2857) result as an example after the test fails, we eventually lead to read failures such as below - but honestly that's a red herring. The real issue is data being left in the reader.

```ERROR tests/test_asyncio/test_timeseries.py::test_get[single-python-parser] - IndexError: pop from empty list```

To me, this spells out the need to properly reset the async side prior to each test, but that's another issue.

**Comment by kristjanvalur** [comment]
Okay, I see.  I will try to figure out what is going on.

**Comment by kristjanvalur** [comment]
Ok, what is happening, is that the 30s timeout occurs as part of the test harness.  The connect call fails, the test is aborted, but the coroutine running the test is never killed.  i.e., this code:

```
with patch.object(asyncio, "open_connection", open_connection):
        await conn.connect()
```
never exits the context manager.  This means that `asyncio.open_connection` remains patched for all subsequent tests.
Now the timeout you see is a weird one, and shouldn't happen.  it comes from base_events.py, here:

```
timeout = None
        if self._ready or self._stopping:
            timeout = 0
        elif self._scheduled:
            # Compute the desired timeout.
            when = self._scheduled[0]._when
            timeout = min(max(0, when - self.time()), MAXIMUM_SELECT_TIMEOUT)

        event_list = self._selector.select(timeout)
```
i.e. we are hitting a hard timeout on select and there is nothing to catch it.   For async, we need async timeouts.
I am also not clear on why you would want this test to suddenly start to timeout like this.   I'll continue to look at what is happening.

**Comment by kristjanvalur** [comment]
Ah, it is the "pytest_timeout".
Yes, if that triggers during an async test, then none of the async cleanup code is executed.  We should never _rely_ on such
a timeout to be part of an async test, we should consider it to be only a _failsafe_

**Comment by kristjanvalur** [comment]
So, yes, we are timing out here:
```python
 try:
            # set the library name and version
            await self.send_command("CLIENT", "SETINFO", "LIB-NAME", "redis-py")
            await self.read_response()
            await self.send_command("CLIENT", "SETINFO", "LIB-VER", get_lib_version())
            await self.read_response()
        except ResponseError:
            pass
```
during the first "self.read_response()".  This is because we have already patched the send/receive stuff.  let me find out a way to fix this test.

chayim:
@kristjanvalur thanks a tonne. I don't think I could have debugged this at all... The byproduct was bizarre - especially since frankly, it seemed completel illogical to me. This is my way of saying :heart: 

@dvora-h can you take a look at merging this into our branch and seeing how it lies.

kristjanvalur:
There is a problem with mocking in 3.7, let me fix that.

kristjanvalur:
I've pushed an additional commit fixing 3.7, feel free to merge/cherry pick it.

chayim (approved):

redis/_parsers/helpers.py
```diff
@@ -628,8 +628,11 @@ def parse_client_info(value):
     Parsing client-info in ACL Log in following format.
     "key1=value1 key2=value2 key3=value3"
     """
+    value = str_if_bytes(value)
+    if value[-1] == "\n":
+        value = value[:-1]
     client_info = {}
-    infos = str_if_bytes(value).split(" ")
+    infos = value.split(" ")
     for info in infos:
```
kristjanvalur, line 636:
```suggestion
    for info in str_if_bytes(value).strip().split():
```

redis/_parsers/helpers.py
```diff
@@ -628,8 +628,11 @@ def parse_client_info(value):
     Parsing client-info in ACL Log in following format.
     "key1=value1 key2=value2 key3=value3"
     """
+    value = str_if_bytes(value)
+    if value[-1] == "\n":
+        value = value[:-1]
```
kristjanvalur, line 633:
```suggestion
```

redis/asyncio/connection.py
```diff
@@ -347,6 +353,19 @@ async def on_connect(self) -> None:
...
+        try:
+            # set the library name and version
+            if self.lib_name:
+                await self.send_command("CLIENT", "SETINFO", "LIB-NAME", self.lib_name)
+                await self.read_response()
+            if self.lib_version:
+                await self.send_command(
+                    "CLIENT", "SETINFO", "LIB-VER", self.lib_version
+                )
+                await self.read_response()
```
kristjanvalur, line 365:
```suggestion
            if self.lib_version:
                await self.send_command(
                    "CLIENT", "SETINFO", "LIB-VER", self.lib_version
                )
                await self.read_response()
                if self.lib_version:
                     await self.read_response()
```

redis/asyncio/connection.py
```diff
@@ -347,6 +353,19 @@ async def on_connect(self) -> None:
...
+        except ResponseError:
+            pass
```
kristjanvalur, line 367:
```suggestion
        # set the library name and version, pipeline for lower startup latency
        if self.lib_name:
            await self.send_command("CLIENT", "SETINFO", "LIB-NAME", self.lib_name)
        if self.lib_version:
            await self.send_command(
                "CLIENT", "SETINFO", "LIB-VER", self.lib_version
            )
        for did_send in (self.lib_name, self.lib_version):
            if did_send:
                try:
                    await self.read_response()
                except ResponseError:
                    pass
```

redis/asyncio/connection.py
```diff
@@ -347,6 +353,19 @@ async def on_connect(self) -> None:
...
+
         # if a database is specified, switch to it
         if self.db:
```
kristjanvalur, line 370: and remove this, since it is part of the pipeline suggestion above.

redis/connection.py
```diff
@@ -360,6 +365,17 @@ def on_connect(self):
             if str_if_bytes(self.read_response()) != "OK":
                 raise ConnectionError("Error setting client name")
 
+        try:
+            # set the library name and version
+            if self.lib_name:
+                self.send_command("CLIENT", "SETINFO", "LIB-NAME", self.lib_name)
+                self.read_response()
+            if self.lib_version:
+                self.send_command("CLIENT", "SETINFO", "LIB-VER", self.lib_version)
```
kristjanvalur, line 374: Pipeline SETINFO and SELECT same as in the async suggestion above.

chayim: Specifically not sent because of the interface - but a suggestion we'd debated previously. At this level, the pipeline shouldn't be used (yet). IMHO once reorganizing connections, all items for connect should be pipelined, agreed.

redis/asyncio/connection.py
chayim: Definitely a cleaner pattern.

vladvildanov (approved):

redis/connection.py
kristjanvalur: Well, you probably debated it out of band, not in this PR :)  What "interface" are you speaking of?

A pipeline isn't an object "to be used", it is just a way of sending a bunch of commands and not waiting for each individual response before sending the next command.  It is always supported by the server.
The spec specifically recommends pipelining startup commands to reduce latency.  We now have two extra roundtrips to the server at startup in addition to the optional one to select a db.

Issue #2831, opened 2023-07-03
4.6.0 - ConnectionError max number of clients reached


nsteinmetz:
**Version**: 4.6.0

**Platform**: Python 3.10 / Docker or Linux

**Server** : Redis 7.0.11

**Description**: 

We have an application using redis in an async context for months now. With the ugrade of redis-py from 4.5.x to 4.6.0, we saw our app on staging environment starts failing with `ConnectionError max number of clients reached`

I first run a review of the code to add the missing `await redis_conn.close()` statements that were missing but the issue still occurs.

It seems that we reach the 10.000 max clients settings in one hour approx whereas we never used to so far. Unfortunately on this environment, I don't track this metric.

On staging last week, I used a workaround to set a client timeout to pass the one hour issue (`config set timeout 60`)

Doing it again a few minutes ago, I could drop connection from 8k to 160 and it's stable then around 160.

On production environment (Python 3.9, redis 4.5.5), we have 25 connected clients on avergage over the last 6 months with a similar activity to our staging environment. And there is no `timeout` configuration set, so it's disabled (default configuration)

On staging, reverting to 4.5.5 version of redis-py and with timeout disabled (aka value to 0), after a few minutes it's stable around 20 connections whereas it was to be 1000+ in the same amount of time with 4.6.0.

My code used to be basically :

```python
from redis.asyncio import Redis
from app import settings

async def get_redis_connection():
    redis = Redis.from_url(settings.REDIS_URL)
    return redis

async def list_queued_job_ids():
    redis_conn = await get_redis_connection()
    ... do something with redis data and related code ...
    await redis_conn.close()
    return something
```

So I don't know if it's a regression in 4.6.0 with the AbstractClass maybe ? or something to add/change in my code but I didn't find any documentation that could help me on the changes to be made.

Let me know if you need further details.

nsteinmetz:
hi @woutdenolf @dvora-h,

As you worked on and validated the PR #2734 about extracting abstract async connection class  would this issue make sense to you ?

nsteinmetz:
Hi,

I'll try before the start of my summer break at the end of week or I'll do once I'm back at the latest. I'll also try to test to see if it's related to the backend/API or one of our workers

But code is not supposed to work this way to hit the 10.000 maxclients limits.

woutdenolf:
Issue https://github.com/redis/redis-py/issues/2814 reports connections not being released to their pool. Maybe related?

nsteinmetz:
[...]

PR #2859, opened 2023-07-19
Fix socket garbage collection

kristjanvalur:
### Pull Request check-list

### Description of change

Issue #2831 shows how connections are kept open when they are simply discarded.
The problem is two-fold:

1. Redis connections opened with `from-url` don't have their `auto_close_connection` boolean set to `True`, as is the default for a pool which is created automatically.  Therefore, when `Redis.close()` was called, the pool would not close its collections.
2. Even so, garbage collection of the `Connection` objects _ought_ to have closed the sockets associated with the underlying `StreamWriter` objects.  This apparently does not reliably happen, so we re-instate a `__del__` handler for the `AbstractConnection` which does the bare minimum required, calling `StreamWriter.close()`.
3. Add an `auto_close_connection_pool` argument to `asyncio.Redis.from_url()` which defaults to true, same as for `asyncio.Redis()`

A fix was added to an unrelated test which caused unit tests to fail on non-linux platforms.

###  Update

I've changed the PR so as to remove the _2_, above.  In general for `asyncio` Python, socket cleanup is _not_ done on garbage collection (since it involves async operations).  Instead, it is assumed that a well written program performs proper disconnects as needed.  `__del__` handlers instead are used to emit _warnings_, that is, if a socket (or other resource) is found to be not closed when purging an object, a warning is emitted.  This is to make sure resources are released properly by the application.

 Fixing of (1) above ensures that the connection pool is indeed cleaned up correctly.  Trying to put cleanup and IO logic into destructors is in general not thought to be a good practice and should be avoided whenever possible.  Correct program operation should not rely on the prompt execution of `__del__` handlers, and implementing them would deprive us of the built-in resource warnings which would otherwise be emitted.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,4 @@
-    * Fix #2831, closing connections and sockets when garbage collected.
+    * Fix #2831, add auto_close_connection_pool=True arg to asyncio.Redis.from_url()
     * Fix incorrect redis.asyncio.Cluster type hint for `retry_on_error`
     * Fix dead weakref in sentinel connection causing ReferenceError (#2767)
     * Fix #2768, Fix KeyError: 'first-entry' in parse_xinfo_stream.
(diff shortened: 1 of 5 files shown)


dvora-h (approved):

kristjanvalur:
@dvora-h you may want to see the latest change.  Arguments intended for the "Redis" class are now explicit in `from_url`, while the `kwargs` get passed to the `ConnectionPool` constructor.  This makes the docstring more correct.

PR #2861, opened 2023-07-23
Adding support for triggered functions (TFUNCTION)

shacharPash:
### Pull Request check-list

### Description of change


redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+        that will be provided to the library on load time,
+        for more information refer to
+        https://github.com/RedisGears/RedisGears/blob/master/docs/function_advance_topics.md#library-configuration
+        ``replace`` - an optional argument, instructs RedisGears to replace the
+        function if its already exists
+
+        For more information see https://redis.io/commands/tfunction-load/
+        # TODO: check link when it will be available
+        """
+        pieces: list[EncodableT] = []
```
chayim, line 2474: Why declare the type of the list and not just do pieces = []

redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+
+        ``lib_name`` the library name to delete.
+
+        For more information see https://redis.io/commands/tfunction-delete/
+        # TODO: check link when it will be available
+        """
+        return self.execute_command("TFUNCTION DELETE", lib_name)
+
+    def tfunction_list(
+        self, with_code: bool = False, verbose: int = 0, lib_name: str = None
```
chayim, line 2494: Same type comment with regard to union

redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+    ) -> ResponseT:
+        """
+        List the functions with additional information about each function.
+
+        ``with_code`` Show libraries code.
+        ``verbose`` output verbosity level, higher number will increase verbosity level
+        ``lib_name`` specifying a library name (can be used multiple times to show multiple libraries in a single command) # noqa
+
+        For more information see https://redis.io/commands/tfunction-list/
+        # TODO: check link when it will be available
```
chayim, line 2504: IMHO this doesn't belong in the source - but you should ping Meir

redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+        """
+        pices: list[EncodableT] = []
```
chayim, line 2506: Variable name should be ```pieces``` not ```pices```
Also - same encodabel comment

redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+        if with_code:
+            pices.append("WITHCODE")
+        if verbose > 0 and verbose < 4:
```
chayim, line 2509: ```if 0 < verbose < 4```

redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+            pices.append("v" * verbose)
+        elif verbose != 0:  # verbose == 0 is the default so no need to throw an error
```
chayim, line 2511: ekse

redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+
+        return self.execute_command("TFUNCTION LIST", *pices)
+
+    def tfcall(
+        self,
+        lib_name: str,
+        func_name: str,
+        keys: KeysT = None,
+        *args: List,
+        _async: bool = False,
```
chayim, line 2525: _async belongs before *args since *args is a list

redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+        Trigger a sync or async (Coroutine) function.
+
+        ``lib_name`` - the library name contains the function.
+        ``func_name`` - the function name to run.
+        ``keys`` - the keys that will be touched by the function.
+        ``args`` - Additional argument to pass to the function.
+        ``_async`` - If True, Invoke an async function (Coroutine.
+
+        For more information see https://redis.io/commands/tfcall/
+        # TODO: check link when it will be available
```
chayim, line 2537: same TODO comment as above

redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+        """
+        pieces: list[EncodableT] = [f"{lib_name}.{func_name}"]
```
chayim, line 2539: Same comment on typing

redis/commands/core.py
```diff
@@ -2454,6 +2454,100 @@ def substr(self, name: KeyT, start: int, end: int = -1) -> ResponseT:
...
+        if keys is not None:
+            pieces.append(len(keys))
+            pieces.extend(keys)
+        else:
+            pieces.append(0)
+        if args is not None:
+            pieces.extend(args)
+        if _async:
+            return self.execute_command("TFCALLASYNC", *pieces)
```
chayim, line 2548: @dvora-h any opinion here? I think this would be the first time we did an overload like this - meaning it's now unclear what command is called from the function name. Maybe users would prefer a tfcall_async rather than a boolean.

If so, IMHO we should have both functions call an _tfcall that takes all of the arguments and handles the validation. WDYT?

@uglide Opinion on the API?

chayim (changes requested):
@shacharPash @dvora-h @uglide any other comments?

redis/commands/core.py
dvora-h: Agree. I more like the approach of 2 seperate functions, we have it in other places in the library e.g. fcall and fcall_ro

shacharPash: @chayim We discussed this when I implemented these commands in [NRedisStack](https://github.com/redis/NRedisStack) and we decided to implement it this way because the two commands have exactly the same input and are pretty much the same command (one is `Invoke a function` and the other is `Invoke an async function (coroutine)`) so you can refer to the command type (`TFCALL` / `TFCALLASYNC`) as an argument.
I can change it but that means I'll have to change it in [NRedisStack](https://github.com/redis/NRedisStack) as well (which would be a Breaking change).

shacharPash: I did 2 separate functions

chayim (approved):

PR #2863, opened 2023-07-25
RESP3 connection examples

chayim:
As part of adding support for RESP3 in 5.x we need to communicate to our users how to enable RESP 3. This PR contains three items:

1. A basic example for the README

2. Updates to the sync examples, to share a redis_url style connection, as well as a basic connection, with RESP3

3. An async example highlighting the protocol=3 usage with async.

While here, I modified the async examples to include a URL based connection example, as one was missing


chayim:
needs #2869 at least

dvora-h (approved):

PR #2864, opened 2023-07-25
Create dt_set.py

sav-norem:
as part of the effort to get tabbed examples in client languages on the redis.io website, this file contains the examples for the set datatype in python with the appropriate short codes to be pulled in


sav-norem:
@uglide just realized I never tagged you on this - here's sets

sav-norem:

PR #2870, opened 2023-08-02
async fixes, remove __del__ and other things

kristjanvalur:
### Pull Request check-list

### Description of change

`__del__` handlers are rarely needed.  in `asyncio` they should not be used to perform IO cleanup, only emit resource usage
warnings.  Code executing at `__del__` time can be very fragile and is best avoided.
This PR fixes those warning emitters present, making them more "shutdown" safe.

- It also removes a redundant `__del__` handler from the async parser.  There is no need to explicitly clear member variables.
- It robustifies the callback system for `PubSub`, eliminating the need for a `__del__` handler there.
- It adds a few fixes to the unittests, when run locally.

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -434,7 +434,9 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
     _DEL_MESSAGE = "Unclosed RedisCluster client"
 
     def __del__(
-        self, _warn: Any = warnings.warn, _grl: Any = asyncio.get_running_loop
+        self,
+        _warn: Any = warnings.warn,
+        _grl: Any = asyncio.get_running_loop,
     ) -> None:
         if hasattr(self, "_initialize") and not self._initialize:
             _warn(f"{self._DEL_MESSAGE} {self!r}", ResourceWarning, source=self)
@@ -972,7 +974,9 @@ class ClusterNode:
     _DEL_MESSAGE = "Unclosed ClusterNode object"
 
     def __del__(
-        self, _warn: Any = warnings.warn, _grl: Any = asyncio.get_running_loop
+        self,
+        _warn: Any = warnings.warn,
+        _grl: Any = asyncio.get_running_loop,
     ) -> None:
         for connection in self._connections:
             if connection.is_connected:
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -433,14 +433,16 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
 
     _DEL_MESSAGE = "Unclosed RedisCluster client"
 
-    def __del__(self) -> None:
+    def __del__(
+        self, _warn: Any = warnings.warn, _grl: Any = asyncio.get_running_loop
+    ) -> None:
         if hasattr(self, "_initialize") and not self._initialize:
-            warnings.warn(f"{self._DEL_MESSAGE} {self!r}", ResourceWarning, source=self)
+            _warn(f"{self._DEL_MESSAGE} {self!r}", ResourceWarning, source=self)
             try:
                 context = {"client": self, "message": self._DEL_MESSAGE}
-                asyncio.get_running_loop().call_exception_handler(context)
+                _grl().call_exception_handler(context)
             except RuntimeError:
-                ...
+                pass
 
     async def on_connect(self, connection: Connection) -> None:
         await connection.on_connect()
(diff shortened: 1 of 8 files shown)


redis/asyncio/cluster.py
```diff
@@ -958,17 +960,18 @@ def __eq__(self, obj: Any) -> bool:
 
     _DEL_MESSAGE = "Unclosed ClusterNode object"
 
-    def __del__(self) -> None:
+    def __del__(
+        self, _warn: Any = warnings.warn, _grl: Any = asyncio.get_running_loop
```
chayim, line 964: Can we split these lines to too (how did black pass this!). I found it.. very difficult to find where __grl_ was defined - so IMHO there's an ergonomic issue here. I fully accept that this is a nit + preference.

redis/asyncio/connection.py
```diff
@@ -222,10 +222,15 @@ def is_connected(self):
...
 
     def register_connect_callback(self, callback):
-        self._connect_callbacks.append(weakref.WeakMethod(callback))
+        wm = weakref.WeakMethod(callback)
+        if wm not in self._connect_callbacks:
+            self._connect_callbacks.append(wm)
 
-    def clear_connect_callbacks(self):
-        self._connect_callbacks = []
+    def deregister_connect_callback(self, callback):
```
chayim, line 229: Given the single use, I'd argue that this isn't a breaking change - but as it's meant to be internal to the library and not reused, let's fix that here. WDYT about an underscore prefix? Yes, I could argue that register_connect_callback might have the same need :(

tests/test_asyncio/test_connection.py
```diff
@@ -254,9 +254,8 @@ async def do_close():
     async def do_read():
         return await conn.read_response()
 
-    reader = mock.AsyncMock()
-    writer = mock.AsyncMock()
-    writer.transport = mock.Mock()
+    reader = mock.Mock(spec=asyncio.StreamReader)
```
chayim, line 257: Teach we swami? Why is this pattern better than AsyncMock?

tests/test_asyncio/test_cwe_404.py
```diff
@@ -49,22 +49,22 @@ def set_delay(self, delay: float = 0.0):
...
 
     async def stop(self):
-        # clean up enough so that we can reuse the looper
+        # shutdown the server
         self.task.cancel()
         try:
             await self.task
         except asyncio.CancelledError:
             pass
+        await self.server.wait_closed()
```
chayim, line 65: definitely a better idea

chayim (commented):
@kristjanvalur I have some questions in the PR - I think they weren't sent because it was draft? Mind checking -I see them

tests/test_asyncio/test_connection.py
kristjanvalur: I think we can keep the AsyncMock the key here is the "spec".    When a spec is used, they become synonymous.
The use of 'spec' is essential when mocking classes with some methods sync, ad some async, since it will create the 
correct mock methods.

redis/asyncio/connection.py
kristjanvalur: fair point

redis/asyncio/cluster.py
kristjanvalur: sure.   This is a pattern seen in the standard lib, the idea is to capture some methods into the function so that they are available even during module teardown.  Pretty ugly.

kristjanvalur: I can't split the args, black formats it straight back to one line :(

tests/test_asyncio/test_connection.py
kristjanvalur: Oh, this went away when I rebased.  But yes, when using "spec", to mock a class instance, it will create async/sync methods as appropritate.  Here is the salient text:

> Setting the spec of a [Mock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock), [MagicMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock), or [AsyncMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock) to a class with asynchronous and synchronous functions will automatically detect the synchronous functions and set them as [MagicMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock) (if the parent mock is [AsyncMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock) or [MagicMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock)) or [Mock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock) (if the parent mock is [Mock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock)). All asynchronous functions will be [AsyncMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock).

redis/asyncio/connection.py
kristjanvalur: Done, also for non-async to keep consistency.

redis/asyncio/cluster.py
Fogapod: you need to add comma after last arg to make black put them on separate lines. same for list, tuples etc. final comma splits line

dvora-h (approved):

dvora-h:
@kristjanvalur Conflicts here too...

dvora-h:
@kristjanvalur The CI failed

kristjanvalur:
TestPubSubAutoReconnect , I guess this test could do with a review, I think there is a race condition there, causing sporadic failure.

dvora-h:
@kristjanvalur conflicts again...

PR #2872, opened 2023-08-03
RESP 3 feature documentation

chayim:
As part of the 5.0 release, we need to add some more documentation surrounding RESP 3. This pull requests introduces more connection examples, and fixes some broken ones. Along the way a base example for sharded pubsub and push notifications are included.

Further, in order to make it possible to catch some of our sillier documentation mistakes, such as spelling - I've included CI for spellchecking our documentation.


docs/advanced_features.rst
```diff
@@ -421,14 +421,32 @@ supported:
...
+The following, is a simplified example:
+
+.. code:: python
+
+    >>> from redis.cluster import RedisCluster, ClusterNode
+    >>> r = RedisCluster(startup_nodes=[ClusterNode('localhost', 6379), ClusterNode('localhost', 6380)])
+    >>> p = r.pubsub()
+    >>> p.ssubscribe('foo')
+    >>> # assume someone sends a message along the channel via a publish
+    >>> message = p.get_sharded_message()
```
dvora-h, line 440: I think it will be good to add also an example of hoe to use `get_sharded_message` on specific node and to explain what is the difference between the options.

vladvildanov (approved):

PR #2882, opened 2023-08-08
SETINFO on sync

chayim:
Adding SETINFO to the sync client, and pulling in changes from the 5.x release for async. Better approach


chayim:
Abandoning, 5.x only.

PR #2898, opened 2023-08-17
Provide aclose() / close() for classes requiring lifetime management

kristjanvalur:
### Pull Request check-list

### Description of change

The canonical name for an asynchronous `close()` method in async python is `aclose()`.  Providing this method allows support for standard programming patterns, including the use of the `aclosing` context manager.  Conversely, a `close()` method should be _synchronous_, for the same reason.

This PR:

- renames `close()` to  `aclose()` method to several async classes, keeping the old `close()` for backwards compatibility.
- renames a `reset()` to `aclose()`, keeping `close()` and `reset()` for compatibility in the `asyncio.PubSub` class`.
- Adds an `aclose()` method to the async `Pipeline` class, and a `close()` method to the `Pipeline` class.
- Adds an `aclose()` method to the async `ConnectionPool` class and the `close()` method to the sync `ConnectionPool`.

Having `close()` / `aclose()` allows the use of the `closing` / `aclosing` context managers.  In particular, they can now be used to manage a `ConnectionPool` instance.

**Note:** We _keep_ the old asynchronous `close()` methods as alternatives, for backwards compatibility.  Maybe we should actively deprecate them?

### DIFF
--- a/docs/examples/asyncio_examples.ipynb
+++ b/docs/examples/asyncio_examples.ipynb
@@ -15,7 +15,7 @@
     "\n",
     "## Connecting and Disconnecting\n",
     "\n",
-    "Utilizing asyncio Redis requires an explicit disconnect of the connection since there is no asyncio deconstructor magic method. By default, a connection pool is created on `redis.Redis()` and attached to this `Redis` instance. The connection pool closes automatically on the call to `Redis.close` which disconnects all connections."
+    "Utilizing asyncio Redis requires an explicit disconnect of the connection since there is no asyncio deconstructor magic method. By default, a connection pool is created on `redis.Redis()` and attached to this `Redis` instance. The connection pool closes automatically on the call to `Redis.aclose` which disconnects all connections."
    ]
   },
   {
@@ -39,9 +39,9 @@
    "source": [
     "import redis.asyncio as redis\n",
     "\n",
-    "connection = redis.Redis()\n",
-    "print(f\"Ping successful: {await connection.ping()}\")\n",
-    "await connection.aclose()"
+    "client = redis.Redis()\n",
+    "print(f\"Ping successful: {await client.ping()}\")\n",
+    "await client.aclose()"
    ]
   },
   {
@@ -60,8 +60,8 @@
     "import redis.asyncio as redis\n",
     "\n",
     "pool = redis.ConnectionPool.from_url(\"redis://localhost\")\n",
-    "connection = redis.Redis.from_pool(pool)\n",
-    "await connection.close()"
+    "client = redis.Redis.from_pool(pool)\n",
+    "await client.close()"
    ]
   },
   {
@@ -91,11 +91,11 @@
     "import redis.asyncio as redis\n",
     "\n",
     "pool = redis.ConnectionPool.from_url(\"redis://localhost\")\n",
-    "connection1 = redis.Redis(connection_pool=pool)\n",
-    "connection2 = redis.Redis(connection_pool=pool)\n",
-    "await connection1.aclose()\n",
-    "await connection2.aclose()\n",
-    "await pool.disconnect()"
+    "client1 = redis.Redis(connection_pool=pool)\n",
+    "client2 = redis.Redis(connection_pool=pool)\n",
+    "await client1.aclose()\n",
+    "await client2.aclose()\n",
+    "await pool.aclose()"
    ]
   },
   {
@@ -113,9 +113,9 @@
    "source": [
     "import redis.asyncio as redis\n",
     "\n",
-    "connection = redis.Redis(protocol=3)\n",
-    "await connection.aclose()\n",
-    "await connection.ping()"
+    "client = redis.Redis(protocol=3)\n",
+    "await client.aclose()\n",
+    "await client.ping()"
    ]
   },
   {
--- a/docs/examples/asyncio_examples.ipynb
+++ b/docs/examples/asyncio_examples.ipynb
@@ -41,7 +41,7 @@
     "\n",
     "connection = redis.Redis()\n",
     "print(f\"Ping successful: {await connection.ping()}\")\n",
-    "await connection.close()"
+    "await connection.aclose()"
    ]
   },
   {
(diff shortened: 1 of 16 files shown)


chayim:
@kristjanvalur You mentioned deprecating in your PR (and thank you). I'm 100% on board for marking as deprecated the old functions - that way that can be safely removed in _future version unknown_. Can you?

I might argue that we (@dvora-h) should define in our CONTRIBUTING.md or elsewhere how and when we'll deprecate an interface

docs/examples/asyncio_examples.ipynb
```diff
@@ -41,7 +41,7 @@
     "\n",
     "connection = redis.Redis()\n",
     "print(f\"Ping successful: {await connection.ping()}\")\n",
-    "await connection.close()"
+    "await connection.aclose()"
```
chayim, line 44: Thank you for taking the time to update the examples too!

redis/asyncio/client.py
```diff
@@ -516,6 +515,12 @@ async def close(self, close_connection_pool: Optional[bool] = None) -> None:
         ):
             await self.connection_pool.disconnect()
 
+    async def close(self, close_connection_pool: Optional[bool] = None) -> None:
```
chayim, line 518: i.e deprecating here - as yo mentioned

redis/asyncio/client.py
```diff
@@ -723,13 +728,18 @@ async def __aenter__(self):
...
 
     async def __aexit__(self, exc_type, exc_value, traceback):
-        await self.reset()
+        await self.aclose()
 
     def __del__(self):
         if self.connection:
             self.connection.clear_connect_callbacks()
 
-    async def reset(self):
```
chayim, line 732: This now removes a public method named reset - something we've shied away from doing. WDYT about keeping it defined, and also marking as deprecated?  At the very least we should mark as breakingchange and then not merge to 5.1 otherwise.

redis/asyncio/client.py
```diff
@@ -741,13 +751,13 @@ async def reset(self):
...
             self.pending_unsubscribe_patterns = set()
 
-    def close(self) -> Awaitable[NoReturn]:
-        # In case a connection property does not yet exist
-        # (due to a crash earlier in the Redis() constructor), return
-        # immediately as there is nothing to clean-up.
-        if not hasattr(self, "connection"):
-            return
-        return self.reset()
+    async def close(self) -> None:
```
chayim, line 754: Good catch. This never should have returned Awaitable[NoReturn].

For the curious readers [see this](https://mypy.readthedocs.io/en/latest/more_types.html). Note that only a function that unconditionally raises an exception should do so.

redis/asyncio/client.py
```diff
@@ -741,13 +751,13 @@ async def reset(self):
...
+        """Alias for aclose(), for backwards compatibility"""
+        await self.aclose()
+
+    async def reset(self) -> None:
```
chayim, line 758: IMHO the None in implied. It's a nit, do you feel differently?

redis/asyncio/client.py
```diff
@@ -741,13 +751,13 @@ async def reset(self):
...
+        """alias for aclose(), for backwards compatibility"""
```
chayim, line 759: s/alias/Alias

tests/test_asyncio/test_pubsub.py
```diff
@@ -217,6 +216,44 @@ async def test_subscribe_property_with_patterns(self, pubsub):
...
+
+    async def test_context_manager(self, r: redis.Redis):
+        p = r.pubsub()
+        async with p:
+            assert p.subscribed is False
+            await p.subscribe("foo")
+            assert p.subscribed is True
+        assert p.subscribed is False
+
+    async def test_close_is_aclose(self, r: redis.Redis):
```
chayim, line 235: awesome, thank you :heart:

tests/test_asyncio/test_pubsub.py
```diff
@@ -217,6 +216,44 @@ async def test_subscribe_property_with_patterns(self, pubsub):
...
+        Test backwards compatible close method
+        """
+        p = r.pubsub()
+        assert p.subscribed is False
+        await p.subscribe("foo")
+        assert p.subscribed is True
+        await p.close()
+        assert p.subscribed is False
+
+    async def test_reset_is_aclose(self, r: redis.Redis):
```
chayim, line 246: and again

chayim:
@dvora-h WDYT? Want to take it from here?

kristjanvalur:
> I'm 100% on board for marking as deprecated the old functions - that way that can be safely removed in _future version unknown_. Can you?

Sure, but what is the proper way do do that?

redis/asyncio/client.py
kristjanvalur: sure, we can keep it around.  If you can provide me with a guideline on how to mark methods and attributes as deprecated, and how we can schedule those for future removal, that would be awesome.

dvora-h:
@kristjanvalur Yes, we have a [deprecated_function](https://github.com/redis/redis-py/blob/ba186d210f4f1e5d16fb27a1ee0afcb930f16a8d/redis/utils.py#L113) decorator. you can add it to the function.

redis/asyncio/client.py
dvora-h: @kristjanvalur  see my comment bellow about how to mark function as deprecated. for attributes, you can't use the decorator, so you just need to import warnings and warn deprecated if the user try to use this attribute, like we did [here](https://github.com/redis/redis-py/blob/ba186d210f4f1e5d16fb27a1ee0afcb930f16a8d/redis/client.py#L203).

dvora-h (approved):

dvora-h:
@kristjanvalur conflicts... 🙂

dvora-h:
@kristjanvalur The CI failed

kristjanvalur:
Bummer. Will investigate in a bit

kristjanvalur:
Oops, looks like I pushed the wrong branch today!  Fixing.

PR #2902, opened 2023-08-20
Fix redis 7.2.0 tests

dvora-h:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -267,6 +267,7 @@ class AbstractRedisCluster:
                 "READONLY",
                 "CLUSTER INFO",
                 "CLUSTER MEET",
+                "CLUSTER MYSHARDID"
                 "CLUSTER NODES",
                 "CLUSTER REPLICAS",
                 "CLUSTER RESET",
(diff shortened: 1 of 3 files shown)


vladvildanov (approved):

redis/cluster.py
```diff
@@ -267,6 +267,7 @@ class AbstractRedisCluster:
                 "READONLY",
                 "CLUSTER INFO",
                 "CLUSTER MEET",
+                "CLUSTER MYSHARDID",
```
sazzad16, line 270: Is this command implemented?

sazzad16 (commented):
About removed tests, is possible just to ignore those tests? I.e. with tag? So that when we have environment with WAITAOF support, we can enable it.

redis/cluster.py
dvora-h: Yes
https://github.com/redis/redis-py/blob/28cc65c18cc4fb37ef14497c963eb181dba8d25d/redis/commands/cluster.py#L638

PR #2904, opened 2023-08-21
Fix: unsubscribe/punsubscribe should be async functions in async client

newspacer:
`execute_command()` is an `async` function; since `unsubscribe()` and `punsubscribe()` call it, they should be `async` functions, and should `await` `execute_command()`.

---

### Pull Request check-list

### Description of change

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -899,7 +899,7 @@ class PubSub:
         self.pending_unsubscribe_patterns.difference_update(new_patterns)
         return ret_val
 
-    async def punsubscribe(self, *args: ChannelT) -> Awaitable:
+    async def punsubscribe(self, *args: ChannelT) -> int:
         """
         Unsubscribe from the supplied patterns. If empty, unsubscribe from
         all patterns.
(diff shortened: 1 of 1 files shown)


kristjanvalur:
The pattern you are seeing is a common, but ugly one.  Basically this is correct:

```python
# innermost awaitable
async def foo() -> int:
    return 1
    
def bar() -> Awaitable[int]:
    # do stuff, then call foo
    return foo()
    
async def doit() -> int:
    value = await bar()
    return value + 1  # cannot use "tail recursion" here.
```

Doing this, forwarding the awaitable in a tail-recursion like manner, is fairly common.  I guess people think of it as performing an optimization.  But it is confusing and one of the problems is that it produces incorrect call stacks when debugging.  I'm also fairly certain that chaining multiple "await" statements in recursion is not a performance problem in modern python.

newspacer:
Thanks for the feedback.  I have updated the type hints as pointed out.  Indeed I think it's confusing to return the awaitable, especially since that pattern doesn't match the `(p)subscribe()` counterpart functions.  However, I can see that making this change therefore would break some client code, so it's understandable if this PR is eventually not merged.

PR #2909, opened 2023-08-23
Add aliases for nonatomic versions of mget/mset

daveisfera:
### Pull Request check-list

- [ ] Does `$ tox` pass with this change (including linting)? `Not sure how to run this but I believe it's ok`
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)? `Not sure how to run these`
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? `Don't believe there's a needed change here`
- [ ] Is there an example added to the examples folder (if applicable)? `Existing example should suffice`

### Description of change

RedisCluster provides functions for handling when multiple keys are used with separate key slots, so alias the single server versions so code can use them without caring which type of server is running

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -54,6 +54,7 @@
     * Fix for Unhandled exception related to self.host with unix socket (#2496)
     * Improve error output for master discovery 
     * Make `ClusterCommandsProtocol` an actual Protocol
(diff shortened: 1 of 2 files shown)


daveisfera:
I believe that this is still an outstanding issue

Issue #2910, opened 2023-08-24
Asyncio and fork() does not mix


kristjanvalur:
The asyncio `ConnectionPool`  contains a lot of code, inherited from the synchronous part, which tries to play safe with sockets after a `fork()` call. 

However, python asyncio and fork does not play well together at all:
https://github.com/python/cpython/issues/66197

Basically, one should never fork a process with a running event loop.  As such, the code comparing PIDs is not helping with anything for async.  

I propose that these rudiments be removed from the async code for clarity and simplicity.

PR #2911, opened 2023-08-24
Remove process-id checks from asyncio.  Asyncio and fork() does not mix.

kristjanvalur:
### Pull Request check-list

### Description of change

Remove `PID` checks from `asyncio` connections, since `asyncio` and `fork()` does not mix.  See  issue #2910 

### Details:

Python does not support `fork()` on a process with a running `asyncio` event loop.  Therefore supporting a connection pool of connection from a different process is not necessary.

Similarly, connection pools don't need to deal with different threads either.

We remove the checks for process ids, intended to ignore _inherited_ file descriptors  (even though these should probably have been closed on fork anyway).

We remove _lenience_ when returning a connection to a pool, it should __always__ be returned to the pool it originated from.  Not doing that would be a resource leak.

We simplify `BlockingConnectionPool` since it does not need to do any multi-threading.  It now just adds a simple wait mechanism on top of the `ConnectionPool`, reducing code and overhead.  The `QueueClass` argument is now ignored.  It always was a weird implementation leak anyway

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1068,13 +1068,9 @@ class ConnectionPool:
     async def release(self, connection: AbstractConnection):
         """Releases the connection back to the pool"""
         async with self._lock:
-            try:
-                self._in_use_connections.remove(connection)
-            except KeyError:
-                # Gracefully fail when a connection is returned to this pool
-                # that the pool doesn't actually own
-                pass
-
+            # Connections should always be returned to the correct pool,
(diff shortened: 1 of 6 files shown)


chayim:
@kristjanvalur Just checking in - is this one something you wanted reviewed?

@dvora-h see the test failures. I'm surprised that some of these (cluster) ran.

kristjanvalur:
This was all clean not so long ago, let me check it out

chayim (approved):

PR #2913, opened 2023-08-24
Add `Redis.from_pool()` class method, for explicitly owning and closing a ConnectionPool

### LINKED ISSUE #2901: `auto_close_connection_pool` ignored
This defect pertains to issues discovered with #2831 

In async mode, it is important that a `Redis` object which holds its own pool, can be told to close it when the `Redis` object
is closed, to free the user from such hassles.  Examples in the code include
``` python
client = Redis(connection_pool=BlockingConnectionPool())
```
when this client is closed, the internal connection pool remains open, i.e. its connections are not shut down.

The `Redis` class includes a `auto_close_connection_pool` argument.  However, this is _ignored_ when a connection pool
is provided to the constructor.  Presumably the thinking is that the caller will manage that pool himself.
But that makes the above kind of pattern, also used within the code, unusable.

```python
client = Redis(connection_pool=BlockingConnectionPool(), auto_close_connection_pool=True)
assert client.auto_close_connection_pool=False
```

Instead, it is necessary to have code like this, which is very poor practice:
```python
client = Redis(connection_pool=BlockingConnectionPool(), auto_close_connection_pool=True)
client.auto_close_connection_pool=True
```

I propose, that we either:

- Always honour the `auto_close_connection_pool` argument.  
- add a `own_connection_pool:ConnectionPool` argument, to be used instead of `connection_pool`.
- rename `auto_close_connection_pool` to `own_pool`, or something less cumbersome.
- add a `construct_connection_pool` argument which takes a callable which returns a connection pool, one which Redis considers its own.

At any rate, it should be easy to create a connection pool, and hand it to the `Redis` constructor for it to "own" without
having to fudge an internal property afterwards.

In addition, I propose that the same mechanics be added to the non-`async` parts of the library.  Currently the `redis.Redis` objects do not close connection pools, instead the library relies on the prompt execution of `__del__()` methods, or explicit `ConnectionPool.disconnect()` calls.  
Relying on garbage collection to free resources is considered bad programming practice in Python, where resource management should in general be more explicit.  It is particularly frowned upon for async code, since executing asynchronous code there is not a good idea(*).  the synchronous code should receive whatever solution we choose for the above as well, so that pools owned by a `Redis` client object are also disconnected when those Redis objects are closed.

> (*)  `__del__()` methods are not async.  While it is possible to use something like `asyncio.run` within such a function, it will __block__ the rest of the program.  Also, `__del__()` methods are often invoked during program cleanup when it is simply not possible to do anything and trying to do so will produce errors and confusion.

### Update

In pr #2913 I have updated a `Redis.from_pool()` method to use when handing over a `ConnectionPool` to be _owned_ by the `Redis` instance.  I think this is a much nicer way than to heap on more constructor arguments to `Redis`

kristjanvalur:
### Pull Request check-list

### Description of change

Traditionally, synchronous `redis-py` has relied on `__del__` to disconnect any `ConnectionPool` owned by a `Redis` object.
Explicitly closing those has been possible, via `client.connection_pool.disconnect()`, but this does not work well with
using context managers and other approaches.  Relying on `__del__()` to clean up sockets is not considered good practice since it relies on the peculiarities of the particular implementation, garbage collection and such things.

In async redis, a mechanism was added to make an `asyncio.Redis` object _own_ a connection and close it, to provide
better connection over connection lifetime.  This was done here:
https://github.com/aio-libs-abandoned/aioredis-py/pull/1256
This mechanism carried over when that library was added to redis-py, and we increasinly rely on it to properly clean up sockets used in asynchronous mode.  However, the mechanism added was cumbersome to use and overly verbose.  In particular, it made it awkward to create a custom `ConnectionPool` instance, hand it to a new `Redis` object, and expect it to close it automatically.

This PR, aims to:

1. fix #2901, adding a new `from_pool()` class method to use instead of the `auto_close_connection_pool` constructor argument.
2. apply the same semantics to the synchronous Redis.
3. Discourage the use of the `auto_close_connection_pool` argument.

The following now applies for both _synchronous_ and _asynchronous_ redis:

- A `Redis` object created without providing a `ConnectionPool` instance, e.g. via `Redis()` or `Redis.from_url()`, will now automatically call `connection_pool.disconnect()` when `call()` is closed.
- A `Redis` object created `Redis.from_pool()`  similarly _takes ownership_ of the pool, and will disconnect the pool.
- As before, if the `connection_pool` argument is used by `Redis.__init__()` this is __not__ done, instead it is assumed the caller
will close the pool.
- A `Sentinel` object, will use a `from_pool()` class method constructing a `Redis` object to return from the `master_for()` and `slave_for()` methods.
- An `auto_close_connection_pool` `Redis.__init__()` argument is not provided for `redis.Redis`, but kept in place for backwards compatibility for `redis.asyncio.Redis`

The above behaviour is controlled by the `auto_close_connection_pool` property of `Redis`.

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -159,10 +159,28 @@ class Redis(
 
         """
         connection_pool = ConnectionPool.from_url(url, **kwargs)
-        return cls(
-            from_pool=connection_pool,
+        client = cls(
+            connection_pool=connection_pool,
             single_connection_client=single_connection_client,
         )
+        client.auto_close_connection_pool = True
+        return client
+
+    @classmethod
+    def from_pool(
+        cls: Type["Redis"],
+        connection_pool: ConnectionPool,
+    ) -> "Redis":
+        """
+        Return a Redis client from the given connection pool.
+        The Redis client will take ownership of the connection pool and
+        close it when the Redis client is closed.
+        """
+        client = cls(
+            connection_pool=connection_pool,
+        )
+        client.auto_close_connection_pool = True
+        return client
 
     def __init__(
         self,
@@ -176,7 +194,6 @@ class Redis(
         socket_keepalive: Optional[bool] = None,
         socket_keepalive_options: Optional[Mapping[int, Union[int, bytes]]] = None,
         connection_pool: Optional[ConnectionPool] = None,
-        from_pool: Optional[ConnectionPool] = None,
         unix_socket_path: Optional[str] = None,
         encoding: str = "utf-8",
         encoding_errors: str = "strict",
@@ -213,7 +230,7 @@ class Redis(
         """
         kwargs: Dict[str, Any]
 
-        if not connection_pool and not from_pool:
+        if not connection_pool:
             # Create internal connection pool, expected to be closed by Redis instance
             if not retry_on_error:
                 retry_on_error = []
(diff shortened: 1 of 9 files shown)


redis/asyncio/client.py
```diff
@@ -114,7 +114,6 @@ def from_url(
         cls,
         url: str,
         single_connection_client: bool = False,
-        auto_close_connection_pool: bool = True,
```
chayim, line 117: This breaks the interface...

chayim:
Here, my concern is primarly around breaking the connection interface. I realize that async is in flux, and frankly many things will be - but this interface is public and exposed, and not internal. Maybe there's another way - without breaking the user contract?

kristjanvalur:
I understand.  I think we can leave this attribute in place.  I'm sure you agree that it feels cludgy, and long.  
Do we have a standard schedule and mechanism to deprecate functionality / arguments in redis-py?

chayim:
I definitely agree is kludgey... 

I generally dislike these interface (sync and asycn.. heck even cluster). But we should break accordingly in the future.

dvora-h (approved):

PR #2919, opened 2023-08-29
Cluster determine slot command name need upper

yangbodong22011:
The judgment of the name is all uppercase, for example:
https://github.com/redis/redis-py/blob/master/redis/cluster.py#L970
```
if command in ("EVAL", "EVALSHA"):
```

Repro code
```
#!/usr/bin/env python
#-*- coding: utf-8 -*-
from redis.cluster import RedisCluster
host = '127.0.0.1'
port = 30001
user = 'default'
pwd = ""
rc = RedisCluster(host=host, port=port, username=user, password=pwd)
print(rc.execute_command('EVAL', 'return', '0', 'hello'))
```

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -967,11 +967,11 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands):
         # redis server to parse the keys. Besides, there is a bug in redis<7.0
         # where `self._get_command_keys()` fails anyway. So, we special case
         # EVAL/EVALSHA.
-        if command in ("EVAL", "EVALSHA"):
+        if command.upper() in ("EVAL", "EVALSHA"):
             # command syntax: EVAL "script body" num_keys ...
             if len(args) <= 2:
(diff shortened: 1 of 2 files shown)


redis/cluster.py
```diff
@@ -956,7 +956,7 @@ def determine_slot(self, *args):
             determine what slots to map the command to; or, if the keys don't
             all map to the same key slot.
         """
-        command = args[0]
+        command = args[0].upper()
```
dvora-h, line 959: It's better to do this in this line:
`if command in ("EVAL", "EVALSHA"):`
because in other cases we call `self._get_command_keys(*args)`, that needs the lower command.

yangbodong22011: you mean `if command.upper() in ("EVAL", "EVALSHA"):` ?

dvora-h: @yangbodong22011 Yes

yangbodong22011: done

dvora-h (approved):

PR #2928, opened 2023-09-04
Partial clean up of Python 3.7 compatibility

akx:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `$ tox` pass with this change (including linting)?
  - No, Tox is no longer a thing 
- [x] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
  - See https://github.com/akx/redis-py/pull/2 
- [x] Is the new or changed code fully tested?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Is there an example added to the examples folder (if applicable)?
- [ ] Was the change added to CHANGES file?
  - No, since CHANGES is unspeakably out-of-date at this point (see #2833).
 
### Description of change

Since 5.0.0 was the last version to support Python 3.7, this PR starts cleaning up some now-unnecessary compatibility code.

PR #2388, opened 2022-09-19
Switch to PEP 517 packaging using hatchling

akx:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [ ] Does `$ tox` pass with this change (including linting)?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [x] Is the new or changed code fully tested?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Is there an example added to the examples folder (if applicable)?
  - not applicable 
- [ ] Was the change added to CHANGES file?
  - not yet
 
### Description of change

This PR shifts redis-py from legacy `setup.py` packaging to [PEP 517](https://peps.python.org/pep-0517/) compliant packaging [using `hatchling` from PyPA](https://hatch.pypa.io/latest/).

Essentially, this does what was discussed in #1658 but with the increasingly standard `hatchling` tool instead of poetry's idiosyncracies.

Refs #1316 (which initially did this, with tools from that age)
Refs #1649 (which reverted that for no good reason)

I wasn't able to test all of these changes yet, but `python -m build` does create a working wheel with all modules included.

Issue #1658, opened 2021-10-28
Discussion: Migrating to setup.py to poetry


chayim:
As redis-py 4.0.0 now targets python versions > 3.6 we can discuss new ways of packaging. Poetry makes dependnecy management for both development and the library itself simpler - reducing the need to manually track the dev_requirements.txt.  It easily integrates with tox via the [tox-poetry](https://pypi.org/project/tox-poetry/) plugin, and makes publishing packages to Pypi easier that twine.

I'm openly declaring that my bias is to move from setup.py to poetry, including some of the pain points that comes with it. I'd love to hear what the community thinks.  Please tag others so that they can weigh in if they so desire.

chayim:
I mean I have a tonne of experience with poetry, but I wanted to not bias.  Since you asked, here's what I like (in short):

- The pyproject.toml is PEP (508, 516, 517) compliant. This means that we can specify building the package, it's dependencies, pypi fields, etc in one place.
	- The outcome of this is that one can still pip install directly from the repo, and pypi
- Separation of project dependencies, optional dependencies, and developer dependencies
- Dependency resolution that can be python specific (though only one poetry.lock)
- Parallel dependency fetching
- Eliminates the pip freeze need / issue - where we track non-project dependencies, due to the nature of pip.
- Manages virtualenvs, and can also not (I choose not personally, but that's a local setting)
- Built-in package publishing

barshaul:
I've never worked with poetry before, but it sounds interesting. Will this migration have any downsides?

chayim:
The only issue is that it _can_ be more difficult to work on multiple versions of python in parallel - given the poetry.lock file, should versions need to be semantically different per python. However, redis-py has no actual dependencies (just dev deps, and an optional hiredis), so IMHO there isn't a realistic downside. 

@AvitalFineRedis I realize I didn't include you - what do you think?

ddevault:
As a distribution maintainer: please no. Poetry is great for workstations and awful for distributions.

chayim:
@ddevault Can you help me understand how - I'm all good with not, but I'd love to know how it's bad for distributions, since I've never experienced this. Thank you in advance!

ddevault:
[...]

PR #1649, opened 2021-10-26
Fixing the package to include commands

chayim:
Fixes #1645

### Pull Request check-list

_Please make sure to review and check all of these items:_

- [ ] Does `$ tox` pass with this change (including linting)?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

_Please provide a description of the change here._

PR #2930, opened 2023-09-04
Switch to PEP 517 packaging using hatchling

### LINKED ISSUE #3463: Remove dependency files and add dependencies to setup.py
To avoid issues like https://github.com/redis/redis-py/issues/3461 which are not picked up in CI but break production.

akx:
### Pull Request check-list

- [ ] Do tests pass?

### Description of change

* Refs #1316
* Refs #1649
* ♻️ Remake of #2388, a year later
*❗Contains dependency changes from #2928 
* Closes #3463

Quoting the description from #2388, a year later...

This PR shifts redis-py from legacy `setup.py` packaging to [PEP 517](https://peps.python.org/pep-0517/) compliant packaging [using `hatchling` from PyPA](https://hatch.pypa.io/latest/).

Essentially, this does what was discussed in #1658 but with the increasingly standard `hatchling` tool instead of poetry's idiosyncracies.

As with #2388, `python -m build` does create a working wheel with all modules included.


akx:
Rebased.

akx:
Rebased. Cc @gerzse @vladvildanov.

vladvildanov:
@akx Thanks for your contribution! I'll review this PR in near time

akx:
@vladvildanov Good to hear. Rebased once again.

redis/__init__.py
```diff
@@ -44,16 +42,9 @@ def int_or_str(value):
         return value
 
 
-try:
-    __version__ = metadata.version("redis")
-except metadata.PackageNotFoundError:
-    __version__ = "99.99.99"
-
+__version__ = "5.1.0b7"
```
vladvildanov, line 45: Should be changed to 5.1.1

vladvildanov (changes requested):

akx:
@vladvildanov Rebased.

akx:
Btw: the tests failing in

```
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_client_setinfo[single] - AssertionError: assert '5.3.0b5' == '5.2.1'
    
    - 5.2.1
    + 5.3.0b5
```

is _very_ curious – given the string "5.3.0b5" does not appear in the source here at all, it almost looks like the CI pipeline is installing an entirely different version of the package...

and indeed:

```
  Downloading redis-5.3.0b5-py3-none-any.whl (271 kB)
```

IOW, the CI pipeline seems to be confused about which libraries it's using...

tasks.py
```diff
@@ -97,4 +97,4 @@ def clean(c):
 @task
 def package(c):
     """Create the python packages"""
-    run("python setup.py sdist bdist_wheel")
```
petyaslavova, line 100: You need to add 'build' dependency to the dev_requirements.txt - its existence in Python is not guaranteed. For example, in my new virtual environment, it wasn't installed...

.github/actions/run-tests/action.yml
```diff
@@ -39,6 +39,8 @@ runs:
         pip install -U setuptools wheel
         pip install -r requirements.txt
```
petyaslavova, line 40: Do we still need to support the requirements.txt file? In the README.md the dependency installation section states that we should install the dependencies using the command "pip install -e ." . In this case the information in requirements.txt and pyproject.toml becomes redundant.

.github/actions/run-tests/action.yml
```diff
@@ -39,6 +39,8 @@ runs:
         pip install -U setuptools wheel
```
petyaslavova, line 39: wheel is already part of the dev_requirements dependencies, and since you are adding hatchling as build backend, probably setuptools is also not needed. Would you please check if this dependency installation can also be skipped?

pyproject.toml
```diff
@@ -0,0 +1,65 @@
...
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+    "Programming Language :: Python",
+]
+dependencies = [
+    'async-timeout>=4.0.3; python_full_version<"3.11.3"',
```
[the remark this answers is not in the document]
akx: Each dependency here is on a single line.

`; python_full_version<"3.11.3"` is a marker that specifies that this requirement should only be installed in an environment where this marker is fulfilled.

This is the same as it is in current `master`:

https://github.com/redis/redis-py/blob/d8676605ef05c786bb705257467a83f963cd1d04/setup.py#L40

See https://github.com/redis/redis-py/issues/3174 for discussion about why the marker is added.

tasks.py
akx: [`build`](https://pypi.org/project/build/) isn't the only PEP517 build frontend – one could use, for instance, `uv build` instead. But sure, I'll add that.

.github/actions/run-tests/action.yml
akx: It's unlikely that `wheel` is required either.

> **Historical note**
> This project used to contain the implementation of the setuptools bdist_wheel command, but as of setuptools v70.1, it no longer needs wheel installed for that to work. Thus, you should install this only if you intend to use the wheel command line tool!

– https://pypi.org/project/wheel/

pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+Code = "https://github.com/redis/redis-py"
+Documentation = "https://redis.readthedocs.io/en/latest/"
+Homepage = "https://github.com/redis/redis-py"
+"Issue tracker" = "https://github.com/redis/redis-py/issues"
+
+[tool.hatch.version]
+path = "redis/__init__.py"
+
+[tool.hatch.build.targets.sdist]
+include = [
```
petyaslavova, line 64: Can you please add tests, pytest.ini and dev_requirements.txt in the built package. These items exist in the current package.

akx: Sure.

* Moved [pytest configuration to pyproject.toml](https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml) so the INI file isn't needed.
* Added tests and dev_requirements to the sdist targets.

I compared the dists built from `master` and this PR and aside from some expected changes, they're identical.

Screenshot 2025-02-25 at 10 48 44

petyaslavova (approved):
LGTM

vladvildanov (approved):

akx:
Thanks for getting this finally merged!

1 year, 5 months 3 weeks – could easily be a personal record for PR perseverance for me.

vladvildanov:
@akx Thanks for your efforts, I'm so sorry that It took so long from our side

Issue #2839, opened 2023-07-10
query object is not outputing the `TIMEOUT` argument in case of 0 ( 0 means unlimited for RediSearch TIMEOUT parameter )


filipecosta90:
**Version**: 4.6.0 or any above 4.4.1 given the timeout parameter was introduced there (https://github.com/redis/redis-py/releases/tag/v4.4.1)

**Description**: I was checking why we where not seeing the timeout parameter on a RediSearch query monitor with timeout(0) which means unlimited execution time on RediSearch.
Given the following:

```
q = (
            Query(
                f"{prefilter_condition}=>[KNN $K @vector $vec_param EF_RUNTIME $EF AS vector_score]"
            )
            .sort_by("vector_score", asc=False)
            .paging(0, top)
            .return_fields("vector_score")
            .dialect(2)
            .timeout(0)
        )
```

notice that on the output of q.get_args() we're missing the `TIMEOUT 0`:
```
['*=>[KNN $K @vector $vec_param EF_RUNTIME $EF AS vector_score]', 'RETURN', 1, 'vector_score', 'SORTBY', 'vector_score', 'DESC', 'DIALECT', 2, 'LIMIT', 0, 100]
```
checking the code we see that the integer 0 will be interpreted as False on the condition https://github.com/redis/redis-py/blob/master/redis/commands/search/query.py#L197
```
if self._timeout:
   args += ["TIMEOUT", self._timeout]
```

PR #2934, opened 2023-09-12
Support timeout = 0 in search query

meiravgri:
In RediSearch query timeout = 0 means unlimited timeout. In the current implementation, the query timeout is only updated `if timeout` which in case of 0, translates to false. Since the default timeout of the `query` class is None, replacing the condition with `if self._timeout is not None` will also work for 0. If the parameter is not a positive integer, redis server will raise an exception.

related issue: #2839 
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/commands/search/query.py
+++ b/redis/commands/search/query.py
@@ -135,7 +135,7 @@ class Query:
     def timeout(self, timeout):
         """overrides the timeout parameter of the module"""
         if isinstance(timeout, int) and timeout >= 0:
-            self._timeout = timeou
+            self._timeout = timeout
         else:
             raise AttributeError("TIMEOUT requires a non negative integer.")
         return self
--- a/redis/commands/search/query.py
+++ b/redis/commands/search/query.py
@@ -194,8 +194,10 @@ class Query:
             args += self._ids
         if self._slop >= 0:
             args += ["SLOP", self._slop]
-        if self._timeout is not None:
+        if isinstance(self._timeout, int) and self._timeout >= 0:
             args += ["TIMEOUT", self._timeout]
+        else:
+            raise AttributeError("TIMEOUT requires a non negative integer.")
         if self._in_order:
             args.append("INORDER")
         if self._return_fields:
--- a/redis/commands/search/query.py
+++ b/redis/commands/search/query.py
@@ -194,7 +194,7 @@ class Query:
             args += self._ids
         if self._slop >= 0:
             args += ["SLOP", self._slop]
-        if self._timeout:
+        if self._timeout is not None:
             args += ["TIMEOUT", self._timeout]
         if self._in_order:
             args.append("INORDER")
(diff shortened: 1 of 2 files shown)


dvora-h (approved):

redis/commands/search/query.py
```diff
@@ -194,7 +194,7 @@ def _get_args_tags(self):
             args += self._ids
         if self._slop >= 0:
             args += ["SLOP", self._slop]
-        if self._timeout:
+        if self._timeout is not None:
```
chayim, line 197: If we need positive timeouts... let's enforce @meiravgri

meiravgri: Done and also updates the test

GuyAv46: We could have used string until this change, no? Do we want to handle strings here?

redis/commands/search/query.py
```diff
@@ -134,7 +134,10 @@ def slop(self, slop):
 
     def timeout(self, timeout):
         """overrides the timeout parameter of the module"""
-        self._timeout = timeout
+        if isinstance(timeout, int) and timeout >= 0:
+            self._timeout = timeou
```
GuyAv46, line 138:
```suggestion
            self._timeout = timeout
```

redis/commands/search/query.py
meiravgri: Can’t we just rely on the server to validate the input (like it was up until today)?
becuase today we support both non negative int & **strings** of non negative ints
It makes the validation in the client more **complicated** and **slow**
Also, we don’t need to **make sure we are aligned** with redis on the definition of valid input (lets say we will want to change unlimited timeout to -1 instead of 0 (in redis), it can take time and pain until we will notice the client needs to be fixed too)

GuyAv46: I agree. Also IMHO in the cases where these errors are well-handled in the server, we should assume that most requests won't have syntax errors and the flow should be optimized for them. In the worst case, the syntax error will be caught anyway on the server side and will raise an error containing the problem.

GuyAv46: And regardless, I think that this PR was meant to fix a bug. Let's make more significant changes separately.

chayim: We have varying states in the library where we do this - and in general it would be great if a low-level library provided this support. I'm of the opinion that if there's a requirement for a positive integer, let's give the user a useful exception, as [we do for ACL](https://github.com/redis/redis-py/blob/1596ac6a0bbc1e00e57c6a48e255a8f917498e39/redis/commands/core.py#L347C10-L347C10).

The clients need to be better than this.

meiravgri: @dvora-h What do you think ? :)

dvora-h: I'm with @meiravgri and @GuyAv46.
@chayim In most cases we do not validate the input, it's a different discussion if we want to change it or not and out of the scope here.

And yes, @meiravgri it's important to support also string inputs so you need to change/delete (IMO delete) the validation check

GuyAv46 (approved):
LGTM

dvora-h (approved):

dvora-h:
closes #2839

PR #2940, opened 2023-09-13
create the JSON python examples

sav-norem:
adding the python code to support the tabbed examples for the JSON data type page


sav-norem:
@chayim if ya could take a look that would be great!

PR #2947, opened 2023-09-16
Add fake redis server to properly test connection handshake

kristjanvalur:
### Pull Request check-list

### Description of change

This PR provides a transport agnostic RESP parser and encoder for the unit tests.
It also provides a simple, transport agnostic, RESP server.  This allows us to share code and logic
between sync and async code.
A parametrized test is added to test the connection handshake for different client settings against different fake REDIS servers,
both sync and async.

- the RESP parser / encoder is simple, assumes utf8 encoding, and always does decoding (from bytes to string.)
-  Decoding uses the 'surrogateescape' convention, so that if binary data was intended (a blob) it can always be recovered by re-encoding.
- Decoding supports most RESP 3 data types
- Encoding assumes utf8, and chooses RESP3 types if protocol 3 is enabled, when appropriate.
- RESP server receives resp command arrays and has handlers to respond.  Can be expanded to perform more extensive Redis emulation
- RESP server can be plugged into a sync or async request handler to simplify testing.

### DIFF
--- a/tests/resp.py
+++ b/tests/resp.py
@@ -1,0 +1,93 @@
+import itertools
+from types import NoneType
+from typing import Any, Optional
+
+
+class RespEncoder:
+    """
+    A class for simple RESP protocol encodign for unit tests
+    """
+
+    def __init__(self, protocol: int = 2, encoding: str = "utf-8") -> None:
+        self.protocol = protocol
+        self.encoding = encoding
+
+    def encode(self, data: Any, hint: Optional[str] = None) -> bytes:
+        if isinstance(data, dict):
+            if self.protocol > 2:
+                result = f"%{len(data)}\r\n".encode()
+                for key, val in data.items():
+                    result += self.encode(key) + self.encode(val)
+                return result
+            else:
+                # Automatically encode dicts as flattened key, value arrays
+                mylist = list(
+                    itertools.chain(*((key, val) for (key, val) in data.items()))
+                )
+                return self.encode(mylist)
+
+        elif isinstance(data, list):
+            result = f"*{len(data)}\r\n".encode()
+            for val in data:
+                result += self.encode(val)
+            return result
+
+        elif isinstance(data, set):
+            if self.protocol > 2:
+                result = f"~{len(data)}\r\n".encode()
+                for val in data:
+                    result += self.encode(val)
+                return result
+            else:
+                return self.encode(list(data))
+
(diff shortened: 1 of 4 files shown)


tests/resp.py
```diff
@@ -0,0 +1,532 @@
...
+        return f"Attribute({super().__repr__()})"
+
+
+class RespEncoder:
+    """
+    A class for simple RESP protocol encoder for unit tests
+    """
+
+    def __init__(
+        self, protocol: int = 2, encoding: str = "utf-8", errorhander="strict"
```
Fogapod, line 66: why no type hint for `errorhander` here?

tests/resp.py
```diff
@@ -0,0 +1,532 @@
...
+        return value
+
+    def get_unparsed(self) -> bytes:
+        return b"".join(self.consumed)
+
+    def close(self) -> None:
+        if self.generator is not None:
+            self.generator.close()
+            self.generator = None
+        del self.consumed[:]
```
Fogapod, line 390: maybe do `self.consumed.clear()` here to make it more readable? same for all other places

tests/resp.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def escape_crln(self, value: bytes) -> bytes:
+        """remove any cr or nl from the value"""
+        return value.replace(b"\r", b"\\r").replace(b"\n", b"\\n")
+
+    def encode(self, data: Any, hint: Optional[str] = None) -> bytes:
+        if isinstance(data, dict):
+            if self.protocol > 2:
+                code = "|" if isinstance(data, Attribute) else "%"
+                result = f"{code}{len(data)}\r\n".encode()
```
Fogapod, line 87: maybe use `CRNL` variable here and in all other lines below since its already defined?

kristjanvalur: no, CRNL is a bytes object, just a shorthand for parsing, not emitting where we use str.

tests/resp.py
kristjanvalur: fair point

tests/resp.py
Fogapod: sorry, i commented on wrong line. there are still a few insurances of `b"\r\n"` in code in addition to strings

kristjanvalur: Yes, I like it like that since it makes the code more consistent and explicit where we are writing RESP responses.

PR #2955, opened 2023-09-20
Replace `clear_connect_callbacks` with `_deregister_connect_callback`

dvora-h:
### Pull Request check-list

### Description of change

This change seems to have been missed in #2870 (@kristjanvalur)

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -783,7 +783,7 @@ class PubSub:
 
     def __del__(self):
         if self.connection:
-            self.connection.clear_connect_callbacks()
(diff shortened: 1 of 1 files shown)


chayim (approved):
CI...

kristjanvalur:
How did that happen?

kristjanvalur:
Ah, it is in a __del__ handler.  In fact, this code is not required and the __del__ handler can be removed, since the callbacks are weak andyway, the reason being to precisely not need __del__.
But I _did_ grep for the function, so my bad on that.

PR #2982, opened 2023-10-05
Fix parsing resp3 dicts

### LINKED ISSUE #2981: resp3 parser async parser incorrectly parses maps
Issue discovered in pr #2947

the evaulation order of `key:val` in `{key():val() for foo in bar}` dict comprehension is undefined in Python.  It mostly is left-right,
but some versions, notably recent 3.7 versions, don't behave that way.  It is best to use explicit ordering when reading from  a stream.

This had already been fixed for __non-async__ parsing, but in pr #2947 caused some scheduling behaviour which triggered this to happen. A fix for __async__ has been added there.

kristjanvalur:
### Pull Request check-list

### Description of change

Fixes #2981

### DIFF
--- a/redis/_parsers/resp3.py
+++ b/redis/_parsers/resp3.py
@@ -96,8 +96,9 @@ class _RESP3Parser(_RESPBase):
                 pass
         # map response
         elif byte == b"%":
-            # we use this approach and not dict comprehension here
-            # because this dict comprehension fails in python 3.7
+            # We cannot use a dict-comprehension to parse stream.
+            # Evaluation order of key:val expression in dict comprehension only
+            # became defined to be left-right in version 3.8
             resp_dict = {}
             for _ in range(int(response)):
                 key = self._read_response(disable_decoding=disable_decoding)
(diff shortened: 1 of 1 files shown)


redis/asyncio/connection.py
```diff
@@ -493,6 +493,7 @@ async def read_response(
                 and self.protocol in ["3", 3]
                 and not HIREDIS_AVAILABLE
             ):
+                read_timeout = None
```
dvora-h, line 496: why do you need this?

redis/_parsers/resp3.py
```diff
@@ -225,12 +226,16 @@ async def _read_response(
                 pass
         # map response
         elif byte == b"%":
-            response = {
-                (await self._read_response(disable_decoding=disable_decoding)): (
-                    await self._read_response(disable_decoding=disable_decoding)
+            # we use this approach and not dict comprehension here
+            # because this dict comprehension fails in python 3.7
+            # (evaluation order of key:val is _undefined_ in Python)
```
dvora-h, line 231: Do you know if it's _undefined_ also in python 3.8+?

redis/asyncio/connection.py
kristjanvalur: ah, debugging code which accidentally crept in, good catch!

redis/_parsers/resp3.py
dvora-h: I'm not sure that we need to fix it because python 3.7 is EOL and we are going to drop support on it in 5.1

kristjanvalur: but it is still supported, though.  as it is, 3.7 will fail parsing resp3.  There is already a fix for the "sync" version, why not fix the async one?

kristjanvalur: Not pushing for it, though.  just putting in an argument :).   For my other PR where this came up, I can certainly put in a skip for 3.7

dvora-h (approved):

PR #2994, opened 2023-10-09
Fixes Redis.from_url() return type

### LINKED ISSUE #2971: from_url() typed incorrectly
**Version**: 5.0.1

**Platform**: All

**Description**: A recent commit that (very helpfully) introduced type hints to the library has an incorrect return type defined for `Redis.from_url()`. Notably, it declares a return type of `None` rather than `Redis` as it should be.

This was introduced in this commit: https://github.com/redis/redis-py/pull/2952/files#diff-3161b005adfa1f5266a9736bffb5ede9d14e2ab40937cb241bc16ca5647df79dR97

davemcphee:
### Pull Request check-list

- [√] Do tests and lints pass with this change?
- [√] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?

### Description of change

Fixes return type of Redis.from_url() (was None, now "Redis") 

fixes #2971

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -94,7 +94,7 @@ class Redis(RedisModuleCommands, CoreCommands, SentinelCommands):
     """
 
     @classmethod
-    def from_url(cls, url: str, **kwargs) -> None:
+    def from_url(cls, url: str, **kwargs) -> "Redis":
         """
         Return a Redis client object configured from the given URL
 


dvora-h:
Fixed in #2963

rafales:
Why not type it as `Self` which would have been more correct (works better with subclasses)? I see that `typing_extensions` is in dependencies, so it can be used to support pre 3.11 versions of python.

Issue #2995, opened 2023-10-10
"No connection available" errors since 5.0.1


DPham-X:
**Version**: 5.0.1

**Platform**:
Python 3.11.5(3.11.5 (main, Sep  4 2023, 15:30:52) [GCC 10.2.1 20210110])

**Description**:

Since 5.0.1, we noticed a significant increase in `No connection available.` errors from redis. We are using cashews (6.3.0) + redis-py for request caching for our fastapi application. We don't experience this with 5.0.0.

Stack trace:
```
CancelledError: null
  File "redis/asyncio/connection.py", line 1170, in get_connection
    async with self._condition:
  File "asyncio/locks.py", line 15, in __aenter__
    await self.acquire()
  File "asyncio/locks.py", line 114, in acquire
    await fut
TimeoutError: null
  File "redis/asyncio/connection.py", line 1169, in get_connection
    async with async_timeout(self.timeout):
  File "asyncio/timeouts.py", line 111, in __aexit__
    raise TimeoutError from exc_val
ConnectionError: No connection available.
  File "cashews/backends/redis/client.py", line 26, in execute_command
    return await super().execute_command(command, *args, **kwargs)
  File "redis/asyncio/client.py", line 601, in execute_command
    conn = self.connection or await pool.get_connection(command_name, **options)
  File "redis/asyncio/connection.py", line 1174, in get_connection
    raise ConnectionError("No connection available.") from err
```

paul-finary:
Same here, I have cases where "No connection available is raised" but the pool has 4 _available_connections, and only 5 _in_use_connections out of 500 max_connections.

kristjanvalur:
async Redis client and connection pools now need to be properly closed.  
The bug is in cachews.  While the Redis backend there does call `await self._client.close()`, that does not close the connection pool.

cachews needs to ensure that the Redis instance _owns_ the pool.  Or, keep a separate pool somewhere and close it.
```python
async def init(self):
        self._client = self._client_class(connection_pool=self._pool_class.from_url(self._address, **self._kwargs))
        await self._client.initialize()
        self.__is_init = True
```
should become
```python
async def init(self):
        pool = self._pool_class.from_url(self._address, **self._kwargs)
        self._client = self._client_class.from_pool(pool)
        await self._client.initialize()
        self.__is_init = True
```

This ensures that the created instance _takes_ownership_ of the pool and closes it when itself is closed.  The current syntax
assumes that the caller will take care of closing the connection pool.   

[...]

PR #2999, opened 2023-10-12
Better deal with "lost" connections for async Redis

kristjanvalur:
ConnectionPool keeps a WeakSet of in_use connections, allowing lost ones to be collected. Collection produces a warning and closes the underlying transport.

### Pull Request check-list

### Description of change

Recent changes, which removed the __del__ handling of lost connections, have illustrated how fragile many applications
are.  This PR attempts to improve on this, by better tracking "lost" connections.
See #2995 for various manifestations of this problem.

- `ConnectionPool` objects now keep a `WeakSet` of `in_use_connections`.  This allows the garbage collector to find such
  objects if they are not returned to the pool for some reason.
- Connection objects now have a `__del__` method which outputs a `ResourceWarning` in the same way as other IO classes
  in the standard library.  For some reason, the underlying "stream" objects do not get garbage collected promptly.  The
  `Connection` will perform a non-async `close()` call on these connections.  This may or may not work, since there may or may
  not be an active event loop in place, but in that case, an error is emitted.  In both cases, the developer should become aware
  of there being a problem.
- `Redis` objects call `_close()` on their connection objects if _they_ get garbage collected.

NOTE:  To see the `ResourceWarnings`, they need to be enabled.  By default they are suppressed.  Add, for instance
the `-W default::ResourceWarning` to the python command line.

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -546,6 +546,7 @@ class Redis(
                 _grl().call_exception_handler(context)
             except RuntimeError:
                 pass
+            self.connection._close()
 
     async def aclose(self, close_connection_pool: Optional[bool] = None) -> None:
         """
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -5,6 +5,7 @@ import inspect
 import socket
 import ssl
 import sys
+import warnings
 import weakref
 from abc import abstractmethod
 from itertools import chain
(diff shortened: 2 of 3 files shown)


dvora-h (approved):

XieJiSS:
Hi @kristjanvalur, I just noticed that this PR changes ConnectionPool's `self._in_use_connections` to `WeakSet` in `def reset()`, but leave `self._in_use_connections = set()` in `def __init__()`. Is this intended? Thanks!

kristjanvalur:
> Hi @kristjanvalur, I just noticed that this PR changes ConnectionPool's `self._in_use_connections` to `WeakSet` in `def reset()`, but leave `self._in_use_connections = set()` in `def __init__()`. Is this intended? Thanks!

Probably not. It's been a long time since I made this, but I suspect that I intended for a weak set to be used. This is precisely the kind of thing weak sets are useful for.

XieJiSS:
Got it. Thanks for the explanation!

PR #3001, opened 2023-10-12
Update repr of important classes with module name and recommended "< …

kristjanvalur:
### Pull Request check-list

### Description of change
the `__repr__` of important classes was missing the name of the `module`, which is often important for
disambiguation.  Also, the format should typically be "< ... >", according to the documentation, for consistency.
Update repr and tests accordingly

The motivation for this is to get more informative `ResourceWarning` errors, e.g. via PR #2999

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -337,7 +337,7 @@ class Redis(
         self._single_conn_lock = asyncio.Lock()
 
     def __repr__(self):
-        return f"{self.__class__.__name__}<{self.connection_pool!r}>"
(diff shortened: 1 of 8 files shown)


chayim:
@dvora-h I think this is a 5.1 change if it's the sort of thing we want to consider.

kristjanvalur:
Sure.  the changes are basically:

1. give the module name as well as the class, to avoid ambiguity.
2. Put the entire thing in <> as is the Python convention:  https://docs.python.org/3/reference/datamodel.html?highlight=__repr__#object.__repr__

The reason I touched this was to make sure that any `ResourceWarning` errors would provide as useful info as possible.

dvora-h (approved):

PR #3004, opened 2023-10-13
Fix incorrect asserts in test and ensure connections are closed

kristjanvalur:
### Pull Request check-list

### Description of change

Various error message assertions were incorrectly placed and were not testing anything.
In addition, sockets were being leaked in the tests.


dvora-h (commented):
Just to my understanding - why is this the correct way to test it?

dvora-h:
@kristjanvalur I think you missed my question here, if you can, I would like to understand it.

kristjanvalur:
> @kristjanvalur I think you missed my question here, if you can, I would like to understand it.

I did indeed, sorry :)

```python
with pytest.raises(ConnectionError) as e:
            sslclient.ping()
            assert "SSL: CERTIFICATE_VERIFY_FAILED" in str(e)
```
The above assert will never be hit.  sslclient.ping() raises a connection error and the `pytest.raises` context manager is exited, before execution ever gets to `assert`.
The correct way to examine the raised exception is to examine the contents of `e.value` _after_ exiting the context manager.  If just grepping for a string, then `str(e)` is good enough too. 

see https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=raises#pytest.raises

kristjanvalur:
seems unit tests runners are shaky again.

dvora-h (approved):

PR #3029, opened 2023-10-31
Fix typing of `HashCommand.hdel`

dmkulazhenko:

### DIFF
--- a/redis/commands/core.py
+++ b/redis/commands/core.py
@@ -4919,7 +4919,7 @@ class HashCommands(CommandsProtocol):
     see: https://redis.io/topics/data-types-intro#redis-hashes
     """
 
-    def hdel(self, name: str, *keys: List) -> Union[Awaitable[int], int]:
+    def hdel(self, name: str, *keys: str) -> Union[Awaitable[int], int]:
         """
         Delete ``keys`` from hash ``name``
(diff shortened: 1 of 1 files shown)


redis/commands/core.py
```diff
@@ -4919,7 +4919,7 @@ class HashCommands(CommandsProtocol):
     see: https://redis.io/topics/data-types-intro#redis-hashes
     """
 
-    def hdel(self, name: str, *keys: List) -> Union[Awaitable[int], int]:
+    def hdel(self, name: str, *keys: str) -> Union[Awaitable[int], int]:
```
chayim, line 4922: Perhaps this should be List[str] ([source](https://redis.io/commands/hdel/))?

dmkulazhenko: no, it's not. Select something from:
- `hdel(self, name: str, keys: List[str])`
- `hdel(self, name: str, *keys: str)`

python-unpacking and List typing cannot be used same time, because right now `*keys: List[str]` -- means that in function it will be `keys: tuple[List[str], ...]`.

https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

dvora-h (approved):

Issue #3032, opened 2023-11-02
get_primaries returns an empty list before any commands are executed against Cluster 


eoghanmurray:
**Version**: 5.0.1

**Platform**: linux

**Description**: 

```
        from redis.asyncio.cluster import RedisCluster
        red = RedisCluster(
            REDIS_CX[0], REDIS_CX[1], password=REDIS_PWD,
            decode_responses=decode_responses,  # utf-8 is implied if True
        )
        print(red.get_primaries())
```
prints out an empty list
the expected output can be achieved after making an async request:

```
        await red.get('xxx-does-not-exist')
        print(red.get_primaries())
```

I'd expect that `get_primaries` throws an error if it's been called before the connection has properly set up (or ideally, there would be some sort of iniitalization along the way — although I appreciate that that would likely require an async call)

dvora-h:
Hi @eoghanmurray In asyncio mode, you must call `initialize` before any other function to initialize the cluster,
so I think this code should work for you:
```
        from redis.asyncio.cluster import RedisCluster
        red = RedisCluster(
            REDIS_CX[0], REDIS_CX[1], password=REDIS_PWD,
            decode_responses=decode_responses,  # utf-8 is implied if True
        )
        await red.initialize()
        print(red.get_primaries())
```

eoghanmurray:
Hmmm, this isn't what I was saying; I was able to work around it fine once I discovered the problem.
The problem is that `get_primaries` silently returns an empty list when called before any `await` calls (which presumably call `red.initialize` as a byproduct).  So the empty list can go unnoticed after e.g. a code reorganization resulting in hard to find bugs.

PR #3040, opened 2023-11-13
Get primaries throw error

eoghanmurray:
### Description of change
See #3032


eoghanmurray:
I was thinking of moving all the checks into `NodesManager` but `nodes_cache` is accessed directly from outside of that.

PR #3042, opened 2023-11-14
Use `disable_decoding` in async `read_response` with hiredis parser.

### LINKED ISSUE #3035: `disable_decoding` not used in async version of `read_response`.
**Version**: redis-py version - 5.0.0 (I still see it on main)

**Description**:
Is there a reason why `disable_decoding` is not being used in async version of `read_response`? 
It seems like there's no async way to get raw data (`NEVER_DECODE`) from Redis.
Are there any plans of adding the `NEVER_DECODE` support for async `read_response`?
We want to save pickled data (not compatible with `UTF-8`) in Redis, and for now we will use the workaround (as in https://github.com/redis/redis-py/issues/1006#issuecomment-1100506453 ), but it's not a very clean solution.

redis/_parsers/hiredis.py:204
https://github.com/redis/redis-py/blob/d3a3ada03e080f39144807c9fbe44876c40e0548/redis/_parsers/hiredis.py#L192-L217

MrDenkoV:
### Pull Request check-list

- [x] ~~Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?~~
- [x] ~~Is there an example added to the examples folder (if applicable)?~~

### Description of change

Fix #3035 
Use `read_response` in async `read_response`.

### DIFF
--- a/redis/_parsers/hiredis.py
+++ b/redis/_parsers/hiredis.py
@@ -198,10 +198,16 @@ class _AsyncHiredisParser(AsyncBaseParser):
         if not self._connected:
             raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR) from None
 
-        response = self._reader.gets()
+        if disable_decoding:
+            response = self._reader.gets(False)
+        else:
+            response = self._reader.gets()
         while response is False:
             await self.read_from_socket()
-            response = self._reader.gets()
+            if disable_decoding:
+                response = self._reader.gets(False)
+            else:
+                response = self._reader.gets()
 
         # if the response is a ConnectionError or the response is a list and
         # the first item is a ConnectionError, raise it as something bad
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,4 @@
+    * Fix async `read_response` to use `disable_decoding`.
     * Add 'aclose()' methods to async classes, deprecate async close().
     * Fix #2831, add auto_close_connection_pool=True arg to asyncio.Redis.from_url()
     * Fix incorrect redis.asyncio.Cluster type hint for `retry_on_error`


dvora-h (approved):

redis/_parsers/hiredis.py
```diff
@@ -198,10 +198,16 @@ async def read_response(
         if not self._connected:
             raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR) from None
 
-        response = self._reader.gets()
+        if disable_decoding:
+            response = self._reader.gets(False)
+        else:
+            response = self._reader.gets()
```
kristjanvalur, line 204: why not `response=self._reader.gets(not disable_decoding)`
or
```python
enable_decoding = not disable decoding
response=self._reader.gets(enable_decoding)
```

MrDenkoV: It's been a while so I don't exactly remember, especially that there wasn't any sophisticated reason.
I think I wanted to make it consistent with the sync version + I had a problem to get to the `gets` definition, and didn't want to take much more time digging deeper.

kristjanvalur: Yes, well, I had a look at the source:
```c
 self->shouldDecode = 1;
    if (!PyArg_ParseTuple(args, "|i", &self->shouldDecode)) {
        return NULL;
    }
```
Basically, a `shouldDecode` argument can just be a Python bool.
so a `should_decode = not disable_decoding` would have been simpler and more readable.  

Anyway, just came across this, don't worry.

MrDenkoV: Thank you for pointing that out, and linking the code! 😄

PR #3046, opened 2023-11-17
Fix `MEMORY USAGE key` in cluster 

eoghanmurray:
Hard code determining that there is a single key argument for MEMORY USAGE so that the Cluster can determine the correct node to query.

Tested against redis 5.0.3, so maybe the newer redis servers have better info in the COMMAND return values


eoghanmurray:
My force push didn't resolve the conflicts, so created #3047 instead

PR #3047, opened 2023-11-17
Fix `MEMORY USAGE key` in cluster

eoghanmurray:
Hard code determining that there is a single key argument for MEMORY USAGE so that the Cluster can determine the correct node to query

Tested against redis 5.0.3, so maybe the newer redis servers have better info in the COMMAND return values


eoghanmurray:
Ah I see the COMMAND return values in latest redis returns lots of subcommand info ... I'll leave it up to the maintainers to determine whether this sort of thing is worth backporting; my implementation is added as a fallback so shouldn't interfere with new information coming in from Redis.

PR #3054, opened 2023-11-27
update hll to include example ID

sav-norem:
the example ID line was forgotten, meaning this is not displayed on the site. update to fix


guyroyse (approved):
LGTM

PR #3063, opened 2023-12-03
Fix parsing of `FT.PROFILE` result

dvora-h:
### Description of change

Fix the parsing of ft.profile result - before this fix, if there are more than 2 child iterators some of them are skipped (because the loop is in pairs), and the result is wrong

### DIFF
--- a/redis/commands/helpers.py
+++ b/redis/commands/helpers.py
@@ -64,6 +64,11 @@ def parse_list_to_dict(response):
     for i in range(0, len(response), 2):
         if isinstance(response[i], list):
             res["Child iterators"].append(parse_list_to_dict(response[i]))
+            try:
+                if isinstance(response[i + 1], list):
+                    res["Child iterators"].append(parse_list_to_dict(response[i + 1]))
+            except IndexError:
+                pass
         elif isinstance(response[i + 1], list):
             res["Child iterators"] = [parse_list_to_dict(response[i + 1])]
         else:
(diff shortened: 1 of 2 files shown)


redis/commands/helpers.py
```diff
@@ -64,6 +64,11 @@ def parse_list_to_dict(response):
     for i in range(0, len(response), 2):
         if isinstance(response[i], list):
             res["Child iterators"].append(parse_list_to_dict(response[i]))
+            try:
```
chayim, line 67: Can we have a test that exploits this?

redis/commands/helpers.py
```diff
@@ -64,6 +64,11 @@ def parse_list_to_dict(response):
...
+                if isinstance(response[i + 1], list):
+                    res["Child iterators"].append(parse_list_to_dict(response[i + 1]))
```
chayim, line 69: What happens with parse_list_to_dict if instead of res["Child iterators"] we pass in a res.get("Child iterators", []) (or None) as a default? Do we avoid the try/except cost?

redis/commands/helpers.py
dvora-h: Added

redis/commands/helpers.py
dvora-h: The try/except block is not for the res but for the response (response[i +1] can raise IndexError).

chayim (approved):

PR #3069, opened 2023-12-06
unused example ID causes entire file to be shown

sav-norem:
xadd_7 is a redis 7 feature and had not yet been implemented - will update when it is but currently it leads to unexpected behavior on the website to have it missing

### DIFF
--- a/doctests/dt_stream.py
+++ b/doctests/dt_stream.py
@@ -76,6 +76,10 @@ except redis.exceptions.ResponseError as e:
     print(e)  # >>> WRONGID
 # STEP_END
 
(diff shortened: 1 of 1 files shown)


dwdougherty:
@sav-norem I noticed that return value variable numbering is off between Java and Python, and I think it's because there is no "res*" variable/value for the missing example, the one this PR addresses. I don't know that it's a big deal, but you might increment all the "res*" variables following the placeholder example. The missing example would be "res11" and the next one (xrange) would be "res12", etc.

dvora-h (approved):

PR #3076, opened 2023-12-16
PythonRespSerializer: better chunking

VadimPushtaev:
### Pull Request check-list

### Description of change

Previous implementation made no difference between memoryviews, that must be sent separately to avoid copying, and just reaching buffer_cutoff. That resulted in some single values being sent separately for no reason, so a chunk could be just a single encoded integer, even though the buffer_cutoff is pretty large.


VadimPushtaev:
I randomly noticed that lists like `[1, 2, ..., 100]` are separated to `[1, ..., 50] + [51] + [52, ..., 100]` instead of just `[1, ..., 50] + [51, ..., 10]`. This is a fix for that.

PR #3078, opened 2023-12-18
skip `ssl` import if not available

dicej:
### Pull Request check-list

### Description of change

`utils.py` has support for detecting whether the `ssl` module is available, and we can use this to omit SSL-specific funcionality while still providing other features (e.g. unencrypted connections).

Prior to this patch, the `connection.py` modules both triggered an `ImportError` due to unconditional imports of the `ssl` module.  Now, we check `utils.SSL_AVAILABLE` prior to attempting the import and only raise an error later if (and only if) the application requests an encrypted connection.  This helps support platforms such as `wasm32-wasi` where the `ssl` module is not built by default.

Ideally, this would be tested in CI using a Python built without `ssl` support.  I'm not sure how feasible that is, though; I'm open to guidance on that aspect.

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -68,14 +68,12 @@ from redis.utils import (
     deprecated_function,
     get_lib_version,
     safe_str,
-    str_if_bytes
+    str_if_bytes,
 )
 
 if SSL_AVAILABLE:
-    import ssl
     from ssl import TLSVersion
 else:
-    ssl = None
     TLSVersion = None
 
 TargetNodesT = TypeVar(
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -2,7 +2,6 @@ import asyncio
 import collections
 import random
 import socket
-import ssl
 import warnings
 from typing import (
     Any,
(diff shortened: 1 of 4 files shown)


petyaslavova:
Hi @dicej, can you please update your code to follow the pattern set in https://github.com/redis/redis-py/blob/master/redis/client.py for ssl import and type hints and fix the conflicts?

dicej:
> Hi @dicej, can you please update your code to follow the pattern set in https://github.com/redis/redis-py/blob/master/redis/client.py for ssl import and type hints and fix the conflicts?

Done, and tested on `wasm32-wasi` (which does not have an `ssl` module).

redis/asyncio/cluster.py
```diff
@@ -64,7 +63,20 @@
...
+    get_lib_version,
+    safe_str,
+    str_if_bytes
+)
+
+if SSL_AVAILABLE:
+    import ssl
+    from ssl import TLSVersion
+else:
+    ssl = None
```
petyaslavova, line 78: In this case, I don't think you need to import and add any initialization of the ssl. 
In the end, you are using just the TLSVersion, and you can handle just this import and default init in case ssl is not available. The same applies for client and cluster files.

petyaslavova (changes requested):
@dicej please fix the linter errors and apply requested changes

redis/asyncio/connection.py
```diff
@@ -27,6 +27,16 @@
 )
 from urllib.parse import ParseResult, parse_qs, unquote, urlparse
 
+from ..utils import SSL_AVAILABLE
+
+if TYPE_CHECKING and SSL_AVAILABLE:
```
petyaslavova, line 32: It seems that the TYPE_CHECKING test here breaks the import of ssl when it really needs to be imported. Can you please remove it +  the import of TYPE_CHECKING?

petyaslavova (approved):

PR #3099, opened 2024-01-06
Support client side caching with ConnectionPool

dvora-h:
This pull request extends the single mode, client-side-caching support to ConnectionPools.

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -25,6 +25,12 @@ from typing import (
     cast,
 )
 
+from redis._cache import (
+    DEFAULT_BLACKLIST,
+    DEFAULT_EVICTION_POLICY,
+    DEFAULT_WHITELIST,
+    _LocalCache,
+)
 from redis._parsers.helpers import (
     _RedisCallbacks,
     _RedisCallbacksRESP2,
@@ -39,12 +45,6 @@ from redis.asyncio.connection import (
 )
 from redis.asyncio.lock import Lock
 from redis.asyncio.retry import Retry
-from redis.cache import (
-    DEFAULT_BLACKLIST,
-    DEFAULT_EVICTION_POLICY,
-    DEFAULT_WHITELIST,
-    _LocalCache,
-)
 from redis.client import (
     EMPTY_RESPONSE,
     NEVER_DECODE,
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -67,7 +67,7 @@ from redis.exceptions import (
     TimeoutError,
     WatchError,
 )
-from redis.typing import ChannelT, EncodableT, KeysT, KeyT, ResponseT
+from redis.typing import ChannelT, EncodableT, KeyT
 from redis.utils import (
     HIREDIS_AVAILABLE,
     _set_info_logger,
@@ -294,6 +294,13 @@ class Redis(
                 "lib_version": lib_version,
                 "redis_connect_func": redis_connect_func,
                 "protocol": protocol,
+                "cache_enable": cache_enable,
+                "client_cache": client_cache,
+                "cache_max_size": cache_max_size,
+                "cache_ttl": cache_ttl,
+                "cache_eviction_policy": cache_eviction_policy,
+                "cache_blacklist": cache_blacklist,
+                "cache_whitelist": cache_whitelist,
             }
             # based on input, setup appropriate connection args
             if unix_socket_path is not None:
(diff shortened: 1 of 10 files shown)


redis/asyncio/client.py
```diff
@@ -294,6 +294,13 @@ def __init__(
                 "lib_version": lib_version,
                 "redis_connect_func": redis_connect_func,
                 "protocol": protocol,
+                "cache_enable": cache_enable,
```
chayim, line 297: Let's have you, @vladvildanov and @uglide sync on these just to finalize before the **next** release. Getting them into the specs finally too - WDYT?

redis/connection.py
```diff
@@ -426,6 +454,9 @@ def disconnect(self, *args):
         except OSError:
             pass
 
+        if self.client_cache:
```
chayim, line 457: happy :D

redis/connection.py
```diff
@@ -578,6 +609,58 @@ def _is_socket_empty(self):
...
+        """
+        If the command is in the local cache, return the response
+        """
+        if (
+            self.client_cache is None
+            or command[0] in self.cache_blacklist
+            or command[0] not in self.cache_whitelist
+        ):
+            return None
+        while not self._is_socket_empty():
```
chayim, line 637: nit: _socket_is_empty instead?

tests/conftest.py
```diff
@@ -356,6 +356,7 @@ def _gen_cluster_mock_resp(r, response):
     connection = Mock(spec=Connection)
     connection.retry = Retry(NoBackoff(), 0)
     connection.read_response.return_value = response
+    connection._get_from_local_cache.return_value = None
```
chayim, line 359: good catch

tests/test_asyncio/test_cache.py
```diff
@@ -2,36 +2,38 @@
...
 import redis.asyncio as redis
+from redis._cache import _LocalCache
 from redis.utils import HIREDIS_AVAILABLE
 
 
 @pytest.mark.skipif(HIREDIS_AVAILABLE, reason="PythonParser only")
 async def test_get_from_cache():
-    r = redis.Redis(cache_enable=True, single_connection_client=True, protocol=3)
+    cache = _LocalCache()
+    r = redis.Redis(protocol=3, client_cache=cache)
```
chayim, line 12: Since we keep doing these, maybe a resp3client helper would be easier, then it can just be passed in repeatedly - similar to _r_

chayim (changes requested):
Small ones - looks good. Let's close the loop in the spec.

redis/asyncio/connection.py
```diff
@@ -205,6 +221,14 @@ def __init__(
             if p < 2 or p > 3:
                 raise ConnectionError("protocol must be either 2 or 3")
             self.protocol = protocol
+        if cache_enable:
```
chayim, line 224: Should we do an enforce check on RESP version, like others did - just for CSC?

redis/asyncio/connection.py
```diff
@@ -429,6 +458,9 @@ async def disconnect(self, nowait: bool = False) -> None:
             raise TimeoutError(
                 f"Timed out closing connection after {self.socket_connect_timeout}"
             ) from None
+        finally:
+            if self.client_cache:
```
chayim, line 462: :D

redis/asyncio/connection.py
```diff
@@ -650,6 +682,58 @@ def _is_socket_empty(self):
...
+    def _cache_invalidation_process(
+        self, data: List[Union[str, Optional[List[str]]]]
+    ) -> None:
+        """
+        Invalidate (delete) all redis commands associated with a specific key.
+        `data` is a list of strings, where the first string is the invalidation message
+        and the second string is the list of keys to invalidate.
+        (if the list of keys is None, then all keys are invalidated)
+        """
+        if data[1] is not None:
```
chayim, line 694: Maybe we can return, and then do the after the fact?

```
if data[1] is None:
    return
```

redis/client.py
```diff
@@ -268,6 +267,13 @@ def __init__(
                 "redis_connect_func": redis_connect_func,
                 "credential_provider": credential_provider,
                 "protocol": protocol,
+                "cache_enable": cache_enable,
```
chayim, line 270: @sjpotter @vladvildanov @ofekshenawa Can you all align on these, with whatever is right _collectively_? Then, we can get this into the specification.

chayim (changes requested):

chayim (approved):

redis/client.py
```diff
@@ -572,63 +548,25 @@ def _disconnect_raise(self, conn, error):
...
     def execute_command(self, *args, **options):
         """Execute a command and return a parsed response"""
         command_name = args[0]
         keys = options.pop("keys", None)
-        response_from_cache = self._get_from_local_cache(args)
+        pool = self.connection_pool
+        conn = self.connection or pool.get_connection(command_name, **options)
+        response_from_cache = conn._get_from_local_cache(args)
         if response_from_cache is not None:
             return response_from_cache
```
Cruuncher, line 560: Isn't this a connection leak if we pull from cache? 

We're getting a connection from the pool and never releasing it in this code path.

dvora-h: Yes, it was, but fixed already in #3160

PR #3102, opened 2024-01-08
Support client side caching with RedisCluster

dvora-h:
This pull request extends the client-side-caching support to RedisCluster.

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -18,6 +18,12 @@ from typing import (
     Union,
 )
 
+from redis._cache import (
+    DEFAULT_BLACKLIST,
+    DEFAULT_EVICTION_POLICY,
+    DEFAULT_WHITELIST,
+    _LocalCache,
+)
 from redis._parsers import AsyncCommandsParser, Encoder
 from redis._parsers.helpers import (
     _RedisCallbacks,
@@ -267,6 +273,13 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
         ssl_keyfile: Optional[str] = None,
         protocol: Optional[int] = 2,
         address_remap: Optional[Callable[[str, int], Tuple[str, int]]] = None,
+        cache_enable: bool = False,
+        client_cache: Optional[_LocalCache] = None,
+        cache_max_size: int = 100,
+        cache_ttl: int = 0,
+        cache_eviction_policy: str = DEFAULT_EVICTION_POLICY,
+        cache_blacklist: List[str] = DEFAULT_BLACKLIST,
+        cache_whitelist: List[str] = DEFAULT_WHITELIST,
     ) -> None:
         if db:
             raise RedisClusterException(
@@ -310,6 +323,14 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
             "socket_timeout": socket_timeout,
             "retry": retry,
             "protocol": protocol,
+            # Client cache related kwargs
+            "cache_enable": cache_enable,
+            "client_cache": client_cache,
+            "cache_max_size": cache_max_size,
+            "cache_ttl": cache_ttl,
+            "cache_eviction_policy": cache_eviction_policy,
+            "cache_blacklist": cache_blacklist,
+            "cache_whitelist": cache_whitelist,
         }
 
         if ssl:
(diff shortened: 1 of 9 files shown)


tests/test_asyncio/test_cache.py
```diff
@@ -1,140 +1,177 @@
...
-    assert await r.get("foo4") == b"bar4"
-    # test the eviction policy
-    assert len(cache.cache) == 3
-    assert cache.get(("GET", "foo")) == b"bar"
-    assert cache.get(("GET", "foo2")) is None
-
-    await r.flushdb()
-    await r.aclose()
+@pytest_asyncio.fixture
+async def r(request, create_redis):
```
chayim, line 10: great

tests/test_cache.py
```diff
@@ -4,131 +4,177 @@
...
-    r.set("foo4", "bar4")
-    assert r.get("foo4") == b"bar4"
-    # test the eviction policy
-    assert len(cache.cache) == 3
-    assert cache.get(("GET", "foo")) == b"bar"
-    assert cache.get(("GET", "foo2")) is None
-
-    r.flushdb()
+@pytest.fixture()
+def r(request):
```
chayim, line 11: also glad to see

redis/cluster.py
```diff
@@ -167,6 +167,13 @@ def parse_cluster_myshardid(resp, **options):
     "ssl_password",
     "unix_socket_path",
     "username",
+    "cache_enable",
+    "client_cache",
+    "cache_max_size",
```
chayim, line 172: IMHO I think we have enough duplication with all these. Part of another issue to clean up / tech debt?

redis/cluster.py
```diff
@@ -1147,14 +1154,18 @@ def _execute_command(self, target_node, *args, **kwargs):
...
-
-                connection.send_command(*args)
-                response = redis_node.parse_response(connection, command, **kwargs)
-                if command in self.cluster_response_callbacks:
-                    response = self.cluster_response_callbacks[command](
-                        response, **kwargs
-                    )
-                return response
+                response_from_cache = connection._get_from_local_cache(args)
+                if response_from_cache is not None:
```
chayim, line 1158: :+1:

chayim (approved):
A question that IMHO belongs elsewhere - looks like our others, and clean.

redis/asyncio/cluster.py
```diff
@@ -267,6 +273,13 @@ def __init__(
         ssl_keyfile: Optional[str] = None,
         protocol: Optional[int] = 2,
         address_remap: Optional[Callable[[str, int], Tuple[str, int]]] = None,
+        cache_enable: bool = False,
+        client_cache: Optional[_LocalCache] = None,
```
vladvildanov, line 277: `client_cache` is an implementation of cache that could be provided by user?

redis/asyncio/connection.py
```diff
@@ -227,6 +227,10 @@ def __init__(
             _cache = None
         self.client_cache = client_cache if client_cache is not None else _cache
         if self.client_cache is not None:
+            if self.protocol not in [3, "3"]:
+                raise RedisError(
+                    "client caching is only supported with protocol version 3 or higher"
```
vladvildanov, line 232: @chayim As I remember we agreed to force RESP3 instead of raising an error? Error should be thrown after it

redis/cluster.py
```diff
@@ -1147,14 +1154,18 @@ def _execute_command(self, target_node, *args, **kwargs):
...
+                    return response_from_cache
+                else:
+                    connection.send_command(*args)
+                    response = redis_node.parse_response(connection, command, **kwargs)
+                    if command in self.cluster_response_callbacks:
```
vladvildanov, line 1163: I think according to RESP3 specification callback should be provided per push notification type and not per command.

`Clients should normally react to the publication of a push data by invoking a callback associated with the push data type. Synchronous clients may instead enter a loop and return every new message or command reply.`

https://github.com/antirez/RESP3/blob/master/spec.md#push-type

redis/asyncio/cluster.py
```diff
@@ -1039,16 +1059,24 @@ async def parse_response(
...
-        # Execute command
-        await connection.send_packed_command(connection.pack_command(*args), False)
-
-        # Read response
-        try:
-            return await self.parse_response(connection, args[0], **kwargs)
-        finally:
-            # Release connection
+        response_from_cache = await connection._get_from_local_cache(args)
+        if response_from_cache is not None:
```
vladvildanov, line 1065: Are you also check a command against a list of blacklisted/whitelisted commands before check it against cache?

dvora-h: Yes 
https://github.com/redis/redis-py/blob/b5d4d29409f83e9a0056c6ed12414bf58e1ea1c0/redis/connection.py#L637

redis/cluster.py
dvora-h: It's unrelated to push notifications and even not to RESP3, it's general response callbacks for all the commands.

PR #3104, opened 2024-01-09
Fix possible pipeline connections leak

ING-XIAOJIAN:
### Pull Request check-list

### Description of change
When Executing "n.write()" may generate some unknown errors(e.g. DataError), which could result in the connection not being released, finally lead to connection leak.

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -2170,13 +2170,13 @@ class ClusterPipeline(RedisCluster):
             node_commands = nodes.values()
             for n in node_commands:
                 n.write()
-    
+
             for n in node_commands:
                 n.read()
         finally:
             for n in nodes.values():
                 n.connection_pool.release(n.connection)
-    
+
         # release all of the redis connections we allocated earlier
         # back into the connection pool.
         # we used to do this step as part of a try/finally block,
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -2174,29 +2174,26 @@ class ClusterPipeline(RedisCluster):
             for n in node_commands:
                 n.read()
         finally:
+            # release all of the redis connections we allocated earlier
+            # back into the connection pool.
+            # we used to do this step as part of a try/finally block,
+            # but it is really dangerous to
+            # release connections back into the pool if for some
+            # reason the socket has data still left in it
+            # from a previous operation. The write and
+            # read operations already have try/catch around them for
+            # all known types of errors including connection
+            # and socket level errors.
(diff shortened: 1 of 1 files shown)


dvora-h:
@ING-XIAOJIAN thanks for suggestion, looks good, but WDYT about:
```
    try:
        node_commands = nodes.values()
        for n in node_commands:
            n.write()

        for n in node_commands:
            n.read()
    finally:
        for n in nodes.values():
            n.connection_pool.release(n.connection)
```

ING-XIAOJIAN:
> @ING-XIAOJIAN thanks for suggestion, looks good, but WDYT about:
> 
> ```
>     try:
>         node_commands = nodes.values()
>         for n in node_commands:
>             n.write()
> 
>         for n in node_commands:
>             n.read()
>     finally:
>         for n in nodes.values():
>             n.connection_pool.release(n.connection)
> ```
I naively think "n.read" never generates some unknown errors,
I will take your suggestion.

dvora-h:
I agree that there is less chance to get an error in the reading, but I prefer to cover that as well just to be safe

ING-XIAOJIAN:
> @ING-XIAOJIAN Can you fix the linters or wants me to do it?

I already fixed those linters

dvora-h:
Thanks!

redis/cluster.py
```diff
@@ -2190,8 +2195,8 @@ def _send_cluster_commands(
         # buffered result back from a previous command and
         # every single request after to that connection will always get
         # a mismatched result.
-        for n in nodes.values():
-            n.connection_pool.release(n.connection)
+        # for n in nodes.values():
```
dvora-h, line 2198: Can you delete this lines from the comment and move the explanations of this lines before them?

ING-XIAOJIAN: I already fixed this. could you review it again?

redis/cluster.py
```diff
@@ -2166,32 +2166,34 @@ def _send_cluster_commands(
...
+            node_commands = nodes.values()
+            for n in node_commands:
+                n.write()
+
+            for n in node_commands:
+                n.read()
+        finally:
+            # release all of the redis connections we allocated earlier
+            # back into the connection pool.
+            # we used to do this step as part of a try/finally block,
```
bcmills, line 2179: This comment seems out-of-date — isn't it explaining why _not_ to do exactly what this change is doing?

It should probably be update to explain why this _is_ safe, not why it _isn't_. 😅

PR #3127, opened 2024-01-30
Allow to control the minimum SSL version

poiuj:
### Description of change

Allow to control the minimum SSL version. It's useful for applications that has strict security requirements. For example to make TLS 1.3 mandatory. 

The change introduces `ssl_min_version` parameter. That is passed onto the SSL context.


dvora-h (approved):

rmacian:
Prior this change, which TLS version was being used?

Issue #3148, opened 2024-02-15
RedisCluster initialization is too slow, can we skip it or optimize it?


willianmrs:
**Version**: 5.0.1

**Platform**: Python 3.8, Ubuntu 22.04

**Description**: 

Hello guys, I'm upgrading the library from a very old version, from redis-py-cluster. I notice that when creating the client, for instance:
```python
RedisCluster(
        host=host_configuration["host"],
        port=int(host_configuration["port"]),
        decode_responses=True,
        skip_full_coverage_check=True,
        socket_timeout=timeout,
        socket_connect_timeout=timeout,
        username=host_configuration.get("user"),
        password=host_configuration.get("pass")
    )

```

the initialization is taking too long because it sends a command INFO to check if the cluster is enabled, [references](https://github.com/redis/redis-py/blob/master/redis/cluster.py#L1528):
```   python          
  # Make sure cluster mode is enabled on this node
  if bool(r.info().get("cluster_enabled")) is False:
      raise RedisClusterException(
          "Cluster mode is not enabled on this node"
      )
  cluster_slots = str_if_bytes(r.execute_command("CLUSTER SLOTS"))
  startup_nodes_reachable = True
```

Since my cluster is big, this command is taking too long, that is not a big issue, because it runs only in the initialization, but the problem is that my requests are very fast, we use a pipeline and get so the majority of requests are below 5ms and I wanna configure a timeout for something like 100ms or even 50ms, but since the timeout parameter is configured only in the initialization of the cluster and it is taking a few seconds to initialize, I need to inform a bigger timeout on the constructor. 

Saying all that, I have two questions, (1) is it possible to change the timeout after creating the client? If not, we can change the way it validates if the cluster is enabled to a faster command or even skip this validation?

PR #3158, opened 2024-02-21
Optimizing cluster initialization changing the checks for cluster-enabled flag

willianmrs:
### Pull Request check-list

### Description of change

In this PR, I propose an optimized approach to RedisCluster initialization, focusing on improving the speed and efficiency for clusters. The key change involves replacing the initial INFO command check for cluster mode enablement with a direct attempt to execute "CLUSTER SLOTS". This method assumes that successful execution of "CLUSTER SLOTS" implies cluster mode is active, thus streamlining the initialization process. This optimization aims to reduce startup time without compromising the integrity of cluster checks, making it particularly beneficial for environments where rapid deployment and high performance are critical.

To compare an INFO command takes an average of 2 to 3 seconds to complete and REDIS SLOTS takes only a few milliseconds. This enables the timeout being used in the constructor, not being necessary to add a bigger timeout just to initialize the cluster. 

Linked issue:
https://github.com/redis/redis-py/issues/3148


dvora-h (approved):

dvora-h:
@willianmrs Thanks for this PR! I like this approach.
Why do you think in async it can result in strange behavior? In my opinion the same should work for async too...

dvora-h:
@willianmrs also - notice the linters failure

willianmrs:
@dvora-h great. About async, I'm working on the async cluster changes. I guess in the first look at the Async connection I understand wrong. But it does the same as the sync cluster, so you are right, the same should work for it. 


And thanks about the lint, I will submit the fix with the async changes.

PR #3160, opened 2024-02-26
Fix bug: client side caching causes unexpected disconnections

dvora-h:
### Pull Request check-list

### Description of change


redis/client.py
```diff
@@ -558,10 +558,10 @@ def execute_command(self, *args, **options):
         pool = self.connection_pool
         conn = self.connection or pool.get_connection(command_name, **options)
         response_from_cache = conn._get_from_local_cache(args)
-        if response_from_cache is not None:
-            return response_from_cache
-        else:
-            try:
+        try:
```
chayim, line 561: I think a comment is warranted, about the behaviour, even just a link to the bug. My instinct, reading the diff is that I would have though that the original code wouldn't cause this,

chayim (approved):

PR #2989, opened 2023-10-06
Fix SlotNotCoveredError when cluster is resharding

dinispeixoto:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [ ] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?
- [ ] Was the change added to CHANGES file?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

_Please provide a description of the change here._

Fixes https://github.com/redis/redis-py/issues/2988

PR #3182, opened 2024-03-13
Fix get_node_from_slot to handle resharding

### LINKED ISSUE #2988: SlotNotCoveredError when cluster is resharding

- Client: 4.5.5
- Cluster Engine: 6.2.6

- ECS and ElastiCache

When the cluster is up or down scaling, while slots are being migrated to the new shard, the client raises the following error:
```
redis.exceptions.SlotNotCoveredError: Command # 9 (HGETALL ...) of pipeline caused error: ('Slot "4718" not covered by the cluster. "require_full_coverage=True"',)
```

**Full stack trace:**
```
  File "/root/clients/redis_client.py", line 219, in run_queries
    return await pipeline.execute()
  File "/usr/local/lib/python3.8/dist-packages/redis/asyncio/cluster.py", line 1455, in execute
    return await self._execute(
  File "/usr/local/lib/python3.8/dist-packages/redis/asyncio/cluster.py", line 1536, in _execute
    raise result
  File "/usr/local/lib/python3.8/dist-packages/redis/asyncio/cluster.py", line 1520, in _execute
    cmd.result = await client.execute_command(
  File "/usr/local/lib/python3.8/dist-packages/redis/asyncio/cluster.py", line 725, in execute_command
    raise e
  File "/usr/local/lib/python3.8/dist-packages/redis/asyncio/cluster.py", line 696, in execute_command
    ret = await self._execute_command(target_nodes[0], *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/redis/asyncio/cluster.py", line 745, in _execute_command
    target_node = self.nodes_manager.get_node_from_slot(
  File "/usr/local/lib/python3.8/dist-packages/redis/asyncio/cluster.py", line 1174, in get_node_from_slot
    raise SlotNotCoveredError(
redis.exceptions.SlotNotCoveredError: Command # 6 (HGET ... ...) of pipeline caused error: ('Slot "11246" not covered by the cluster. "require_full_coverage=True"',)
```

**Reproduce locally**
It's also possible to reproduce this locally quite easily:
1. Setup a Redis cluster locally (https://github.com/Grokzen/docker-redis-cluster)
2. Create a simple Redis app that leverages a pipeline of read-only commands and run a load test to force multiple reads. I also made sure there was a different client writing non-stop to the local cluster.
3. Force resharding of a few slots (e.g. 5k) with `redis-cli --cluster reshard 0.0.0.0:7000`
4. Redis app should raise `SlotNotCoveredError` quite often while the migration of slots is happening, once it's done it stops.

**Possible root cause**
I tried debugging the library and here's what I could find so far:

1. The slots cache is outdated, the slot we need is in a different shard. Client replaces `self.slots_cache[slot]` with the `redirected_node` (returned by the `MOVED` error) which is **only** the master node for the new shard of the slot.

https://github.com/redis/redis-py/blob/master/redis/asyncio/cluster.py#L1177

```python
else:
    # The new slot owner is a new server, or a server from a different
    # shard. We need to remove all current nodes from the slot's list
    # (including replications) and add just the new node.
    self.slots_cache[e.slot_id] = [redirected_node]
```

2. However at the same time, we might call `NodesManager.get_node_from_slot`. Since we are only reading from read replicas it starts by getting the server index using the RR load balancer. Let's say we have 1 master node and 2 replicas for each shard, it can return an index from 0 to 2. However, as we've seen earlier for this specific slot the slots cache only has a single entry (the master node). This essentially means that if the RR load balancer returns either 1 or 2, it will raise an `IndexError`, thus a `SlotNotCoveredError`.

https://github.com/redis/redis-py/blob/master/redis/asyncio/cluster.py#L1191C1-L1204C14

```python
        try:
            if read_from_replicas:
                # get the server index in a Round-Robin manner
                primary_name = self.slots_cache[slot][0].name
                node_idx = self.read_load_balancer.get_server_index(
                    primary_name, len(self.slots_cache[slot])
                )
                return self.slots_cache[slot][node_idx]
            return self.slots_cache[slot][0]
        except (IndexError, TypeError):
            raise SlotNotCoveredError(
                f'Slot "{slot}" not covered by the cluster. '
                f'"require_full_coverage={self.require_full_coverage}"'
            )
```

3. In order to fix the issue I just patched the method to check whether the index returned by the load balancer is available in the slots cache and in case it isn't just get it from the primary node instead.

```python
                node_idx = self.read_load_balancer.get_server_index(
                    primary_name, len(self.slots_cache[slot])
                )

                node = node_idx if node_idx < len(self.slots_cache[slot]) else 0
                return self.slots_cache[slot][node]
```

Happy to create a PR if this makes sense. Sorry if I'm missing something.

Thanks in advance 🙌

johan-seesaw:
### Pull Request check-list

### Description of change

~Introduce a new Enum and optional flag value to allow reading only from replicas, if the command supports it. It appears the sync version of `RedisCluster` used to support this when `server_type` was passed in to `get_node_from_slot`, but that parameter isn't set anywhere.~

This PR addresses the issue of reading from replicas while resharding, which can cause index failures.
It is an alternative, and more comprehensive solution to https://github.com/redis/redis-py/pull/2989, but in both sync and asyncio implementations. (https://github.com/redis/redis-py/issues/2988). 

This implementation moves the logic to a shared location between the asyncio and the sync versions of the library. I have a follow on PR to introduce additional read-only modes that was initially part of this PR, but has been kept separate to hopefully increase the likelihood that this PR can get merged.

Before the change, we stored the _next replica to read from_ in the `primary_to_idx` cache. After the change we store the _last read replica/primary_. This is important for the following example:

1. Server exists with a slot with 1 primary and 2 replicas.
2. A number of commands are executed. The last command executes on index `1`, setting `primary_to_idx` to value 2.
3. Before another command is executed, the number of replicas changes from 2 to 1. The total number of nodes in that slot is now `2`.
4. `get_node_from_slot` reads the next-node value as `2`, and returns that (and sets the next-node value to `1`, due to `2+1%2 = 1`, skipping the primary on the next run)
5. We try to read the Node at `slot_nodes[2]`, which doesn't exist, and gives us an index exception.

After, we only store the last-read, not the next-value, so the situation would unfold as follows:
1. Server exists with a slot with 1 primary and 2 replicas.
2. A number of commands are executed. The last command executes on index `1`, setting `primary_name_to_last_used_index` to value 1.
3. Before another command is executed, the number of replicas changes from 2 to 1. The total number of nodes in that slot is now `2`.
4. `get_node_from_slot` reads the last-read-node value as `1`, increments, and modulos against length of 2, resulting in the value of `0` being stored in `primary_name_to_last_used_index` and returned
5. We try to read the Node at `slot_nodes[0]`, the primary, as expected, and everything works.

Fixes https://github.com/redis/redis-py/issues/2988


noorul:
@johan-seesaw How did you fix it in the application as this is taking time to get merged?

Issue #3189, opened 2024-03-19
pretty sure sadd should take a KeyT not an str for the key name


earonesty:
This function:

```
    def sadd(self, name: str, *values: FieldT) -> Union[Awaitable[int], int]:
```

differs from all other key functons in that the name is a "str", however there is no documentation that says it must be an str.  all keys are "bytes safe" from the docs.   should probably change all the set functions to KeyT.   Unless anyone disagrees, happy to put up a PR.

petyaslavova:
Fixed with PR #3190

PR #3190, opened 2024-03-20
KeyT for set operations

earonesty:
### Description of change

Change set operation key-types to KeyT

https://github.com/redis/redis-py/issues/3189

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -2350,10 +2350,11 @@ class TransactionStrategy(AbstractStrategy):
         # watching something
         if self._transaction_connection:
             try:
-                # call this manually since our unwatch or
-                # immediate_execute_command methods can call reset()
-                await self._transaction_connection.send_command("UNWATCH")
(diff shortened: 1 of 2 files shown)


dvora-h (approved):

gerzse:
@earonesty, somehow this PR ended up in conflict state. Could you, please, update it, so we merge it? Thanks!

redis/commands/core.py
```diff
@@ -3364,7 +3364,7 @@ def sinterstore(
         return self.execute_command("SINTERSTORE", dest, *args)
```
petyaslavova, line 3364: The `dest` argument of `sinterstore` method should also be changed as it represents the name of a new set.
The same applies for `sdiffstore` method's `dest` argument

petyaslavova (changes requested):

redis/commands/core.py
petyaslavova: Done.

petyaslavova (approved):

PR #3198, opened 2024-04-02
Adding a twine check prior to package release

chayim:
Twine supports a basic check for the package prior to upload. Prior to this the package was built and installed via a different CI.

It is theoretically possible for the associated tools (or a python version) to create broken packages, in between the initial pull request, and an actual release. This at least adds a basic safety check.


chayim:
@dvora-h surprise

Issue #3209, opened 2024-04-19
Add details to connection error message


ebergard:
https://github.com/redis/redis-py/blob/07fc339b4a4088c1ff052527685ebdde43dfc4be/redis/asyncio/connection.py#L808

If the number of exception.args is more than one, the error code is just duplicated instead of showing the error details

petyaslavova:
Covered by PR #3305

PR #3211, opened 2024-04-20
Add details to the asyncio connection error message

dkuser:
### Pull Request check-list

### Description of change

Issue https://github.com/redis/redis-py/issues/3209
- Add details to the asyncio connection error message
- Add test for asyncio.Connection._error_message


redis/asyncio/connection.py
```diff
@@ -817,7 +817,7 @@ def _error_message(self, exception: BaseException) -> str:
         else:
             return (
                 f"Error {exception.args[0]} connecting to {host_error}. "
-                f"{exception.args[0]}."
+                f"{exception}."
```
gerzse, line 820: Thanks for straightening this out.

This rendering of the error message is done in four places, sync and async, unix and network socket. All four slightly different from each other.

Would you care aligning all four places? Some ideas that come on the spot:
- pull out the handling at least into the two abstract connection classes, maybe even reuse a helper method between the two abstract classes;
- if there are no args, return a simple error message, don't even assume it's the "connection reset", there might be other cases too;
- if there is one arg, use that one;
- if there are more then one args, use the first one like it is today, and then put `exception.arg[1:]` at the end of the message, concatenated.

What do you think, does it make sense?

gerzse (changes requested):

redis/asyncio/connection.py
gerzse: I tried to push a commit with my suggestion, but I don't have permissions. I'll merge this as it is, it's already a good step forward. And open a separate PR as continuation.

gerzse (approved):

PR #3212, opened 2024-04-22
Make it possible to customize SSL ciphers

gerzse:
### Pull Request check-list

### Description of change

Given that Python 3.10 changed the default list of SSL ciphers, it is a good idea in general to allow customization of the list of cyphers when using Redis with TLS.

It seems that this works only with TLS 1.2, and with TLS 1.3 it's intentionally not possible to change the ciphers:
https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_ciphers


uglide (approved):
LGTM

PR #3216, opened 2024-04-25
Streamline client side caching API typing

gerzse:
### Pull Request check-list

### Description of change

Streamline the typing of the client side caching API. Some of the methods are defining commands of type `str`, while in reality tuples are being sent for those parameters.

Add client side cache tests for Sentinels. In order to make this work, fix the sentinel configuration in the docker-compose stack.

Add a test for client side caching with a truly custom cache, not just injecting our internal cache structure as custom.

Add a test for client side caching where two different types of commands use the same key, to make sure they invalidate each others cached data.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -1,5 +1,6 @@
 click==8.0.4
 black==24.3.0
+cachetools
 flake8==5.0.4
 flake8-isort==6.0.0
 flynt~=0.69.0
--- a/redis/_cache.py
+++ b/redis/_cache.py
@@ -17,7 +17,7 @@ class EvictionPolicy(Enum):
 
 DEFAULT_EVICTION_POLICY = EvictionPolicy.LRU
 
-DEFAULT_BLACKLIST = [
+DEFAULT_DENY_LIST = [
     "BF.CARD",
     "BF.DEBUG",
     "BF.EXISTS",
@@ -77,7 +77,7 @@ DEFAULT_BLACKLIST = [
     "TTL",
 ]
 
-DEFAULT_WHITELIST = [
+DEFAULT_ALLOW_LIST = [
     "BITCOUNT",
     "BITFIELD_RO",
     "BITPOS",
--- a/redis/_cache.py
+++ b/redis/_cache.py
@@ -8,7 +8,14 @@ from typing import Iterable, List, Union
 
 from redis.typing import KeyT, ResponseT
 
-DEFAULT_EVICTION_POLICY = "lru"
+
+class EvictionPolicy(Enum):
+    LRU = "lru"
+    LFU = "lfu"
+    RANDOM = "random"
+
+
+DEFAULT_EVICTION_POLICY = EvictionPolicy.LRU
 
 DEFAULT_BLACKLIST = [
     "BF.CARD",
@@ -151,12 +158,6 @@ _RESPONSE = "response"
 _KEYS = "keys"
 _CTIME = "ctime"
 _ACCESS_COUNT = "access_count"
-
-
-class EvictionPolicy(Enum):
-    LRU = "lru"
-    LFU = "lfu"
-    RANDOM = "random"
 
 
 class AbstractCache(ABC):
@@ -314,28 +315,28 @@ class _LocalCache(AbstractCache):
         Args:
             command (Union[str, Iterable[str]]): The redis command.
         """
-        if self.eviction_policy == EvictionPolicy.LRU.value:
+        if self.eviction_policy == EvictionPolicy.LRU:
             self.cache.move_to_end(command)
-        elif self.eviction_policy == EvictionPolicy.LFU.value:
+        elif self.eviction_policy == EvictionPolicy.LFU:
             self.cache[command]["access_count"] = (
                 self.cache.get(command, {}).get("access_count", 0) + 1
             )
             self.cache.move_to_end(command)
-        elif self.eviction_policy == EvictionPolicy.RANDOM.value:
+        elif self.eviction_policy == EvictionPolicy.RANDOM:
             pass  # Random eviction doesn't require updates
 
     def _evict(self):
         """Evict a redis command from the cache based on the eviction policy."""
         if self._is_expired(self.commands_ttl_list[0]):
             self.delete_command(self.commands_ttl_list[0])
-        elif self.eviction_policy == EvictionPolicy.LRU.value:
+        elif self.eviction_policy == EvictionPolicy.LRU:
             self.cache.popitem(last=False)
-        elif self.eviction_policy == EvictionPolicy.LFU.value:
+        elif self.eviction_policy == EvictionPolicy.LFU:
             min_access_command = min(
                 self.cache, key=lambda k: self.cache[k].get("access_count", 0)
             )
             self.cache.pop(min_access_command)
-        elif self.eviction_policy == EvictionPolicy.RANDOM.value:
+        elif self.eviction_policy == EvictionPolicy.RANDOM:
             random_command = random.choice(list(self.cache.keys()))
             self.cache.pop(random_command)
 
--- a/redis/_cache.py
+++ b/redis/_cache.py
@@ -167,19 +167,24 @@ class AbstractCache(ABC):
     """
 
     @abstractmethod
-    def set(self, command: str, response: ResponseT, keys_in_command: List[KeyT]):
+    def set(
+        self,
+        command: Union[str, Iterable[str]],
+        response: ResponseT,
+        keys_in_command: List[KeyT],
+    ):
         pass
 
     @abstractmethod
-    def get(self, command: str) -> ResponseT:
+    def get(self, command: Union[str, Iterable[str]]) -> ResponseT:
         pass
 
     @abstractmethod
-    def delete_command(self, command: str):
+    def delete_command(self, command: Union[str, Iterable[str]]):
         pass
 
     @abstractmethod
-    def delete_commands(self, commands):
+    def delete_commands(self, commands: List[Union[str, Iterable[str]]]):
         pass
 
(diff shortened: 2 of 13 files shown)


tests/test_asyncio/test_cache.py
```diff
@@ -228,3 +233,43 @@ async def test_cache_decode_response(self, r):
...
+    async def test_get_from_cache(self, local_cache, master):
+        await master.set("foo", "bar")
+        # get key from redis and save in local cache
+        assert await master.get("foo") == b"bar"
+        # get key from local cache
+        assert local_cache.get(("GET", "foo")) == b"bar"
+        # change key in redis (cause invalidation)
+        await master.set("foo", "barbar")
+        # send any command to redis (process invalidation in background)
+        await master.ping()
```
vladvildanov, line 251: This step is not needed if you check socket each time before execute command, I think this approach will be more safe and usable from user perspective WDYT?

vladvildanov: Well, it's not about GET command or any other, the problem is that we need to send an additional read command to server to make sure that we process an invalidation, which leads to data inconsistency in case you're not interacting with server. Another check each time we're executing command add some overhead but it's the only way to make sure that we work with system in correct state, apart from sending additional command before it which also add overhead.

For example:

```
client1 = new Redis(cache=true)
client2 = new Redis(cache=true)

client1.set("key", "value")
client1.get("key") // Executed against server, value is cached
client1.get("key") // Executed against cache

client2.set("key", "new_value") // Client2 updates key tracked by Client1
client1.get("key") // Client1 does not handle invalidation sent by server, because value in cache consider as valid
```

gerzse: You don't _need_ to send an additional command. Invalidations are handled even before a GET call, so in your example the last line will handle the invalidation from the server. Just that the test cases written so far want to emphasize that invalidations are handled on other commands automatically. I will add a test with your scenario, now that we discuss it makes sense to have one.

vladvildanov: Ah, okay. So it's actually the same what I mean

vladvildanov (approved):

dev_requirements.txt
```diff
@@ -1,5 +1,6 @@
 click==8.0.4
 black==24.3.0
+cachetools
```
dvora-h, line 3: I can't find where do you use this?

redis/_cache.py
```diff
@@ -224,12 +222,17 @@ def __init__(
         self.key_commands_map = defaultdict(set)
         self.commands_ttl_list = []
 
-    def set(self, command: str, response: ResponseT, keys_in_command: List[KeyT]):
+    def set(
+        self,
+        command: Union[str, Iterable[str]],
```
dvora-h, line 227: Why it need to be also Iterable[str]?

tests/test_cache.py
```diff
@@ -11,16 +16,28 @@
...
         # client.flushdb()
 
 
+@pytest.fixture()
+def local_cache():
+    return _LocalCache()
+
+
 @pytest.mark.skipif(HIREDIS_AVAILABLE, reason="PythonParser only")
+@pytest.mark.onlynoncluster
```
dvora-h, line 38: Why? I think we want some of the tests in this class to also run in the cluster

tests/test_cache.py
```diff
@@ -39,8 +56,12 @@ def test_get_from_cache(self, r, r2):
...
         assert r.get("foo") == b"barbar"
 
-    @pytest.mark.parametrize("r", [{"cache": _LocalCache(max_size=3)}], indirect=True)
-    def test_cache_max_size(self, r):
+    @pytest.mark.parametrize(
+        "r",
+        [{"cache": _LocalCache(max_size=3)}],
+        indirect=True,
+    )
+    def test_cache_lru_eviction(self, r):
```
dvora-h, line 63: same as in the async test (naming)

redis/asyncio/connection.py
```diff
@@ -730,6 +730,27 @@ def _add_to_local_cache(
         ):
             self.client_cache.set(command, response, keys)
 
+    def flush_cache(self):
+        try:
+            if self.client_cache:
```
dvora-h, line 735: Why do you need the `if` inside the `try` block?

tests/test_asyncio/test_cache.py
```diff
@@ -36,7 +41,7 @@ async def test_get_from_cache(self, r, r2):
         assert await r.get("foo") == b"barbar"
 
     @pytest.mark.parametrize("r", [{"cache": _LocalCache(max_size=3)}], indirect=True)
-    async def test_cache_max_size(self, r):
+    async def test_cache_lru_eviction(self, r):
```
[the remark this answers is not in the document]
gerzse: There was a discussion about eviction policies. Max size is always bound to an eviction policy. It felt cleaner to express this as "test that the default eviction policy is LRU".

redis/_cache.py
gerzse: Because in all unit tests tuples are being sent for this parameter.

tests/test_cache.py
gerzse: Yes, same reason as above.

dev_requirements.txt
gerzse: I added a test for fully custom cache structure (so we don't even reuse the _LocalCache). For sake of argument I based that structure on `cachetools`.

tests/test_asyncio/test_cache.py
```diff
@@ -14,9 +14,14 @@ async def r(request, create_redis):
...
 
+@pytest_asyncio.fixture()
+async def local_cache():
+    yield _LocalCache()
+
+
+@pytest.mark.onlynoncluster
 @pytest.mark.skipif(HIREDIS_AVAILABLE, reason="PythonParser only")
 class TestLocalCache:
-    @pytest.mark.onlynoncluster
```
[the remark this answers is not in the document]
gerzse: I moved the decorator to class level.

redis/asyncio/connection.py
```diff
@@ -696,7 +696,7 @@ def _cache_invalidation_process(
         and the second string is the list of keys to invalidate.
         (if the list of keys is None, then all keys are invalidated)
         """
-        if data[1] is not None:
+        if data[1] is None:
```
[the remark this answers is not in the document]
gerzse: Surfaced due to the new tests, basically luck.

redis/asyncio/connection.py
gerzse: The `try` is meant to compensate for the `if`? You mean we could leave the `if` without the `try`? I would find that cleaner.

tests/test_cache.py
gerzse: Good point. I'll revert this.

tests/test_asyncio/test_cache.py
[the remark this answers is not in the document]
dvora-h: Could you refer me to this discussion?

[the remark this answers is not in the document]
gerzse: The discussion was mostly online, in the weekly call. Another argument: all tests for eviction strategies implicitly test the max size of the cache.

tests/test_asyncio/test_cache.py
```diff
@@ -142,7 +162,6 @@ async def test_cache_return_copy(self, r):
         check = cache.get(("LRANGE", "mylist", 0, -1))
         assert check == [b"baz", b"bar", b"foo"]
 
-    @pytest.mark.onlynoncluster
```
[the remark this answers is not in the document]
gerzse: Moved back.

tests/test_asyncio/test_cache.py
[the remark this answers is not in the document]
gerzse: Moved back.

tests/test_cache.py
gerzse: Reverted.

redis/_cache.py
gerzse: Not in unit tests, that is for the `get`. In Connection we have this signature for example:

    def _add_to_local_cache(
        self, command: Tuple[str], response: ResponseT, keys: List[KeysT]
    ):

dvora-h: So maybe we need only Iterable[str]? I'm not sure...

gerzse: This is a broader discussion, I think we can leave it with both for the beta release. In general I think we should not expose to the outside world the way we represent commands in the cache. We should only expose the concept of keys externally, and do the mapping to commands and back only internally, and then we have full freedom of representation.

gerzse: I used `Sequence` instead of `Iterable` and `Tuple`. I think it is the most expressive one.

redis/asyncio/connection.py
CBerndt-Work: The if-statement will prevent the call to flush/delete_command/invalidate_key that would cause the AttributeError if client_cache is falsy. The try-catch-block will also catch an AttributeError if client_cache is truthy but doesn't provide the method.
I'm missing to much context to say which is more desirable. But if it can be assumed, that a thruthy client_cache always also provides the methods, then the if-statement and try-catch are redundant and it becomes a question of style. You can read more on that here https://realpython.com/python-lbyl-vs-eafp/
Best regards, a random passerby XD

gerzse: Thanks @CBerndt-Work for your nice explanation. The article is very well written. I realized that I am by now too conditioned towards LBYL, from other languages. Especially here, using `AttributeError`, which signals issues with the code structure, not with data. So I went for the `if`s.

dvora-h (approved):

PR #3218, opened 2024-04-26
Hash field expiration commands

gerzse:
### Pull Request check-list

### Description of change

Hash field expiration commands


redis/commands/core.py
```diff
@@ -5081,6 +5081,485 @@ def hstrlen(self, name: str, key: str) -> Union[Awaitable[int], int]:
...
+        self,
+        name: KeyT,
+        seconds: ExpiryT,
+        *fields: str,
+        nx: bool = False,
+        xx: bool = False,
+        gt: bool = False,
+        lt: bool = False,
+    ) -> ResponseT:
+        """
```
tishun, line 5094: BTW on the topic of docstrings : as mentioned by [this discussion in Stackoverflow](https://stackoverflow.com/questions/3898572/what-are-the-most-common-python-docstring-formats) we can use the [Pyment](https://github.com/dadadel/pyment/blob/master/README.rst) tool to generate new reST documentation or convert existing one from other formats. This is probably going to help in the general direction of consistency.

IMHO it is a good idea to continue with the reST format you are using here chiefly because of two things - we can generate documentation using [Sphinx](http://sphinx-doc.org/) and it is generally recommended by [PEP 287](https://www.python.org/dev/peps/pep-0287)

gerzse: After trying out some more things, I think that the Google style documentation would be better suited. The main reason is that it supports bullet lists in a straightforward way. With reST I could not get a bullet list properly formatted in the PyCharm documentation popup.

Also it seems to me that a great deal of the existing documentation is a flavor of the Google style. It would be the least effort to convert it to the proper Google format.

And, after enabling the `napoleon` preprocessor for Sphinx, this Google style documentation also gets converted well (after some additional fine tuning related to typing).

tishun: Makes sense.

vladvildanov (approved):

PR #3220, opened 2024-04-30
fix scan iter command issued to different replicas

### LINKED ISSUE #3197: scan_iter family commands gives inconsistent result when using Sentinel connection pool
redis-py 4.5.0

Python 3.10

`scan_iter` family commands (`scan_iter`, `sscan_iter`, `hscan_iter`, `zscan_iter`) might give inconsistent result when the client is created using a connection pool, and when there are multiple concurrent requests. 

Assume we have this setup
- 2 replicas, host A and host B
- use `SentinelConnectionPool` to manage connections to different server
- 2 concurrent `scan_iter` commands, in which each will issue multiple `scan` commands. `scan` commands issued by these `scan_iter` commands are labelled `scan (1)` and `scan (2)` below.

What might happen is: 
1. `scan (1)` is issued
2. `scan (1)` gets connection from the pool
     - The pool is empty so it creates a new connection
     - For sentinel connection pool, creating a new connection means getting the next replica in the `connection_pool.rotate_slaves` rotation. 
     - Since this can return any replicas on rotation, let's say it arbitrarily connects to host A
3. `scan (1)` executed at host A
4. `scan (2)` is issued in the meantime
5. `scan (2)` gets connection from the pool
     - The pool is empty (there was 1 connection created but it's still in use)so it creates a new connection
     - Get the next replica in the `connection_pool.rotate_slaves` rotation. 
     - Since this can return any replicas on rotation, let's say it arbitrarily connects to host B
6. `scan (2)` executed on host B 
7. `scan (1)` is finished. Connection to host A is put back to the pool
8. `scan (2)` is finished. Connection to host B is put back to the pool
9. `scan (1)` gets connection from connection pool, it gets the connection to host B (since connection pool will just `pop()` the last element from the available connections) 
10. `scan (1)` is executed on host B

Step 9 is the bug. All `scan` commands coming from the same `scan_iter` command needs to go to the same replica. This is because the 'state' of the `scan_iter` command is stored in the cursor and different replicas will store keys in a different order.
Hence, **if we use the cursor from host A to do a scan on host B, we'll get an inconsistent result.**

There are 3 different base implementations of a connection pool, `ConnectionPool`, `SentinelConnectionPool` and `BlockingConnectionPool`. All of them does something similar when getting a new connection from the pool. It creates a 'dummy' connection object, and call `connection.connect()`, which will actually connect to the intended replica. 

There are 4 different implementations of a connection, `Connection`, `SSLConnection`, `SentinelManagedConnection`, and `SentinelManagedSSLConnection`. 
- For `SentinelManagedConnection` and `SentinelManagedSSLConnection`, this is fixable by making `SentinelConnectionPool` maintaining an id of the scan iter command to the host it has previously issued command to 
- For `Connection` and `SSLConnection`, `connection.connect()`, will depend on the impl of the connection class' `.connect` but by default will connect to `self.host` and `self.port` of the connection.

agnesnatasya:
### Pull Request check-list

   - [x] lint passes
   - [x] test passes on `Ubuntu 22.04`, `python3.11.2`
```
Starting Redis tests
= 2352 passed, 1235 skipped, 849 deselected, 29 xpassed, 347 warnings in 163.65s (0:02:43) =
Waiting for 6 cluster nodes to become available
All nodes are available!
= 1571 passed, 1494 skipped, 1396 deselected, 4 xpassed, 244 warnings in 728.56s (0:12:08) =
```
- [x] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)? all succeeded in my fork
- [x] Is the new or changed code fully tested? added tests coverage
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? bugfix only
  - [x] I added some docstring to `redis.asyncio.sentinel` module. I wanted to check if the rST syntax is correct, but this module is not included in the builddir's index, so I think it's no-op. 
- [x] Is there an example added to the examples folder (if applicable)? bugfix only
- [x] Was the change added to CHANGES file? bugfix only

### fix scan iter command issued to different replicas

Fixes #3197. See linked issue for full description of the bug


agnesnatasya:
Some of the CI tests fails flakily with a segmentation fault - if i rerun the CI in my fork, it would work most of the time. But I don't have permission to rerun it in the main repository.
Do you know whether this is expected? @gerzse 
Thank you!!

redis/asyncio/sentinel.py
```diff
@@ -167,6 +193,92 @@ async def rotate_slaves(self) -> AsyncIterator:
...
+        if server_host is None or server_port is None:
+            try:
+                connection = self._available_connections.pop()
+            except IndexError:
+                connection = self.make_connection()
+        # If this is not the first scan request of the iter command
+        else:
+            # Check from the available connections, if any of the connection
+            # is connected to the host and port that we want
+            for available_connection in self._available_connections.copy():
```
gerzse, line 233: What if there are many connections? This linear search might slow down things.

redis/asyncio/sentinel.py
```diff
@@ -122,6 +147,7 @@ def __init__(self, service_name, sentinel_manager, **kwargs):
         self.sentinel_manager = sentinel_manager
         self.master_address = None
         self.slave_rr_counter = None
+        self._request_id_to_replica_address = {}
```
gerzse, line 150: Would it make sense to use a more general name, e.g. instead of `request_id` use `context_id`? Would it express better the fact that you are basically trying to run several commands in the same context?

redis/asyncio/sentinel.py
```diff
@@ -167,6 +193,92 @@ async def rotate_slaves(self) -> AsyncIterator:
...
+                await connection.connect()
+                if await connection.can_read_destructive():
+                    raise ConnectionError("Connection not ready") from None
+        except BaseException:
+            # Release the connection back to the pool so that we don't
+            # leak it
+            await self.release(connection)
+            raise
+        # Store the connection to the dictionary
+        self._request_id_to_replica_address[iter_req_id] = (
```
gerzse, line 276: When would these entries be removed from the dict? So that it does not grow indefinitely.

gerzse (commented):
Thanks, this is a valid point. I left some comments in the changes, and I have one more general comment: the implementation covers now the async code. Did you leave out the sync code intentionally, or would it make sense to also adapt that one?

gerzse:
> Some of the CI tests fails flakily with a segmentation fault - if i rerun the CI in my fork, it would work most of the time. But I don't have permission to rerun it in the main repository. Do you know whether this is expected? @gerzse Thank you!!

Those segmentation faults have been there for a long time, to be honest I have no idea why they happen. They are so annoying. Eventually I'll spend some time trying to dig deeper.

redis/asyncio/sentinel.py
agnesnatasya: Yeah good point, I created a separate `cleanup` method that will be called at the end of scan iter family commands

agnesnatasya (commented):
Thank you for reviewing this @gerzse, appreciate it! Great points, I've addressed your comments. This is also a bug in the sync code, I've added the fix to the sync code as well!

redis/connection.py
```diff
@@ -1185,23 +1215,7 @@ def get_connection(self, command_name: str, *keys, **options) -> "Connection":
...
-            # (if caching enabled the connection will not always be ready
-            # to send a command because it may contain invalidation messages)
-            try:
-                if connection.can_read() and connection.client_cache is None:
-                    raise ConnectionError("Connection has data")
-            except (ConnectionError, OSError):
-                connection.disconnect()
-                connection.connect()
-                if connection.can_read():
-                    raise ConnectionError("Connection not ready")
```
agnesnatasya, line 1204: I refactored this to `ensure_connection` so that the subclass can just directly call this. There's also a similar block of code in `BlockingConnectionPool`, except that it doesn't have the `and connection.client_cache is None` clause. But it looks to me like there's nothing special with `client_cache` in `BlockingConnectionPool`
Should we refactor that out as well, or is the absence of that clause an intentional distinction for `BlockingConnectionPool`?

tests/test_sentinel_managed_connection.py
```diff
@@ -0,0 +1,209 @@
+from typing import Tuple
+from unittest import mock
+
+import pytest
+from redis import Redis
+from redis.sentinel import Sentinel, SentinelConnectionPool, SentinelManagedConnection
+from redis.utils import HIREDIS_AVAILABLE
+
+pytestmark = pytest.mark.skipif(HIREDIS_AVAILABLE, reason="PythonParser only")
```
agnesnatasya, line 9: I marked these test to ignore Hiredis. I'm not very familiar with Hiredis setup, but it looks like the hiredis' code paths doesn't really work well when we call connection-related operations like `can_read` (and as far as I understand it is unrelated to my change). Let me know if i need to not skip Hiredis

agnesnatasya (commented):
Sorry it took awhile for me to address your comments @gerzse, it had more changes than I expected, but this version should be ready for another round of review! Thank you very much!
I ran the CI workflow on my fork, it's passing except the tests that I believe is being fixed in #3324.

PR #3235, opened 2024-05-16
Fix typing for address_remap (fixes #3234)

### LINKED ISSUE #3234: Typing for address_remap is incorrect
**Version**: 5.0.4

**Platform**: Python 3.12.3 on Debian Bookworm

**Description**: Typing for `address_remap` is as separate args but it's actually a tuple

daveisfera:
This is just a typing change so it shouldn't impact the functionality of the code

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -275,7 +275,7 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
         ssl_min_version: Optional[ssl.TLSVersion] = None,
         ssl_ciphers: Optional[str] = None,
         protocol: Optional[int] = 2,
-        address_remap: Optional[Callable[[str, int], Tuple[str, int]]] = None,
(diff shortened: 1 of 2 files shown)


gerzse:
@daveisfera Thanks for this!

The CI fails on Python 3.8: `TypeError: Callable[args, result]: args must be a list. Got typing.Tuple[str, int]`.

Also from the docs at https://docs.python.org/3/library/typing.html#annotating-callable-objects 

> The subscription syntax must always be used with exactly two values: the argument list and the return type. The argument list must be a list of types, a [ParamSpec](https://docs.python.org/3/library/typing.html#typing.ParamSpec), [Concatenate](https://docs.python.org/3/library/typing.html#typing.Concatenate), or an ellipsis. The return type must be a single type.

Maybe the accurate way to spell it is: `Callable[[Tuple[str, int]], Tuple[str, int]]`?

daveisfera:
You're right. I forgot to put the `Tuple` in a `list`, so that's fixed now

gerzse (approved):

PR #3244, opened 2024-05-24
Add extra tests for GEO search

gerzse:
### Pull Request check-list

### Description of change

Add more tests for GEO search, to cover the query operators `within`, `contains`, `intersects` and `disjoint`, for POINT and POLYGON, i.e. the currently supported shapes and operators.

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -2322,10 +2322,9 @@ def test_geoshape(client: redis.Redis):
     q2 = Query("@geom:[CONTAINS $poly]").dialect(3)
     qp2 = {"poly": "POLYGON((2 2, 2 50, 50 50, 50 2, 2 2))"}
     result = client.ft().search(q1, query_params=qp1)
-    assert len(result.docs) == 1
-    assert result.docs[0]["id"] == "small"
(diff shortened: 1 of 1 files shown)


tests/test_search.py
```diff
@@ -2284,3 +2328,16 @@ def test_geoshape(client: redis.Redis):
...
+    """
+    Make sure the result of a geo search is as expected, taking into account the RESP
+    version being used.
+    """
+    if is_resp2_connection(client):
+        assert set([doc.id for doc in result.docs]) == set(expected_doc_ids)
+        assert result.total == len(expected_doc_ids)
+    else:
+        assert set([doc["id"] for doc in result["results"]]) == set(expected_doc_ids)
+        assert result["total_results"] == len(expected_doc_ids)
```
vladvildanov, line 2343: So for RESP2 the response type is an object and for RESP3 array?

gerzse: While checking this, I found one existing test that was not asserting well for RESP3, I updated it. Most probably the test does not even run with RESP3, since there were no failures in the past. A thing to check, but I'll leave that for a separate PR.

vladvildanov (approved):

PR #3246, opened 2024-05-24
Use standard Redis docker image in CI

gerzse:
Adapt the tests to use the standard Redis docker image where possible, instead of using the Redis Stack image in all places. This way we can run the CI, at least in theory, against different versions of Redis and Redis Stack dockers.

### DIFF
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -2450,6 +2450,7 @@ class TestClusterRedisCommands:
             except Exception:
                 pass
 
+    @pytest.mark.redismod
     @skip_if_server_version_lt("7.1.140")
     def test_tfunction_load_delete(self, r):
         r.gears_refresh_cluster()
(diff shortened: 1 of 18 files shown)


tishun:
LGTM

tests/test_cluster.py
```diff
@@ -2450,6 +2450,7 @@ def try_delete_libs(self, r, *lib_names):
             except Exception:
                 pass
 
+    @pytest.mark.redismod
```
uglide, line 2453: You basically skipped those tests with `redismod` mark, because cluster tests exclude `redismod` https://github.com/redis/redis-py/blob/0d47d6527a10fb70f8fcfdf8df69ae3b11ec92ef/tasks.py#L73

It's not critical for these tests because T&F is deprecated anyway, but be careful with it in the future.

vladvildanov (approved):

PR #3252, opened 2024-05-28
Add typings to the `redis.Redis` class 

bitterteriyaki:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -4,7 +4,7 @@ import threading
 import time
 import warnings
 from itertools import chain
-from typing import Any, Callable, Dict, List, Optional, Type, Union
+from typing import Any, Callable, Dict, List, Optional, Type, Union, Self
 
 from redis._cache import (
     DEFAULT_ALLOW_LIST,
@@ -105,7 +105,7 @@ class Redis(RedisModuleCommands, CoreCommands, SentinelCommands):
     """
 
     @classmethod
-    def from_url(cls, url: str, **kwargs) -> "Redis":
+    def from_url(cls, url: str, **kwargs: Any) -> Self:
         """
         Return a Redis client object configured from the given URL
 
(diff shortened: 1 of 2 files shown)


bitterteriyaki:
In some parameters, I don't know what is the proper type. I'll fix these later.

bitterteriyaki:
> Currently redis-py supports Python 3.8 up to and including 3.11, and support for 3.12 is planned.

Hi, @gerzse. Thank you for the answer. Since Python 3.7 was dropped, It would be nice to update the [documentation](https://redis-py.readthedocs.io/en/stable/index.html).

> In general it's best to use language constructs that work in all these versions. For example Union[str, None] instead of str | None, which is available only since 3.10.

Alright. I'll be changing this.

> The Self construct can be made available in versions earlier than 3.11?

According to the [`typing` documentation](https://docs.python.org/3/library/typing.html#typing.Self), no. I'll also changing this.

bitterteriyaki:
Instead of `typing.Self`, I am using `from __future__ import annotations` which is available for old Python versions and is a library for "future" which is not available for these versions.

redis/client.py
```diff
@@ -105,7 +107,7 @@ class Redis(RedisModuleCommands, CoreCommands, SentinelCommands):
     """
 
     @classmethod
-    def from_url(cls, url: str, **kwargs) -> "Redis":
```
Vulwsztyn, line 108: Any doesn't really help anyone

redis/client.py
```diff
@@ -156,66 +158,61 @@ class initializer. In the case of conflicting arguments, querystring
         return client
 
     @classmethod
-    def from_pool(
```
Vulwsztyn, line 159: why remove the typing?

redis/client.py
```diff
@@ -345,18 +342,18 @@ def __repr__(self) -> str:
             f"({repr(self.connection_pool)})>"
         )
 
-    def get_encoder(self) -> "Encoder":
+    def get_encoder(self) -> Encoder:
         """Get the connection pool's encoder"""
         return self.connection_pool.get_encoder()
 
-    def get_connection_kwargs(self) -> Dict:
+    def get_connection_kwargs(self) -> Dict[str, Any]:
```
Vulwsztyn, line 349: not really any more clear than the previous version

bitterteriyaki:
Hi, @Vulwsztyn. This PR is still a draft and it is not complete.

redis/client.py
Vulwsztyn: ?

petyaslavova:
Closing this PR as it has been marked as draft for nearly a year, and most of the proposed changes have already been incorporated into the codebase.
If you’d like to contribute any remaining changes, feel free to open a new PR.

PR #3271, opened 2024-06-07
DOC-3801 move code samples to master branch

andy-stark-redis:
### Pull Request check-list

### Description of change

The code samples for the main documentation pages are kept in this repo but they were previously in a separate branch (emb-examples). The plan is to move them to the master branch and run tests on them so they get checked frequently.


doctests/search_vss.py
```diff
@@ -0,0 +1,340 @@
...
+from redis.commands.search.query import Query
+from sentence_transformers import SentenceTransformer
+
+# STEP_END
+
+# STEP_START get_data
+URL = ("https://raw.githubusercontent.com/bsbodden/redis_vss_getting_started"
+       "/main/data/bikes.json"
+       )
+response = requests.get(URL, timeout=10)
```
akx, line 33: Maybe this data should be included here instead of being pulled from another repo that can change at any moment. That'd drop the `requests` dependency.

doctests/search_vss.py
```diff
@@ -0,0 +1,340 @@
...
+    queries_table.sort_values(
+        by=["query", "score"], ascending=[True, False], inplace=True
+    )
+    queries_table["query"] = queries_table.groupby("query")["query"].transform(
+        lambda x: [x.iloc[0]] + [""] * (len(x) - 1)
+    )
+    queries_table["description"] = queries_table["description"].apply(
+        lambda x: (x[:497] + "...") if len(x) > 500 else x
+    )
+    queries_table.to_markdown(index=False)
```
akx, line 299: Given that this is "optional" and nothing is actually even _done_ with the markdown return value, this should probably be removed so `pandas` and `tabulate` aren't required to run the tests.

doctests/search_vss.py
andy-stark-redis: @akx Thanks for the review - very much appreciated! I did wonder myself if it would be better to add the JSON to the code. However, I think the [tutorial](https://redis.io/docs/latest/develop/get-started/vector-database/#1-fetch-the-demo-data) gets the data from a URL to illustrate the kind of task the user might want to do in real life. Also, the file on the Github repo was deliberately created for this tutorial, so it shouldn't change unless we want it to. I'd say we should leave it like this for now but I'll find out if we've got any feedback about this page and see if the users have any problems with the HTTP request.

doctests/search_vss.py
andy-stark-redis: @akx Again, I think it's trying to show the user a real world example. The code sample doesn't use the Markdown explicitly but the output Markdown table is actually shown in the [tutorial page](https://redis.io/docs/latest/develop/get-started/vector-database/#2-perform-a-k-nearest-neighbors-knn-query). I think the doc text on the page could probably use some improvement but that's a separate PR on the docs repo.

akx: `to_markdown()` returns the table as a string, so this would just do a lot of work to dataframe-ify things, group them, transform them... and then throw the result away. I really don't think it's a great idea to spend CI CPU cycles (and contribute to our planet burning in a minor way) for that.

I assume these are somehow extracted from something like a notebook session, where the value of the last expression might actually be shown?

andy-stark-redis: OK, I've added a return value to the function and then the main code prints that out when the function is called. The doc page shows the Markdown table as rendered, but it looks like it only shows the first query currently (this is an issue with the doc page, not the code sample - I'll fix that separately).

akx: I'm sorry, I'm not following. Why bother with that either? Is there something asserting on the output?

If there is, if the idea of these being here (and e.g. being run as part of CI?) is to test that "typical" Redis-utilizing code works, then it would break if `pandas` or `tabulate` decide to change their (human-oriented) output.

If there isn't, I don't see a point in doing the work to human-format data that no one will care about.

(On that note, running `sentence-transformers` in CI also sounds like a bad idea, since it will have to download the transformers models, then run embedding computation on CPU, ...)

andy-stark-redis: I think you might be misunderstanding what this code is for :-) The code samples in this folder are included in the docs (eg, https://redis.io/docs/latest/develop/get-started/vector-database/). Their purpose is to explain to the user how to do stuff with Redis and to supply bits of code for copy/paste. The reason for putting them in the client repo as tests is to check that the sample code works against the current version of the client, not to act as real world example tests for the client.

gerzse (approved):

doctests/search_vss.py
gerzse: @akx , these tests were living in a separate branch until now, this is a first step to put them as close as possible to the code they are documenting. Right now there is no CI in this repo for them, I guess we can add something in other PRs. The purpose so far is educational, and we want to decrease maintenance effort. I'll merge this as it is.

PR #3281, opened 2024-06-15
Refactor(Typing): Add and update type annotations for core Redis commands

ElayGelbart:
### Pull Request check-list

### Overview
This pull request introduces a comprehensive update and correction of the type annotations 
for all core Redis commands. The primary focus of this effort was to enhance the precision 
and usefulness of the type information based on the official Redis documentation. 
The task involved meticulously revisiting each command's return type to ensure 
accuracy and completeness.

### Key Changes
- **Standardization of Return Types**: Many core commands lacked proper return type 
  annotations.
  some used overly generic types, while others did not accurately reflect 
  all possible return values. This update corrects these issues, providing a clear and 
  reliable type for each function based on the Redis command specifications.
- **Correction of Types**: Adjustments were made to type annotations to 
  accurately include possible `nil` and other replies, addressing previous oversights.
- **Optional Arguments**: Some function arguments were incorrectly marked as non-optional. 
  This has been corrected, ensuring that optional parameters are appropriately typed.
- **Handling Awaitable Types**: Discrepancies in the typing of awaitable and non-awaitable 
  unions in return types have been standardized - some commands mention it and some not, now all commands is handled in the same direction of union sync and async responses.
- **Usage of 'OK' Type**: The 'OK' type annotation was missing in certain commands 
  where it should have been applied. This has now been consistently integrated where applicable.
- **Class Initializations**: Instances of classes being initialized more than once have 
  been corrected.
- **Compatibility Adjustments**: Replaced unsupported below Python 3.8 'list' type usage with 
  'typing.List' to maintain compatibility across supported versions.

### Challenges and Acknowledgements
* The task involved extensive review and application of type annotations one by one 
from the Redis documentation. While I have made effort to ensure accuracy, some types might still need adjustments. This effort significantly improves the clarity and reliability of the codebase, even though some 
minor discrepancies might remain.

* Binary response of 1\0 is marked as integer, need to be corrected after the progression of this PR.

* Async\Sync - Generic way to handle the differences between async commands and sync commands to get correct typing, for now I didn't want to change logic and only change the typing and continue the same type decision as before.

* core.py - 6000+ lines of code made it hard to edit and refactor, different classes of core commands can be separated to multiple files to enhance future code contributions.

### DIFF
--- a/redis/commands/core.py
+++ b/redis/commands/core.py
@@ -1,12 +1,10 @@
-# from __future__ import annotations
-
 import datetime
 import hashlib
 import warnings
 from typing import (
     TYPE_CHECKING,
+    Any,
     AsyncIterator,
-    Awaitable,
     Callable,
     Dict,
     Iterable,
@@ -16,16 +14,18 @@ from typing import (
     Mapping,
     Optional,
     Sequence,
-    Set,
     Tuple,
     Union,
 )
 
 from redis.exceptions import ConnectionError, DataError, NoScriptError, RedisError
 from redis.typing import (
+    OKT,
     AbsExpiryT,
     AnyKeyT,
+    ArrayResponseT,
     BitfieldOffsetT,
+    BulkStringResponseT,
     ChannelT,
     CommandsProtocol,
     ConsumerT,
@@ -33,6 +33,7 @@ from redis.typing import (
     ExpiryT,
     FieldT,
     GroupT,
+    IntegerResponseT,
     KeysT,
     KeyT,
     PatternT,
@@ -56,7 +57,9 @@ class ACLCommands(CommandsProtocol):
     see: https://redis.io/topics/acl
     """
 
-    def acl_cat(self, category: Union[str, None] = None, **kwargs) -> ResponseT:
+    def acl_cat(
+        self, category: Union[str, None] = None, **kwargs
+    ) -> ResponseT[ArrayResponseT]:
         """
         Returns a list of categories or commands within a category.
 
@@ -66,10 +69,12 @@ class ACLCommands(CommandsProtocol):
 
         For more information see https://redis.io/commands/acl-cat
         """
-        pieces: list[EncodableT] = [category] if category else []
+        pieces: List[EncodableT] = [category] if category else []
         return self.execute_command("ACL CAT", *pieces, **kwargs)
 
-    def acl_dryrun(self, username, *args, **kwargs):
+    def acl_dryrun(
+        self, username, *args, **kwargs
+    ) -> ResponseT[Union[BulkStringResponseT, OKT]]:
         """
         Simulate the execution of a given command by a given ``username``.
 
@@ -77,7 +82,7 @@ class ACLCommands(CommandsProtocol):
         """
         return self.execute_command("ACL DRYRUN", username, *args, **kwargs)
 
-    def acl_deluser(self, *username: str, **kwargs) -> ResponseT:
+    def acl_deluser(self, *username: str, **kwargs) -> ResponseT[IntegerResponseT]:
         """
         Delete the ACL for the specified ``username``s
 
@@ -85,7 +90,9 @@ class ACLCommands(CommandsProtocol):
         """
         return self.execute_command("ACL DELUSER", *username, **kwargs)
 
-    def acl_genpass(self, bits: Union[int, None] = None, **kwargs) -> ResponseT:
+    def acl_genpass(
+        self, bits: Union[int, None] = None, **kwargs
+    ) -> ResponseT[BulkStringResponseT]:
         """Generate a random password value.
         If ``bits`` is supplied then use this number of bits, rounded to
         the next multiple of 4.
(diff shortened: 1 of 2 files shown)


redis/typing.py
```diff
@@ -32,7 +32,14 @@
 PatternT = _StringLikeT  # Patterns matched against keys, fields etc
 FieldT = EncodableT  # Fields within hash tables, streams and geo commands
 KeysT = Union[KeyT, Iterable[KeyT]]
-ResponseT = Union[Awaitable[Any], Any]
+OldResponseT = Union[Awaitable[Any], Any]  # Deprecated
```
gerzse, line 35: This is not used anymore, I would say just get rid of it.

redis/typing.py
```diff
@@ -32,7 +32,14 @@
...
+AnyResponseT = TypeVar("AnyResponseT", bound=Any)
+ResponseT = Union[AnyResponseT, Awaitable[AnyResponseT]]
+OKT = Literal[True]
+ArrayResponseT = List
```
gerzse, line 39: Two aspects here.

First, `List` is itself generic, wouldn't it be cleaner to keep that into account? I guess in most places it would end up being `ArrayResponseT[Any]`, but still it would feel more correct.

Second, why define aliases to single types? You can just use `List` and it would be easier on the reader, I think.

redis/typing.py
```diff
@@ -32,7 +32,14 @@
...
+IntegerResponseT = int
```
gerzse, line 40: Same point here, just use `int`, why alias it? At first sight this is a bit of speculative generality.

redis/typing.py
```diff
@@ -32,7 +32,14 @@
...
+NullResponseT = type(None)
```
gerzse, line 41: This seems not used at all.

redis/typing.py
```diff
@@ -32,7 +32,14 @@
...
+BulkStringResponseT = str
```
gerzse, line 42: Again, why not use `str` directly?

gerzse (changes requested):
This is pretty impressive, @ElayGelbart ! I left some comments, more about principles I think.

ElayGelbart:
@gerzse Thanks for the response and the comments ! 

Thank you for your insights on the PR. I understand the concerns about introducing generics and appreciate the complexities it brings.

The primary intent of transitioning from a generic ANY to **List** types and the introduction of **BulkStringResponseT** and **IntegerResponseT** is to make our responses more predictable and iterables clear, which is crucial for robust type development. This is just a starting point, and I recognize it's far from perfect but necessary for gradual, meaningful improvements.

Regarding the differences between RESP2 and RESP3 and the decode options, this PR lays the groundwork for future enhancements. The use of IntegerResponseT as a Boolean in certain contexts is an interim solution, and I'm open to refining these aspects based on your feedback.

I've invested significant effort in these changes and am truly grateful for your time in reviewing them. If the consensus is that generics are redundant at this stage, I am willing to adjust the approach accordingly. I deeply respect the community’s direction and decisions. Thank you for considering these enhancements.

gerzse:
> @gerzse Thanks for the response and the comments !
> 
> Thank you for your insights on the PR. I understand the concerns about introducing generics and appreciate the complexities it brings.
> 
> The primary intent of transitioning from a generic ANY to **List** types and the introduction of **BulkStringResponseT** and **IntegerResponseT** is to make our responses more predictable and iterables clear, which is crucial for robust type development. This is just a starting point, and I recognize it's far from perfect but necessary for gradual, meaningful improvements.
> 
> Regarding the differences between RESP2 and RESP3 and the decode options, this PR lays the groundwork for future enhancements. The use of IntegerResponseT as a Boolean in certain contexts is an interim solution, and I'm open to refining these aspects based on your feedback.
> 
> I've invested significant effort in these changes and am truly grateful for your time in reviewing them. If the consensus is that generics are redundant at this stage, I am willing to adjust the approach accordingly. I deeply respect the community’s direction and decisions. Thank you for considering these enhancements.

@ElayGelbart This PR is a huge step forward, and personally I think we should merge it. I see nobody else voiced an opinion from the community.

I think I understand your intention for introducing `IntegerResponseT` and `BulkStringResponseT`. My though is: what if `IntegerResponseT` will always be `int`? What if no other type will ever be assimilated to an "integer response"? Does it make sense to have this custom type name introduced at this early stage?

From my point of view generics are definitely not redundant, but the above mentioned type aliases might be. In any case, we must merge this contribution, to take us further in the direction of typing hints alignment. The day when we can run `mypy` and get no errors is far away, but we'll get closer to it.

Besides the RESP2 vs RESP3 differences you mention, another point to consider in the future is the difference between sync and async. In some places the `Union[T, Awaitable[T]]` does not fit either in sync code, for example because `Awaitable` is not iterable, or in async code, because `T` is not awaitable. Big changes, but let's thing about them later.

ElayGelbart:
any updates ?

redis/typing.py
Vulwsztyn: for clarity, it isn't just any `str` it is a `BulkStringResponseT` even better would be a uage of newtype https://docs.python.org/3/library/typing.html#newtype

redis/commands/core.py
```diff
@@ -5016,8 +4998,8 @@ def hset(
         key: Optional[str] = None,
         value: Optional[str] = None,
```
ManuCorrea, line 4999: Due to the encoder class and its encode function:
https://github.com/redis/redis-py/blob/1cbb97ecd7a0bf87aec332a257b589919385d90d/redis/_parsers/encoders.py#L14
 inputs of values can be more than just string, I think something like the following would be more accurate:
`Union[str, float, int, bool, bytes, memoryview]`

Graeme22:
This PR is great! Any change it'll get merged soon/anything I can do to help?
My code is currently littered with `# type: ignore`s on almost every redis function, this would solve it!

PR #3290, opened 2024-06-25
Fixes CAE-333

dmaier-redislabs:
The bug uncovered that the init method of the base class did override the initialization of the socket_timeout parameter.

This was fixed by moving `super().__init__(**kwargs)` before setting the timeout and path properties.

In addition, a test case was added.

### DIFF
--- a/tests/test_connect.py
+++ b/tests/test_connect.py
@@ -98,10 +98,10 @@ def test_tcp_ssl_tls12_custom_ciphers(tcp_address, ssl_ciphers):
     _assert_connect(conn, tcp_address, certfile=certfile, keyfile=keyfile)
 
 
-'''
+"""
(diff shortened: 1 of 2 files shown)


vladvildanov (approved):

tests/test_connect.py
```diff
@@ -98,6 +98,20 @@ def test_tcp_ssl_tls12_custom_ciphers(tcp_address, ssl_ciphers):
     _assert_connect(conn, tcp_address, certfile=certfile, keyfile=keyfile)
 
 
+'''
+Addresses bug CAE-333 which uncovered that the init method of the base
+class did override the initialization of the socket_timeout parameter.
+'''
+def test_unix_socket_with_timeout():
```
vladvildanov, line 105: Two blank-lines required between function definition and doc-blocks, as well as after the test

Issue #2772, opened 2023-05-22
Bytes Vectors from `r.hget` vs Bytes string returned from `r.ft().search(query="*")`


oq-9:
**Redis Python Lib Version**: version 4.5.5

**Redis Stack Version**: version 7.0.0

**Platform**: Python 3.10.6 and Ubuntu 22.04 

**Description**: Description of your issue, stack traces from errors and code that reproduces the issue

After storing a bunch of numpy vectors in bytes in HSETs and creating an index (FT), I am trying to retrieve all of the embeddings using FT.SEARCH with "*" query, however, the vector is returned in a string that differs from the bytes format I get when using HGET. I'll add a few line of code as an example:

```python
import redis
import os
import numpy as np

_redis_match_config = os.getenv("NQAI_REDIS_MATCH_CONFIG")
fake_vec = np.array([0.1,0.2,0.3,0.4])
r = redis.Redis(**_redis_match_config)
expert_hash = {"person_id":1, "vector_emb" : fake_vec.astype(np.float32).tobytes()}
r.hset("person:1", mapping=expert_hash)
index_name = "person"
person_prefix = f"{index_name}:"
vector_search_attributes = {"TYPE": "FLOAT32", "DIM": 4, "DISTANCE_METRIC": "COSINE"}
schema = (
                    TagField("person_id"),
                    VectorField("embeddings_bio", algorithm="HNSW", attributes=vector_search_attributes)
                    )

r.ft(index_name).create_index(fields=schema, definition=IndexDefinition(prefix=[person_prefix], index_type=IndexType.HASH))

byets_person_1 = r.hget("person:1", "vector_emb")
print(byets_person_1)
print(np.frombuffer(byets_person_1, dtype=np.float32))
```
```bash
> output : b"\xcd\xcc\xcc=\xcd\xccL>\x9a\x99\x99>\xcd\xcc\xcc>"
> output : array([0.1, 0.2, 0.3, 0.4], dtype=float32)
```
However, when I do:
```python
query = (
                    Query("*")
                    .return_fields("id", "vector_emb",)
                )
all_of = r.ft(index_name).search(query=query, query_params={}).docs
print(all_of[0]["vector_emb"])
print(all_of[0]["vector_emb"].encode("utf-32"))
print(np.frombuffer(bytes(all_of[0]["vector_emb"].encode("utf-32")), dtype=np.float32))
```
```bash
> output : "=L>>>"
[...]

Issue #2275, opened 2022-07-13
byte vector is incorrectly decoded as utf-8 string in ft result class


AnneYang720:
**Version**: 
```
$ pip3 show redis
Name: redis
Version: 4.3.4
```

**Platform**: 
Python 3.9.2 on Debian GNU/Linux 11

**Description**: The bytes is converted to string in the vector search results and there is an error in this conversion. The bytes including `b'\x80'` is converted to a wrong string.

**Example Code**
```Python
from redis import Redis
from redis.commands.search.field import VectorField
from redis.commands.search.query import Query

r = Redis(host='localhost',port=6379)
schema = (VectorField("v", "HNSW", {"TYPE": "FLOAT32", "DIM": 1, "DISTANCE_METRIC": "L2"}),)
r.ft().create_index(schema)

r.hset(f'{1}',mapping={'v':b'\x80\x00\x00\x00'})

q = Query("*=>[KNN 1 @v $vec AS vector_score]").dialect(2)
results = r.ft().search(q, query_params={"vec": b'\x80\x00\x00\x00'}).docs

for m in results:
    print(m.v)
    print('match emb =', bytes(m.v,'utf-8'))
```

The original bytes `b'\x80\x00\x00\x00'` is converted to string  `'\x00\x00\x00'`.

**Reason**
```Python
# /redis/commands/search/result.py
dict(
    dict(
        zip(
            map(to_string, res[i + fields_offset][::2]),
            map(to_string, res[i + fields_offset][1::2]),
        )
    )
)

# /redis/commands/search/_util.py
def to_string(s):
    if isinstance(s, str):
        return s
    elif isinstance(s, bytes):
        return s.decode("utf-8", "ignore") # here! 
    else:
        return s

```

colibrisson:
@AnneYang720 did you find a workaround?

kamyabzad:
What about using "backslashreplace" mode instead of "ignore"?

gaoyichuan:
@kamyabzad I think in this case, we should get the original bytes as result, rather than try any kind of unicode decoding? Since user may need to convert this back to a numpy array or float array.

I don't see a good solution or workaround under current search result parsing codebase though, maybe we need some ideas from the maintainers.

petyaslavova:
[...]

PR #3309, opened 2024-07-09
Decode search results at field level

uglide:
Fixes: #2772, #2275

### Pull Request check-list

### Description of change


bsbodden (commented):
LGTM! Thanks!

tylerhutcherson (approved):
Thanks team -- looks great

redis/commands/search/result.py
```diff
@@ -39,18 +49,22 @@ def __init__(
...
-                    )
-                    if hascontent
-                    else {}
-                )
+                keys = map(to_string, res[i + fields_offset][::2])
+                values = res[i + fields_offset][1::2]
+
+                for key, value in zip(keys, values):
+                    if field_encodings is None or key not in field_encodings:
+                        fields[key] = to_string(value)
```
gerzse, line 57: This seems like a second line of defaults, to make sure by default we always decode. But the fact that `decode_field` defaults to `True` and `encoding` defaults to `utf8` should achieve the same result, right? And this class is only used from private methods. So I would not keep this logic here, it's too "smart" for what the `Result` class is supposed to do.

tests/test_asyncio/test_search.py
```diff
@@ -1560,3 +1568,53 @@ async def test_query_timeout(decoded_r: redis.Redis):
     q2 = Query("foo").timeout("not_a_number")
     with pytest.raises(redis.ResponseError):
         await decoded_r.ft().search(q2)
+
+
+@pytest.mark.redismod
+@skip_if_resp_version(3)
```
gerzse, line 1574: Can't we run the test for RESP3 also?

tests/test_search.py
```diff
@@ -1705,6 +1707,54 @@ def test_search_return_fields(client):
         assert "telmatosaurus" == total["results"][0]["extra_attributes"]["txt"]
 
 
+@pytest.mark.redismod
+@skip_if_resp_version(3)
```
gerzse, line 1711: Can it also run with RESP3?

tests/test_search.py
```diff
@@ -1705,6 +1707,54 @@ def test_search_return_fields(client):
...
+    )
+
+    assert np.array_equal(
+        decoded_vec_from_search_results, fake_vec
+    ), "The vectors are not equal"
+
+    assert (
+        docs[0]["first_name"] == mixed_data["first_name"]
+    ), "The text field is not decoded correctly"
+
```
gerzse, line 1756: Just an idea, would it be relevant to also query the `first_name` without decoding, to show that it looks different?

redis/commands/search/commands.py
```diff
@@ -926,7 +932,12 @@ async def search(
         """  # noqa
         args, query = self._mk_query_args(query, query_params=query_params)
         st = time.time()
-        res = await self.execute_command(SEARCH_CMD, *args)
+
+        options = {}
+        if get_protocol_version(self.client) not in ["3", 3]:
```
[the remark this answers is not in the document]
uglide: To keep the compatibility with current behavior.  We have a bunch of tests that rely on the decoded RESP3 replies.

tests/test_asyncio/test_search.py
uglide: We can't because we just return plain dict on RESP3 without any decoding and parsing result.

tests/test_search.py
uglide: See my comment above.

redis/commands/search/result.py
uglide: This condition ensures that we preserve current behavior for cases when a query doesn't specify return fields.

tests/test_asyncio/test_search.py
gerzse: We can assert differently depending on RESP2/RESP3, there are already tests doing that.

gerzse: I think I actually get it, it's because of the vector data...

redis/commands/search/result.py
gerzse: Maybe add a comment explaining this?

gerzse (approved):

Issue #3145, opened 2024-02-15
redisCluster can't enable RESP3


owanio1992:
when try enable protocol=3 
get error
```
Segmentation fault
```

**Version**: 5.0.1

**Platform**: Python3.10.12  on ubuntu 22.04

**Description**: can't set args protocol=3

**reproduce**:
```
pip3 install redis==5.0.1
python3
from redis import RedisCluster as Redis
r = Redis.from_url("redis://localhost:7000/0?protocol=2")
r.ping()
r = Redis.from_url("redis://localhost:7000/0?protocol=3")
```
![Screenshot_20240215_155546](https://github.com/redis/redis-py/assets/145188628/33d97352-ad88-449b-9bde-e4a986f5ccd0)

akx:
Also, probably #3074.

dvora-h:
Hi @owanio1992 
What is your processor architecture?

owanio1992:
@dvora-h  my processor architecture is x86-64 (AMD EPYC 7571)

@akx here is output with `PYTHONDEVMODE=1`

```
(nebula-ap_2024-02-16_1708077366) ubuntu@node1:/opt/nebula/venv/nebula-ap_api$ PYTHONDEVMODE=1 python
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from redis import RedisCluster as Redis
>>> r = Redis.from_url("redis://localhost:7000/0?protocol=3")
Fatal Python error: Segmentation fault

Current thread 0x00007f38dff5a000 (most recent call first):
  File "/opt/nebula/venv/venv_src/nebula-ap_2024-02-16_1708077366/lib/python3.10/site-packages/redis/_parsers/hiredis.py", line 128 in read_response
  File "/opt/nebula/venv/venv_src/nebula-ap_2024-02-16_1708077366/lib/python3.10/site-packages/redis/connection.py", line 500 in read_response
  File "/opt/nebula/venv/venv_src/nebula-ap_2024-02-16_1708077366/lib/python3.10/site-packages/redis/client.py", line 553 in parse_response
  File "/opt/nebula/venv/venv_src/nebula-ap_2024-02-16_1708077366/lib/python3.10/site-packages/redis/cluster.py", line 1151 in _execute_command
[...]

PR #3324, opened 2024-07-19
Handle RESP3 sets as Python lists

### LINKED ISSUE #3074: `redis.cluster.RedisCluster` with RESP3 and hiredis causes segmentation fault
**Version**: redis-py 5.0.1 / Redis 7.1 (AWS ElastiCache)

**Platform**: CPython 3.11.7 on macOS Sonoma 14.2 (M2 Pro)

**Description**: `redis.cluster.RedisCluster` instances with protocol=3 (i.e. RESP3 enabled) crash the Python interpreter when performing commands and the `hiredis` extra dependency is installed in the local environment.

Example to reproduce:

```python3
url = "rediss://hostname:6379/0?protocol=3"
client = redis.cluster.RedisCluster.from_url(url)
client.get("test")

Process finished with exit code 139 (interrupted by signal 11:SIGSEGV)
```

gerzse:
### Pull Request check-list

### Description of change

Although the RESP3 protocol defines the set data structure, sometimes the responses from the Redis server contain sets with nested maps, which cannot be represented in Python as sets with nested dicts, because dicts are not hashable.

Versions of HIREDIS before 3.0.0 would cause segmentation fault when parsing such responses. Starting with version 3.0.0 the problem was fixed, with the compromise that RESP3 sets are represented as Python lists.

The embedded RESP3 parser was so far trying to represent RESP3 sets as Python sets, if possible. Only when this was not possible it would switch to the list representation. Arguably this is not the best user experience, not knowing when you will get back a set or a list.

Upgrade the required hiredis-py version to be at least 3.0.0, and change the embedded parser to always represent RESP3 sets as lists. This way we get a consistent experience in all cases.

This is a breaking change.

Fixes #3074 and #3145

### DIFF
--- a/doctests/dt_set.py
+++ b/doctests/dt_set.py
@@ -58,11 +58,11 @@ r.sadd("bikes:racing:france", "bike:1", "bike:2", "bike:3")
 r.sadd("bikes:racing:usa", "bike:1", "bike:4")
 # HIDE_END
 res7 = r.sinter("bikes:racing:france", "bikes:racing:usa")
-print(res7)  # >>> {'bike:1'}
+print(res7)  # >>> ['bike:1']
 # STEP_END
 
 # REMOVE_START
-assert res7 == {"bike:1"}
+assert res7 == ["bike:1"]
 # REMOVE_END
 
 # STEP_START scard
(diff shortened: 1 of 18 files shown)


vladvildanov (approved):

kosuke-zhang:
Hi, when will this be released?

ThiefMaster:
Can this please be fixed by making a release? Right now you get a (partially) broken combination of dependencies when you simply install `redis[hiredis]`. That should never happen. While people have to pin their transitive dependencies, the current state means that even in a new application where you add `redis[hiredis]` and then pin the dependencies, you get a broken combination, which can only fixed by adding an explicit pin for hiredis...

vladvildanov:
@ThiefMaster Hi! New release planned for next week, we will include this

PR #3350, opened 2024-08-08
Client side caching refactoring

vladvildanov:
### Pull Request check-list

### Description of change

Refactor for existing CSC implementation. The core changes is:

1. Single cache instance per connection pool. In RedisCluster single cache instance per all nodes.
2. Caching logic now is on the level of connection, so it can be reusable by different clients.
3. Additional abstractions around caching domain (CacheKey, CacheEntry, EvictionPolicy...)
4. Increased test coverage.
5. Added minimal version restrictions for the feature (>= Redis 7.4)


redis/connection.py
```diff
@@ -734,6 +715,188 @@ def _host_error(self):
...
+
+        with self._cache_lock:
+            # If current command reply already cached
+            # prevent sending data over socket.
+            if self._cache.get(self._current_command_hash):
+                return
+
+            # Set temporary entry as a status to prevent
+            # race condition from another connection.
+            self._cache[self._current_command_hash] = "caching-in-progress"
```
uglide, line 809: `"caching-in-progress"` Should be defined as constant.

redis/connection.py
```diff
@@ -734,6 +715,188 @@ def _host_error(self):
...
+        conf: CacheConfiguration,
+        cache_lock: threading.Lock,
+    ):
+        self.pid = os.getpid()
+        self._conn = conn
+        self.retry = self._conn.retry
+        self.host = self._conn.host
+        self.port = self._conn.port
+        self._cache = cache
+        self._cache_lock = cache_lock
```
uglide, line 741: Guarding cache object calls with an "external" lock makes it impossible to replace the current Cache implementation with something more efficient.

redis/connection.py
```diff
@@ -734,6 +715,188 @@ def _host_error(self):
...
+        self._conf = conf
+        self._current_command_hash = None
+        self._current_command_keys = None
+        self._current_options = None
+        self._keys_mapping = LRUCache(maxsize=10000)
```
uglide, line 746: What is that "magic" number `maxsize=10000`?

redis/connection.py
```diff
@@ -734,6 +715,188 @@ def _host_error(self):
...
+
+    def _on_invalidation_callback(self, data: List[Union[str, Optional[List[str]]]]):
+        with self._cache_lock:
+            # Flush cache when DB flushed on server-side
+            if data[1] is None:
+                self._cache.clear()
+            else:
+                for key in data[1]:
+                    normalized_key = ensure_string(key)
+                    if normalized_key in self._keys_mapping:
```
uglide, line 891: It looks strange that we share cache across connections but don't do the same for `_keys_mapping`. It will lead to partial cache cleanup on invalidation when the same key was accessed by multiple connections, isn't it?

requirements.txt
```diff
@@ -1 +1,3 @@
 async-timeout>=4.0.3
+cachetools
```
uglide, line 2: We should restrict the versions of dependencies. We should also try to avoid adding `apscheduler` because it uses redis-py as an optional dependency.

uglide (changes requested):

redis/cluster.py
```diff
@@ -507,6 +503,11 @@ def __init__(
         dynamic_startup_nodes: bool = True,
         url: Optional[str] = None,
         address_remap: Optional[Callable[[Tuple[str, int]], Tuple[str, int]]] = None,
+        use_cache: bool = False,
+        cache: Optional[Cache] = None,
+        cache_eviction: Optional[EvictionPolicy] = None,
```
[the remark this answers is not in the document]
vladvildanov: Agreed about Cache abstraction, but I would prefer to keep cache configuration as separate options, due to the fact that `redis_url` type of configuration exists and it's always prefered to operate with simple types to support it

requirements.txt
vladvildanov: Fixed

redis/connection.py
vladvildanov: Added constant

redis/connection.py
vladvildanov: Moved lock into the connection itself

redis/connection.py
vladvildanov: Fixed

dmaier-redislabs (commented):
Please see comments in the PR.

redis/cluster.py
```diff
@@ -1051,7 +1056,10 @@ def _parse_target_nodes(self, target_nodes):
             )
         return nodes
 
-    def execute_command(self, *args, **kwargs):
+    def execute_command(self, *args, **options):
```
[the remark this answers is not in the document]
vladvildanov: Sure, I'll fix this

redis/cluster.py
```diff
@@ -1154,19 +1161,13 @@ def _execute_command(self, target_node, *args, **kwargs):
                     connection.send_command("ASKING")
                     redis_node.parse_response(connection, "ASKING", **kwargs)
                     asking = False
-                response_from_cache = connection._get_from_local_cache(args)
```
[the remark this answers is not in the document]
vladvildanov: `RedisCluster` client cannot handle MOVED responses from server and consider this as an error. We're unable to reproduce this test case as the exception will be thrown as soon as failover will be triggered. Here's a statement from documentation:

`However, if the command execution fails due to cluster topology changes, a retry attempt will not be made, since the passed target node/s may no longer be valid, and the relevant cluster or connection error will be returned.`

https://redis-py.readthedocs.io/en/stable/clustering.html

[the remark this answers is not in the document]
vladvildanov: @dmaier-redislabs UPD: Even with the limitation that we already have I was able to test client as if it's handle MOVED errors. The outcome was exactly as you mentioned, we will run into data duplication on caches for node1 and node2.

For now updated RedisCluster client to have a single cache per all nodes

docs/resp3_features.rst
```diff
@@ -67,3 +67,35 @@ This means that should you want to perform something, on a given push notificati
     >> p = r.pubsub(push_handler_func=our_func)
 
 In the example above, upon receipt of a push notification, rather than log the message, in the case where specific text occurs, an IOError is raised. This example, highlights how one could start implementing a customized message handler.
+
+Client-side caching
+-------------------
+
+Client-side caching is a technique used to create high performance services.
+It exploits the memory available on application servers, servers that are usually distinct computers compared to the database nodes, to store some subset of the database information directly in the application side.
```
uglide, line 75: "Exploit" usually has a negative connotation. It is better to reword it to something like this:

> It utilizes the memory on application servers, typically separate from the database nodes, to cache a subset of the data directly on the application side.

redis/cache.py
```diff
@@ -0,0 +1,401 @@
...
+    def get_config(self) -> CacheConfigurationInterface:
+        pass
+
+    @abstractmethod
+    def get_eviction_policy(self) -> EvictionPolicyInterface:
+        pass
+
+    @abstractmethod
+    def get_size(self) -> int:
+        pass
```
uglide, line 119: A more Python idiomatic approach here is to use properties instead of getters like this:
```
    @property
    @abstractmethod
    def size(self) -> int:
        pass
```
You mixed both approaches in this module, so it makes sense to unify it.

redis/cache.py
```diff
@@ -0,0 +1,401 @@
...
+        self.redis_keys = redis_keys
+
+    def get_redis_keys(self) -> tuple:
+        return self.redis_keys
+
+    def __hash__(self):
+        return hash((self.command, self.redis_keys))
+
+    def __eq__(self, other):
+        return hash(self) == hash(other)
```
uglide, line 29: You can simplify the code with frozen dataclass:
```
from dataclasses import dataclass

@dataclass(frozen=True)
class CacheKey:
    command: str
    redis_keys: tuple
```
The same applies to CacheEntry below.

uglide (commented):
See my suggestions in the inline comments.

docs/resp3_features.rst
vladvildanov: Fixed

redis/cache.py
vladvildanov: Updated across for all getters in the class

redis/cache.py
vladvildanov: Fixed for CacheKey, CacheEntry isn't read-only cannot apply the same

uglide (approved):

minhduc0711:
With this PR, is CDC no longer supported for async clients?

iushas:
5.Added minimal version restrictions for the feature (>= Redis 7.4)

Why, is there any problem with lower than 7.4?

PR #3359, opened 2024-08-21
Allows cluster mode connections to wait for free connection when at max.

cavemanpi:
### Pull Request check-list

- [x ] Do tests and lints pass with this change?
- [ x] Is the new or changed code fully tested?
- [x ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ x] Was the change added to CHANGES file?

### Description of change

This change adds a flag to the cluster client which is propagated down to the cluster node object called 'wait_for_connections'. This helps in highly distributed high throughput environments where redis connections need to be capped to avoid running into connection limits, but also we want the request to make it through eventually in the event of some sort of stampede.

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1,5 +1,4 @@
 import asyncio
-import collections
 import random
 import socket
 import ssl
@@ -7,7 +6,6 @@ import warnings
 from typing import (
     Any,
     Callable,
-    Deque,
     Dict,
     Generator,
     List,
@@ -65,9 +63,9 @@ from redis.exceptions import (
     RedisClusterException,
     ResponseError,
     SlotNotCoveredError,
-    TryAgainError,
 )
 from redis.exceptions import TimeoutError as RedisTimeoutError
+from redis.exceptions import TryAgainError
 from redis.typing import AnyKeyT, EncodableT, KeyT
 from redis.utils import (
     deprecated_function,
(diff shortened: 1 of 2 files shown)


vladvildanov:
@cavemanpi This functionality is already covered by `BlockingConnectionPool`, you should use `RedisCluster` with provided configuration `connection_pool_class=BlockingConnectionPool`

https://github.com/redis/redis-py/blob/master/redis/connection.py#L1306

cavemanpi:
> @cavemanpi This functionality is already covered by `BlockingConnectionPool`, you should use `RedisCluster` with provided configuration `connection_pool_class=BlockingConnectionPool`
> 
> https://github.com/redis/redis-py/blob/master/redis/connection.py#L1306

Please take a look again. What you say is valid for the synchronous client, but I don't think it's true for the asyncio client.

cavemanpi:
@vladvildanov I updated the title to make it clearer this is in reference to the asyncio client.

vladvildanov:
@cavemanpi Thanks for clarification! We already have a PR that aims to cover the same functionality, feel free to join as well!

https://github.com/redis/redis-py/pull/3321

cavemanpi:
@vladvildanov, thanks for pointing me at that other PR; I appreciate it, but I find myself a little confused as I think that is trying to solve a different, but related problem.

https://github.com/redis/redis-py/blob/2fb40c10014c0b10fdb5005d0e7b5f2f4cdc1e00/redis/asyncio/cluster.py#L390

From what I can see the asyncio RedisCluster class makes use of a class called NodesManager track cluster nodes and route commands to the correct nodes. There is not a constructor parameter to override what is used for a NodesManager class replacement. This NodesManager class has mappings to ClusterNode objects which conceptually act like both the representation of a node as well as a connection pool. There also does not look to be a public constructor parameter to override the connection pooling behavior.

Is the proper fix here to actually decouple the concept of the cluster node from its connection pool and make the connection pool class configurable like most of the rest of the clients?

vladvildanov:
@cavemanpi For alignment with sync API we have to have `connection_pool_class` argument in constructor, so this way it would be possible to set a type for connection pool. Use sync API as an example

https://github.com/redis/redis-py/blob/2fb40c10014c0b10fdb5005d0e7b5f2f4cdc1e00/redis/cluster.py#L1349

PR #3376, opened 2024-09-10
Fix sentinel connections

garionphx:
When passing args to Redis(), using the connection_kwargs args allows us to use authentication and other connection features.

### Pull Request check-list

### Description of change

Modified the arguments passed to Redis() when forming Sentinel connections

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -240,7 +240,7 @@ class Sentinel(SentinelCommands):
         self.sentinel_kwargs = sentinel_kwargs
 
         self.sentinels = [
-            Redis(hostname, port, **self.sentinel_kwargs)
+            Redis(hostname, port, **connection_kwargs)
             for hostname, port in sentinels
         ]
         self.min_other_sentinels = min_other_sentinels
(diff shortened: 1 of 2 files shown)


garry-t:
great, thank you!

vladvildanov:
@garionphx Thanks for your contribution! Make sure that tests passes

vladvildanov:
@garionphx Since this changes makes `self.sentinel_kwargs` argument completely redundant, it should be removed from constructor properties and it will be a breaking change.

redis/sentinel.py
```diff
@@ -240,7 +240,7 @@ def __init__(
         self.sentinel_kwargs = sentinel_kwargs
 
         self.sentinels = [
-            Redis(hostname, port, **self.sentinel_kwargs)
```
vladvildanov, line 243: As this is the only place where `self.sentinel_kwargs` are used we have to remove it

petyaslavova:
Hi @garionphx,

Thanks for the contribution! We should keep sentinel_kwargs and connection_kwargs as separate arguments. These serve distinct purposes: we may need different connection settings for Sentinel nodes and for Redis servers (both master and slaves). For example, SSL configuration, authentication credentials, or even different users might vary between the two connection types.

Keeping them as separate dictionaries is currently the only flexible and clean way to support these differences. That said, you can still supply usernames and credentials specifically for Sentinel connections via sentinel_kwargs.

For these reasons, I am going to close this PR.

PR #3397, opened 2024-10-01
Avoid stacktrace on process exit in Client.__del__()

### LINKED ISSUE #3014: Stderr backtrace caused by `redis.client.Redis.__del__` on process exit.
**Version**: 5.0.1

**Platform**: Python 3.8.18 on Ubuntu 20.04

**Description**: Backtrace to stderr on process exit due to `redis.client.Redis.__del__()`

Here is the trace:
```
Exception ignored in: <function Redis.__del__ at 0x7fcf29e50ca0>
  File "/venv/lib/python3.8/site-packages/redis/client.py", line 491, in __del__
  File "/venv/lib/python3.8/site-packages/redis/client.py", line 506, in close
  File "/venv/lib/python3.8/site-packages/redis/connection.py", line 1155, in disconnect
  File "/venv/lib/python3.8/site-packages/redis/connection.py", line 1062, in _checkpid
AttributeError: 'NoneType' object has no attribute 'getpid'
```

Occurs due to unloaded globals being referenced in methods called in destructors on process exit. See warning in Python docs [here](https://docs.python.org/3.8/reference/datamodel.html#object.__del__)

This can be fixed in a couple of ways:
1. Add an exception handler to `redis.client.Redis.__del__()` as it appears has been done in many other destructors (https://github.com/redis/redis-py/blob/d3a3ada03e080f39144807c9fbe44876c40e0548/redis/connection.py#L226 for example)
2. Add references to required globals to the `redis.connection.ConnectionPool` class and call via the references. The globals that require references in are `os.getpid` and `chain` (from `itertools`) afaict.

NOTE: It is quite hard to come up with a small piece of code to reproduce this issue as it is caused by a race in the unload order at exit. I'll try to come up with something reliable if req'd.

NOTE: The issue is triggered by https://github.com/redis/redis-py/blob/d3a3ada03e080f39144807c9fbe44876c40e0548/redis/client.py#L506 which was recently introduced in https://github.com/redis/redis-py/commit/012f7cf7d2950e8241eb5705e7cdd83b09370ba4

noirbee:
### Pull Request check-list

### Description of change

Client.close() may call ConnectionPool.release() or ConnectionPool.disconnect(); both methods may end up calling os.getpid() (through ConnectionPool._checkpid() or threading.Lock() (through ConnectionPool.reset()). As mentioned in the Python documentation [1], at interpreter shutdown, module globals (in this case, the os and threading module references) may be deleted or set to None before __del__() methods are called. This causes an AttributeError to be raised when trying to run e.g. os.getpid(); while the error is ignored by the interpreter, the traceback is still printed out to stderr.

Closes #3014

[1] https://docs.python.org/3/reference/datamodel.html#object.__del__


noirbee:
Note that I unfortunately cannot provide an easy way to reproduce the issue; I can only reproduce it using our own codebase. I suspect this is due to the fact that we load a much larger amount of code than what a synthetic / minimal reproduction attempt would.

redis/client.py
```diff
@@ -517,7 +517,18 @@ def __exit__(self, exc_type, exc_value, traceback):
         self.close()
 
     def __del__(self):
-        self.close()
+        try:
```
petyaslavova, line 520: I recommend catching Exception here without raising anything:
```
  try:
      self.close()
  except Exception:
      pass
```
Since __del__ is called during object cleanup, handling this exception is unnecessary.
I also verified that the same pattern is used in the RedisCluster class. Could you update the __del__ method implementation there as well?

noirbee:
> I have requested some changes to be added.

Thank you for reviewing @petyaslavova , I've updated the code to catch all exceptions and just `pass` in both `Redis` and `RedisCluster`, and rebased on the current master.

petyaslavova (approved):

PR #2979, opened 2023-10-05
Python 3.12

chayim:
Updating the trove, and CI to support the Python 3.12 release announcement.

closes #3020

PR #3404, opened 2024-10-06
there is no pyopenssl 23.2.1

dimbleby:
since #2979 this project pins to a version of pyopenssl that does not exist

per https://pypi.org/project/pyOpenSSL/#history, the latest version at time of that commit - as today - is 24.2.2

I've also removed the exact pin, this just invites conflicts with other projects.

While I have not tested this at all except to install it, I suppose that is more than anyone else has done!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,6 @@ setup(
     ],
     extras_require={
         "hiredis": ["hiredis>=3.0.0"],
-        "ocsp": ["cryptography>=36.0.1", "pyopenssl==23.2.1", "requests>=2.31.0"],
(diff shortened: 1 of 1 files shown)


petyaslavova:
Hi @dimbleby, thank you for your contribution! The currently listed dependency for pyOpenSSL specifies version 20.0.1, which is a valid version of the library. Since this has already been addressed, I’ll be closing this PR. Thanks again for your efforts!

dimbleby:
Please also change them constraint to `>=` - as above and per linked issues, an exact requirement invites unnecessary conflicts

PR #3415, opened 2024-10-15
Migrate to clients test image

uglide:
This change introduces a new test matrix based on Redis versions **7.4, 7.2, and 6.2**. The new workflows are Redis 8.0 ready, but 8.0 should be added to test matrix as a separate PR to address broken tests and fix incompatibility with 8.0.
By default, we start integration jobs only to test a change on all Redis versions with **Python 3.8 and 3.12**. On success, all other compatibility test runs kick in to do full validation.


uglide:
CC: @ggivo

vladvildanov (approved):

akx:
@uglide This seems to have introduced errors like `/home/runner/work/_temp/9be3a23c-94d1-45cf-8346-a760182dff2c.sh: line 14: ((: < 7 : syntax error: operand expected (error token is "< 7 ")` in subsequent CI runs.

uglide:
Thanks @akx! Unfortunately, I've missed your message here. Still, late is better than never... so here is the fix https://github.com/redis/redis-py/pull/3642

PR #3424, opened 2024-11-05
Fix inaccurate docstring for unwatch()

djroze:
The 'unwatch()' method of the Redis client, as currently documented, says that it unwatches the value at key "name", but it does not actually take any arguments ("name" or otherwise). According to the latest Redis documentation at the given URL for the UNWATCH command, this command unwatches all previously watched keys for the current transaction. Modified docstring to reflect that this method does not take any arguments and instead (presumably) unwatches all previously watched keys.

### Pull Request check-list

### Description of change


petyaslavova (approved):
LGTM

PR #3437, opened 2024-11-18
Fixing randomly failing test

IlianIliev:
### Pull Request check-list

_NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open._

### Description of change

This test fails randomly depending on whether it runs close to the end of the second or not.

For example, if the test runs at `1527800400.999999`, then the integer value of the timestamp + timedelta of 1 seconds = 1527800401. This way when we call `r.hexpireat` with that value, the combination of the time spent in the python code and the network delay adds enough time to cause the expiration timestamp to be in the past when received by the server.
As a result, the result value of the `hexpireat` is 2 (timestamp in the past) instead of 1. 
```

```

### DIFF
--- a/tests/test_asyncio/test_hash.py
+++ b/tests/test_asyncio/test_hash.py
@@ -129,9 +129,9 @@ async def test_hpexpire_multiple_fields(r):
 async def test_hexpireat_basic(r):
     await r.delete("test:hash")
     await r.hset("test:hash", mapping={"field1": "value1", "field2": "value2"})
-    exp_time = int((datetime.now() + timedelta(seconds=1)).timestamp())
+    exp_time = math.ceil((datetime.now() + timedelta(seconds=1)).timestamp())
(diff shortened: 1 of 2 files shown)


IlianIliev:
I realized there is similar logic in core.py -> https://github.com/redis/redis-py/blob/master/redis/commands/core.py#L5265-L5266

```
if isinstance(unix_time_seconds, datetime.datetime):
    unix_time_seconds = int(unix_time_seconds.timestamp())
```

This causes failures when running `test_hexpireat_with_datetime`.

One option is to change the test to ensure it always rounds down the microseconds. Another is to change the `core.py`, which will change the logic and may result in a non-expected behavior for the user.

vladvildanov:
@IlianIliev Let's go with option 1 for now

vladvildanov:
@IlianIliev As I see pypy 3.10 still failing cause of hexpireat

vladvildanov (approved):

PR #3442, opened 2024-11-26
Avoid the multiprocessing forkserver method in Python 3.14

befeleme:
### Pull Request check-list

### Description of change

This change is essentially a copy of: https://github.com/giampaolo/pyftpdlib/pull/656

Python 3.14 has changed the default multiprocessing method from fork to forkserver which causes failures like this one:

```
____________________ TestMultiprocessing.test_redis_client _____________________

self = <tests.test_multiprocessing.TestMultiprocessing object at 0x7f646b43d250>
r = <redis.client.Redis(<redis.connection.ConnectionPool(<redis.connection.Connection(host=localhost,port=6379,db=0)>)>)>

    def test_redis_client(self, r):
        "A redis client created in a parent can also be used in a child"
        assert r.ping() is True
    
        def target(client):
            assert client.ping() is True
            del client
    
        proc = multiprocessing.Process(target=target, args=(r,))
>       proc.start()

tests/test_multiprocessing.py:163: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib64/python3.14/multiprocessing/process.py:121: in start
    self._popen = self._Popen(self)
/usr/lib64/python3.14/multiprocessing/context.py:224: in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
/usr/lib64/python3.14/multiprocessing/context.py:300: in _Popen
    return Popen(process_obj)
/usr/lib64/python3.14/multiprocessing/popen_forkserver.py:35: in __init__
    super().__init__(process_obj)
/usr/lib64/python3.14/multiprocessing/popen_fork.py:20: in __init__
    self._launch(process_obj)
/usr/lib64/python3.14/multiprocessing/popen_forkserver.py:47: in _launch
    reduction.dump(process_obj, buf)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

obj = <Process name='Process-9' parent=996 initial>
file = <_io.BytesIO object at 0x7f646bbee390>, protocol = None

    def dump(obj, file, protocol=None):
        '''Replacement for pickle.dump() using ForkingPickler.'''
>       ForkingPickler(file, protocol).dump(obj)
E       _pickle.PicklingError: Can't pickle local object <function TestMultiprocessing.test_redis_client.<locals>.target at 0x7f646bb0f530>
E       when serializing dict item '_target'
E       when serializing multiprocessing.context.Process state
E       when serializing multiprocessing.context.Process object
```

I verified this makes tests pass with Python 3.14.0a2 in Fedora's build environment

### DIFF
--- a/tests/test_multiprocessing.py
+++ b/tests/test_multiprocessing.py
@@ -1,6 +1,5 @@
 import contextlib
 import multiprocessing
-import sys
 
 import pytest
 import redis
@@ -8,9 +7,6 @@ from redis.connection import Connection, ConnectionPool
 from redis.exceptions import ConnectionError
 
 from .conftest import _get_client
-
-if sys.platform == "darwin":
-    multiprocessing.set_start_method("fork", force=True)
 
 
 @contextlib.contextmanager
@@ -22,6 +18,16 @@ def exit_callback(callback, *args):
 
 
 class TestMultiprocessing:
+    # On macOS and newly non-macOS POSIX systems (since Python 3.14),
+    # the default method has been changed to forkserver.
+    # The code in this module does not work with it,
+    # hence the explicit change to 'fork'
+    # See https://github.com/python/cpython/issues/125714
+    if multiprocessing.get_start_method() == "forkserver":
+        _mp_context = multiprocessing.get_context(method="fork")
+    else:
+        _mp_context = multiprocessing.get_context()
+
     # Test connection sharing between forks.
     # See issue #1085 for details.
 
@@ -45,7 +51,7 @@ class TestMultiprocessing:
             assert conn.read_response() == b"PONG"
             conn.disconnect()
 
-        proc = multiprocessing.Process(target=target, args=(conn,))
+        proc = self._mp_context.Process(target=target, args=(conn,))
         proc.start()
         proc.join(3)
         assert proc.exitcode == 0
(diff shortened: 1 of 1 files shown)


tests/test_multiprocessing.py
```diff
@@ -22,6 +22,15 @@ def exit_callback(callback, *args):
 
 
 class TestMultiprocessing:
+
```
petyaslavova, line 25: Can you please remove the code on lines 12 and 13, as it also handles a similar issue but for macOS only,  and update your comment to include this case as well?

petyaslavova (changes requested):

petyaslavova:
@befeleme please check linters failure

befeleme:
Hopefully mitigated.

petyaslavova:
Hi @befeleme, the linters are failing. You can run them locally to check if there is something else than the one reported from the pipeline, the instructions are available in the [CONTRIBUTING.md](https://github.com/redis/redis-py/blob/master/CONTRIBUTING.md) file

petyaslavova (approved):

PR #3447, opened 2024-11-28
Add dynamic_startup_nodes parameter to async RedisCluster

### LINKED ISSUE #2472: async redis cluster should use initial startup nodes during reinitialization in case of failover
**Version**: 
redis-py: 4.3.3
redis: 7.0.2

**Platform**: 
python 3.7

**Description**: 
Async implementation of `RedisCluster` overwrites initial startup nodes during first initialization. This is caused by [this](https://github.com/redis/redis-py/blob/master/redis/asyncio/cluster.py#L1271) line.

When deployed in dynamic systems like Kubernetes your shards may sometimes change their IP addresses.
This likely leads to a situation when python client ends up with a pool of invalid (outdated) startup nodes and can not perform reinitialization, since `NodeManager` uses raw IP addresses.

For example, let's say we have cluster with 3 masters in k8s and use python client this way:
```
from redis.asyncio import RedisCluster

client = RedisCluster.from_url('redis://redis-cluster-headless:6379/0')
print(client.nodes_manager.startup_nodes)

# outputs {'redis-cluster-headless:6379': [host=redis-cluster-headless, port=6379, name=redis-cluster-headless:6379, server_type=None]}

await client # initialization
print(client.nodes_manager.startup_nodes)

# outputs something like this
# {'<ip1>:6379': [host=<ip1>, port=6379, name=<ip1>:6379, server_type=primary],
#  '<ip2>:6379': [host=<ip2>, port=6379, name=<ip2>:6379, server_type=primary],
#  '<ip3>:6379': [host=<ip3>, port=6379, name=<ip3>:6379, server_type=primary]}
```
now image your python client is idle for some time without requests, and during that period K8S have recreated each redis container due to whatever reason (crash, or moved to another node by autoscaler for example).

Then, next time python client will try to perform a redis call it will find out that all ClusterNodes are unavaliable, will try to reinitialize NodeManager and end up with no connectable node to get new cluster configuration from.
This would not be an issue if initial headless URL was still in the pool of startup nodes.

Sync implementation of `RedisCluster` preserves initial node address just as suggested above, so i believe async implementation should behave the same.

For now I use the following patch, but I can't call it an elegant way to get around the problem. It must be solved through changes to NodeManager.
```
class AsyncRedisCluster(RedisCluster):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._initial_startup_nodes = dict(self.nodes_manager.startup_nodes)

    async def initialize(self):
        rv = await super().initialize()
        self.nodes_manager.startup_nodes.update(self._initial_startup_nodes)
        return rv
```

P.S. sync implementation has the opposite problem, `NodesManager` doesn't pop unused addresses from startup_nodes storage => it may grow indefinitely as soon as your containers change IPs.

Kakadus:
### Pull Request check-list

- [X] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [X] Is there an example added to the examples folder (if applicable)?

### Description of change

The async version of the RedisCluster lacked the `dynamic_startup_nodes` parameter. This PR adds the parameter to be activated when dynamic DNS endpoints for startup nodes are in use. 

Enabling `dynamic_startup_nodes` fixes #2472, which describes the problem in detail.

https://github.com/redis/redis-py/pull/3111 also addresses said issue, but their logic is different to the sync version, while this approach moves the behaviour of the async version closer to the sync version.


julianhille:
any news here? this looks good to me

petyaslavova:
Hi @Kakadus would you have some time to complete this PR. There are some conflicts with the master branch that need to be resolved.

petyaslavova:
Hi @Kakadus,
I tried to complete the changes for this PR, but while resolving conflicts and merging with master, I accidentally included the master commits as new ones. To avoid risking any issues with the Git history, I’ve recreated your changes in a new PR: #3646.
Apologies for the mess — and thank you for your understanding!
I’ll go ahead and close this PR now.

Issue #3453, opened 2024-12-09
`Wrong number of response items from pipeline execution` error on pipeline when client-side-caching is enabled


iRhonin:
**Version**: `redis-py==5.2.0` `redis==7.4.1`

**Platform**: `Python==3.8.18` `Ubuntu==24.04`

**Description**:

I am getting `Wrong number of response items from pipeline execution` error when `client-side-caching` is enabled. The error occurs when using pipeline and at least one of the keys in the pipeline is cached in client.    

**Steps to reproduce:**

```python
r = Redis(protocol=3, cache_config=CacheConfig())

pipe = r.pipeline()
pipe.set('foo', 5)
pipe.set('bar', 18.5)
pipe.execute()

r.get('foo')

pipe = r.pipeline()
pipe.set('foo', 5)
pipe.set('bar', 18.5)
pipe.execute()
```

**Error:**

```python
ResponseError                             Traceback (most recent call last)
Cell In[43], line 4
      2 pipe.set('foo', 5)
      3 pipe.set('bar', 18.5)
----> 4 pipe.execute()

File ~/.pyenv/versions/3.8.18/lib/python3.8/site-packages/redis/client.py:1530, in Pipeline.execute(self, raise_on_error)
   1527     self.connection = conn
   1529 try:
-> 1530     return conn.retry.call_with_retry(
   1531         lambda: execute(conn, stack, raise_on_error),
   1532         lambda error: self._disconnect_raise_reset(conn, error),
   1533     )
   1534 finally:
   1535     self.reset()

File ~/.pyenv/versions/3.8.18/lib/python3.8/site-packages/redis/retry.py:62, in Retry.call_with_retry(self, do, fail)
     60 while True:
     61     try:
---> 62         return do()
     63     except self._supported_errors as error:
     64         failures += 1

File ~/.pyenv/versions/3.8.18/lib/python3.8/site-packages/redis/client.py:1531, in Pipeline.execute.<locals>.<lambda>()
   1527     self.connection = conn
   1529 try:
   1530     return conn.retry.call_with_retry(
-> 1531         lambda: execute(conn, stack, raise_on_error),
   1532         lambda error: self._disconnect_raise_reset(conn, error),
   1533     )
   1534 finally:
   1535     self.reset()

File ~/.pyenv/versions/3.8.18/lib/python3.8/site-packages/redis/client.py:1410, in Pipeline._execute_transaction(self, connection, commands, raise_on_error)
   1408 if len(response) != len(commands):
[...]

PR #3456, opened 2024-12-12
Fix incorrect attribute reuse in redis.connection.CacheProxyConnection

zs-neo:
### Pull Request check-list

### Description of change

Attemp to adress https://github.com/redis/redis-py/issues/3453

`CacheProxyConnection` adds caching capabilities to the `send_command` and `read_response` methods, but does not clear the `_current_command_cache_key` attribute. `Pipeline` and `ClusterPipeline` use `send_packed_command` to send requests, which may cause the wrong `_current_command_cache_key` to be reused during read_response.

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -499,9 +499,9 @@ class TestUnitCacheProxyConnection:
         )
         proxy_connection.send_command(*["GET", "foo"], **{"keys": ["foo"]})
         assert proxy_connection.read_response() == b"bar"
+        assert proxy_connection._current_command_cache_key is None
         assert proxy_connection.read_response() == b"bar"
 
-        mock_connection.read_response.assert_called_once()
         mock_cache.set.assert_has_calls(
             [
                 call(
@@ -525,9 +525,6 @@ class TestUnitCacheProxyConnection:
 
         mock_cache.get.assert_has_calls(
             [
-                call(CacheKey(command="GET", redis_keys=("foo",))),
-                call(CacheKey(command="GET", redis_keys=("foo",))),
-                call(CacheKey(command="GET", redis_keys=("foo",))),
                 call(CacheKey(command="GET", redis_keys=("foo",))),
                 call(CacheKey(command="GET", redis_keys=("foo",))),
                 call(CacheKey(command="GET", redis_keys=("foo",))),
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -904,9 +904,11 @@ class CacheProxyConnection(ConnectionInterface):
                 and self._cache.get(self._current_command_cache_key).status
                 != CacheEntryStatus.IN_PROGRESS
             ):
-                return copy.deepcopy(
+                res = copy.deepcopy(
                     self._cache.get(self._current_command_cache_key).cache_value
                 )
+                self._current_command_cache_key = None
+                return res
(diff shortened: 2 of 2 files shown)


vladvildanov:
@zs-neo Thanks for your contribution! Will mark this as a bug and will be included into next patch release

vladvildanov:
@zs-neo Few unit tests are still failing

tests/test_connection.py
```diff
@@ -468,24 +468,6 @@ def test_read_response_returns_cached_reply(self, mock_cache, mock_connection):
                 status=CacheEntryStatus.IN_PROGRESS,
                 connection_ref=mock_connection,
             ),
-            CacheEntry(
-                cache_key=CacheKey(command="GET", redis_keys=("foo",)),
-                cache_value=b"bar",
-                status=CacheEntryStatus.VALID,
-                connection_ref=mock_connection,
```
vladvildanov, line 475: Why do you need to remove those entries?

zs-neo: In the code below, `mock_cache.get` is only called three times, so I removed it. I hope my understanding is correct.

zs-neo:
@vladvildanov Thanks to the useful comments in CacheProxyConnection, which helped me understand the relevant logic. I tried to fix the remaining problems, please take a look at this PR again, any suggestions are welcome.

tests/test_connection.py
vladvildanov: `mock_cache.get` called multiple times because of cache logic, there's a "soft" lock to avoid race condition so when you're sending a command a temporary cache entry is created with status "IN_PROGRESS", then when the actual response is received the same instance becomes associated with value and status changes to "VALID". So that's why status of first instance is different from consequent calls. Also additional calls is just to verify that value is retrieved from cache

vladvildanov:
@zs-neo Thank you for your input! Let's finish it up and I will include it into next 5.3 GA

tests/test_connection.py
zs-neo: Thank you for your patience in answering my questions, now I fully understand！

zs-neo:
`tests.test_auth.test_token_manager.TestTokenManager.test_async_success_token_renewal` failed, I'll check again.
```
2024-12-23T12:43:09.9515473Z >       assert len(tokens) == tokens_refreshed
2024-12-23T12:43:09.9516084Z E       assert 3 == 2
2024-12-23T12:43:09.9517386Z E        +  where 3 = len([<redis.auth.token.SimpleToken object at 0x0000000009d708e0>, <redis.auth.token.SimpleToken object at 0x0000000009d70918>, <redis.auth.token.SimpleToken object at 0x0000000009d70950>])
```

tests/test_connection.py
vladvildanov: Could you please revert those changes? And we're good to go

zs-neo: done

vladvildanov:
@zs-neo It's a flacky test, I'll fix and you could rebase a branch

zs-neo:
> @zs-neo It's a flacky test, I'll fix and you could rebase a branch

Thanks a lot, I tried to fix it but I'm not too familiar with TokenManager, it would be great if you could fix it.

PR #3458, opened 2024-12-17
Fix AttributeError when client.get_default_node() returns None

bacchuswng:
### Pull Request check-list

### Description of change
The default node is sometimes None. It's not obvious from the code how this happens (maybe an Exception that isn't being handled correctly, or a connection-related object is being closed while pipeline execution is still in progress). Regardless, we've observed `AttributeError: 'NoneType' object has no attribute 'name'` in practice. 

Adding this defensive check ensures that we don't run into this issue.

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1568,18 +1568,22 @@ class ClusterPipeline(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterComm
                         result.args = (msg,) + result.args[1:]
                         raise result
 
-            default_node = nodes.get(client.get_default_node().name)
-            if default_node is not None:
-                # This pipeline execution used the default node, check if we need
-                # to replace it.
-                # Note: when the error is raised we'll reset the default node in the
-                # caller function.
-                for cmd in default_node[1]:
-                    # Check if it has a command that failed with a relevant
-                    # exception
-                    if type(cmd.result) in self.__class__.ERRORS_ALLOW_RETRY:
-                        client.replace_default_node()
-                        break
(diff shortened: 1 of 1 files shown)


zs-neo:
Can you provide more information? How do you use redis-py? We are also trying related features and want to make sure we don’t encounter the same error.

petyaslavova (commented):
Hi @bacchuswng, can you please share more details about your code where you hit that issue - how do you use the RedisCluster client. I tried to reproduce something like it, but with no luck... 
Before merging some fix, we should know how exactly we get into the problematic case and validate that this is the correct fix and not just hiding another problem...

bacchuswng:
> Hi @bacchuswng, can you please share more details about your code where you hit that issue - how do you use the RedisCluster client. I tried to reproduce something like it, but with no luck... Before merging some fix, we should know how exactly we get into the problematic case and validate that this is the correct fix and not just hiding another problem...

@petyaslavova Understood. Unfortunately, this specific issue is extremely hard to repro and only happened under load with a lot of concurrency. 

Staring at the code, it is generally possible that `client.get_default_node()` returns `None`, but I haven't been able to find a smoking gun where, in this specific case, this call would return `None`. As a result, I view this PR as more a layer of defense that mostly preserves the spirit of what the code is doing: replacing the default node, if there is one.

petyaslavova (changes requested):
The requested change is posted previously. This update is just to have the corresponding status in the PRs list.

bacchuswng:
> The requested change is posted previously. This update is just to have the corresponding status in the PRs list.

Done!

petyaslavova (approved):

PR #3466, opened 2024-12-23
convert to pyproject.toml

### LINKED ISSUE #3463: Remove dependency files and add dependencies to setup.py

woutdenolf:
Closes #3463


woutdenolf:
pyproject.toml is not yet supported for dependency audit: https://github.com/pypa/pip-audit/issues/694. Would that be a show stopper?

woutdenolf:
Ok then I'll close this one so only one of us needs to spend time rebasing.

PR #3469, opened 2024-12-30
Added Redis 8.0 to test matrix

vladvildanov:
### Pull Request check-list

### Description of change


tests/test_search.py
```diff
@@ -2073,6 +2093,96 @@ def test_profile(client):
...
+    client.ft().create_index((TextField("t"),))
+    client.ft().client.hset("1", "t", "hello")
+    client.ft().client.hset("2", "t", "world")
+
+    # check using Query
+    q = Query("hello|world").no_content()
+    if is_resp2_connection(client):
+        res, det = client.ft().profile(q)
+        det = det.info
+        assert det[0] == "Shards"
```
uglide, line 2108: Arguably, we shouldn't test FT.PROFILE reply structure, since we are testing the client here, not the server. I'd suggest trimming most of the checks and replace them with something like 
```python
assert typeof(det) is dict
assert len(det) >= X
```

redis/commands/search/commands.py
```diff
@@ -66,8 +67,10 @@ class SearchCommands:
     """Search commands."""
 
     def _parse_results(self, cmd, res, **kwargs):
-        if get_protocol_version(self.client) in ["3", 3]:
+        if get_protocol_version(self.client) in ["3", 3] and cmd != "FT.PROFILE":
```
uglide, line 70: Better to do something like this to hardcode FT.PROFILE only once

```
if get_protocol_version(self.client) in ["3", 3]:
    return ProfileInformation(res) if cmd == "FT.PROFILE" else res
```

uglide (changes requested):
Some minor changes are required. See my inline comments

redis/commands/search/commands.py
```diff
@@ -5,12 +5,13 @@
 from redis.client import NEVER_DECODE, Pipeline
 from redis.utils import deprecated_function
 
-from ..helpers import get_protocol_version, parse_to_dict
```
uglide, line 8: It seems that `parse_to_dict` is a dead/unused code now. It should be removed if this is the case with the test case for it.

uglide (commented):
One more comment

redis/commands/search/profileInformation.py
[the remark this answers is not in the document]
vladvildanov: Sorry I was mislead by another file `indexDefinition.py`, I will rename both of them

uglide (approved):
LGTM

PR #3475, opened 2025-01-08
Moved ClusterParser exceptions to BaseParser class

vladvildanov:
### Pull Request check-list

### Description of change

If RESP3 is enabled, the ClusterParser becomes immediately overridden with _RESP3Parser which leads to the fact that all additional exceptions isn't handled by cluster implementation.

ClusterParser class shouldn't exist and inherit other parsers just to extend a list of exceptions that should be associated with given error codes. Parsing errors and associate them to exception classes is a responsibility of parsers itself regardless of server topology, potentially you could get any error code as a response from server and parser should be aware of it.

### DIFF
--- a/docs/clustering.rst
+++ b/docs/clustering.rst
@@ -17,8 +17,8 @@ Nodes <#specifying-target-nodes>`__ \| `Multi-key
 Commands <#multi-key-commands>`__ \| `Known PubSub
 Limitations <#known-pubsub-limitations>`__
 
-Creating clusters
------------------
+Connecting to cluster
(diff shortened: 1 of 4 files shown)


uglide (approved):
LGTM, but why do you think it's a breaking change? I see no changes affecting public API

vladvildanov:
@uglide Well, from my POV this isn't, just the fact that public class was removed makes me feel it would be wise to consider this as a breaking change

uglide:
Ah, I see. Good then.

PR #3477, opened 2025-01-10
Fix connection health check for protocol != 2 when auth credentials are provided and health check interval is configured

Yard1:
### Pull Request check-list

### Description of change

Fixes `PING` failing in health check due to async connection `on_connect` if protocol != 2 by following the pattern already applied in other if...else branches in `on_connect`.


redis/asyncio/connection.py
```diff
@@ -363,7 +363,9 @@ async def on_connect(self) -> None:
                 self._parser.on_connect(self)
             if len(auth_args) == 1:
                 auth_args = ["default", auth_args[0]]
-            await self.send_command("HELLO", self.protocol, "AUTH", *auth_args)
+            # avoid checking health here -- PING will fail if we try
+            # to check the health prior to the AUTH
+            await self.send_command("HELLO", self.protocol, "AUTH", *auth_args, check_health=False)
```
petyaslavova, line 368: @Yard1  The same pattern is used in the sync connection code. Please update that one as well.

petyaslavova (commented):
The problem is reproducible when a value >0 is set for health_check_interval argument when the sync and async clients are created.

redis/asyncio/connection.py
petyaslavova: Done.

petyaslavova (approved):

PR #3480, opened 2025-01-18
Fix invalid return type annotation

birthdaysgift:
`or` cannot be used in type annotations, `Union[]` or `|` should be used instead depending on the python version


petyaslavova (approved):
LGTM

PR #3490, opened 2025-01-30
Update Python imports (index_definition => indexDefinition)

dwdougherty:
### Pull Request check-list

### Description of change

A customer tried to run some of the query_* tests and got tripped up on the import for `IndexDefinition`. This PR fixes those imports.

Fixes https://github.com/redis/docs/issues/1097

### DIFF
--- a/doctests/query_agg.py
+++ b/doctests/query_agg.py
@@ -6,7 +6,7 @@ from redis.commands.json.path import Path
 from redis.commands.search import Search
 from redis.commands.search.aggregation import AggregateRequest
 from redis.commands.search.field import NumericField, TagField
-from redis.commands.search.index_definition import IndexDefinition, IndexType
+from redis.commands.search.indexDefinition import IndexDefinition, IndexType
 import redis.commands.search.reducers as reducers
 
 r = redis.Redis(decode_responses=True)
--- a/doctests/query_combined.py
+++ b/doctests/query_combined.py
@@ -6,7 +6,7 @@ import redis
 import warnings
 from redis.commands.json.path import Path
 from redis.commands.search.field import NumericField, TagField, TextField, VectorField
-from redis.commands.search.index_definition import IndexDefinition, IndexType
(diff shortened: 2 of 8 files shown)


dwdougherty:
Hi @petyaslavova. I'm sorry, but I'm not following you. Are you saying that this PR isn't correct and can't be merged? Currently, the code examples on redis/docs are broken because of bad imports. The only way to fix this is to fix the doctests in this repo. What are you advising?

Cc: @uglide

dwdougherty:
Is it likely that customers will use redis-py this way? I'd expect that most folks would install redis-py via the PYPI market place.

petyaslavova:
Hi @dwdougherty, You are right about that. To make the doctests compatible with the latest released library version, this change should be merged. However, we should keep in mind that the rename should be applied again after the library version containing the latest code is released.

petyaslavova (approved):
LGTM.

dwdougherty:
Thank you, @petyaslavova! Much appreciated.

uglide:
@petyaslavova @dwdougherty I think this is the wrong way of fixing this. Instead, we should "pin" the version of client repos to the latest stable version.

uglide:
See my comment here https://github.com/redis/docs/issues/1097#issuecomment-2634271684

dwdougherty:
I didn't realize these imports had already been fixed on the 5.3 branch. I can update the JSON file in our repo as you suggested, @uglide, but I'll need to be informed if it needs to change again in the future.

uglide:
Imports will be changed in the future version, which has not been released yet. So the right way to keep docs aligned is to pin all client repos to the latest stable tag, so when we make any breaking changes to the master/unstable branch in client library, it doesn't change the docs. Once new stable version released, we will bump the version of the client in the docs (this repo)

dwdougherty:
Cc: @andy-stark-redis. Andy: see Igor's comment above.

PR #3495, opened 2025-02-03
Add types to ConnectionPool.from_url

patrick91:
Exactly like the async version of connection pool ^^

If we had typing extensions we could use "Self" instead of `_CP`,
but that works too.

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -9,7 +9,7 @@ from abc import abstractmethod
 from itertools import chain
 from queue import Empty, Full, LifoQueue
 from time import time
-from typing import Any, Callable, Dict, List, Optional, Type, Union
+from typing import Any, Callable, Dict, List, Optional, Type, TypeVar, Union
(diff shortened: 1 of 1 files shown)


redis/connection.py
```diff
@@ -1278,7 +1281,7 @@ class ConnectionPool:
     """
 
     @classmethod
-    def from_url(cls, url, **kwargs):
+    def from_url(cls: Type[_CP], url: str, **kwargs) -> _CP:
```
vladvildanov, line 1284: I think currently `cls` arg is used as str, it will be a breaking change

patrick91: @vladvildanov not sure I understood, where is it used as string? 😊

I copied the signature from here: https://github.com/redis/redis-py/blob/ceb12fe8b52f488bd1e0a42b9dc53161e77514d0/redis/asyncio/connection.py#L1005

PR #3505, opened 2025-02-11
Replacing the redis and redis-stack-server images with redis-libs-tests image in test infrastructure

petyaslavova:
### Pull Request check-list

### Description of change

* Replacing the `redis` and `redis-stack-server` images with the `redis-libs-tests` image in test infrastructure - change affects docker containers for redis-stack, replica and sentinel.
* Updating the version in setup.py to the current stable version (5.2.1)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ setup(
     long_description_content_type="text/markdown",
     keywords=["Redis", "key-value store", "database"],
     license="MIT",
-    version="5.1.1",
+    version="5.3.0",
     packages=find_packages(
         include=[
             "redis",
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ setup(
     long_description_content_type="text/markdown",
     keywords=["Redis", "key-value store", "database"],
     license="MIT",
-    version="5.3.0",
+    version="5.2.1",
     packages=find_packages(
         include=[
             "redis",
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -8,7 +8,7 @@
     long_description_content_type="text/markdown",
     keywords=["Redis", "key-value store", "database"],
     license="MIT",
-    version="5.1.1",
+    version="5.3.0",
```
vladvildanov, line 11: Version should be 5.2.1 as master always refers to latest stable version. 5.3.0 is in beta so it's only for now in 5.3 version branch

vladvildanov:
@petyaslavova Could we also replace current official `redis-stack-server` image with the one from `client-libs-test` as a part of this PR?

petyaslavova:
> @petyaslavova Could we also replace current official `redis-stack-server` image with the one from `client-libs-test` as a part of this PR?

I will try - I'm not sure what will happen with the older versions that don't have modules, but I will check if/how it can be replaced. ..

vladvildanov:
@petyaslavova There's all versions of redis-stack images that includes modules, should be an easy replacement. Here's a stack images:
rs-7.2.0-v14
rs-6.2.6-v18
rs-7.4.0-v2
https://hub.docker.com/r/redislabs/client-libs-test/tags

setup.py
petyaslavova: Fixed.

vladvildanov (approved):

PR #3502, opened 2025-02-11
test: Updated CredentialProvider test infrastructure

vladvildanov:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [ ] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?
- [ ] Was the change added to CHANGES file?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

_Please provide a description of the change here._

PR #3521, opened 2025-02-21
Isolate redis-entraid dependency for tests

akx:
### Pull Request check-list

### Description of change

This PR follows up on https://github.com/redis/redis-py/pull/3502#issuecomment-2673756540.

It:

* moves the duplicated redis-entraid specific testing code to a separate module
* defers the imports for redis-entraid (so tests can be run even without redis-entraid installed)
* fixes the CI to uninstall the PyPI redis-py version that installing redis-entraid installs, and instead installs the working copy version as editable (so e.g. `metadata.version()` returns the correct version).


akx:
The "CI / Redis 7.4.2; Python 3.8; RESP Parser:plain; EL:uvloop (pull_request)" failure seems to be a flake (where a wait takes 0.202 seconds instead of 0.2).

The failures in "Validate building and installing the package" (which shouldn't be running tests in the first place if you ask me, since there are tests for everything already) seem related to the search module, which this clearly doesn't touch.

```
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_acl_modules_cat_with_category[single] - redis.exceptions.ResponseError: Unknown category 'search'
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_acl_modules_cat_with_category[pool] - redis.exceptions.ResponseError: Unknown category 'search'
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_acl_modules_commands[single] - redis.exceptions.ResponseError: unknown command 'FT.CREATE', with args beginning with: 'idx' 'SCHEMA' 'txt' 'TEXT' 'WEIGHT' '1.0'
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_acl_modules_commands[pool] - redis.exceptions.ResponseError: unknown command 'FT.CREATE', with args beginning with: 'idx' 'SCHEMA' 'txt' 'TEXT' 'WEIGHT' '1.0'
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_acl_modules_category_commands[single] - redis.exceptions.ResponseError: Error in ACL SETUSER modifier '+@search': Unknown command or category name in ACL
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_acl_modules_category_commands[pool] - redis.exceptions.ResponseError: Error in ACL SETUSER modifier '+@search': Unknown command or category name in ACL
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_config_get_for_modules[single] - AssertionError: assert 'search-timeout' in {}
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_config_get_for_modules[pool] - AssertionError: assert 'search-timeout' in {}
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_config_set_for_search_module[single] - KeyError: 'search-default-dialect'
FAILED tests/test_asyncio/test_commands.py::TestRedisCommands::test_config_set_for_search_module[pool] - KeyError: 'search-default-dialect'
```

akx:
@petyaslavova @vladvildanov Rebased post #3518.

petyaslavova (approved):
LGTM.

tests/test_asyncio/test_credentials.py
```diff
@@ -17,10 +17,14 @@
 from redis.credentials import CredentialProvider, UsernamePasswordCredentialProvider
 from redis.exceptions import ConnectionError
 from redis.utils import str_if_bytes
-from redis_entraid.cred_provider import EntraIdCredentialsProvider
 from tests.conftest import get_endpoint, skip_if_redis_enterprise
 from tests.test_asyncio.conftest import get_credential_provider
 
+try:
```
petyaslavova, line 23: The dependency is already added as just a dev dependency. 
To have confidence that the code is working as expected, everyone who needs/wants to make changes in the code should run as many tests as possible. 
Having that in mind, the EntraId tests that do not require specific setup(the tests that do are already marked with cp_integration and skipped from the standard test execution) should also be run - even for people that won't use the functionality, and we should not risk to silently just skip them.
So could you please revert this import to be mandatory?

tests/test_credentials.py
```diff
@@ -16,14 +16,18 @@
...
-from redis_entraid.cred_provider import EntraIdCredentialsProvider
 from tests.conftest import (
     _get_client,
     get_credential_provider,
     get_endpoint,
     skip_if_redis_enterprise,
 )
 
+try:
+    from redis_entraid.cred_provider import EntraIdCredentialsProvider
```
petyaslavova, line 27: ditto

tests/conftest.py
```diff
@@ -623,124 +600,21 @@ def mock_identity_provider() -> IdentityProviderInterface:
...
-        app_kwargs=kwargs,
-    )
-
-
 def get_credential_provider(request) -> CredentialProvider:
     cred_provider_class = request.param.get("cred_provider_class")
     cred_provider_kwargs = request.param.get("cred_provider_kwargs", {})
 
-    if cred_provider_class != EntraIdCredentialsProvider:
+    if not cred_provider_class:
```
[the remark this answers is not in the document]
akx: IMO, the developer should be capable enough to see that the test they just wrote was skipped.

I did it like this to avoid a more intrusive patch to how the tests are currently written (`parametrize`d over lists of credential providers and their args, which there are currently only one).

The other option is of course to generate those credential provider lists so that a `None` value doesn't appear there at all – I can do that, but that'll be a slightly bigger patch. WDYT?

tests/conftest.py
```diff
@@ -623,124 +600,21 @@ def mock_identity_provider() -> IdentityProviderInterface:
...
-    token_mgr_config = TokenManagerConfig(
-        expiration_refresh_ratio=expiration_refresh_ratio,
-        lower_refresh_bound_millis=lower_refresh_bound_millis,
-        token_request_execution_timeout_in_ms=DEFAULT_TOKEN_REQUEST_EXECUTION_TIMEOUT_IN_MS,  # noqa
-        retry_policy=RetryPolicy(
-            max_attempts=max_attempts,
-            delay_in_ms=delay_in_ms,
-        ),
-    )
+    from tests.entraid_utils import get_entra_id_credentials_provider
```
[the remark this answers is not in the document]
akx: No, I can't, because this file can't be imported unless `redis-entraid` is installed.

tests/test_asyncio/test_credentials.py
akx: > The dependency is already added as just a dev dependency.

Unfortunately so, yes. IMO, `redis-entraid` should not be a mandatory development dependency for the core Redis client package at all. 

For one, `redis-entraid` is currently unlicensed, meaning it can't really legally be used by anyone else than its copyright holders for any purpose.'

Having larger mandatory dev dependency chains also spells headache for, say, downstream packagers like Debian or Ubuntu, where the policy is, as far as I know, that packages required for testing a package will also need to be packaged as Debian packages.

Secondly, when someone somewhere decides you'll need first-party support for some AWS-specific authentication, you'd be pulling in that dependency too (and all of the AWS dependencies), etc, and then the same for Google, Oracle, Scaleway, whatever's-the-next-hot-thing. (IMO, TBH, these _tests_ shouldn't be here in the core library at all either...)

The CI process _here_ could pull in every extra dependency required to achieve full coverage, but it doesn't seem worth it for every developer who wants to touch the library (which is, as you probably know yourself, sprawling as it is).

> So could you please revert this import to be mandatory?

That would defeat the whole point of this PR.

tests/conftest.py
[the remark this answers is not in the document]
petyaslavova: It is not always so obvious, especially if you are not familiar with the code base and you are just following the Contributing guide... I suggest failing the test if it is not properly configured.

tests/conftest.py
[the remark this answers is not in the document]
vladvildanov: @akx @petyaslavova is right, redis-entraid package is a dev dependency and it shouldn't be any problems with it's import. We only bring it for test purposes

tests/test_asyncio/test_credentials.py
vladvildanov: @akx You're right about licensing, but there's few exceptions here. First, the package is under `redis-developer` and was created by me, according to laws of intellectual property this project belongs to Redis LTD for now. Seconds is that we still in the beta now, we will add a license file as a part of GA release of `redis-entraid`.

I would say pulling an additional dependency for development shouldn't be really a big deal and it doesn't worth this maintenance efforts to make it optional dev dependency.

This PR is still valuable as it isolates EntraID specifics, in a separate place which kind of makes sense to me.

akx: > First, the package is under redis-developer and was created by me, according to laws of intellectual property this project belongs to Redis LTD for now.

I don't know what you mean with "under redis-developer". The point here is that as it is, any developer who wants to run tests on their machines, or wherever, will now need to pull in code that they have no license to use. That is a problem in my books, and I would assume in the books of the aforementioned downstream packagers such as Debian.

> I would say pulling an additional dependency for development shouldn't be really a big deal 

Installing `redis==5.3.0b5` installs pyjwt and redis (and a future beta won't even install PyJWT thanks to #3518).

Installing `redis-entraid` installs certifi, cffi, charset-normalizer, cryptography, idna, msal, pycparser, pyjwt, redis, edis-entraid, requests, and urllib3.

vladvildanov: As I mentioned, the license issue will be solved until GA. Okay, let's keep it optional. @petyaslavova any other concerns about it?

vladvildanov (approved):

tests/test_asyncio/test_credentials.py
petyaslavova: @vladvildanov, @akx I'm ok with the skipped import error.

tests/conftest.py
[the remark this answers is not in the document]
petyaslavova: Hi @akx, this is the only unresolved thread. Once fixed, we can merge the PR. Considering that not everyone runs tests locally, it's much safer to have failing tests due to improper configuration than to silently skip them.

petyaslavova (approved):

PR #3523, opened 2025-02-21
WIP Harden the replicas check

apolukhin:
Harden the checks

### Pull Request check-list

### Description of change

On MacOS in the `get_replicas()` call returns empty list, even though the `cluster_nodes()` returns primaries and replicas and `RedisCluster` was created after the replicas negotiated with masters. This PR attempts to harden the checks, to make sure that replicas exist.

### DIFF
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -352,7 +352,9 @@ class TestRedisClusterObj:
         Test command execution with nodes flag PRIMARIES
         """
         primaries = r.get_primaries()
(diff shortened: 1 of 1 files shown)


apolukhin:
Discovered that `CLUSTER NODES` may report that everything is fine, however `CLUSTER SLOTS` still does not know about some replicas. Started to use `CLUSTER SLOTS` in our startup scripts and now `get_replicas()` is not empty

PR #3531, opened 2025-02-27
feat: adds option not to raise exception when leaving context manager after lock expiration

julianolm:
Fixes: #3532 

### Pull Request check-list

### Description of change

I was having trouble using async locks with the context manager because it was failing when exiting the context manager after timeout had expired (even if my code executed successfully).

I would really like not to raise a `LockNotOwnedError` in the exit of the context manager (if I try to release a lock and it was already "released" then great!).

I understand just changing the behavior would be critical, so I added an option to just behave this way when desired instead.

Issue: #3532

### DIFF
--- a/redis/asyncio/lock.py
+++ b/redis/asyncio/lock.py
@@ -178,7 +178,9 @@ class Lock:
         except LockError:
             if self.raise_on_release_error:
                 raise
-            logger.warning("Lock was unlocked or no longer owned when exiting context manager.")
+            logger.warning(
+                "Lock was unlocked or no longer owned when exiting context manager."
+            )
 
     async def acquire(
         self,
--- a/redis/asyncio/lock.py
+++ b/redis/asyncio/lock.py
@@ -179,6 +179,10 @@ class Lock:
             if self.raise_on_release_error:
                 raise
             logger.warning("Lock was no longer owned when exiting context manager.")
+        except LockError:
+            if self.raise_on_release_error:
+                raise
+            logger.warning("Lock was unlocked when exiting context manager.")
 
     async def acquire(
         self,
--- a/redis/asyncio/lock.py
+++ b/redis/asyncio/lock.py
@@ -1,6 +1,7 @@
 import asyncio
 import threading
 import uuid
+import logging
 from types import SimpleNamespace
 from typing import TYPE_CHECKING, Awaitable, Optional, Union
 
@@ -9,6 +10,8 @@ from redis.typing import Number
 
 if TYPE_CHECKING:
     from redis.asyncio import Redis, RedisCluster
+
+logger = logging.getLogger(__name__)
 
 
 class Lock:
@@ -85,6 +88,7 @@ class Lock:
         blocking: bool = True,
         blocking_timeout: Optional[Number] = None,
         thread_local: bool = True,
+        raise_on_release_error: bool = True,
     ):
         """
         Create a new Lock instance named ``name`` using the Redis client
@@ -127,6 +131,11 @@ class Lock:
                      token is *not* stored in thread local storage, then
                      thread-1 would see the token value as "xyz" and would be
                      able to successfully release the thread-2's lock.
+        
+        ``raise_on_release_error`` indicates whether to raise an exception when
+        the lock is no longer owned when exiting the context manager. By default,
+        this is True, meaning an exception will be raised. If False, the warning
+        will be logged and the exception will be suppressed.
 
         In some use cases it's necessary to disable thread local storage. For
         example, if you have code where one thread acquires a lock and passes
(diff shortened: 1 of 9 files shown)


julianolm:
PTAL ?
@andymccurdy @dvora-h @chayim @petyaslavova @vladvildanov

redis/asyncio/lock.py
```diff
@@ -162,7 +172,13 @@ async def __aenter__(self):
         raise LockError("Unable to acquire lock within the time specified")
 
     async def __aexit__(self, exc_type, exc_value, traceback):
-        await self.release()
+        try:
+            await self.release()
+        except LockNotOwnedError as e:
+            if self.raise_on_release_error:
+                raise e
```
akx, line 179:
```suggestion
                raise
```
to retain the original traceback?

julianolm: Oh interesting. Thanks!
`raise e` would not?

petyaslavova:
Hi @julianolm, thank you for your contribution! Your change will be reviewed soon.

petyaslavova:
Hi @julianolm, can you please check the reported errors from code linters?

julianolm:
Hey @petyaslavova , updated!!! Thanks!

petyaslavova:
@julianolm  you can run the linters locally - there are instructions in the contributing guide. 
Instead of using the "invoke all-tests" command, you can just call "invoke linters" and you will see all reported errors and changes that are needed.

julianolm:
Oh, great. Sorry I missed it.
Now its ok.
Thanks again @petyaslavova

julianolm:
@petyaslavova wdyt?

redis/asyncio/lock.py
```diff
@@ -162,7 +172,12 @@ async def __aenter__(self):
         raise LockError("Unable to acquire lock within the time specified")
 
     async def __aexit__(self, exc_type, exc_value, traceback):
-        await self.release()
+        try:
+            await self.release()
+        except LockNotOwnedError:
```
petyaslavova, line 177: I think that here you should catch LockError - it can also raise a LockError if for example, release has already been called (you can see PR #3534 for more details). The same applies to the sync lock.py code. And with this change, a new unit test can also be added, covering the LockError case.

petyaslavova:
Hi @julianolm, you have conflicting change with the current master code and small change requested. Once these are resolved, we are ready to merge your change.

julianolm:
Hey @petyaslavova thanks! Gonna address those today asap!

redis/asyncio/lock.py
vladvildanov: +1

julianolm:
Hello, @petyaslavova . Sorry for the delay. Busy day at work today.

Requests addressed. I felt it was better to catch both `LockNotOwnedError` and `LockError` separetely just so we can log more precise warnings on each situation.

But I will leave an opened suggestion for a single catch (of parent class `LockError`) if you think it is better. Please feel free to commit or discard it.

Thanks again!

redis/asyncio/lock.py
julianolm: Addressed here 1bfb7e03bed8994e44436a21a39aa515dee01b1d

redis/asyncio/lock.py
```diff
@@ -163,7 +173,16 @@ async def __aenter__(self):
...
+            if self.raise_on_release_error:
+                raise
+            logger.warning("Lock was no longer owned when exiting context manager.")
+        except LockError:
+            if self.raise_on_release_error:
+                raise
+            logger.warning("Lock was unlocked when exiting context manager.")
```
julianolm, line 185: Here the suggestion I mentioned in my last comment (please see). Feel free to commit or discard it as you prefer

```suggestion
        except LockError:
            if self.raise_on_release_error:
                raise
            logger.warning("Lock was unlocked or no longer owned when exiting context manager.")
```

redis/lock.py
```diff
@@ -168,7 +178,16 @@ def __exit__(
...
+        try:
+            self.release()
+        except LockNotOwnedError:
+            if self.raise_on_release_error:
+                raise
+            logger.warning("Lock was no longer owned when exiting context manager.")
+        except LockError:
+            if self.raise_on_release_error:
+                raise
+            logger.warning("Lock was unlocked when exiting context manager.")
```
julianolm, line 190: Here the suggestion I mentioned in my last comment (please see). Feel free to commit or discard it as you prefer

```suggestion
        except LockError:
            if self.raise_on_release_error:
                raise
            logger.warning("Lock was unlocked or no longer owned when exiting context manager.")
```

julianolm:
PS: I ran the tests with above suggestions applied and it works just fine as well

petyaslavova:
> > But I will leave an opened suggestion for a single catch (of parent class `LockError`) if you think it is better. Please feel free to commit or discard it.
> 
> I left one warning - in case you just want to print a warning instead of failing, I don't think having different detailed messages will be so important.

@julianolm And now unfortunately, linters are failing ...

julianolm:
Hello @petyaslavova linter problems solved

PR #3539, opened 2025-03-04
Fix client_list with multiple client ids

three-kinds:
### Pull Request check-list

### Description of change

* The current client_list function will cause `ResponseError: Invalid client ID` when there are multiple client ids.
* In the current unit test, since the length of `clients` is 1 and `client_id = clients[:-1]`, `client_id` is an empty list, so it does not play the role of testing multiple client ids.
* The fix is simple, please check the code.


petyaslavova:
Hi @three-kinds, thank you for your contribution! Your change will be reviewed soon.

petyaslavova (approved):
LGTM.

Issue #3161, opened 2024-02-26
Increase supported version of pyOpenSSL


libre-man:
Thanks for wanting to report an issue you've found in redis-py. Please delete this text and fill in the template below.  
It is of course not always possible to reduce your code to a small test case, but it's highly appreciated to have as much data as possible. Thank you!

**Version**: 5.0.1

**Platform**: Any platform

**Description**: The redis-py package pins the supported version of pyOpenSSL to a old version, 20, when using the ocsp extras while never versions are already available. Please make this dependency pin less string.

PR #3541, opened 2025-03-04
allow more recent pyopenssl

dimbleby:
per https://github.com/redis/redis-py/pull/1873#issuecomment-1336493561, #3161, #3404 pinning to an exact version of pyopenssl is undesirable, it makes conflicts much more likely in projects that have their own requirement on that library.

pinning to 20.0.1 is especially bizarre: that release was five years ago, and there have been more than a dozen since.

relax the requirement


petyaslavova:
Fixes issue #3161

petyaslavova (approved):

Issue #3555, opened 2025-03-12
Retry Mechanism Fails When Redis Container is Paused


ManelCoutinhoSensei:
### Expected behavior

When the Redis container is **paused** (not stopped), the connection attempt should fail, triggering the retry mechanism. The retry number should increase monotonically until the specified maximum number of retries is reached.

#### Example logs of expected behavior:

```logs
INFO - Attempt 1/5. Backing off for 0.5 seconds
INFO - Attempt 2/5. Backing off for 1.0 seconds
INFO - Attempt 3/5. Backing off for 2.0 seconds
INFO - Attempt 4/5. Backing off for 4.0 seconds
```

The print statement was added at the end of the following `except`  block:
https://github.com/redis/redis-py/blob/ea01a303ab54e9698689b796dc5c67644425ba51/redis/retry.py#L60-L70

### Actual behavior

Instead of progressing through the retry attempts, the retry mechanism gets stuck at the first attempt, repeating indefinitely.

#### Example logs of actual behavior:
```logs
INFO - Attempt 1/5. Backing off for 0.5 seconds
INFO - Attempt 1/5. Backing off for 0.5 seconds
INFO - Attempt 1/5. Backing off for 0.5 seconds
INFO - Attempt 1/5. Backing off for 0.5 seconds
```

### Root Cause

The issue occurs because the `sock.connect` (line 575 of the `_connect` method) succeeds even when the container is `paused`. However, subsequent read operations fail with `Timeout`.
https://github.com/redis/redis-py/blob/ea01a303ab54e9698689b796dc5c67644425ba51/redis/connection.py#L728-L763 

### Possible solution

To properly detect when the connection is truly established, we can send a `PING` command immediately after `connect()` and verify the response.
Add the following after `sock.connect()` to ensure the connection is functional:
```python
ping_parts = self._command_packer.pack("PING")
for part in ping_parts:
    sock.sendall(part)

    response = sock.recv(7)

    if not str_if_bytes(response).startswith("+PONG"):
        raise OSError(f"Redis handshake failed: unexpected response {response!r}")
```

### Additional Comments

- There may be a better way to handle the read operation for the `PING` response using existing methods, but calling `_send_ping` directly does not work in this case.

- This issue also affects the **asynchronous version** of `redis-py`.

-----

Let me know if you'd like a clearer example to reproduce the behavior.

vladvildanov:
Hi! We would have a look on this in meantime

[...]

PR #3559, opened 2025-03-14
(fix): Separated retries for read and write operations

### LINKED ISSUE #3554: Potential Command Duplication in _send_command_parse_response Retry Mechanism
Hi there,

I have a conceptual problem with how the `_send_command_parse_response` is being grouped:

https://github.com/redis/redis-py/blob/ea01a303ab54e9698689b796dc5c67644425ba51/redis/client.py#L590-L595

My issue is that this method is called inside a retry mechanism:

https://github.com/redis/redis-py/blob/ea01a303ab54e9698689b796dc5c67644425ba51/redis/client.py#L623-L628

Which means that if the Redis container goes down after `send_command` and before `parse_response`, the entire `_send_command_parse_response` will be retried. This means that any command with side effects (such as `XADD`) could be executed twice instead of just once, leading to unintended duplication.

Would love to hear thoughts on potential fixes for this issue. Let me know if you need any additional details!

vladvildanov:
### Pull Request check-list

### Description of change

Separate retries for read and write operations to prevent duplicated writes on read failure.

Closes #3554

### DIFF
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -91,7 +91,7 @@ async def test_single_connection():
             await asyncio.gather(r.set("a", "b"), r.set("c", "d"))
 
     assert init_call_count == 1
-    assert command_call_count == 2
+    assert command_call_count == 4
     r.connection = None  # it was a Mock
     await r.aclose()
 
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -3,7 +3,7 @@ import socket
 import sys
 import types
 from errno import ECONNREFUSED
-from unittest.mock import patch
+from unittest.mock import patch, AsyncMock
 
 import pytest
 import redis
@@ -317,7 +317,6 @@ async def test_pool_auto_close(request, from_url):
 
 
 @pytest.mark.onlynoncluster
-@pytest.mark.skipif(sys.version_info < (3, 9), reason="requires python > 3.8")
 async def test_client_do_not_retry_write_on_read_failure(mock_connection, mock_pool):
     mock_connection.send_command.return_value = True
     mock_connection.read_response.side_effect = [
@@ -327,7 +326,7 @@ async def test_client_do_not_retry_write_on_read_failure(mock_connection, mock_p
     ]
     mock_connection.retry = Retry(ExponentialBackoff(), 3)
     mock_connection.retry_on_error = (ConnectionError,)
-    mock_pool.get_connection.return_value = mock_connection
+    mock_pool.get_connection = AsyncMock(return_value=mock_connection)
     mock_pool.connection_kwargs = {}
 
     r = Redis(
@@ -343,7 +342,6 @@ async def test_client_do_not_retry_write_on_read_failure(mock_connection, mock_p
 
 
 @pytest.mark.onlynoncluster
-@pytest.mark.skipif(sys.version_info < (3, 9), reason="requires python > 3.8")
 async def test_pipeline_immediate_do_not_retry_write_on_read_failure(
     mock_connection, mock_pool
 ):
@@ -355,7 +353,7 @@ async def test_pipeline_immediate_do_not_retry_write_on_read_failure(
     ]
     mock_connection.retry = Retry(ExponentialBackoff(), 3)
     mock_connection.retry_on_error = (ConnectionError,)
-    mock_pool.get_connection.return_value = mock_connection
+    mock_pool.get_connection = AsyncMock(return_value=mock_connection)
     mock_pool.connection_kwargs = {}
 
     r = Redis(
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -2,7 +2,7 @@ import asyncio
 import socket
 import types
 from errno import ECONNREFUSED
-from unittest.mock import patch
+from unittest.mock import patch, call
 
 import pytest
 import redis
@@ -20,7 +20,7 @@ from redis.asyncio.connection import (
     parse_url,
 )
 from redis.asyncio.retry import Retry
-from redis.backoff import NoBackoff
+from redis.backoff import NoBackoff, ExponentialBackoff
 from redis.exceptions import ConnectionError, InvalidResponse, TimeoutError
 from redis.utils import HIREDIS_AVAILABLE
 from tests.conftest import skip_if_server_version_lt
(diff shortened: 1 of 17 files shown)


tests/test_asyncio/test_connection.py
```diff
@@ -315,6 +315,65 @@ async def get_redis_connection():
...
+    mock_pool.connection_kwargs = {}
+
+    r = Redis(
+        connection_pool=mock_pool,
+        retry=Retry(ExponentialBackoff(), 3),
+        single_connection_client=True,
+    )
+    await r.set("key", "value")
+
+    # If read from socket fails, writes won't be executed.
```
petyaslavova, line 338: Shouldn't you check here for the read_response function that it is called at least once and add another check that send_command has been executed exactly once?

vladvildanov: Read response has a side effect assigned, so we can be sure that it called exactly 3 times. For send_command it makes sense additionally check that it was called exactly once

vladvildanov: Anyway, I added additional call count for read_response as well

petyaslavova (approved):

ManelCoutinhoSensei (changes requested):
This approach doesn't seem to work:
1. There is no reconnect mechanism inside the `parse_reponse` method. If Redis goes down between `send_command` and the parse retry, it will get stuck retrying `parse_response` indefinitely.
2. Even if you add a guard to `AbstractConnection.read_response`, similar to the one in `can_read` :
https://github.com/redis/redis-py/blob/9251650769f67351bfb51c2da49e26b1b1bf1abe/redis/connection.py#L592-L594
it will still send some commands, such as those in the `on_connect` method, which generate new responses. As a result, the response to the original command (e.g., `XADD`) may no longer be available or the correct one.

Unfortunately, fixing this seems to be more complex than it appears.

vladvildanov:
@ManelCoutinhoSensei I think we're talking here about different issues, the one that you refers is #3555. This one fixes potential duplicate writes if read fails and it does fix that

ManelCoutinhoSensei:
@vladvildanov, That might actually be a better approach than the one I was describing. However, I believe it still encounters the issue I mentioned in my second point:
Calling `AbstractConnection.connect`, will eventually trigger `AbstractConnection.on_connect`, which sets several configurations by sending commands like these:
https://github.com/redis/redis-py/blob/9251650769f67351bfb51c2da49e26b1b1bf1abe/redis/connection.py#L498-L509
As a result, by the time you attempt to read the response to the original command, it will no longer be available.

vladvildanov:
@ManelCoutinhoSensei Well you're right, the existing logic that consider disconnect on retry makes it impossible to separate retries for write and read, I did some investigation in other clients and they also stick to this "transactional" logic when doing retries, so to be consistent we assume that command is called at least once, but you may live with duplicate. Other way you need to handle retries in the application.

https://redis.github.io/lettuce/advanced-usage/#failures-and-at-least-once-execution

ManelCoutinhoSensei:
Hey @vladvildanov,

Before closing this PR and the [issue](https://github.com/redis/redis-py/issues/3554) completely, I wanted to check if an alternative approach could address the duplicate problem while still maintaining the at least once assumption used by other libraries.

Wouldn't something like this work?
```python
try:
    conn.retry.call_with_retry(
        lambda: conn.send_command(*args, **options),
        lambda error: self._disconnect_raise(conn, error),
    )
    return self.parse_response(conn, command_name, **options)
except ParseResponseError as e:
    raise CustomErrorSayingThatReadResponseFailed
finally:
    ...
```

This keeps the changes minimal, avoids the duplication caused by read issues, preserves the original assumption (considering retries are already in place—without them, the guarantee of at least once doesn't hold anyway). and, instead of requiring users to track whether the command was sent once or twice, they would only need to check whether it was successful in case of an error.

I know other clients aren't doing this, but I believe we can do better 😉. This seems like a reasonable compromise between the ideal solution and taking no action.

What do you think?

vladvildanov:
@ManelCoutinhoSensei Currently our users have an expectations that retry happens on timeout errors, your approach doesn't take that into a count. If you disable the retries you would achieve exactly what you mentioned "only need to check whether it was successful in case of an error".

vladvildanov:
@ManelCoutinhoSensei Let's say this way, we can support an additional retry strategy, something like "at most once", but the existing one should be kept

ManelCoutinhoSensei:
@vladvildanov, I have bad news and good news. 

The bad news is that, as it stands, the retry mechanism does not work properly with non-idempotent operations. I believe this should be made clearer in the documentation. In the worst-case scenario, if the Redis connection is unstable, the same command could be sent up to `N_RETRIES` times, with the end user only being aware of the last attempt.

The good news is that since every `execute_command` call is funneled through a single method, I believe we can modify my previous suggestion into a more flexible solution. This would allow the system to support both "**at least once**" and "**at most once**" execution strategies, making the retry mechanism more useful for non-idempotent operations.

Here’s a rough draft of how it could work that would help you jump start this small implementation:
```python

    def _execute_command_with_strategy(self, conn, command_name, *args, **options):
        if self.at_least_once: # Or retrieve it from the retry object (e.g., conn.retry.at_least_once)
            # This behaves as before, ensuring "at least once" execution.
            return self._send_command_parse_response(conn, command_name, *args, **options)

        # "At most once" case:
        # - Send the command with retries
        # - If an exception occurs during response parsing, leave it to the user to handle
        conn.send_command(*args, **options)
        try:
            return self.parse_response(conn, command_name, **options)
        except Exception as e:
            raise InternalException # Decide the exception to send here: if internal or smt already defined for the final user


    def _execute_command(self, *args, **options):
        """Execute a command and return a parsed response"""
        pool = self.connection_pool
        command_name = args[0]
        conn = self.connection or pool.get_connection(command_name, **options)
        try:
            return conn.retry.call_with_retry(
                lambda: self._execute_command_with_strategy(conn, command_name, *args, **options),
                lambda error: self._disconnect_raise(conn, error),

            )
```
This way, we don't have to discard the retry mechanism entirely. Instead, we can provide a configurable approach that better supports non-idempotent operations.

What do you think? 🚀

vladvildanov:
@ManelCoutinhoSensei Looks good to me, I'll take it into an action as a separate PR, will add you as a reviewer

vladvildanov:
@petyaslavova Would appreciate your thoughts on this :arrow_up::arrow_up:

petyaslavova:
> @petyaslavova Would appreciate your thoughts on this ⬆️⬆️

It sounds reasonable. When we have a broken connection, retrying the response reading won't bring any value.
Combined with a new configuration for strategy selection, it would be a working solution for users who don't care how many times the write has been executed—they won't need to change their implementation. At the same time, we would add flexibility for use cases where writing at most once is critical. In both cases, the time spent on reconnection will be saved.

ManelCoutinhoSensei:
@vladvildanov Let me try to rephrase your proposal to make sure I understood it correctly:

You're suggesting the following:
- **At-least-once**: Keep it as I described earlier.
- **At-most-once**: Slightly adjust the logic so that we can still retry `read_response` **if** the exception is either `redis.exceptions.TimeoutError` or `socket.timeout`. In those specific cases, the error handler wouldn’t disconnect, and we'd pass something like `lambda error: None` to skip the disconnect logic.


If that's what you're proposing, I think there are two things worth considering:
1. The `_supported_errors` in the `Retry` object are user-configurable, so modifying the behavior for specific error types might interfere with user expectations. (Also, redoing or wrapping a `Retry` object manually could introduce its own complications—but maybe I'm overthinking this.)
2. I'm not sure if there are edge cases where reading from a socket that has timed out could return stale or out-of-date data (like in some `failover` transition state). That might be worth double-checking.

vladvildanov:
@ManelCoutinhoSensei Yeah, you're get it right! Let me elaborate a bit on your worries:

1. Instead of setting ` lambda error: None` for `read_response` I'm planning to add a condition inside retry callback itself, something like `if error in [TimeoutError, socket.Timeout] conn.disconnect()`, this way we don't need to worry about user-defined errors cause even if it will be fully overridden the initial behaviour will be kept and socket will be closed.

2. That's something that makes sense to check.

vladvildanov:
@ManelCoutinhoSensei Well, I don't get it, but then let's restrict it for `TimeoutError` only for now

ManelCoutinhoSensei:
@vladvildanov I think `TimeoutError` is just a wrapper for the `socket.timeout`. I think that if you are going to do one, do both (or neither)

vladvildanov:
@ManelCoutinhoSensei It's still a question how do we want to handle failures on read in case of "at-most-once" strategies and `ConnectionError`, cause it's impossible to distinguish if error thrown because or read or write. Also it's not that obvious how do we need to behave in case of pipelines and transactions, this thing turns to be more complicated than initially looked like, I started an internal discussion about retries and let you know how do we proceed with it

ManelCoutinhoSensei:
@vladvildanov, Any news regarding the internal discussion?

vladvildanov:
@ManelCoutinhoSensei Unfortunately, it's still in the backlog.

ManelCoutinhoSensei:
@vladvildanov do you have any ETA?

vladvildanov:
@ManelCoutinhoSensei Nothing on my plate now, there's multiple features in development right now

ManelCoutinhoSensei:
@vladvildanov Let me know if you need a hand or if you want to discuss any of this. I'd be more than happy to help!! 😊

Issue #3552, opened 2025-03-11
read_from_replicas=True Still Sends Requests to Master Nodes in RedisCluster


akshitsaini1:
We are using `RedisCluster` with read_from_replicas=True to offload read operations from the master nodes to replicas. However, we have observed that even with this setting enabled, RedisCluster is still sending some read requests to master nodes instead of strictly using replicas.

Upon checking the source code, we found that the LoadBalancer class is responsible for distributing the requests across both master and replica nodes. This behavior is unexpected as the expectation with  `read_from_replicas=True` is that all read operations should strictly go to the replicas.

Impact on Our Setup
1. We handle heavy read/write operations with 100Cr+ (10 billion) reads and writes daily.
2. This issue increases load on master nodes, affecting both write and read latencies.
3. Replicas are not being fully utilized as expected.

Expected Behavior
When read_from_replicas=True is set, all read requests should only go to replica nodes, and master nodes should be used strictly for write operations.

Observed Behavior
Reads are distributed between both master and replica nodes, likely due to how the LoadBalancer class assigns nodes.

This issue affects high-scale production workloads where strict read/write separation is necessary for performance optimization.

akshitsaini1:
Hello @vladvildanov 

Thank you for the quick response! I appreciate the clarification regarding the current strategy. That makes sense.

Could you please share if there’s a planned version where this change might be introduced? This would help us plan accordingly on our end. Looking forward to the update!

petyaslavova:
Fixed with PR #3563.
The fix doesn't change the exposed property meaning or effect, but 'read_from_replicas' is now deprecated.
With this PR, we are introducing another property allowing the user to select which concrete load-balancing strategy should be used.

PR #3563, opened 2025-03-17
Adding load balancing strategy configuration to cluster clients(replacement for 'read_from_replicas' config)

petyaslavova:
### Pull Request check-list

### Description of change

Adding load balancing strategy configuration to cluster clients(replacement for 'read_from_replicas' config)
The idea of this setting is to replace the currently existing 'read_from_replicas' configuration - its name is confusing. It leaves the impression that all the reads will be sent to replica nodes, while it actually chooses between the primary and replicas in a round-robin manner.
There was one specific situation in which random replica selection was also exposed - with this change, this configuration will be exposed through the RedisCluster objects for sync and async clients.

- Adding a new configuration field for Redis Cluster objects - load_balancing_strategy
- Adding deprecation for one of the current configs - read_from_replicas
- Adding a new enum for the available load-balancing strategies (LoadBalancingStrategy)
- Adding unit tests for the cluster and cluster pipeline operations with the new load-balancing strategies
- Adding unit test coverage for the added load-balancing strategies
- Adding some setup fixes for multiptocessing unit tests (the issue resurfaced for Mac users with a recent merge)

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -124,13 +124,14 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
           https://redis.io/docs/manual/scaling/#redis-cluster-configuration-parameters
     :param read_from_replicas:
         | @deprecated - please use load_balancing_strategy instead
-        | Enable read from replicas in READONLY mode. You can read possibly stale data.
+        | Enable read from replicas in READONLY mode.
           When set to true, read commands will be assigned between the primary and
           its replications in a Round-Robin manner.
+          The data read from replicas is eventually consistent with the data in primary nodes.
     :param load_balancing_strategy:
         | Enable read from replicas in READONLY mode and defines the load balancing
           strategy that will be used for cluster node selection.
-          You can read possibly stale data (when reading from replicas).
+          The data read from replicas is eventually consistent with the data in primary nodes.
     :param reinitialize_steps:
         | Specifies the number of MOVED errors that need to occur before reinitializing
           the whole cluster topology. If a MOVED error occurs and the cluster does not
@@ -236,7 +237,7 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
         startup_nodes: Optional[List["ClusterNode"]] = None,
         require_full_coverage: bool = True,
         read_from_replicas: bool = False,
-        load_balancing_strategy: Union[LoadBalancingStrategy, None] = None,
+        load_balancing_strategy: Optional[LoadBalancingStrategy] = None,
         reinitialize_steps: int = 5,
         cluster_error_retry_attempts: int = 3,
         connection_error_retry_attempts: int = 3,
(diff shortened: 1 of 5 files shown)


petyaslavova:
Fixes issue #3552

redis/cluster.py
```diff
@@ -520,11 +527,16 @@ def __init__(
...
+             @deprecated - please use load_balancing_strategy instead
              Enable read from replicas in READONLY mode. You can read possibly
              stale data.
              When set to true, read commands will be assigned between the
              primary and its replications in a Round-Robin manner.
-         :param dynamic_startup_nodes:
+        :param load_balancing_strategy:
+             Enable read from replicas in READONLY mode and defines the load balancing
+             strategy that will be used for cluster node selection.
+             You can read possibly stale data (when reading from replicas).
```
vladvildanov, line 538: Just an opinion, maybe it makes sense to use a phrase "Eventually consistent data" and not "Stale data". Eventual consistence is something that defined by replication mechanism itself and not something that our client affects

tests/test_cluster.py
```diff
@@ -3333,6 +3403,45 @@ def test_readonly_pipeline_from_readonly_client(self, request):
...
+            load_balancing_strategy=load_balancing_strategy,
+        )
+        key = "bar"
+        ro.set(key, "foo")
+        import time
+
+        time.sleep(0.2)
+
+        with ro.pipeline() as readonly_pipe:
+            mock_all_nodes_resp(ro, "MOCK_OK")
```
vladvildanov, line 3432: I think it makes sense to add an integration tests as well. Very generic one just to prove that writes and reads are working with all strategies as expected. And make it for pipelines as well as standard mode

vladvildanov (changes requested):

redis/asyncio/cluster.py
```diff
@@ -224,6 +236,7 @@ def __init__(
         startup_nodes: Optional[List["ClusterNode"]] = None,
         require_full_coverage: bool = True,
         read_from_replicas: bool = False,
+        load_balancing_strategy: Union[LoadBalancingStrategy, None] = None,
```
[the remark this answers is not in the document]
petyaslavova: Done.

redis/cluster.py
petyaslavova: Changed the working in both files.

tests/test_cluster.py
petyaslavova: I was thinking the same when I started the tests, but in those cases, we won't be able to see the changes in the indexes of replica nodes. 
Our setup is defined to have 6 nodes - 3 primaries with one replica each.
In order to validate the flows properly, we should change it to have at least two replicas for each primary node.
@vladvildanov are you ok with this change in the infra? Tt will also lead to changes in cluster readiness detection in the test setup.

vladvildanov: I totally get the point, we can't track the actual requests and how they distributed. But even simple `set, get` test with enabled balancing (each of them) will be beneficial so we will be sure that requests are still routed to the correct shards of the cluster and load balancing doesn't break any of existing logic.

petyaslavova: Done.

vladvildanov (approved):

Issue #3593, opened 2025-04-10
Cannot use IntEnum as ex value


phi-friday:
**Describe the bug**

Cannot use IntEnum as ex value.

**To reproduce**

```python
from __future__ import annotations

import asyncio
from enum import IntEnum

from redis.asyncio import Redis


class SomeEnum(IntEnum):
    EX = 60


async def main() -> None:
    client = Redis.from_url("redis://localhost:6379/0")
    await client.set("key", "value", ex=SomeEnum.EX)


if __name__ == "__main__":
    asyncio.run(main())
```

output:
```bash
❯ uv run python ./test.py
Traceback (most recent call last):
  File "***/./test.py", line 19, in <module>
    asyncio.run(main())
    ~~~~~~~~~~~^^^^^^^^
  File "/home/phi/.local/share/uv/python/cpython-3.13.2-linux-x86_64-gnu/lib/python3.13/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/home/phi/.local/share/uv/python/cpython-3.13.2-linux-x86_64-gnu/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/phi/.local/share/uv/python/cpython-3.13.2-linux-x86_64-gnu/lib/python3.13/asyncio/base_events.py", line 725, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "***/./test.py", line 15, in main
    await client.set("key", "value", ex=SomeEnum.EX)
  File "***/.venv/lib/python3.13/site-packages/redis/asyncio/client.py", line 616, in execute_command
    return await conn.retry.call_with_retry(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
    )
    ^
  File "***/.venv/lib/python3.13/site-packages/redis/asyncio/retry.py", line 59, in call_with_retry
    return await do()
           ^^^^^^^^^^
  File "***/.venv/lib/python3.13/site-packages/redis/asyncio/client.py", line 590, in _send_command_parse_response
    return await self.parse_response(conn, command_name, **options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3594, opened 2025-04-10
Fix: Allow IntEnum as ex, px arguments.

phi-friday:
### Pull Request check-list

### Description of change

Closes: #3593

### DIFF
--- a/redis/utils.py
+++ b/redis/utils.py
@@ -276,7 +276,7 @@ def extract_expire_flags(
         if isinstance(ex, datetime.timedelta):
             exp_options.append(int(ex.total_seconds()))
         elif isinstance(ex, int):
-            exp_options.append(ex)
+            exp_options.append(int(ex))
         elif isinstance(ex, str) and ex.isdigit():
             exp_options.append(int(ex))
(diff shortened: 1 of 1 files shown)


petyaslavova:
Hi @phi-friday, thank you for your contribution — I really appreciate your effort!

In this particular case, I have some reservations about merging it. The issue lies in the type of the property in question: the expected types are limited to int or timedelta, and IntEnum values are not included in that specification.

Since the type hint is explicit, we should avoid adding additional parsing logic to handle other types that fall outside the intended scope.

phi-friday:
> Hi @phi-friday, thank you for your contribution — I really appreciate your effort!
> 
> In this particular case, I have some reservations about merging it. The issue lies in the type of the property in question: the expected types are limited to int or timedelta, and IntEnum values are not included in that specification.
> 
> Since the type hint is explicit, we should avoid adding additional parsing logic to handle other types that fall outside the intended scope.

An instance of IntEnum is an instance of int, so there is no problem with type hint, but I understand the decision not to support subclasses of int.

PR #3557, opened 2025-03-12
Fixed infinitely recursive health checks

excitoon:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [ ] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [x] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?
- [ ] Was the change added to CHANGES file?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

Fixed infinite recursion in health checks (once and for all).

Fixes #2866

Issue #3203, opened 2024-04-10
Bug in retry-handling in the PubSub-client


sondrfos:
**Version**: 
redis-py: 5.0.3
redis: 7.2.4

**Platform**: 
Python 3.12.2 running on Debian bookworm

**Description**: 
First off I want to say thanks for maintaining this library :) 

So to the issue at hand:
It seems like the retry-functionality in the PubSub-client (both async and sync) is not working correctly. 
To simulate a crash of redis I am closing and restarting redis while our system is running. Even though our connection is set up with an infinite amount of retry, the PubSub-client always crashes and never recovers.

Having dug though the source code I have found the pathways causing the crash.
https://github.com/redis/redis-py/blob/1784b37f548076aca16d7236882c085428f9b23e/redis/asyncio/client.py#L885-L896
The `_execute`-function runs periodicially, and whenever redis is closed the command fails.
This causes the `_disconnect_raise_connect`-function to be called:
https://github.com/redis/redis-py/blob/1784b37f548076aca16d7236882c085428f9b23e/redis/asyncio/client.py#L870-L883
`_disconnect_raise_connect` closes the connection, `conn.retry_on_error` is `True` making the function skip raising error. 
This is followed by an attempt at reconnecting.
https://github.com/redis/redis-py/blob/1784b37f548076aca16d7236882c085428f9b23e/redis/asyncio/connection.py#L269-L284
Since we are not currently connected, the first if-statement is `False`. A call with retry is made to the `_connect`-function.
https://github.com/redis/redis-py/blob/1784b37f548076aca16d7236882c085428f9b23e/redis/asyncio/connection.py#L678-L700
Inside the `_connect`-function an attempt is made to connect to redis. Since redis is unavailable an `OSError` is created. `OSError` is not in the list of  `supported_errors` causing it to bypass the error handling in the `call_with_retry`-function and is propagated to the `connect`-function. Here it is converted to a `ConnectionError` and is propagated all the way to the top unhandled. 

Is this intended behaviour?
Is this something that could be handled in a different way?

We have temporarily fixed this by adding `OSError` to the list of `supported_errors`, but this feels like quite a hacky solution.

ManelCoutinhoSensei:
Hey,

Based on your description, I believe I've encountered the same issue with the synchronous version (as you said).

### Minimal Reproducible Example

The following pseudo-code snippet helps to reproduce the issue:
```python
def event_handler(msg):
    print(msg)

[...]

PR #3596, opened 2025-04-11
Add Redis readiness verification (#3555)

### LINKED ISSUE #3555: Retry Mechanism Fails When Redis Container is Paused

ManelCoutinhoSensei:
### Pull Request check-list

### Description of change

Fixes #3555 (specially the [comment](https://github.com/redis/redis-py/issues/3555#issuecomment-2729673443) on that issue)

Please check if you'd rather run the tests with `check_ready=True` by default (I also tested that) and if you think that any other tests deserve a second version with this flag set.


ManelCoutinhoSensei:
@petyaslavova you need to rerun the failed tests due to the test suit flakiness

redis/asyncio/connection.py
```diff
@@ -933,7 +968,6 @@ async def _connect(self):
             reader, writer = await asyncio.open_unix_connection(path=self.path)
         self._reader = reader
         self._writer = writer
-        await self.on_connect()
```
petyaslavova, line 964: Why do you remove this call?

ManelCoutinhoSensei:
Hi @petyaslavova , any news about this one?

petyaslavova:
Hi @ManelCoutinhoSensei, I have started the review, but couldn't complete it. 
We have some urgent changes we need to complete, and the PR reviews will be delayed by a week or two. 
Sorry for the delay!

ManelCoutinhoSensei:
Should I solve the conflicts because the PR reviews are going to be done this week or should I postpone it?

ManelCoutinhoSensei:
Hey @petyaslavova! 🙌 Just checking in to see if there are any updates or an estimated timeline for moving forward with this PR. Totally understand things can get busy — just wanted to make sure it’s still on your radar.
Thanks again for your time and support! 😊

ManelCoutinhoSensei:
cyring-in-the-rain

redis/connection.py
```diff
@@ -378,12 +380,35 @@ def connect(self):
         "Connects to the Redis server if not already connected"
         self.connect_check_health(check_health=True)
 
+    def _connect_check_ready(self):
+        sock = self._connect()
+
+        # Doing handshake since connect and send operations work even when Redis is not ready
+        if self.check_ready:
+            try:
+                ping_parts = self._command_packer.pack("PING")
```
petyaslavova: I think it would be better to replace this block with a call to `self._send_ping()`, and refactor that method to accept an additional argument—something like `accept_no_auth_response=False` by default. This way, you can reuse the existing logic without duplicating the socket write/read operations, making the code cleaner and easier to maintain.

redis/client.py
```diff
@@ -203,6 +203,7 @@ def __init__(
         encoding: str = "utf-8",
         encoding_errors: str = "strict",
         decode_responses: bool = False,
+        check_ready: bool = False,
```
petyaslavova: Please add details about this setting to the pydocs. Also, I would suggest changing the name so that it would be easier to understand its purpose just by looking at the name, something like `check_server_ready`, or `on_connect_check_server_ready`

redis/connection.py
```diff
@@ -378,12 +380,35 @@ def connect(self):
         "Connects to the Redis server if not already connected"
         self.connect_check_health(check_health=True)
 
+    def _connect_check_ready(self):
```
petyaslavova: Please move this method after `connect_check_health `

redis/client.py
ManelCoutinhoSensei: I'm not 100% sure what you mean by `pydocs`.
I added to the `docstrings` of both methods, let me know if that was what was intended.

redis/connection.py
ManelCoutinhoSensei: I understand the desired but in order to use the `read_response` part of the `_send_ping`, the parsers should already be initialized (which only happens on `on_connect_check_health` that is called after the `_connect_check_server_ready`).
Either way I refactored the `_connection` method so that i could use the `send_command` command part, let me know if this suffices or if we should consider refactoring the connection flow to accommodate an earlier parser initialization

redis/asyncio/connection.py
ManelCoutinhoSensei: Two reasons:
1. Make the connection flow consistent between sync and async versions
2. It wasn't present in the normal `Connection` making it inconsistent between the different async connection objects

---
This is the connection flow:
```
connect → connect_check_health:
         - _connect_check_ready (with retries) → _connect
         - on_connect_check_health(True)
-----
on_connect → on_connect_check_health(True)
```
This means that adding `on_connect` to the end of the `UnixDomainSocket._connect` will only duplicate the `on_connect_check_health`  call (but I might be missing something)

ManelCoutinhoSensei:
Hey @petyaslavova, just updated this PR after the merge of #3601 
One of the failing test-suits is the same problem as before and the other seems to be flaky (since it passed in my [run](https://github.com/ManelCoutinhoSensei/redis-py/actions/runs/16597016131?pr=2)).
Let me know if there are any follow ups needed to your comments or any new changes that you'd like to see made

ManelCoutinhoSensei:
Hey @petyaslavova, have you had the chance of reviewing the changes you asked?

petyaslavova:
> Hey @petyaslavova, have you had the chance of reviewing the changes you asked?

Hi @ManelCoutinhoSensei, I was in vacation for a while and I didn't have a chance to look at this. I'm trying to catch up with the PRs this and next week.

ManelCoutinhoSensei:
Hey @petyaslavova, I updated it and it is passing all the tests on my side

petyaslavova:
> Hey @petyaslavova, I updated it and it is passing all the tests on my side

Hi @ManelCoutinhoSensei, there are some details in this PR that can't be merged into the current master state - there is a new functionality in the client that handles specific push notifications, which can be received with the command responses. If you read directly from the socket, you won't be able to use the automatic handling of those messages, which happens in the parser. 
Can't it be transformed to use `read_response` function(the parser is already initialized when those requests are sent)?

ManelCoutinhoSensei:
Hey @petyaslavova,

>  (...) Can't it be transformed to use read_response function(the parser is already initialized when those requests are sent)?

does [this comment](https://github.com/redis/redis-py/pull/3596#discussion_r2176235275) from July 1st address what you are asking? Or is it no longer up to date?

petyaslavova:
> Hey @petyaslavova,
> 
> > (...) Can't it be transformed to use read_response function(the parser is already initialized when those requests are sent)?
> 
> does [this comment](https://github.com/redis/redis-py/pull/3596#discussion_r2176235275) from July 1st address what you are asking? Or is it no longer up to date?

I have to admit I’d forgotten about this one, but yes — this is exactly the right solution here.

ManelCoutinhoSensei:
> > > (...) Can't it be transformed to use read_response function(the parser is already initialized when those requests are sent)?
> > 
> > 
> > does [this comment](https://github.com/redis/redis-py/pull/3596#discussion_r2176235275) from July 1st address what you are asking? Or is it no longer up to date?
> 
> I have to admit I’d forgotten about this one, but yes — this is exactly the right solution here.

Which part is the right solution? Are you saying that using the `send_command` is sufficient or that we should refactor the connection flow to initialize the parser earlier?

petyaslavova:
> > > > (...) Can't it be transformed to use read_response function(the parser is already initialized when those requests are sent)?
> > > 
> > > 
> > > does [this comment](https://github.com/redis/redis-py/pull/3596#discussion_r2176235275) from July 1st address what you are asking? Or is it no longer up to date?
> > 
> > 
> > I have to admit I’d forgotten about this one, but yes — this is exactly the right solution here.
> 
> Which part is the right solution? Are you saying that using the `send_command` is sufficient or that we should refactor the connection flow to initialize the parser earlier?

The parser is initialized in the init method when resp3 is used - https://github.com/redis/redis-py/blob/8403ddcfcf16c95baab330076b4905483e44fbb7/redis/connection.py#L429C25-L429C37
This logic can be extended to set the parser when protocol is 2 as well, and then you will be able to use directly the `_send_ping` command to validate server is already ready to respond to requests.

ManelCoutinhoSensei:
> The parser is initialized in the init method when resp3 is used - https://github.com/redis/redis-py/blob/8403ddcfcf16c95baab330076b4905483e44fbb7/redis/connection.py#L429C25-L429C37
> This logic can be extended to set the parser when protocol is 2 as well,

I might be wrong but even when using resp3, the parser is only initialized (meaning only the `socket_read_size` is set). For the `read_response` method to work, we need to call the `Parser.on_connect`  method (that initializes the `_buffer` and the `enconder`) which is only called afterwards on `on_connect_check_health` (like previously explained)

petyaslavova:
> > The parser is initialized in the init method when resp3 is used - https://github.com/redis/redis-py/blob/8403ddcfcf16c95baab330076b4905483e44fbb7/redis/connection.py#L429C25-L429C37
> > This logic can be extended to set the parser when protocol is 2 as well,
> 
> I might be wrong but even when using resp3, the parser is only initialized (meaning only the `socket_read_size` is set). For the `read_response` method to work, we need to call the `Parser.on_connect` method (that initializes the `_buffer` and the `enconder`) which is only called afterwards on `on_connect_check_health` (like previously explained)

You’re absolutely right. 🙂

I finally had time to carefully review (and debug) the code related to the reported issue and this PR. My conclusion is that this change would introduce additional complexity without fully resolving the core problem.

With the current approach, we’d force a PING on all connections, even though we already have health checks available. When health checks are enabled, a PING is executed just before the first real command as part of the initial client setup.
The issue is that some clients of the library may prefer not to enable health checks because they add extra command executions.

In my opinion, the correct solution is to move the initial health-check logic (executed after the socket connect) into the retry block. That way, we can retry the entire connection-establishment flow—not just the socket connect.

Something like the following will handle the retry of the whole connection establishment logic:
`    
def connect(self):
        "Connects to the Redis server if not already connected"
        self.retry.call_with_retry(
            lambda: self.connect_check_health(
                check_health=True, retry_socket_connect=False
            ),
            lambda error: self.disconnect(error),
        )
`
This will try once the socket connect and run the commands in the `connect_check_health` method, in case of failure of any of those two - this whole flow will be retried after disconnect is called as error handler.

@ManelCoutinhoSensei, what do you think about this change in direction?

ManelCoutinhoSensei:
Hey @petyaslavova,

Thanks a lot for taking a close look at this issue! 🙏

The problem that I originally flagged still exists (`_connect` doesn't guarantee _connectedness_ in cases of idle containers). That said, the issue seems largely solved by [fixing the recursion bug](https://github.com/redis/redis-py/pull/3557) - at least in my toy examples. It might reappear in more complex scenarios.

What I was proposing was essentially the current approach (sending a `PING` after attempting to connect to confirm the connection). So both your suggestion and mine introduce a bit of extra complexity, without much additional benefit.
I think that as long as we keep in mind that `_connect` doesn't guarantee a fully working connection, we're fine to close this branch (and related [issue](https://github.com/redis/redis-py/issues/3555#issuecomment-3301630330)) with one caveat: <ins>authenticated requests</ins>.

### Authenticated requests
If you look at the authentication-related code ([650-652](https://github.com/redis/redis-py/blob/8f56b52162fbf636cad3bd861d06b79cf7a71b48/redis/connection.py#L650-L652), [661](https://github.com/redis/redis-py/blob/8f56b52162fbf636cad3bd861d06b79cf7a71b48/redis/connection.py#L661) and [670](https://github.com/redis/redis-py/blob/8f56b52162fbf636cad3bd861d06b79cf7a71b48/redis/connection.py#L670)) we're hardcoding `check_health=False` to avoid error when `PING` fails. I think that we could/should change this to use the same pattern as the other commands (`check_health=check_health`) and then change `_send_ping` so it accepts `AuthenticationError` as valid outcome (since this method is only used to check server health and this does, indeed, mean that the server is alive and healthy).

That would require setting `disconnect_on_error=False` for `read_response`, but those cases are already handled properly. In the end, something like this should be close to what we need:
```python
def _send_ping(self):
    """Send PING, expect PONG in return"""
    self.send_command("PING", check_health=False)
    try:
        if str_if_bytes(self.read_response(disconnect_on_error=False)) != "PONG":
            raise ConnectionError("Bad response from PING health check")
    except AuthenticationError:
        # That's fine -> server is healthy
        # Reset _parser due to the rewind on error mechanic that would make it read the same answer twice
        # The rational behind rewind on error when disconnect_on_error is deactivated should be rethought perhaps
        self._parser.on_disconnect()
        self._parser.on_connect(self)
```


Let me know what you think! 😊

ManelCoutinhoSensei:
Hi again @petyaslavova 👋,

I’ve been experimenting with [this other issue](https://github.com/redis/redis-py/issues/3203) (better explained in [this comment](https://github.com/valkey-io/valkey-py/issues/169#issuecomment-2704939737)) and realized that my previous explanation wasn’t entirely accurate.

Here’s the situation:

* If the connection is successfully established but the container later goes idle, we attempt to reconnect.
* However, since the last `check_health` may have been performed very recently, that step will be skipped.

This was something I didn’t account for in my [previous comment](https://github.com/redis/redis-py/pull/3596#issuecomment-3352915216).

To address this, I updated the synchronous version with a proposal where the `PING` command is moved **unconditionally** outside of `send_commands`. In other words, regardless of the last ping or the `health_check_interval`, we always check the connection when starting a new one.

I’ve also removed the previous flag since, as mentioned earlier, the library’s default behavior already includes sending this `PING`. This change keeps that same behavior (though I can re-add the flag if you prefer).

On top of that, this approach also resolves the #3203 (mentioned before) - at least in the way I’ve been reproducing it.

Eager to hear your thoughts! Should we move forward with this solution and implement it for the async version as well? 🚀

ManelCoutinhoSensei:
Hi @petyaslavova , I have you add the chance to read my comments?
I would hope to solve this ASAP so that it doesn't happen the same as it happened last time (new changes were added and the solution became obsolete)

ManelCoutinhoSensei:
Sorry for pinning you again, but did you have the chance to look at the idea behind this new PR @petyaslavova ? It can potentially solve 2 open PRs...

petyaslavova:
> Sorry for pinning you again, but did you have the chance to look at the idea behind this new PR @petyaslavova ? It can potentially solve 2 open PRs...

Hi @ManelCoutinhoSensei,

Unfortunately, it’s been a pretty intense schedule lately, and I haven’t had the chance to review the new details yet. Since this change impacts quite a lot, it’ll need some focused attention. I still have the PR open in my browser, so I won’t forget to get back to it. 🙂

petyaslavova:
Thank you again, @ManelCoutinhoSensei, for the work you put into this PR and for the thorough investigation behind it. Your analysis was extremely helpful — in fact, it directly led us to discover gaps in our retry logic that weren’t addressed elsewhere.

We’ve incorporated the necessary fixes in a separate PR, which expands the retry handling to cover both the socket connection and the initial handshake. Because of that, I’m going to close this PR now.

That said, your contribution was genuinely valuable, both in surfacing the underlying issue and in helping shape the final solution. Really appreciate the time and care you put into this — thank you! 🙏

If you spot anything else or want to continue the discussion, feel free to reopen or open a new PR anytime.

PR #3614, opened 2025-04-23
Updating default retry strategy for standalone clients. 3 retries with ExponentialWithJitterBackoff become the default config.

petyaslavova:
### Pull Request check-list

### Description of change

Updating default retry strategy for standalone clients. 
3 retries with ExponentialWithJitterBackoff become the default config.

The default retry configuration for the sync and async standalone clients are affected. 
This change affects the cluster nodes Redis clients so when those clients are created a retry object with 0 retries and NoBackOff strategy is provided.


redis/asyncio/retry.py
```diff
@@ -43,6 +43,18 @@ def update_supported_errors(self, specified_errors: list):
             set(self._supported_errors + tuple(specified_errors))
         )
 
+    def get_retries_count(self) -> int:
```
vladvildanov, line 46: Unused methods?

redis/retry.py
```diff
@@ -44,6 +44,12 @@ def update_supported_errors(
             set(self._supported_errors + tuple(specified_errors))
         )
 
+    def update_retries_count(self, retries: int) -> None:
```
vladvildanov, line 47: Unused method?

vladvildanov (changes requested):

redis/asyncio/retry.py
petyaslavova: Same as in the other client - will remove it from this commit and reintroduce it with next change.

redis/asyncio/client.py
```diff
@@ -929,19 +947,14 @@ async def connect(self):
             )
         )
 
-    async def _disconnect_raise_connect(self, conn, error):
+    async def _reconnect(self, conn):
         """
         Close the connection and raise an exception
```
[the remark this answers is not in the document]
petyaslavova: Done.

vladvildanov (approved):

ManelCoutinhoSensei:
Hey @petyaslavova, can you justify this change here?
I'm not sure why adding retries by default is a desirable behavior but at the very least I think you failed to consider the Sentinel case.

If we do smt like this:
```python
sentinel = Sentinel(
        [("localhost", 26370)],
        encoding="utf-8",
        decode_responses=True,
        socket_keepalive=True,
        socket_timeout=1,
        socket_connect_timeout=1,
        health_check_interval=5,
        retry=Retry(ExponentialBackoff(10, 0.5), 5),
    )
```
we'll have 5 x 3 retries on every command because the internal sentinels are create like:
https://github.com/redis/redis-py/blob/7081047e5db0b9e7ba3ef52f187619fa2a9d9abc/redis/sentinel.py#L247-L250

it seems weird that you have to do this in order to turn off this (wrong) behavior:

```python
sentinel = Sentinel(
        [("localhost", 26370)],
        encoding="utf-8",
        decode_responses=True,
        socket_keepalive=True,
        socket_timeout=1,
        socket_connect_timeout=1,
        health_check_interval=5,
        retry=Retry(ExponentialBackoff(10, 0.5), 5),
        sentinel_kwargs={"retry": None}
    )
```
Specially because now the user doesn't know if it should add the "socket_" options to that `sentinel_kwargs` as well or not do to the way that the default behavior works:
https://github.com/redis/redis-py/blob/7081047e5db0b9e7ba3ef52f187619fa2a9d9abc/redis/sentinel.py#L239-L253

ManelCoutinhoSensei:
Hey @petyaslavova, can you justify this change here?
I'm not sure why adding retries by default is a desirable behavior but at the very least I think you failed to consider the Sentinel case.

If we do smt like this:
```python
sentinel = Sentinel(
        [("localhost", 26370)],
        encoding="utf-8",
        decode_responses=True,
        socket_keepalive=True,
        socket_timeout=1,
        socket_connect_timeout=1,
        health_check_interval=5,
        retry=Retry(ExponentialBackoff(10, 0.5), 5),
    )
```
we'll have 5 x 3 retries on every command because the internal sentinels are create like:
https://github.com/redis/redis-py/blob/7081047e5db0b9e7ba3ef52f187619fa2a9d9abc/redis/sentinel.py#L247-L250

it seems weird that you have to do this in order to turn off this (wrong) behavior:

```python
sentinel = Sentinel(
        [("localhost", 26370)],
        encoding="utf-8",
        decode_responses=True,
        socket_keepalive=True,
        socket_timeout=1,
        socket_connect_timeout=1,
        health_check_interval=5,
        retry=Retry(ExponentialBackoff(10, 0.5), 5),
        sentinel_kwargs={"retry": None}
    )
```
Specially because now the user doesn't know if it should add the "socket_" options to that `sentinel_kwargs` as well or not do to the way that the default behavior works:
https://github.com/redis/redis-py/blob/7081047e5db0b9e7ba3ef52f187619fa2a9d9abc/redis/sentinel.py#L239-L253

PR #3626, opened 2025-04-30
Changing the default value for ssl_check_hostname to True, to ensure security validations  are not skipped by default

petyaslavova:
### Pull Request check-list

### Description of change

Changing the default value for ssl_check_hostname to True, to ensure security validations  are not skipped by default

### DIFF
--- a/docs/examples/ssl_connection_examples.ipynb
+++ b/docs/examples/ssl_connection_examples.ipynb
@@ -34,9 +34,10 @@
     "import redis\n",
     "\n",
     "r = redis.Redis(\n",
-    "    host='localhost', \n",
-    "    port=6666, \n",
-    "    ssl=True, \n",
+    "    host='localhost',\n",
+    "    port=6666,\n",
+    "    ssl=True,\n",
+    "    ssl_check_hostname=False,\n",
     "    ssl_cert_reqs=\"none\",\n",
     ")\n",
     "r.ping()"
@@ -68,7 +69,7 @@
    "source": [
     "import redis\n",
     "\n",
-    "r = redis.from_url(\"rediss://localhost:6666?ssl_cert_reqs=none&decode_responses=True&health_check_interval=2\")\n",
+    "r = redis.from_url(\"rediss://localhost:6666?ssl_cert_reqs=none&ssl_check_hostname=False&decode_responses=True&health_check_interval=2\")\n",
     "r.ping()"
    ]
   },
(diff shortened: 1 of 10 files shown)


docs/examples/ssl_connection_examples.ipynb
```diff
@@ -34,9 +34,10 @@
     "import redis\n",
     "\n",
     "r = redis.Redis(\n",
-    "    host='localhost', \n",
-    "    port=6666, \n",
-    "    ssl=True, \n",
+    "    host='localhost',\n",
+    "    port=6666,\n",
+    "    ssl=True,\n",
+    "    ssl_check_hostname=False,\n",
```
vladvildanov, line 40: Just for confirmation, we need to set this to False because otherwise it wouldn't work?

vladvildanov (changes requested):

tests/test_asyncio/test_connect.py
```diff
@@ -68,6 +68,7 @@ async def test_tcp_ssl_tls12_custom_ciphers(tcp_address, ssl_ciphers):
         socket_timeout=10,
         ssl_min_version=ssl.TLSVersion.TLSv1_2,
         ssl_ciphers=ssl_ciphers,
+        ssl_check_hostname=False,
```
elena-kolevska, line 71: Can we please update the tests so they pass with `ssl_check_hostname=True`?

tests/test_ssl.py
```diff
@@ -37,7 +37,13 @@ def test_ssl_with_invalid_cert(self, request):
     def test_ssl_connection(self, request):
         ssl_url = request.config.option.redis_ssl_url
         p = urlparse(ssl_url)[1].split(":")
-        r = redis.Redis(host=p[0], port=p[1], ssl=True, ssl_cert_reqs="none")
+        r = redis.Redis(
+            host=p[0],
+            port=p[1],
+            ssl=True,
+            ssl_check_hostname=False,
```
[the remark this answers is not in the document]
elena-kolevska: I would say we should be testing with both false and true, but at least with `ssl_check_hostname=True`

tests/test_asyncio/test_connect.py
petyaslavova: The certificates we have are not appropriate and the hostname validation fails. I will add the suggested comments why this is done for the tests.

tests/test_ssl.py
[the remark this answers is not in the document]
petyaslavova: Fixed where possible.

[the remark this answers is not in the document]
petyaslavova: In this exact case the ssl_check_hostname is False, because we have ssl_cert_reqs="none". 
ssl_cert_reqs="none" is not compatible with ssl_check_hostname=True

redis/connection.py
```diff
@@ -1028,7 +1028,7 @@ def __init__(
         ssl_cert_reqs="required",
         ssl_ca_certs=None,
         ssl_ca_data=None,
-        ssl_check_hostname=False,
+        ssl_check_hostname=True,
```
harpener, line 1031: @petyaslavova FYI, you left the docstring a bit misleading. :-)
> ssl_check_hostname: If set, match the hostname during the SSL handshake. Defaults to False.

PR #3637, opened 2025-05-06
Fix SSL verification with ssl_cert_reqs="none" and ssl_check_hostname=True

omerfeyzioglu:
### Pull Request check-list

### Description of change

# Fix SSL verification with `ssl_cert_reqs="none"` and `ssl_check_hostname=True`

## Problem

When connecting to a Redis cluster with SSL enabled, the following error occurs when trying to set `ssl_check_hostname=True` along with `ssl_cert_reqs="none"`:
Fail to establish redis connection: Redis Cluster cannot be connected. Please provide at least one reachable node: Cannot set verify_mode to CERT_NONE when check_hostname is enabled.

This is because of a limitation in Python's SSL library that prevents `check_hostname=True` from being used with `verify_mode=ssl.CERT_NONE`.

## Solution

This PR fixes the issue by:

1. Automatically setting `check_hostname=False` when `verify_mode=ssl.CERT_NONE` in both:
   - `redis/asyncio/connection.py` for async clients
   - `redis/connection.py` for sync clients

2. Adding tests to verify the fix works correctly in both sync and async modes.

## Notes

This issue is particularly relevant after the 6.0.0 release which changed the default value of `ssl_check_hostname` to `True` for improved security. This fix ensures a smoother transition for users who need to use `ssl_cert_reqs="none"` while maintaining compatibility with this new default setting.

The fix ensures that when users set `ssl_cert_reqs="none"` (or use `ssl.CERT_NONE`), hostname verification will be automatically disabled to prevent this error, while still respecting explicit `ssl_check_hostname=False` settings when certificate verification is enabled.


petyaslavova (changes requested):
The logic to `set check_hostname=False` when `cert_reqs` is None makes sense, but it doesn't fully address the underlying issue—it mostly masks it.

To resolve this properly, we should ensure that `ssl_check_hostname` passed through `kwargs` is propagated to the connection initialization. In cluster.py, there's a tuple called `REDIS_ALLOWED_KEYS` - adding `ssl_check_hostname` there would prevent it from being stripped from `kwargs`.

Also, since this change affects behavior demonstrated in the examples, could you please take a look at the documentation updated in PR #3626 and remove the explicit ssl_check_hostname=False setting if it’s no longer necessary?
It will also be good if you could change the tests in which cert_reqs is set to none not to provide explicitly ssl_check_hostname=False.

petyaslavova (approved):

redis/asyncio/connection.py
```diff
@@ -868,7 +868,7 @@ def __init__(
         cert_reqs: Optional[Union[str, ssl.VerifyMode]] = None,
         ca_certs: Optional[str] = None,
         ca_data: Optional[str] = None,
-        check_hostname: bool = False,
+        check_hostname: bool = True,
```
sigio, line 871: This change in defaults bit us this week ;)

dbarbosapn: +1 🤣

petyaslavova: I somehow overlooked that line in the changes 😞. Apologies for the miss! We can fix it in the next minor release, which will be published soon.

sigio: Thanks for the quick reply. In the meantime we explicitly set it to false instead of relying on defaults in our app-code.

petyaslavova: Will be fixed with PR #3655

PR #3647, opened 2025-05-14
Add support for Trio via AnyIO

agronholm:
### Pull Request check-list

### Description of change

This adds an alternate, [AnyIO](https://pypi.org/project/anyio/) based asynchronous client. This is necessary in order to make it possible for users of the [Trio](https://pypi.org/project/trio/) async framework to use Redis asynchronously. But the potential does not end there. Given that AnyIO supports asyncio as well, it would be possible to later replace the asyncio client with the AnyIO one.

A minor improvement would also be to remove `pytest-asyncio` from test dependencies and just use the AnyIO test plugin for testing both the AnyIO and asyncio clients.

A few things to note:
* Given that `redis-entraid` is a separate library with asyncio-only async code, I had to limit tests involving that to only run on asyncio. I could also migrate that library as follow-up work.
* This new client must be used with context management due to structured concurrency (task groups being used behind the scenes)
* If the general idea gets a green light, I will make the necessary documentation changes
* Further deduplication of code is possible and I am open to it as follow-up work


agronholm:
What is the Python version support policy in this project? I see that v6.0.0 was released with Python 3.8 support even though it reached its end of life last year.

agronholm:
> Hi @agronholm, thank you for your contribution! We will have a look at it soon. We will discuss the 3.8 support internally and will update you here.

Any updates on that?

Graeme22:
Would love to see this get added!

agronholm:
@petyaslavova It seems like the project has dropped Python 3.8 support, so I've updated my PR accordingly.

Can we establish some communication channel to discuss ways to move forward with this? An important matter is async tests and whether you want to keep using `pytest-asyncio` or if you'd like all async tests to use `anyio` instead (allowing you to drop `pytest-asyncio` as a test dependency). I realize that this PR is a beast to review, but I could sort out the test plugin issue in a separate PR that would be merged first.

arthur-tacca:
not stale

agronholm:
Frankly I don't expect this to ever get a review, consider the sheer scale of changes here. I just wish the maintainers were up-front about it.

petyaslavova:
Hey @agronholm!

You are right that we owe you a clear answer, and I am sorry it took this long. Here it is.

As of now we can't afford to add a third client stack to redis-py. Today we maintain two: the sync clients and the asyncio ones. This PR adds a third one that is close to a full copy of `redis/asyncio` (about 30k added lines, including a duplicated test suite). After that, every new command, every bug fix, and every new feature would have to be written and tested three times, and we do not have the capacity for that.

The problem is already visible in this PR. Since the branch was last updated in August 2025, `redis/asyncio` received 71 commits and several new modules: multi-database (active-active) support, maintenance notifications, keyspace notifications, and observability. The `legacy_responses` option for unified responses was added too. None of that exists in `redis/anyio`. So a third stack would either slow the whole library down or ship without the newer features. That already happens from time to time with our asyncio clients, and we do not want to make it worse.

If you want to take this further, a separate package that depends on redis-py is the path we can support. Your PR shows it is realistic: `redis/anyio` already reuses our shared command mixins (`redis/commands`) and response parsing (`redis/_parsers`), and only the connection layer is really AnyIO specific. We are happy to keep those shared parts importable, and to discuss small changes on our side that would make such a package easier to build and keep in sync.

I am closing this PR now. Thank you again for the work you put into it, and for your patience!

agronholm:
For any casual visitors, [coredis](https://github.com/alisaifee/coredis) became a thing while waiting, and I also wrote my own [redis-anyio](https://github.com/agronholm/redis-anyio) which I ended up not publishing.

PR #3653, opened 2025-05-22
fix: avoid instantiating a connection on _repr__

robertosantamaria-scopely:
### Pull Request check-list

### Description of change

Changing the `__repr__` methods in connection pool classes to avoid instantiating a connection object just to have its string representation.


redis/connection.py
```diff
@@ -1424,10 +1424,16 @@ def __init__(
         self._lock = threading.Lock()
         self.reset()
 
-    def __repr__(self) -> (str, str):
+    def __repr__(self) -> str:
```
petyaslavova, line 1427: I suggest something like the following:
`connection_kwargs_repr = ",".join(
            [f"{k}={v}" for k, v in self.connection_kwargs.items()]
        )
        return (
            f"<{self.__class__.__module__}.{self.__class__.__name__}"
            f"({self.connection_class.__module__}.{self.connection_class.__name__}"
            f"({connection_kwargs_repr})))>"
        )`

With this representation, it will be obvious what the Connection class is and from which module(it can be custom).
We will also be able to see all the provided kwargs when a new connection is instantiated.

Although this covers most of the cases, with the above code, the default args set to the connection are still missing and it would require fixing some unit tests.

robertosantamaria-scopely: Will do, thanks.

robertosantamaria-scopely: I updated this but the checks that fail do not show any specific error, right? Do I need to do something else for them to run @petyaslavova ?

robertosantamaria-scopely: Oh that's right, I did not see the collapsed line numbers 🤦 . Feel free to go ahead, I did not expect to affect that many unit tests. This should not really affect much, it just got i the way of debugging when trying to log objects, but if it does not alter the overall behavior of the driver, I'm OK with discarding this MR - up to you guys!

vladvildanov (approved):

redis/asyncio/connection.py
```diff
@@ -1112,9 +1112,11 @@ def __init__(
             self._event_dispatcher = EventDispatcher()
 
     def __repr__(self):
+        conn_kwargs = ",".join([f"{k}={v}" for k, v in self.connection_kwargs.items()])
```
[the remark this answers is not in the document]
petyaslavova: When the keys are sorted the order is too different from the one in the previous implementation and not very intuitive, so I'll skip this suggestion.

PR #3654, opened 2025-05-23
fix: Fixed potential deadlock from unexpected __del__ call

### LINKED ISSUE #3289: Possible deadlock in class `Pipeline`
**Version**: 4.3.4, but the code that causes the issue still exists in master

**Platform**: Python 3.11

**Description**: In rare cases, there's a deadlock in class `Pipeline`. The class resets it's connection in the `__del__` method (in case the connection wasn't closed beforehand). `__del__` is called when the object is deleted, and can be inserted by the GC anywhere in the code. In my cause, `__del__` was inserted in method `get_connection` of `ConnectionPool` after it acquired `self._lock`. The `__del__`'s reset method calls `self.connection_pool.release`, which tries to acquire the same lock (`self._lock`), and everything is stuck in a deadlock.

One way to solve this issue would be to use an rlock.

vladvildanov:
### Pull Request check-list

### Description of change

Fixes potential deadlock when GC takes action and `__del__` invoked in the middle of `get_connection()` or `release()` methods. `ClusterPipeline` object releases connection on `__del__ ` method which leads to a deadlock. 

This PR introduces only non-breaking changing fixes, the follow up PR with backward incompatible fixes will be released as a part of next major release

Closes #3289

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -369,6 +369,8 @@ class Redis(RedisModuleCommands, CoreCommands, SentinelCommands):
         ]:
             raise RedisError("Client caching is only supported with RESP version 3")
 
+        # TODO: To avoid breaking changes during the bug fix, we have to keep non-reentrant lock.
+        # TODO: Remove this before next major version (7.0.0)
         self.single_connection_lock = threading.Lock()
         self.connection = None
         self._single_connection_client = single_connection_client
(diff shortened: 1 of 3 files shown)


petyaslavova (approved):
LGTM.

PR #3658, opened 2025-05-25
fix(doc) : update Python print output

JCornat:
### Description

The Pyhton output should be number instead of a string


dmaier-redislabs:
@elena-kolevska Can you take a look please?

elena-kolevska (approved):

elena-kolevska:
Nice catch @JCornat ! And thanks for the fix.
@petyaslavova pls include this in the next release.

PR #3628, opened 2025-05-02
add equality and hashability to ``Retry`` and backoff classes

terencehonles:
### Description of change

This change adds equality and hashability to ``Retry`` and backoff classes.

This broke in DjangoRQ (fixed) after https://github.com/redis/redis-py/pull/3622 / https://github.com/redis/redis-py/pull/3614 and I provided a workaround in https://github.com/rq/django-rq/pull/706, but this allows other downstream libraries to compare two retry instances.

### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Do tests and lints pass with this change?
- [x] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [x] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?
- [ ] Was the change added to CHANGES file?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

PR #3668, opened 2025-06-03
add async Retry ``__eq__`` and ``__hash__`` & fix ExponentialWithJitterBackoff ``__eq__``

terencehonles:
### Description of change

This change fixes a typo in `ExponentialWithJitterBackoff`'s `__eq__` method from https://github.com/redis/redis-py/pull/3628, and allows async `Retry` objects to also support `__eq__` and `__hash__`.

### Pull Request check-list


redis/retry.py
```diff
@@ -10,18 +10,17 @@
     from redis.backoff import AbstractBackoff
 
 
-class Retry:
+class AbstractRetry:
     """Retry a specific number of times after a failure"""
 
+    __slots__ = "_backoff", "_retries", "_supported_errors"
```
petyaslavova: I think that we don't need to use `__slots__` for the `retry` object. We are not keeping that many instances of it, and the saved memory is not going to be significant. I see that we've had both patterns in the `sync` and `async` versions, and I suggest sticking to the one that has been used by the sync `Retry` class.

redis/retry.py
```diff
@@ -10,18 +10,17 @@
...
     def __init__(
         self,
         backoff: "AbstractBackoff",
         retries: int,
-        supported_errors: Tuple[Type[Exception], ...] = (
-            ConnectionError,
-            TimeoutError,
-            socket.timeout,
-        ),
+        supported_errors: Union[Tuple[Type[Exception], ...], None] = None,
```
petyaslavova: I think it would be better here if you extract the list as a constant and assign this list as the default value - it will be more visible what will be assigned to the list in case the user doesn't provide a value.

redis/retry.py
```diff
@@ -32,10 +31,11 @@ def __init__(
         """
         self._backoff = backoff
         self._retries = retries
-        self._supported_errors = supported_errors
+        if supported_errors:
+            self._supported_errors = supported_errors
 
     def __eq__(self, other: Any) -> bool:
-        if not isinstance(other, Retry):
+        if not isinstance(other, AbstractRetry):
```
petyaslavova: I think we should have `__eq__` implemented in each class(sync and async) and check if it is an instance of the concrete classes.

redis/retry.py
terencehonles: done

redis/retry.py
terencehonles: done

redis/retry.py
terencehonles: I was about to update the PR to include this, but I cannot do that without re-implementing `__init__` in each of the concrete base classes. Is this part of your request? I would have to make each concrete base class call `super().__init__` passing the list of supported errors that it expects, and that would require duplicating the whole signature for what I see as minimal value (you may disagree).

Otherwise, the abstract class would not be able to select the appropriate supported errors to use without using the pattern I provided.

petyaslavova: Hi @terencehonles,
The change sounds good — having __init__ methods in each subclass makes the supported errors for the retry objects much clearer.
Yes, there will be a bit of duplication, but considering we only have two subclasses with three arguments each, it seems reasonable.

terencehonles: Just coming back from vacation, but I added the `__init__` methods to the sub-classes.

redis/retry.py
```diff
@@ -34,22 +34,14 @@ def __init__(
...
     def __eq__(self, other: Any) -> bool:
-        if not isinstance(other, Retry):
-            return NotImplemented
-
-        return (
-            self._backoff == other._backoff
-            and self._retries == other._retries
-            and set(self._supported_errors) == set(other._supported_errors)
-        )
+        return NotImplemented
```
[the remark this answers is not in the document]
terencehonles: `NotImplemented` is a valid value for `__eq__` so it is effectively a concrete implementation.

redis/retry.py
```diff
@@ -69,6 +61,34 @@ def update_retries(self, value: int) -> None:
...
+        supported_errors: Tuple[Type[Exception], ...] = (
+            ConnectionError,
+            TimeoutError,
+            socket.timeout,
+        ),
+    ):
+        super().__init__(backoff, retries, supported_errors)
+
+    __init__.__doc__ = AbstractRetry.__init__.__doc__
+
```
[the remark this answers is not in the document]
terencehonles: I removed both of these, but I'm not sure this is actually correct. This is supposed to be the case for `class` docstrings, but this is a class method and testing locally (I wasn't sure of this before the review and tested myself) it did not inherit the docstring (at least when using the `help` method).

petyaslavova (approved):

Issue #3640, opened 2025-05-09
Deadlock in `PubSub` class after update to 5.3.0


ManelCoutinhoSensei:
Hi again!

I just updated redis-py from 5.2.1 to 5.3.0 and I noticed that while adding the `Dispatchers` there was deadlock introduced.

## Summary
When using a client with `Retries` enabled and subscribing to a pattern using `psubscribe`, a deadlock can occur under certain conditions:
- If a `punsubscribe` is attempted and a disconnection (such as a broken pipe) happens during this process, a reconnection is triggered (via the `PubSub._execute` error handling).
- During reconnection, the client tries to re-subscribe to the previously subscribed patterns, but this process can result in a deadlock.

## Full explanation
There was a new lock introduced before each `PubSub` command execution on commit [`40e5fc1`](https://github.com/redis/redis-py/commit/40e5fc131b457edc368637c1034424567553ba55), which is part of the 5.3.0 release:
https://github.com/redis/redis-py/blob/7130e1ad913f51abc0de1886a52cfd6338e1beee/redis/client.py#L874-L875

When a command execution fails, if `Retries` are enabled, there will be a reconnection attempt:
https://github.com/redis/redis-py/blob/7130e1ad913f51abc0de1886a52cfd6338e1beee/redis/client.py#L895-L916

However, when executing the `connect` we get to the final lines where the resubscriptions callbacks are triggered for `pubsub`:
https://github.com/redis/redis-py/blob/7130e1ad913f51abc0de1886a52cfd6338e1beee/redis/connection.py#L406-L413

This will hit the same lock again a cause a deadlock.

---

Let me know if you need any other information. 🙂

evios:
+1

petyaslavova:
Hi @ManelCoutinhoSensei, thanks for reporting this! We will have a look at it shortly.

matejsp:
+1

matejsp:
@Darek07 did you try just released 6.2.0 ? It says in release notes that it fixes one deadlock.

ManelCoutinhoSensei:
@matejsp I don't know which is the lock that causes the issue on `django-celery` but the one mentioned in the released 6.2.0 is the cluster one and not the PubSub that i mention here

Darek07:
> [@Darek07](https://github.com/Darek07) did you try just released 6.2.0 ? It says in release notes that it fixes one deadlock.

Unfortunately it doesn't help. With redis 6.2.0 the problem still appears.

[...]

PR #3677, opened 2025-06-13
Removing the threading.Lock locks and replacing them with RLock objects to avoid deadlocks.

petyaslavova:
### Pull Request check-list

### Description of change

Usage of `threading.Lock` has introduced deadlocks in the client. 
With this change, I'm removing all places that can cause such issues.
**The change is breaking, because it changes the interfaces of several objects and cannot  be released before the next major release, which will be 7.0.0**

### DIFF
--- a/redis/client.py
+++ b/redis/client.py
@@ -368,9 +368,7 @@ class Redis(RedisModuleCommands, CoreCommands, SentinelCommands):
         ]:
             raise RedisError("Client caching is only supported with RESP version 3")
 
-        # TODO: To avoid breaking changes during the bug fix, we have to keep non-reentrant lock.
-        # TODO: Remove this before next major version (7.0.0)
-        self.single_connection_lock = threading.Lock()
+        self.single_connection_lock = threading.RLock()
         self.connection = None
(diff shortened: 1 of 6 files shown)


vladvildanov (approved):

petyaslavova:
This PR fixes issue #3640

yanivtoledano:
@petyaslavova - appreciate this is a breaking change but given the deadlocks introduced has broken production code for us (and, based on feedback on various OS repos, many others), do you have an estimate of when a v7.0.0 will be released?

To avoid the deadlocks we had to downgrade from 6.x to 5.2.1 and I guess everyone on >5.2.1,<7.0.0 may potentially also have very hard to debug issues going forward (it took us forever to triangulate the issue to deadlocks in redis)?

junhaoliao:
Thanks @petyaslavova @vladvildanov for the fix! I noticed it’s already included in 6.4.0, but it wasn’t mentioned in [the changelog](https://github.com/redis/redis-py/releases/tag/v6.4.0). Could you add it there so other users are aware as well?

PR #3683, opened 2025-06-19
`LoadBalancer` keyed on slot instead of primary node, not reset on `NodesManager.initialize()`

### LINKED ISSUE #3681: Round robin load balancing isn't working as expected if primary node goes down for Redis cluster mode
Using `redis-py==6.2.0`. I have a 6 node Redis Cluster running in EKS (3 primaries / 3 replicas), and whenever one of the primary nodes goes down, my service will raise `TimeoutError` despite the replica being available, having sufficient `Retry` on the client, and using client-side load balancing as `LoadBalancingStrategy.ROUND_ROBIN`. 

I've boiled it down to the following (so I can better observe what's happening in the `RedisCluster._internal_execute_command` [retry loop](https://github.com/redis/redis-py/blob/1a59471870658c7fc661a4705541a0762acd8d85/redis/cluster.py#L1156-L1188)):

```
import os
from redis.cluster import LoadBalancingStrategy, RedisCluster
rc = RedisCluster(
  host=os.getenv('REDIS_HOST'),
  port=os.getenv('REDIS_PORT'),
  password=os.getenv('REDIS_PASSWORD'),
  load_balancing_strategy=LoadBalancingStrategy.ROUND_ROBIN,
  socket_connect_timeout=1.0,
  require_full_coverage=False,
  decode_responses=True,
)
rc.set("foo", "bar")  # True
slot = rc.determine_slot("GET", "foo")  # 12182
lbs = rc.load_balancing_strategy  # ROUND_ROBIN
for i in range(5):
    print(f"\nAttempt {i + 1}")
    try:
        primary_name = rc.nodes_manager.slots_cache[slot][0].name
        n_slots = len(rc.nodes_manager.slots_cache[slot])
        node_idx = rc.nodes_manager.read_load_balancer.get_server_index(primary_name, n_slots, lbs)
        node = rc.nodes_manager.slots_cache[slot][node_idx]
        print(f"idx: {node_idx} | node: {node.name} | type: {node.server_type}")
        rc._execute_command(node, "GET", "foo")
    except Exception as e:
        print(f"Exception: {e}")
```

With a healthy cluster, this will output:

```
Attempt 1
idx: 0 | node: 100.66.97.179:6379 | type: primary
'bar'

Attempt 2
idx: 1 | node: 100.66.106.241:6379 | type: replica
'bar'

Attempt 3
idx: 0 | node: 100.66.97.179:6379 | type: primary
'bar'

Attempt 4
idx: 1 | node: 100.66.106.241:6379 | type: replica
'bar'

Attempt 5
idx: 0 | node: 100.66.97.179:6379 | type: primary
'bar'
```

If I kill the primary node in EKS (`kubectl delete pod redis-node-3` where this was the `100.66.97.179` pod), and running the loop again, I get the following (until EKS gets `redis-node-3` back up and running):

```
Attempt 1
idx: 1 | node: 100.66.106.241:6379 | type: replica
'bar'

Attempt 2
idx: 0 | node: 100.66.97.179:6379 | type: primary
Exception: Timeout connecting to server

Attempt 3
idx: 0 | node: 100.66.97.179:6379 | type: primary
Exception: Timeout connecting to server

Attempt 4
idx: 0 | node: 100.66.97.179:6379 | type: primary
Exception: Timeout connecting to server

Attempt 5
idx: 0 | node: 100.66.97.179:6379 | type: primary
Exception: Timeout connecting to server
```

Basically, as soon as I get the `TimeoutError` from the primary node, the load balancer gets stuck and stops bouncing between the primary and the replica, but keeps trying the primary over and over.

If I instead kill the replica node in EKS, I get exactly what I'd expect, where the load balancer still round robins between both nodes:

```
Attempt 1
idx: 0 | node: 100.66.111.89:6379 | type: primary
'bar'

Attempt 2
idx: 1 | node: 100.66.106.241:6379 | type: replica
Exception: Timeout connecting to server

Attempt 3
idx: 0 | node: 100.66.111.89:6379 | type: primary
'bar'

Attempt 4
idx: 1 | node: 100.66.106.241:6379 | type: replica
Exception: Timeout connecting to server

Attempt 5
idx: 0 | node: 100.66.111.89:6379 | type: primary
'bar'
```

drewfustin:
### Pull Request check-list

- [x] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)? [link](https://github.com/drewfustin/redis-py/actions/runs/15764094627)

### Description of change

- Fixes #3681
- `LoadBalancer` now uses `slot_to_idx` instead of `primary_to_idx`, using slot as the key instead of primary node name.
- When `NodesManager` resets in `initialize`, it no longer runs `LoadBalancer.reset()` which would clear the `slot_to_idx` dictionary.
- Tests `TestNodesManager.test_load_balancer` updated accordingly.

As noted in #3681, reseting the load balancer on `NodesManager.initialize()` causes the index associated with the primary node to reset to 0. If a `ConnectionError` or `TimeoutError` is raised by an attempt to connect to a primary node, `NodesManager.initialize()` is called, and the the load balancer's index for that node will reset to 0. Therefore, the next attempt in the retry loop will not move on from the primary node to a replica node (with index > 0) as expected, but will instead retry the primary node again (and presumably raise the same error).

Since `NodesManager.initialize()` being called on `ConnectionError` or `TimeoutError` is the valid strategy, and since the primary node's host will often be replaced in tandem with events that cause these errors (e.g. when a primary node is deleted and then recreated in Kubernetes), keying the `LoadBalancer` dictionary on the primary node's name (`host:port`) doesn't feel appropriate. Instead, keying the dictionary on the Redis Cluster's slot seems to be a better strategy. As such, the `server_index` corresponding to key `slot` doesn't need to be reset to 0 on `NodesManager.initialize()` as the `slot` isn't expected to change and need to be reset, only the `host:port` would require such. Instead, the `slot` can maintain its "state" even when the `NodesManager` is reinitialized, thus resolving #3681.

With the fix in this PR implemented, the output of the loop from #3681 becomes what is expected when the primary node goes down (the load balancer continues to the next node on a `TimeoutError`):

```
Attempt 1
idx: 2 | node: 100.66.151.143:6379 | type: replica
'bar'

Attempt 2
idx: 0 | node: 100.66.122.229:6379 | type: primary
Exception: Timeout connecting to server

Attempt 3
idx: 1 | node: 100.66.151.143:6379 | type: replica
'bar'

Attempt 4
idx: 2 | node: 100.66.106.241:6379 | type: replica
'bar'

Attempt 5
idx: 0 | node: 100.66.122.229:6379 | type: primary
Exception: Error 113 connecting to 100.66.122.229:6379. No route to host.
```

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -1835,7 +1835,6 @@ class NodesManager:
 
     def reset(self):
         pass
-    
 
     def remap_host_port(self, host: str, port: int) -> Tuple[str, int]:
         """
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -1835,6 +1835,7 @@ class NodesManager:
 
     def reset(self):
         pass
+    
 
     def remap_host_port(self, host: str, port: int) -> Tuple[str, int]:
         """
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -1403,12 +1403,12 @@ class LoadBalancer:
     """
 
     def __init__(self, start_index: int = 0) -> None:
-        self.primary_to_idx = {}
+        self.slot_to_idx = {}
         self.start_index = start_index
 
     def get_server_index(
         self,
-        primary: str,
+        slot: int,
         list_size: int,
         load_balancing_strategy: LoadBalancingStrategy = LoadBalancingStrategy.ROUND_ROBIN,
     ) -> int:
@@ -1416,26 +1416,26 @@ class LoadBalancer:
             return self._get_random_replica_index(list_size)
         else:
             return self._get_round_robin_index(
-                primary,
+                slot,
                 list_size,
                 load_balancing_strategy == LoadBalancingStrategy.ROUND_ROBIN_REPLICAS,
             )
 
     def reset(self) -> None:
-        self.primary_to_idx.clear()
+        self.slot_to_idx.clear()
 
     def _get_random_replica_index(self, list_size: int) -> int:
         return random.randint(1, list_size - 1)
 
     def _get_round_robin_index(
-        self, primary: str, list_size: int, replicas_only: bool
+        self, slot: int, list_size: int, replicas_only: bool
     ) -> int:
-        server_index = self.primary_to_idx.setdefault(primary, self.start_index)
+        server_index = self.slot_to_idx.setdefault(slot, self.start_index)
         if replicas_only and server_index == 0:
             # skip the primary node index
             server_index = 1
         # Update the index for the next round
-        self.primary_to_idx[primary] = (server_index + 1) % list_size
+        self.slot_to_idx[slot] = (server_index + 1) % list_size
         return server_index
 
 
@@ -1575,9 +1575,8 @@ class NodesManager:
 
         if len(self.slots_cache[slot]) > 1 and load_balancing_strategy:
             # get the server index using the strategy defined in load_balancing_strategy
-            primary_name = self.slots_cache[slot][0].name
             node_idx = self.read_load_balancer.get_server_index(
-                primary_name, len(self.slots_cache[slot]), load_balancing_strategy
+                slot, len(self.slots_cache[slot]), load_balancing_strategy
             )
         elif (
             server_type is None
@@ -1835,11 +1834,7 @@ class NodesManager:
                 node.redis_connection.close()
 
     def reset(self):
-        try:
-            self.read_load_balancer.reset()
-        except TypeError:
-            # The read_load_balancer is None, do nothing
-            pass
+        pass
 
     def remap_host_port(self, host: str, port: int) -> Tuple[str, int]:
         """
--- a/tests/test_asyncio/test_cluster.py
+++ b/tests/test_asyncio/test_cluster.py
@@ -2444,33 +2444,31 @@ class TestNodesManager:
             slot_1: [node_1, node_2, node_3],
             slot_2: [node_4, node_5],
         }
-        primary1_name = n_manager.slots_cache[slot_1][0].name
-        primary2_name = n_manager.slots_cache[slot_2][0].name
         list1_size = len(n_manager.slots_cache[slot_1])
         list2_size = len(n_manager.slots_cache[slot_2])
 
         # default load balancer strategy: LoadBalancerStrategy.ROUND_ROBIN
         # slot 1
-        assert lb.get_server_index(primary1_name, list1_size) == 0
-        assert lb.get_server_index(primary1_name, list1_size) == 1
-        assert lb.get_server_index(primary1_name, list1_size) == 2
-        assert lb.get_server_index(primary1_name, list1_size) == 0
+        assert lb.get_server_index(slot_1, list1_size) == 0
+        assert lb.get_server_index(slot_1, list1_size) == 1
+        assert lb.get_server_index(slot_1, list1_size) == 2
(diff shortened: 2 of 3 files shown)


petyaslavova:
Hi @drewfustin,
First of all, thank you for the time and effort you've put into this PR — it's much appreciated!

That said, I have some concerns about the proposed approach. With your change, the replica indexes would be rotated per slot, which could lead to repeatedly hitting the same cluster node over an extended period, instead of distributing requests across replicas or across both replicas and primaries.

In a Redis Cluster with 16,384 slots, an application that doesn’t deliberately target specific slots might end up consistently hitting a single node. This behavior doesn't align with the intended behavior of the round-robin algorithm, which is designed to spread requests across both primaries and replicas or just through the replicas to ensure better load distribution.

redis/cluster.py
```diff
@@ -1403,39 +1403,39 @@ class LoadBalancer:
...
 
     def _get_random_replica_index(self, list_size: int) -> int:
         return random.randint(1, list_size - 1)
 
     def _get_round_robin_index(
-        self, primary: str, list_size: int, replicas_only: bool
+        self, slot: int, list_size: int, replicas_only: bool
     ) -> int:
-        server_index = self.primary_to_idx.setdefault(primary, self.start_index)
+        server_index = self.slot_to_idx.setdefault(slot, self.start_index)
```
Sanjays2402, line 1433: This can return an out-of-range index after a topology refresh shrinks the replica list: `slot_to_idx` may retain 2 from a three-node list, then `list_size=2` updates the stored value modulo 2 but still returns 2, and `get_node_from_slot` indexes the two-element list with it. Normalize `server_index` by `list_size` before returning it.

eeshsaxena (commented):
The direction here makes sense, keying the round robin state on the slot and letting it survive `initialize()` is a real improvement over keying on the primary name and wiping it on every reinit. But the combination of "no longer reset" plus "index is not re-bounded on read" looks like it can hand back an out of range index after a topology change.

`_get_round_robin_index` does:

```python
server_index = self.slot_to_idx.setdefault(slot, self.start_index)
...
self.slot_to_idx[slot] = (server_index + 1) % list_size
return server_index
```

The `% list_size` only bounds the value stored for next time, not the value returned now. The returned `server_index` is whatever was stored on the previous call, and that was bounded by the previous call's `list_size`. Since `NodesManager.reset()` is now a no-op, that stored value now carries across `initialize()`.

So if the number of nodes for a slot shrinks (a replica dropped during resharding or failover), the persisted index can be larger than the new list:

- slot S has 3 nodes, round robin advances until `slot_to_idx[S]` is 2
- a replica for S goes away, topology reinitializes, `slots_cache[S]` now has 2 nodes
- next read to S: `get_server_index(S, 2)` returns the stored 2, and `get_node_from_slot` does `self.slots_cache[slot][2]` on a 2 element list, which raises `IndexError`

The old code happened to avoid this because `initialize()` reset the balancer, so a stale index never outlived a topology change. Now that it persists, re-bounding on read would keep the persistence benefit without the risk, e.g.:

```python
server_index = self.slot_to_idx.setdefault(slot, self.start_index) % list_size
```

(or clamp `node_idx` in `get_node_from_slot`). Worth a test that advances the index, then re-runs with a smaller `list_size` for the same slot and asserts the returned index stays in range, since the current `test_load_balancer` only exercises a stable node count.

PR #3690, opened 2025-07-02
Add support for new BITOP operations: DIFF, DIFF1, ANDOR, ONE

ofekshenawa:
### Summary
This PR adds support for four new BITOP operations that will be available in Redis 8.2+: DIFF, DIFF1, ANDOR, and ONE. These operations extend Redis bitmap functionality to support common set operations that previously required multiple commands or Lua scripts.

### New Operations
BITOP DIFF: Returns members of the first bitmap that are not members of any other bitmaps (X ∧ ¬(Y1 ∨ Y2 ∨ ...))
BITOP DIFF1: Returns members of any bitmap except the first that are not members of the first bitmap (¬X ∧ (Y1 ∨ Y2 ∨ ...))
BITOP ANDOR: Returns members of the first bitmap that are also members of one or more other bitmaps (X ∧ (Y1 ∨ Y2 ∨ ...))
BITOP ONE: Returns members of exactly one of the given bitmaps (generalized XOR for multiple keys)
### Changes Made
Tests: Added comprehensive test coverage for all four new operations
Synchronous tests in tests/test_commands.py
Asynchronous tests in tests/test_asyncio/test_commands.py
Tests include bitwise logic verification, edge cases, and return value validation
All tests gated with @skip_if_server_version_lt("8.2.0") for compatibility

### DIFF
--- a/tests/test_asyncio/test_commands.py
+++ b/tests/test_asyncio/test_commands.py
@@ -882,103 +882,78 @@ class TestRedisCommands:
     @pytest.mark.onlynoncluster
     @skip_if_server_version_lt("8.2.0")
     async def test_bitop_diff(self, r: redis.Redis):
-        """Test BITOP DIFF operation"""
-        # Set up test data: a=0b11110000, b=0b11000000, c=0b10000000
-        await r.set("a", b"\xf0")  # 11110000
-        await r.set("b", b"\xc0")  # 11000000
-        await r.set("c", b"\x80")  # 10000000
+        await r.set("a", b"\xf0")
+        await r.set("b", b"\xc0")
+        await r.set("c", b"\x80")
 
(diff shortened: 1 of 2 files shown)


tests/test_asyncio/test_commands.py
```diff
@@ -879,6 +879,103 @@ async def test_bitop_string_operands(self, r: redis.Redis):
         assert int(binascii.hexlify(await r.get("res2")), 16) == 0x0102FFFF
         assert int(binascii.hexlify(await r.get("res3")), 16) == 0x000000FF
 
+    @pytest.mark.onlynoncluster
+    @skip_if_server_version_lt("8.2.0")
```
petyaslavova: The server still returns 8.1.xxx as server version, and those new tests will be skipped. I suggest to define this rule as @skip_if_server_version_lt("8.1.224"). It can be updated later, when server version is update.

PR #3696, opened 2025-07-04
MultiDbClient implementation

vladvildanov:
### Pull Request check-list

### Description of change

Added initial implementation of MultiDBClient that allows to operate on Active-Active databases


redis/data_structure.py
```diff
@@ -0,0 +1,73 @@
...
+from typing import List, Any, TypeVar, Generic, Union
+
+T = TypeVar('T')
+
+class WeightedList(Generic[T]):
+    """
+    Thread-safe weighted list.
+    """
+    def __init__(self):
+        self._items: List[tuple[Any, Union[int, float]]] = []
```
petyaslavova: You can use the custom `Number` type defined in typing instead of `Union[int, float]`

redis/data_structure.py
```diff
@@ -0,0 +1,73 @@
...
+            # Find insertion point using binary search
+            left, right = 0, len(self._items)
+            while left < right:
+                mid = (left + right) // 2
+                if self._items[mid][0] < weight:
+                    right = mid
+                else:
+                    left = mid + 1
+
+            self._items.insert(left, (weight, item))
```
petyaslavova: According to the type definition of `self._items`, the order of the elements should be `(weight, item)`:

`self._items: List[tuple[Any, Union[int, float]]]`

redis/data_structure.py
petyaslavova, line 1: I notice a slight inconsistency in the file regarding the order of `item` and `weight` throughout the methods in the file. For better clarity and consistency, it's best to use a single format throughout—either `(weight, item)` or `(item, weight)`. Personally, I find `(item, weight)` more intuitive, as the item functions more like a key, with the weight acting as an associated value.

redis/event.py
```diff
@@ -77,17 +85,33 @@ def __init__(self):
...
+            listeners = self._event_listeners_mapping.get(type(event), [])
 
-        for listener in listeners:
-            await listener.listen(event)
+            for listener in listeners:
+                await listener.listen(event)
+
+    def register_listeners(self, event_listeners: Dict[Type[object], List[EventListenerInterface]]):
+        with self._lock:
+            for event in event_listeners:
```
petyaslavova: I think the logic will be clearer if the `event` field is renamed to `event_type`

redis/data_structure.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def get_top_n(self, n: int) -> List[tuple[Any, Union[int, float]]]:
+        """Get top N the highest weighted items"""
+        with self._lock:
+            return [(item, weight) for weight, item in self._items[:n]]
+
+    def update_weight(self, item, new_weight: float):
+        with self._lock:
+            """Update weight of an item"""
+            old_weight = self.remove(item)
```
petyaslavova, line 56: We have a weight property defined in the Database object as well. Shouldn't we update this one also, or is it handled somewhere else?

redis/multidb/failover.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+
+    def set_databases(self, databases: Databases) -> None:
+        self._databases = databases
+
+    def _get_active_database(self) -> AbstractDatabase:
+        for database, _ in self._databases:
+            if database.circuit.state == CBState.CLOSED:
+                return database
+            else:
```
petyaslavova: You can just skip the else here.

redis/multidb/client.py
```diff
@@ -0,0 +1,220 @@
...
+        # Initial databases check to define initial state
+        self._check_databases_health()
+
+        # Starts recurring health checks on the background.
+        self._bg_scheduler.run_recurring(
+            self._health_check_interval,
+            self._check_databases_health,
+        )
+
+        is_active_db = False
```
petyaslavova: This property can be renamed to something like `is_active_db_found` or `found_active_db` - it will be easier to follow the code and understand the logic.

redis/multidb/client.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def set_active_database(self, database: AbstractDatabase) -> None:
+        """
+        Promote one of the existing databases to become an active.
+        """
+        exists = None
+
+        for existing_db, _ in self._databases:
+            if existing_db == database:
+                exists = True
```
petyaslavova, line 101: You can break the cycle once the db is found.

redis/multidb/client.py
```diff
@@ -0,0 +1,220 @@
...
+            self._command_executor.active_database = highest_weighted_db
+
+    def update_database_weight(self, database: AbstractDatabase, weight: float):
+        """
+        Updates a database from the database list.
+        """
+        exists = None
+
+        for existing_db, _ in self._databases:
+            if existing_db == database:
```
petyaslavova, line 156: You can break the cycle after db is found.

redis/multidb/client.py
```diff
@@ -0,0 +1,220 @@
...
+                exists = True
+
+        if not exists:
+            raise ValueError('Given database is not a member of database list')
+
+        highest_weighted_db, highest_weight = self._databases.get_top_n(1)[0]
+        self._databases.update_weight(database, weight)
+
+        if weight > highest_weight and database.circuit.state == CBState.CLOSED:
+            database.state = DBState.ACTIVE
```
petyaslavova: This block of code repeats on several places and can be extracted in a helper method.

redis/multidb/command_executor.py
```diff
@@ -0,0 +1,164 @@
...
+            self._schedule_next_fallback()
+
+        try:
+            return self._active_database.client.execute_command(*args, **options)
+        except (ConnectionError, TimeoutError, socket.timeout) as e:
+            # Register command failure
+            self._event_dispatcher.dispatch(OnCommandFailEvent(args, e, self.active_database.client))
+
+            # Retry until failure detector will trigger opening of circuit
+            return self.execute_command(*args, **options)
```
petyaslavova: Don't we risk here to have the execution blocked on the same command for too long? The circle will open after 100 failures by default. If we have a command for which just the parsing of the response fails for some reason - it will be executed a hundred times.

redis/data_structure.py
vladvildanov: Good to know, fixed!

redis/data_structure.py
vladvildanov: Fixed

redis/data_structure.py
vladvildanov: You're absolutely right, I changed the whole format to `(item, wight)`

redis/event.py
vladvildanov: Fixed

redis/data_structure.py
vladvildanov: Yes, it's in another place. The idea for this data structure is to be generic (that's why it's under `redis/` package and not `redis/multidb`), so any manupulkation with `item` should be outside of this structure. In our case it's here:

https://github.com/redis/redis-py/pull/3696/files#diff-e1d7fdc9293a30cdfd5fa4a4670c76ea0e597275664899ac6edc297e63c22e3aR138

redis/multidb/client.py
vladvildanov: Done

redis/multidb/client.py
vladvildanov: Done

redis/multidb/client.py
vladvildanov: Done

redis/multidb/command_executor.py
vladvildanov: Yes, this part is unclear to me (how does retries should work with failure detection?) and it will be a topic on tomorrow discussion with David and other guys working on AA. I don't understand how they should be combined and what the logic is, cause if we raise an exception it might be that the process will be killed (if the app is not backed by GUnicorn, Uvicorn or Hypercorn) and we will lose a track of failures so it wouldn't work.

vladvildanov: @petyaslavova retry logic was changed according to the discussion

redis/multidb/config.py
```diff
@@ -0,0 +1,95 @@
...
+    event_dispatcher: EventDispatcherInterface = field(default_factory=default_event_dispatcher)
+
+    def databases(self) -> Databases:
+        databases = WeightedList()
+
+        for database_config in self.databases_config:
+            if database_config.client_kwargs.get("retry", None) is not None:
+                # The retry object is not used in the lower level clients, so we can safely remove it.
+                # We rely on command_retry in terms of global retries.
+                database_config.client_kwargs.update({"retry": Retry(retries=0, backoff=NoBackoff())})
```
petyaslavova: I think if the user has provided concrete retry configuration we should keep it and add zero retry config only when the retry is not in the kwargs.

redis/multidb/failure_detector.py
```diff
@@ -51,12 +58,16 @@ def register_failure(self, database, exception: Exception, cmd: tuple) -> None:
...
+
+    def set_command_executor(self, command_executor) -> None:
+        self._command_executor = command_executor
 
-    def _check_threshold(self, database):
+    def _check_threshold(self):
         with self._lock:
             if len(self._failures_within_duration) >= self._threshold:
-                database.circuit.state = CBState.OPEN
+                if self._command_executor and self._command_executor.active_database:
```
petyaslavova: Is it possible to have self._command_executor = None here? Or not having an active database?

redis/multidb/config.py
vladvildanov: In my opinion this nesting level of retries creates incorrect expectations, as inner level retries are not registered by FailureDetector that operate on the level of MultiDbClient. In terms of communication MultiDbClient not that different from usual client, so if it faces the same error there's no point to retry it on different layers.

redis/multidb/failure_detector.py
vladvildanov: Theoretically yes, practically no as we always set it in runtime. I'll remove this checks.

redis/multidb/config.py
petyaslavova: Ok. I still have one concern -  in the current code, the case when the user has not provided a retry config is not handled and the default config of the client will be used - which is with 3 retries.

vladvildanov: Not sure if I get you, we always rely on `command_retry` parameter which is required and has some default values. The default values is what you concerned?

petyaslavova: The `command_retry` is used for multi-DB client configuration and the command execution in the context of eventually changing different Redis DB instances. 
Here, each `database_config` contains the configuration for each separate Redis DB instance - each such configuration creates an object of type Redis, RedisCluster, or Sentinel, where we have a default of 3 retries in case `retry` is not provided in the `kwargs`. By the way, if a pool is provided directly via `from_pool`, the retry configuration of the pool is not changed, so in this case, we might end up with unexpected additional retries as well.

Issue #3700, opened 2025-07-07
deprecated_args decorator strips typing


mharding-hpe:
After moving to use `redis` v6 from v5, `mypy` began complaining that `redis.Redis()` was returning type `Any`. I investigated and pinned down that up through `v6.0.0b2`, it correctly knew the return type was `redis.client.Redis`, and starting in `v6.0.0`, it was `Any`.

I noticed that the `@deprecated_args` decorator was added to the `redis.client.Redis.__init__()` method with [this commit](https://github.com/redis/redis-py/commit/fe13ed10c5330245050ebbc14142bb1c9786f119) that fell in the gap between those two versions. I verified that removing that decorator solved the problem. I then looked at the definition of that decorator, and I can see why it is losing the type information. I have tested a fix for this, and will be opening a PR for consideration.

petyaslavova:
Hi @mharding-hpe, thank you for taking the time to report the issue and implement a code change to fix it!

mharding-hpe:
@petyaslavova Happy to help when I can!

PR #3701, opened 2025-07-07
Annotate deprecated_args decorator to preserve wrapped function type signature

mharding-hpe:
### Pull Request check-list

Everything passes except one standalone test: `tests/test_commands.py::TestRedisCommands::test_psync`

However, I ran the tests on the `master` branch of the repo and got the same result, so I assume that test failure is unrelated to my changes.

https://github.com/mharding-hpe/redis-py-dev/pull/2

N.A

N/A

### Description of change

Opened to resolve issue #3700 

Added type annotations to the `deprecated_args` decorator, to clarify that the type of callable will not be modified by anything inside the different functions (thus preserving its original signature). Without this, `mypy` interprets the type of `redis.Redis()` to be `Any`.

This issue was introduced between `v6.0.0b2` and `v6.0.0`. So ideally this fix would be backported to the `v6.0` and `v6.1` branches, but I don't know the policy for that in this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Start a redis via docker (for Redis versions < 8.0):
 
 ``` bash
 docker run -p 6379:6379 -it redis/redis-stack:latest
-
+```
 To install redis-py, simply:
 
 ``` bash


petyaslavova:
Hi @mharding-hpe, thank you for your contribution! It will be reviewed shortly.
Regarding backporting to previous versions: as per our policy, we only backport critical fixes, so this change will be included in newer versions only.

redis/utils.py
```diff
@@ -1,9 +1,10 @@
 import datetime
 import logging
 import textwrap
+from collections.abc import Callable
```
petyaslavova, line 4: @mharding-hpe can you please change this one to be imported from typing module for consistency with the other typing imports?

mharding-hpe: I can if you want, but importing it from `typing` has been [deprecated since Python 3.9](https://docs.python.org/3/library/typing.html#typing.Callable). No plans have yet been announced to remove the alias from the `typing` module, but it has been deprecated for a long time now, so who knows? 

Anyway, please let me know if you still want me to change it in light of this information, @petyaslavova , and I'll do so.

petyaslavova: @mharding-hpe No, lets's leave it as it is. I just didn't know it is deprecated :)

PR #3713, opened 2025-07-17
Hitless upgrade: Support initial implementation for synchronous Redis client - no handshake, no failing over notifications support.

petyaslavova:
### Pull Request check-list

### Description of change
Hitless upgrade support implementation for synchronous Redis client.

### DIFF
--- a/redis/_parsers/base.py
+++ b/redis/_parsers/base.py
@@ -1,7 +1,7 @@
 import sys
 from abc import ABC
 from asyncio import IncompleteReadError, StreamReader, TimeoutError
-from typing import Callable, List, Optional, Protocol, Union
+from typing import Awaitable, Callable, List, Optional, Protocol, Union
 
 from redis.maintenance_events import (
     NodeMigratedEvent,
@@ -243,8 +243,8 @@ class AsyncPushNotificationsParser(Protocol):
 
     pubsub_push_handler_func: Callable
     invalidation_push_handler_func: Optional[Callable] = None
-    node_moving_push_handler_func: Optional[Callable] = None
-    maintenance_push_handler_func: Optional[Callable] = None
+    node_moving_push_handler_func: Optional[Callable[..., Awaitable[None]]] = None
+    maintenance_push_handler_func: Optional[Callable[..., Awaitable[None]]] = None
 
     async def handle_pubsub_push_response(self, response):
         """Handle pubsub push responses asynchronously"""
--- a/redis/client.py
+++ b/redis/client.py
@@ -954,12 +954,13 @@ class PubSub:
         patterns we were previously listening to
         """
 
+        if conn.should_reconnect():
+            self._reconnect(conn)
+
         response = conn.retry.call_with_retry(
             lambda: command(*args, **kwargs),
             lambda _: self._reconnect(conn),
         )
-        if conn.should_reconnect():
-            self._reconnect(conn)
 
         return response
 
@@ -1172,6 +1173,7 @@ class PubSub:
                 return None
 
         response = self.parse_response(block=(timeout is None), timeout=timeout)
+
         if response:
             return self.handle_message(response, ignore_subscribe_messages)
         return None
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -445,6 +445,8 @@ class AbstractConnection(ConnectionInterface):
                 if orig_socket_connect_timeout
                 else self.socket_connect_timeout
             )
+        else:
+            self._maintenance_event_connection_handler = None
         self._should_reconnect = False
         self.maintenance_state = maintenance_state
 
@@ -511,8 +513,8 @@ class AbstractConnection(ConnectionInterface):
             maintenance_event_pool_handler.handle_event
         )
 
-        # Initialize maintenance event connection handler if it doesn't exist
-        if not hasattr(self, "_maintenance_event_connection_handler"):
+        # Update maintenance event connection handler if it doesn't exist
+        if not self._maintenance_event_connection_handler:
             self._maintenance_event_connection_handler = (
                 MaintenanceEventConnectionHandler(
                     self, maintenance_event_pool_handler.config
@@ -520,6 +522,10 @@ class AbstractConnection(ConnectionInterface):
             )
             self._parser.set_maintenance_push_handler(
                 self._maintenance_event_connection_handler.handle_event
+            )
+        else:
+            self._maintenance_event_connection_handler.config = (
+                maintenance_event_pool_handler.config
             )
 
     def connect(self):
@@ -1983,20 +1989,8 @@ class ConnectionPool:
         :param reset_relax_timeout: Whether to reset the relax timeout to the original timeout.
         :param include_free_connections: Whether to include free/available connections.
         """
-        for conn in self._in_use_connections:
-            if self.should_update_connection(
-                conn, address_type_to_match, matching_address
-            ):
-                self.update_connection_settings(
-                    conn,
-                    state=state,
-                    relax_timeout=relax_timeout,
-                    reset_host_address=reset_host_address,
-                    reset_relax_timeout=reset_relax_timeout,
-                )
-
-        if include_free_connections:
-            for conn in self._available_connections:
+        with self._lock:
+            for conn in self._in_use_connections:
                 if self.should_update_connection(
                     conn, address_type_to_match, matching_address
                 ):
@@ -2007,6 +2001,19 @@ class ConnectionPool:
                         reset_host_address=reset_host_address,
                         reset_relax_timeout=reset_relax_timeout,
                     )
+
+            if include_free_connections:
+                for conn in self._available_connections:
+                    if self.should_update_connection(
+                        conn, address_type_to_match, matching_address
+                    ):
+                        self.update_connection_settings(
+                            conn,
+                            state=state,
+                            relax_timeout=relax_timeout,
+                            reset_host_address=reset_host_address,
+                            reset_relax_timeout=reset_relax_timeout,
+                        )
 
     def update_connection_kwargs(
         self,
@@ -2036,11 +2043,12 @@ class ConnectionPool:
         :param tmp_relax_timeout: The relax timeout to use for the connection.
         :param moving_address_src: The address of the node that is being moved.
         """
-        for conn in self._in_use_connections:
-            if self.should_update_connection(conn, "connected", moving_address_src):
-                self._update_connection_for_reconnect(
-                    conn, tmp_host_address, tmp_relax_timeout
-                )
+        with self._lock:
+            for conn in self._in_use_connections:
+                if self.should_update_connection(conn, "connected", moving_address_src):
+                    self._update_connection_for_reconnect(
+                        conn, tmp_host_address, tmp_relax_timeout
+                    )
 
     def disconnect_and_reconfigure_free_connections(
         self,
@@ -2058,12 +2066,12 @@ class ConnectionPool:
         :param tmp_relax_timeout: The relax timeout to use for the connection.
         :param moving_address_src: The address of the node that is being moved.
         """
-
-        for conn in self._available_connections:
-            if self.should_update_connection(conn, "connected", moving_address_src):
-                self._disconnect_and_update_connection_for_reconnect(
-                    conn, tmp_host_address, tmp_relax_timeout
-                )
+        with self._lock:
+            for conn in self._available_connections:
+                if self.should_update_connection(conn, "connected", moving_address_src):
+                    self._disconnect_and_update_connection_for_reconnect(
+                        conn, tmp_host_address, tmp_relax_timeout
+                    )
 
     def _update_connection_for_reconnect(
         self,
@@ -2340,22 +2348,9 @@ class BlockingConnectionPool(ConnectionPool):
         """
         Override base class method to work with BlockingConnectionPool's structure.
         """
-        if include_free_connections:
-            for conn in tuple(self._connections):
-                if self.should_update_connection(
-                    conn, address_type_to_match, matching_address
-                ):
-                    self.update_connection_settings(
-                        conn,
-                        state=state,
-                        relax_timeout=relax_timeout,
-                        reset_host_address=reset_host_address,
-                        reset_relax_timeout=reset_relax_timeout,
-                    )
-        else:
-            connections_in_queue = {conn for conn in self.pool.queue if conn}
-            for conn in self._connections:
-                if conn not in connections_in_queue:
+        with self._lock:
+            if include_free_connections:
+                for conn in tuple(self._connections):
                     if self.should_update_connection(
                         conn, address_type_to_match, matching_address
                     ):
@@ -2366,6 +2361,20 @@ class BlockingConnectionPool(ConnectionPool):
                             reset_host_address=reset_host_address,
                             reset_relax_timeout=reset_relax_timeout,
                         )
+            else:
+                connections_in_queue = {conn for conn in self.pool.queue if conn}
+                for conn in self._connections:
+                    if conn not in connections_in_queue:
+                        if self.should_update_connection(
+                            conn, address_type_to_match, matching_address
+                        ):
+                            self.update_connection_settings(
+                                conn,
+                                state=state,
+                                relax_timeout=relax_timeout,
+                                reset_host_address=reset_host_address,
+                                reset_relax_timeout=reset_relax_timeout,
+                            )
 
     def update_active_connections_for_reconnect(
         self,
@@ -2409,15 +2418,16 @@ class BlockingConnectionPool(ConnectionPool):
         :param tmp_relax_timeout: The relax timeout to use for the connection.
         :param moving_address_src: The address of the node that is being moved.
         """
-        existing_connections = self.pool.queue
+        with self._lock:
+            existing_connections = self.pool.queue
 
-        for conn in existing_connections:
-            if conn:
-                if moving_address_src and conn.getpeername() != moving_address_src:
-                    continue
-                self._disconnect_and_update_connection_for_reconnect(
-                    conn, tmp_host_address, tmp_relax_timeout
-                )
+            for conn in existing_connections:
+                if conn:
+                    if moving_address_src and conn.getpeername() != moving_address_src:
+                        continue
+                    self._disconnect_and_update_connection_for_reconnect(
+                        conn, tmp_host_address, tmp_relax_timeout
+                    )
 
     def _update_maintenance_events_config_for_connections(
         self, maintenance_events_config
--- a/tests/test_maintenance_events_handling.py
+++ b/tests/test_maintenance_events_handling.py
@@ -503,7 +503,7 @@ class TestMaintenanceEventsHandlingSingleProxy:
 
         # Verify that maintenance events are initially disabled
         assert existing_conn._parser.node_moving_push_handler_func is None
-        assert not hasattr(existing_conn, "_maintenance_event_connection_handler")
+        assert existing_conn._maintenance_event_connection_handler is None
         assert existing_conn._parser.maintenance_push_handler_func is None
 
         # Create a new enabled configuration and set up pool handler


redis/maintenance_events.py
```diff
@@ -0,0 +1,434 @@
...
+                        # take care for the active connections in the pool
+                        # mark them for reconnect after they complete the current command
+                        self.pool.update_active_connections_for_reconnect(
+                            tmp_host_address=event.new_node_host,
+                            tmp_relax_timeout=self.config.relax_timeout,
+                        )
+
+                        # take care for the inactive connections in the pool
+                        # delete them and create new ones
+                        start_time_2 = time.time()
```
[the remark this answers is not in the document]
petyaslavova: Fixed

redis/connection.py
```diff
@@ -1442,6 +1628,42 @@ def get_protocol(self):
         """
         return self.connection_kwargs.get("protocol", None)
 
+    def maintenance_events_pool_handler_enabled(self):
+        """
+        Returns:
+            True if the maintenance events pool handler is enabled, False otherwise.
+        """
+        maintenance_events_config = self.connection_kwargs.get(
```
petyaslavova, line 1694: Default return value should be None

redis/connection.py
```diff
@@ -221,6 +232,66 @@ def set_re_auth_token(self, token: TokenInterface):
...
+    @abstractmethod
+    def should_reconnect(self):
+        """
+        Returns True if the connection should be reconnected.
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def maintenance_state(self) -> MaintenanceState:
```
petyaslavova: maintenance_state  methods should be moved above mark_for_reconnect

redis/connection.py
```diff
@@ -1559,12 +1784,18 @@ def make_connection(self) -> "ConnectionInterface":
             raise ConnectionError("Too many connections")
         self._created_connections += 1
 
+        # Pass current maintenance_state to new connections
```
petyaslavova: Check if this is needed here: state should be saved in the kwargs when editing the data on moving notification handling, in the other cases it should already be None by default.

redis/connection.py
```diff
@@ -1794,25 +2277,129 @@ def release(self, connection):
...
+                connection.disconnect()
+        finally:
+            if self._locked:
+                try:
+                    self._lock.release()
+                except Exception:
+                    pass
+                self._locked = False
+
+    def update_active_connections_for_reconnect(
```
petyaslavova, line 2379: Add docstring

redis/connection.py
```diff
@@ -1794,25 +2277,129 @@ def release(self, connection):
...
+            for conn in self._connections:
+                if conn not in connections_in_queue:
+                    self._update_connection_for_reconnect(
+                        conn,
+                        tmp_host_address,
+                        tmp_relax_timeout,
+                        skip_original_data_update,
+                    )
+
+    def disconnect_and_reconfigure_free_connections(
```
petyaslavova, line 2405: add docstring

redis/connection.py
```diff
@@ -1794,25 +2277,129 @@ def release(self, connection):
...
+    ):
+        existing_connections = self.pool.queue
+
+        for conn in existing_connections:
+            if conn:
+                self._disconnect_and_update_connection_for_reconnect(
+                    conn, tmp_host_address, tmp_relax_timeout, skip_original_data_update
+                )
+
+    def update_connections_current_timeout(
```
petyaslavova: add docstring

redis/connection.py
```diff
@@ -1794,25 +2277,129 @@ def release(self, connection):
...
+    def _update_maintenance_events_config_for_connections(
+        self, maintenance_events_config
+    ):
+        for conn in tuple(self._connections):
+            conn.maintenance_events_config = maintenance_events_config
+
+    def _update_maintenance_events_configs_for_connections(
+        self, maintenance_events_pool_handler
+    ):
+        """Override base class method to work with BlockingConnectionPool's structure."""
```
petyaslavova: add docstring

redis/connection.py
```diff
@@ -1794,25 +2277,129 @@ def release(self, connection):
...
+        """
+        Override base class method to work with BlockingConnectionPool's structure.
+
+        Restore original settings from temporary configuration for all connections in the pool.
+        """
+        for conn in tuple(self._connections):
+            conn.reset_tmp_settings(reset_host_address=True, reset_relax_timeout=True)
+
+    def set_in_maintenance(self, in_maintenance: bool):
+        """Set the maintenance mode for the connection pool."""
```
petyaslavova: add more info - why is this needed

redis/maintenance_events.py
```diff
@@ -0,0 +1,484 @@
...
+
+            with self.pool._lock:
+                if (
+                    self.config.proactive_reconnect
+                    or self.config.is_relax_timeouts_enabled()
+                ):
+                    if getattr(self.pool, "set_in_maintenance", False):
+                        self.pool.set_in_maintenance(True)
+
+                    prev_moving_in_progress = False
```
petyaslavova: Add comment with info why this is needed.

redis/_parsers/base.py
```diff
@@ -158,48 +163,117 @@ async def read_response(
...
+            *_INVALIDATION_MESSAGE,
+            *_MAINTENANCE_MESSAGES,
+            *_MOVING_MESSAGE,
+        ):
             return self.pubsub_push_handler_func(response)
-        if self.invalidation_push_handler_func:
+        if msg_type in _INVALIDATION_MESSAGE and self.invalidation_push_handler_func:
             return self.invalidation_push_handler_func(response)
+        if msg_type in _MOVING_MESSAGE and self.node_moving_push_handler_func:
+            if msg_type in _MOVING_MESSAGE:
```
vladvildanov: Duplicated if condition comparing to previous line, looks redundant

tests/test_maintenance_events.py
```diff
@@ -0,0 +1,546 @@
...
+        """Test is_expired returns False for non-expired event."""
+        with patch("time.monotonic", return_value=1000):
+            event = NodeMovingEvent(
+                id=1, new_node_host="localhost", new_node_port=6379, ttl=10
+            )
+
+        with patch("time.monotonic", return_value=1005):  # 5 seconds later
+            assert not event.is_expired()
+
+    def test_is_expired_true(self):
```
vladvildanov: This test could be merged with previous using `@pytest.fixture.parametrize`

redis/connection.py
```diff
@@ -249,6 +320,12 @@ def __init__(
         protocol: Optional[int] = 2,
         command_packer: Optional[Callable[[], None]] = None,
         event_dispatcher: Optional[EventDispatcher] = None,
+        maintenance_events_pool_handler: Optional[MaintenanceEventPoolHandler] = None,
+        maintenance_events_config: Optional[MaintenanceEventsConfig] = None,
+        orig_host_address: Optional[str] = None,
```
vladvildanov, line 345: Isn't this parameter on creation time just equals `host`? As I understand this is original host, so you can just assign default host to keep original value. Why someone would need to set it separately?

redis/maintenance_events.py
```diff
@@ -0,0 +1,484 @@
...
+
+class MaintenanceEventPoolHandler:
+    def __init__(
+        self,
+        pool: Union["ConnectionPool", "BlockingConnectionPool"],
+        config: MaintenanceEventsConfig,
+    ) -> None:
+        self.pool = pool
+        self.config = config
+        self._processed_events = set()
```
vladvildanov, line 325: Just to clarify, you keep processed events to de-duplicate push notifications received by all connections in pool, whereas you need to react only on the first one?

redis/connection.py
petyaslavova: Hm... That a very good idea :) I can set it when the client is initiated and never update it.. I'll update the code according to that logic.

redis/maintenance_events.py
petyaslavova: Yes, when the moving comes only the first reading should apply changes. The notification will be sent to all open connections. The first connection/parser that reads it, updates all other connections in the pool. Any subsequent reads of the same notification should simply be ignored.

redis/_parsers/base.py
petyaslavova: Fixed.

redis/connection.py
petyaslavova: done.

redis/connection.py
petyaslavova: Done.

redis/connection.py
petyaslavova: Done.

tests/test_maintenance_events.py
petyaslavova: done.

redis/connection.py
petyaslavova: Removed - works without it as well - it has a default value when it is not provided.

redis/connection.py
petyaslavova: Done.

redis/connection.py
petyaslavova: done

redis/connection.py
petyaslavova: done

redis/connection.py
petyaslavova: done.

redis/connection.py
petyaslavova: done.

redis/connection.py
```diff
@@ -1646,6 +1815,154 @@ def re_auth_callback(self, token: TokenInterface):
             for conn in self._in_use_connections:
                 conn.set_re_auth_token(token)
 
+    def update_connection_kwargs_with_tmp_settings(
```
vladvildanov: Could you make it more generic? Something like `update_connection_kwargs(**kwargs)`

redis/connection.py
```diff
@@ -1646,6 +1815,154 @@ def re_auth_callback(self, token: TokenInterface):
...
+        This new address will be used to create new connections until the old node is decomissioned.
+
+        :param tmp_host_address: The temporary host address to use for the connection.
+        :param tmp_relax_timeout: The relax timeout to use for the connection.
+                              If -1 is provided - the relax timeout is disabled, so the tmp property is not set
+        """
+        self.connection_kwargs.update({"tmp_host_address": tmp_host_address})
+        self.connection_kwargs.update({"tmp_relax_timeout": tmp_relax_timeout})
+
+    def update_connections_tmp_settings(
```
vladvildanov: The same here, could it be more generic?

redis/connection.py
```diff
@@ -1646,6 +1815,154 @@ def re_auth_callback(self, token: TokenInterface):
...
+        :param tmp_host_address: The temporary host address to use for the connection.
+        :param tmp_relax_timeout: The relax timeout to use for the connection.
+        """
+
+        for conn in self._available_connections:
+            self._disconnect_and_update_connection_for_reconnect(
+                conn, tmp_host_address, tmp_relax_timeout
+            )
+
+    def update_connections_current_timeout(
```
vladvildanov: This will be not needed if `update_connections_tmp_settings` will be more generic

redis/asyncio/connection.py
```diff
@@ -1308,6 +1308,8 @@ def __init__(
         )
         self._condition = asyncio.Condition()
         self.timeout = timeout
+        self._in_maintenance = False
```
vladvildanov: Is it needed? The PR mentions that it's synchronous client only

redis/connection.py
```diff
@@ -2146,66 +2346,173 @@ def release(self, connection):
...
+                        continue
+                else:
+                    if matching_address and conn.host != matching_address:
+                        continue
+                conn.update_current_socket_timeout(relax_timeout)
+        else:
+            connections_in_queue = {conn for conn in self.pool.queue if conn}
+            for conn in self._connections:
+                if conn not in connections_in_queue:
+                    if address_type_to_match == "connected":
```
vladvildanov: There's a lot of duplicate of this part:

```
                    if address_type_to_match == "connected":
                        if matching_address and conn.getpeername() != matching_address:
                            continue
```

Maybe it makes to create a method that filter connections by this criteria and returns filtered?

vladvildanov (changes requested):

redis/asyncio/connection.py
petyaslavova: There were failures while testing and this one fixed it, but now that I removed it, everything is still working.. So I'll remove this change from the PR.

redis/maintenance_events.py
```diff
@@ -0,0 +1,491 @@
...
+                        # take care for the inactive connections in the pool
+                        # delete them and create new ones
+                        self.pool.disconnect_and_reconfigure_free_connections(
+                            tmp_host_address=event.new_node_host,
+                            tmp_relax_timeout=self.config.relax_timeout,
+                            moving_address_src=moving_address_src,
+                        )
+                        if getattr(self.pool, "set_in_maintenance", False):
+                            self.pool.set_in_maintenance(False)
+            print(f"Starting timer for {event} for {event.ttl} seconds")
```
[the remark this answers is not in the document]
petyaslavova: done

tests/test_maintenance_events_handling.py
```diff
@@ -0,0 +1,1770 @@
...
+            Helpers.validate_free_connections_state(
+                pool=test_redis_client.connection_pool,
+                expected_host_address=DEFAULT_ADDRESS.split(":")[0],
+                expected_socket_timeout=None,
+                expected_socket_connect_timeout=None,
+                expected_orig_host_address=DEFAULT_ADDRESS.split(":")[0],
+                expected_orig_socket_timeout=None,
+                expected_orig_socket_connect_timeout=None,
+                should_be_connected_count=1,
+                connected_to_tmp_addres=True,
```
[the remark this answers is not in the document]
petyaslavova: Fixed.

tests/test_maintenance_events_handling.py
```diff
@@ -0,0 +1,1770 @@
...
+                pool=test_redis_client.connection_pool,
+                expected_state=MaintenanceState.MOVING,
+                expected_host_address=AFTER_MOVING_ADDRESS.split(":")[0],
+                expected_socket_timeout=self.config.relax_timeout,
+                expected_socket_connect_timeout=self.config.relax_timeout,
+                expected_orig_host_address=DEFAULT_ADDRESS.split(":")[0],
+                expected_orig_socket_timeout=None,
+                expected_orig_socket_connect_timeout=None,
+                should_be_connected_count=1,
+                connected_to_tmp_addres=True,
```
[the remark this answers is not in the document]
petyaslavova: Fixed.

tests/test_maintenance_events_handling.py
```diff
@@ -0,0 +1,1770 @@
...
+                    assert (
+                        connection._sock.getpeername()[0] == expected_current_peername
+                    )
+            assert connection.maintenance_state == expected_state
+
+    @staticmethod
+    def validate_free_connections_state(
+        pool,
+        should_be_connected_count=0,
+        connected_to_tmp_addres=False,
```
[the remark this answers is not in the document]
petyaslavova: Fixed.

tests/test_maintenance_events_handling.py
```diff
@@ -0,0 +1,1770 @@
...
+                expected_orig_host_address=DEFAULT_ADDRESS.split(":")[0],
+                expected_orig_socket_timeout=None,
+                expected_orig_socket_connect_timeout=None,
+                expected_current_socket_timeout=self.config.relax_timeout,
+                expected_current_peername=DEFAULT_ADDRESS.split(":")[0],
+            )
+            Helpers.validate_free_connections_state(
+                pool=test_redis_client.connection_pool,
+                should_be_connected_count=1,
+                connected_to_tmp_addres=True,
```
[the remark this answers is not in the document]
petyaslavova: Fixed.

tests/test_maintenance_events_handling.py
```diff
@@ -0,0 +1,1770 @@
...
+                    expected_orig_socket_timeout=None,
+                    expected_orig_socket_connect_timeout=None,
+                    expected_current_socket_timeout=self.config.relax_timeout,
+                    expected_current_peername=orig_after_moving.split(":")[0],
+                )
+                # print(test_redis_client.connection_pool._available_connections)
+                Helpers.validate_free_connections_state(
+                    test_redis_client.connection_pool,
+                    should_be_connected_count=1,
+                    connected_to_tmp_addres=True,
```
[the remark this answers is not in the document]
petyaslavova: Fixed.

tests/test_maintenance_events_handling.py
```diff
@@ -0,0 +1,1770 @@
...
+            expected_orig_host_address=DEFAULT_ADDRESS.split(":")[0],
+            expected_orig_socket_timeout=None,
+            expected_orig_socket_connect_timeout=None,
+            expected_current_socket_timeout=self.config.relax_timeout,
+            expected_current_peername=DEFAULT_ADDRESS.split(":")[0],
+        )
+        Helpers.validate_free_connections_state(
+            pool=pool,
+            should_be_connected_count=0,
+            connected_to_tmp_addres=False,
```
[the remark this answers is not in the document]
petyaslavova: Fixed.

tests/test_maintenance_events_handling.py
```diff
@@ -0,0 +1,1770 @@
...
+            expected_orig_host_address=DEFAULT_ADDRESS.split(":")[0],
+            expected_orig_socket_timeout=None,
+            expected_orig_socket_connect_timeout=None,
+            expected_current_socket_timeout=None,
+            expected_current_peername=DEFAULT_ADDRESS.split(":")[0],
+        )
+        Helpers.validate_free_connections_state(
+            pool=pool,
+            should_be_connected_count=0,
+            connected_to_tmp_addres=False,
```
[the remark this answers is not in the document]
petyaslavova: Fixed.

redis/connection.py
petyaslavova: I wanted to iterate over the different lists once and not create additional lists holding references to the connection objects that match the filter, so I think it will be better to create a new helper method that just checks if the connection should be updated or not. You are right - it is a lot of code duplication :)

redis/connection.py
petyaslavova: @vladvildanov I didn’t quite understand this one — do you mean having a single generic method that handles all possible updates to object properties, including the socket timeout update for already connected connections?

I intentionally separated them to make it easier to follow what exactly gets updated during notification handling — specifically, whether we’re only updating properties used during `connect`, or actually changing the behavior of already established socket connections.

redis/connection.py
petyaslavova: Now I see why I didn't understand the next comment :) This is an old function that has been removed and split into two separate functions. 
I'll see how will be appropriate to combine some of the into more generic ones or which of the newly added ones can be transformed into more generic ones(like the one mentioned above --> update_connection_kwargs)

redis/connection.py
petyaslavova: Done.

redis/connection.py
```diff
@@ -732,6 +868,60 @@ def re_auth(self):
...
+            self._sock.settimeout(timeout)
+            self.update_parser_buffer_timeout(timeout)
+
+    def update_parser_buffer_timeout(self, timeout: Optional[float] = None):
+        if self._parser and self._parser._buffer:
+            self._parser._buffer.socket_timeout = timeout
+
+    def set_tmp_settings(
+        self,
+        tmp_host_address: Optional[Union[str, object]] = SENTINEL,
```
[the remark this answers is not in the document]
petyaslavova: It is defined in the file. Code is executed successfully.

redis/connection.py
petyaslavova: I refactored the code, and now there are two more generic methods: `update_connection_kwargs` and `update_connections_settings`, that can be further extended if needed.

redis/connection.py
petyaslavova: Done.

redis/connection.py
```diff
@@ -1646,6 +1906,171 @@ def re_auth_callback(self, token: TokenInterface):
...
+        moving_address_src: Optional[str] = None,
+    ):
+        """
+        Disconnect all free/available connections.
+        This is used when a cluster node is migrated to a different address.
+
+        When this method is called the pool will already be locked, so getting the pool lock inside is not needed.
+
+        :param orig_host_address: The temporary host address to use for the connection.
+        :param orig_relax_timeout: The relax timeout to use for the connection.
```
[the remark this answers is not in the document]
petyaslavova: fixed.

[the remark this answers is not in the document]
petyaslavova: fixed

tests/test_maintenance_events_handling.py
```diff
@@ -0,0 +1,1770 @@
...
+                # MIGRATED push message before SET key_receive_migrated_X response
+                # Format: >1\r\n$8\r\nMIGRATED\r\n (1 element: MIGRATED)
+                migrated_push = ">1\r\n$8\r\nMIGRATED\r\n"
+                response = migrated_push.encode() + response
+            elif b"key_receive_moving_" in data:
+                # MOVING push message before SET key_receive_moving_X response
+                # Format: >3\r\n$6\r\nMOVING\r\n:15\r\n+localhost:6379\r\n (3 elements: MOVING, ttl, host:port)
+                # Note: Using + instead of $ to send as simple string instead of bulk string
+                moving_push = f">3\r\n$6\r\nMOVING\r\n:{MOVING_TIMEOUT}\r\n+{AFTER_MOVING_ADDRESS}\r\n"
+                response = moving_push.encode() + response
```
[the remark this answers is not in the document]
petyaslavova: Will leave it this way for now.

redis/_parsers/base.py
```diff
@@ -158,48 +164,122 @@ async def read_response(
...
+        self.maintenance_push_handler_func = maintenance_push_handler_func
+
 
 class AsyncPushNotificationsParser(Protocol):
     """Protocol defining async RESP3-specific parsing functionality"""
 
     pubsub_push_handler_func: Callable
     invalidation_push_handler_func: Optional[Callable] = None
+    node_moving_push_handler_func: Optional[Callable] = None
+    maintenance_push_handler_func: Optional[Callable] = None
```
vladvildanov: Just for the better typing, makes sense to make them as Awaitable or Coroutines, to clarify that the handler func should be async

redis/client.py
```diff
@@ -911,10 +953,15 @@ def _execute(self, conn, command, *args, **kwargs):
         called by the # connection to resubscribe us to any channels and
         patterns we were previously listening to
         """
-        return conn.retry.call_with_retry(
+
+        response = conn.retry.call_with_retry(
             lambda: command(*args, **kwargs),
             lambda _: self._reconnect(conn),
         )
+        if conn.should_reconnect():
```
vladvildanov: Hmm, what about moving it above the command execution? Makes sense if it needs to be reconnected before rather than after, taking into a count a fact that connection may be anyway reconnected in case of error.

redis/connection.py
```diff
@@ -375,6 +503,25 @@ def set_parser(self, parser_class):
...
+    def set_maintenance_event_pool_handler(
+        self, maintenance_event_pool_handler: MaintenanceEventPoolHandler
+    ):
+        maintenance_event_pool_handler.set_connection(self)
+        self._parser.set_node_moving_push_handler(
+            maintenance_event_pool_handler.handle_event
+        )
+
+        # Initialize maintenance event connection handler if it doesn't exist
+        if not hasattr(self, "_maintenance_event_connection_handler"):
```
vladvildanov: Can't we always initialise this property as None and make check for None? Is there's any specific reason why we want to not have this property at all?

redis/connection.py
```diff
@@ -1646,6 +1920,173 @@ def re_auth_callback(self, token: TokenInterface):
...
+
+        :param state: The maintenance state to set for the connection.
+        :param relax_timeout: The relax timeout to set for the connection.
+        :param matching_address: The address to match for the connection.
+        :param address_type_to_match: The type of address to match.
+        :param reset_host_address: Whether to reset the host address to the original address.
+        :param reset_relax_timeout: Whether to reset the relax timeout to the original timeout.
+        :param include_free_connections: Whether to include free/available connections.
+        """
+        for conn in self._in_use_connections:
```
vladvildanov: We should apply locking here, otherwise `in_use_connections` is dynamic and may lead to deadlocks

redis/connection.py
```diff
@@ -1646,6 +1920,173 @@ def re_auth_callback(self, token: TokenInterface):
...
+                self.update_connection_settings(
+                    conn,
+                    state=state,
+                    relax_timeout=relax_timeout,
+                    reset_host_address=reset_host_address,
+                    reset_relax_timeout=reset_relax_timeout,
+                )
+
+        if include_free_connections:
+            for conn in self._available_connections:
```
vladvildanov: Same here

redis/connection.py
```diff
@@ -1646,6 +1920,173 @@ def re_auth_callback(self, token: TokenInterface):
...
+        Mark all active connections for reconnect.
+        This is used when a cluster node is migrated to a different address.
+
+        When this method is called the pool will already be locked, so getting the pool lock inside is not needed.
+
+        :param tmp_host_address: The temporary host address to use for the connection.
+        :param tmp_relax_timeout: The relax timeout to use for the connection.
+        :param moving_address_src: The address of the node that is being moved.
+        """
+        for conn in self._in_use_connections:
```
vladvildanov: Same here

redis/connection.py
```diff
@@ -1646,6 +1920,173 @@ def re_auth_callback(self, token: TokenInterface):
...
+        This is used when a cluster node is migrated to a different address.
+
+        When this method is called the pool will already be locked, so getting the pool lock inside is not needed.
+
+        :param tmp_host_address: The temporary host address to use for the connection.
+        :param tmp_relax_timeout: The relax timeout to use for the connection.
+        :param moving_address_src: The address of the node that is being moved.
+        """
+
+        for conn in self._available_connections:
```
vladvildanov: Same here

redis/connection.py
```diff
@@ -1707,16 +2150,27 @@ def __init__(
...
+                    break
 
-        # Keep a list of actual connection instances so that we can
-        # disconnect them later.
-        self._connections = []
+            # Keep a list of actual connection instances so that we can
+            # disconnect them later.
+            self._connections = []
+        finally:
+            if self._locked:
```
vladvildanov, line 2176: I think this flag is redundant. If you try to release a lock that wasn't acquired by this thread it will throw a RuntimeException, if you try to release a lock that was acquired by another thread it will be the same. Since, you anyway surround `release()` method with try catch, you will either release lock if it's possible or suppress any exception and move on

redis/maintenance_events.py
```diff
@@ -0,0 +1,496 @@
...
+        self._processed_events = set()
+        self._lock = threading.RLock()
+        self.connection = None
+
+    def set_connection(self, connection: "ConnectionInterface"):
+        self.connection = connection
+
+    def remove_expired_notifications(self):
+        with self._lock:
+            for notification in tuple(self._processed_events):
```
vladvildanov, line 334: Why do you need to convert it into a tuple? Set is also iterable

redis/maintenance_events.py
```diff
@@ -0,0 +1,496 @@
...
+                if notification.is_expired():
+                    self._processed_events.remove(notification)
+
+    def handle_event(self, notification: MaintenanceEvent):
+        self.remove_expired_notifications()
+
+        if isinstance(notification, NodeMovingEvent):
+            return self.handle_node_moving_event(notification)
+        else:
+            logging.error(f"Unhandled notification type: {notification}")
```
vladvildanov, line 344: If we don't expect to handle any other events then `NodeMovingEvent` why does signature allows it's parent class to be passed (`MaintenanceEvent`)

vladvildanov (changes requested):

redis/_parsers/base.py
petyaslavova: Done.

redis/client.py
petyaslavova: Both approaches work here in the PubSub. 
We don't return the connection to the pool, so the restriction that we should return "handled" connections to the pool doesn't apply.
I'll move it above.

redis/connection.py
petyaslavova: Done.

redis/connection.py
petyaslavova: That's true, but it will be slower. 
We will be in the True case very rarely - only when we have active moving maintenance. 
If I remove the check, the exception will be raised all the time, and it is a much more resource-consuming operation than the self._locked check.

redis/maintenance_events.py
petyaslavova: Because I'm changing the elements if any of the existing inside notifications have expired - I can't iterate on an iterable object and change its elements inside the cycle.

redis/maintenance_events.py
petyaslavova: Because we don't know if this will be the only one - it is for now.

redis/connection.py
petyaslavova: Done. 
I was relying on the lock in the handlers, but you'e absolutely right - it is a public function.. It should be safe to  execute it in another context as well.

redis/connection.py
petyaslavova: The lock covers all the operations - we want to complete all connection updates before releasing the lock. Otherwise, we might skip some connections.

redis/connection.py
petyaslavova: Done.

redis/connection.py
petyaslavova: Done.

PR #3716, opened 2025-07-24
Adding handling of FAILING_OVER and FAILED_OVER events/push notifications

petyaslavova:
### Pull Request check-list

### Description of change

Adding handling of FAILING_OVER and FAILED_OVER events/push notifications

### DIFF
--- a/redis/maintenance_events.py
+++ b/redis/maintenance_events.py
@@ -12,6 +12,7 @@ class MaintenanceState(enum.Enum):
     NONE = "none"
     MOVING = "moving"
     MIGRATING = "migrating"
+    FAILING_OVER = "failing_over"
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


tests/test_maintenance_events_handling.py
```diff
@@ -16,9 +16,11 @@
 from redis.maintenance_events import (
     MaintenanceEventsConfig,
     NodeMigratingEvent,
+    NodeMigratedEvent,
+    NodeFailingOverEvent,
+    NodeFailedOverEvent,
     MaintenanceEventPoolHandler,
```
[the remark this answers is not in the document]
petyaslavova: Fixed linter errors

redis/maintenance_events.py
```diff
@@ -465,32 +565,37 @@ def __init__(
 
     def handle_event(self, event: MaintenanceEvent):
         if isinstance(event, NodeMigratingEvent):
-            return self.handle_migrating_event(event)
+            return self.handle_maintenance_start_event(MaintenanceState.MIGRATING)
         elif isinstance(event, NodeMigratedEvent):
-            return self.handle_migration_completed_event(event)
+            return self.handle_maintenance_completed_event()
```
vladvildanov: What about creating a separate mapping like this: `type(event) : callback`? This way you don't need to scale this if..else if tree and it will be better performance wise as it will be O(1) instead of O(N) complexity.

```
mapping = {
   NodeMigratingEvent: self.handle_maintenance_start_event
   ...
   NodeFailedOverEvent: self.handle_maintenance_completed_event
}

handler = self.mapping.get(type(event), None)

if handler is None:
    logging.error(f"Unhandled event type: {event}")
    return

handler()
```

petyaslavova: Great idea for simplification :) I was able to do something similar. Removed the two separate states - we react the same way on both, so we don't need to have both constants, and now it is easier to handle the two cases with fewer checks.

redis/maintenance_events.py
```diff
@@ -457,40 +557,54 @@ def handle_node_moved_event(self, event: NodeMovingEvent):
...
         self.config = config
 
     def handle_event(self, event: MaintenanceEvent):
-        if isinstance(event, NodeMigratingEvent):
-            return self.handle_migrating_event(event)
-        elif isinstance(event, NodeMigratedEvent):
-            return self.handle_migration_completed_event(event)
-        else:
+        # get the event type by checking its class in the _EVENT_TYPES dict
+        event_type = self._EVENT_TYPES.get(event.__class__)
```
vladvildanov: get(event.__class__, None) or get will raise an exception on missing type

redis/maintenance_events.py
vladvildanov: Looks good!

redis/maintenance_events.py
petyaslavova: fixed.

PR #3722, opened 2025-07-30
Added support for Pub/Sub mode in MultiDbClient

vladvildanov:
### Pull Request check-list

### DIFF
--- a/redis/multidb/command_executor.py
+++ b/redis/multidb/command_executor.py
@@ -3,13 +3,13 @@ from abc import ABC, abstractmethod
 from datetime import datetime, timedelta
 from typing import List, Union, Optional, Callable
 
-from redis.client import Pipeline
+from redis.client import Pipeline, PubSub, PubSubWorkerThread
 from redis.exceptions import ConnectionError, TimeoutError
 from redis.event import EventDispatcherInterface, OnCommandsFailEvent
 from redis.multidb.config import DEFAULT_AUTO_FALLBACK_INTERVAL
 from redis.multidb.database import Database, AbstractDatabase, Databases
 from redis.multidb.circuit import State as CBState
-from redis.multidb.event import RegisterCommandFailure
+from redis.multidb.event import RegisterCommandFailure, ActiveDatabaseChanged, ResubscribeOnActiveDatabaseChanged
 from redis.multidb.failover import FailoverStrategy
 from redis.multidb.failure_detector import FailureDetector
 
@@ -35,7 +35,7 @@ class CommandExecutor(ABC):
 
     @property
     @abstractmethod
-    def active_database(self) -> Union[Database, None]:
+    def active_database(self) -> Optional[Database]:
         """Returns currently active database."""
         pass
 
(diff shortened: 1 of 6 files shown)


redis/multidb/command_executor.py
```diff
@@ -170,6 +204,39 @@ def callback():
...
+    def execute_pubsub_run_in_thread(
+        self,
+        pubsub,
+        sleep_time: float = 0.0,
+        daemon: bool = False,
+        exception_handler: Optional[Callable] = None,
+    ) -> "PubSubWorkerThread":
+        def callback():
+            return self._active_pubsub.run_in_thread(
+                sleep_time, daemon=daemon, exception_handler=exception_handler, pubsub=pubsub
```
petyaslavova, line 235: Why do we need to provide the pubsub reference here? If a failover has occurred, shouldn’t _active_pubsub already contain all the necessary information? By passing in this reference, don’t we risk holding onto a pubsub object from a failed instance and potentially trying to work with it?

tests/test_scenario/test_active_active.py
```diff
@@ -203,4 +214,90 @@ def callback(pipe: Pipeline):
...
+        # Execute pipeline before network failure
+        while not event.is_set():
+            r_multi_db.spublish('test-channel', data)
+            sleep(0.1)
+
+        # Execute pipeline after network failure
+        for _ in range(3):
+            r_multi_db.spublish('test-channel', data)
+            sleep(0.1)
+
```
petyaslavova, line 306: What do we expect to have happened in the pubsub_thread? Would it be worth validating whether it completed successfully or failed?

redis/multidb/command_executor.py
vladvildanov: We need currently active pub/sub reference to be able to execute any commands on it in any point of time (f.e subscribe to more channels if needed or unsubscribe), if MultiDbClient is in Pub/Sub mode. Also we keep it to be able to re-initialise new pub/sub during failover with the same configuration and re-subscribe to the same channels.

We always update reference when failover happens, so we do not risk to keep stale reference.

https://github.com/redis/redis-py/pull/3722/files#diff-77af8b3b216196f2b0569279fe1c6d9db146e355eba207eeb84021ccf54c3c0eR54

tests/test_scenario/test_active_active.py
vladvildanov: In pubsub_thread we're validating if we receive expected message from subscribed channel

https://github.com/redis/redis-py/pull/3722/files/bee15d943e3c381ad1b0e9cebe6061cbce973d16#diff-eccf1b0c83ba8b8e9d8dc7e837ba57e529ec23df09f3d6f43f354fdf5584fcf7R281

Do you think we need to verify something else? Like counter to prove that it was called > 0 times?

petyaslavova: Maybe a check of the thread object results. The thread might have raised an exception due to this assert failure, but unless the results of the thread object are checked, we might never see that something has failed.

redis/multidb/command_executor.py
petyaslavova: Yes, and it handles the state of `self._active_pubsub` for which we call `run_in_thread`, but then a `pubsub` object is also provided as an argument - I was asking about this one.

vladvildanov: Ah, now I get it. The whole point that `run_in_thread` method will call `get_message()` not in the Redis client's Pub/Sub object, but on the MultiDBClient's Pub/Sub object, because only this way it will be able to handle failover. Lower-level client knows nothing about other DB's, it only operate on the one it supposed to (which I believe is the correct way), whereas MultiDbClient is able failover between DB's

tests/test_scenario/test_active_active.py
vladvildanov: Okay, I'll add additional check

redis/multidb/command_executor.py
vladvildanov: @petyaslavova This might answer your question

https://github.com/redis/redis-py/pull/3722/files#diff-e1d7fdc9293a30cdfd5fa4a4670c76ea0e597275664899ac6edc297e63c22e3aR438

tests/test_scenario/test_active_active.py
vladvildanov: @petyaslavova Done

PR #3728, opened 2025-08-02
blocking_timeout should default to self.timeout

jvanasco:
### Pull Request check-list

- [NA] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [NA] Is there an example added to the examples folder (if applicable)?

### Description of change

Default the `blocking_timeout` on `Lock.acquire` to `self.timeout`.

I ran into a platform specific error, which prevented redis-py from acquiring a Lock. Although the `timeout` was set, there was no `blocking_timeout`.  This led to a test suite hanging on an infinite loop, minutes longer than the explicit timeout.  There is no reason for a blocking_timeout to ever be longer than the normal timeout.

If the acquire succeeds after the Lock.timeout expires, the lock will already fail.  This change simply makes that failure instant, and prevents potentially infinite loops.

### DIFF
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -219,7 +219,7 @@ class Lock:
         if blocking is None:
             blocking = self.blocking
         if blocking_timeout is None:
-            blocking_timeout = self.blocking_timeout
+            blocking_timeout = self.blocking_timeout or self.timeout
         stop_trying_at = None
         if blocking_timeout is not None:
             stop_trying_at = mod_time.monotonic() + blocking_timeout


jvanasco:
All working tests pass in my repo, except for the same test that is currently failing on the main repo.

petyaslavova:
Hi @jvanasco,
Thank you for the time and effort you've put into this PR — it's much appreciated!

Unfortunately, I won't be able to merge this change. According to the documentation for the `lock`, the `timeout` and `blocking_timeout` parameters serve distinct purposes and are not intended to be used interchangeably. Specifically, `timeout` defines how long the lock will be held after it is acquired, whereas `blocking_timeout` controls how long the system should wait when attempting to acquire the lock. Mixing the two could lead to unintended behavior.

Thanks again for your contribution, and I’m happy to discuss further if you have a different approach in mind.

petyaslavova:
Closing this PR due to the last comment.

jvanasco:
Sorry for the late reply, I was away.  You are correct, and this could lead to unintended circumstances when there is expected long blocking.  I will be filing bugs with the libraries I used that did not accept/persist the blocking timeout.

Can I suggest a breaking change for the next major version, in which there is a default `blocking_timeout` of 60 seconds?  My concern here is ensuring test and CI systems do not indefinitely hang - which happens when redis-py is unable to obtain a lock.

petyaslavova:
> Sorry for the late reply, I was away. You are correct, and this could lead to unintended circumstances when there is expected long blocking. I will be filing bugs with the libraries I used that did not accept/persist the blocking timeout.
> 
> Can I suggest a breaking change for the next major version, in which there is a default `blocking_timeout` of 60 seconds? My concern here is ensuring test and CI systems do not indefinitely hang - which happens when redis-py is unable to obtain a lock.

@jvanasco thanks for the suggestion, we will discuss it internally :)

PR #3751, opened 2025-08-25
Hitless-Upgrade: Add handling of MOVING push notification with "null" host:port info.

petyaslavova:
### Pull Request check-list

### Description of change

Adding handling for the cases when push notification doesn't contain new host and port pair - it just contains "null" txt.
For this case we are not supposed to reconnect immediately  - we should wait half of the ttl in the notification and then trigger reconnects to the originally set host address.

### DIFF
--- a/redis/_parsers/base.py
+++ b/redis/_parsers/base.py
@@ -203,38 +203,38 @@ class MaintenanceNotificationsParser:
         return NodeMovingEvent(id, host, port, ttl)
 
 
-_INVALIDATION_MESSAGE = (b"invalidate", "invalidate")
-_MOVING_MESSAGE = (b"MOVING", "MOVING")
-_MIGRATING_MESSAGE = (b"MIGRATING", "MIGRATING")
-_MIGRATED_MESSAGE = (b"MIGRATED", "MIGRATED")
-_FAILING_OVER_MESSAGE = (b"FAILING_OVER", "FAILING_OVER")
-_FAILED_OVER_MESSAGE = (b"FAILED_OVER", "FAILED_OVER")
+_INVALIDATION_MESSAGE = "invalidate"
+_MOVING_MESSAGE = "MOVING"
+_MIGRATING_MESSAGE = "MIGRATING"
+_MIGRATED_MESSAGE = "MIGRATED"
+_FAILING_OVER_MESSAGE = "FAILING_OVER"
+_FAILED_OVER_MESSAGE = "FAILED_OVER"
 
 _MAINTENANCE_MESSAGES = (
-    *_MIGRATING_MESSAGE,
-    *_MIGRATED_MESSAGE,
-    *_FAILING_OVER_MESSAGE,
-    *_FAILED_OVER_MESSAGE,
+    _MIGRATING_MESSAGE,
+    _MIGRATED_MESSAGE,
+    _FAILING_OVER_MESSAGE,
+    _FAILED_OVER_MESSAGE,
 )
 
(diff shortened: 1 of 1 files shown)


redis/_parsers/base.py
```diff
@@ -214,7 +216,11 @@ def handle_push_response(self, response, **kwargs):
                 # Expected message format is: MOVING <seq_number> <time> <endpoint>
                 id = response[1]
                 ttl = response[2]
-                host, port = response[3].decode().split(":")
+                if response[3] in [b"null", "null"]:
+                    host, port = None, None
```
vladvildanov: Does it makes sense to continue execution flow if host and port are null? How would it affect overall event handling?

redis/_parsers/base.py
```diff
@@ -229,6 +235,13 @@ def handle_push_response(self, response, **kwargs):
                 elif msg_type in _MIGRATED_MESSAGE:
                     id = response[1]
                     notification = NodeMigratedEvent(id)
+                elif msg_type in _FAILING_OVER_MESSAGE:
```
vladvildanov: The approach I recommended earlier (dictionary where keys are message types and values are handling callback) would work here better because of clean code practices + O(1) complexity

vladvildanov (changes requested):

redis/_parsers/base.py
```diff
@@ -214,7 +216,11 @@ def handle_push_response(self, response, **kwargs):
                 # Expected message format is: MOVING <seq_number> <time> <endpoint>
                 id = response[1]
                 ttl = response[2]
-                host, port = response[3].decode().split(":")
+                if response[3] in [b"null", "null"]:
+                    host, port = None, None
+                else:
+                    host, port = response[3].decode().split(":")
```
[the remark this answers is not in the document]
petyaslavova: done

redis/_parsers/base.py
```diff
@@ -286,7 +299,11 @@ async def handle_push_response(self, response, **kwargs):
                 # push notification from enterprise cluster for node moving
                 id = response[1]
                 ttl = response[2]
-                host, port = response[3].split(":")
+                if response[3] in [b"null", "null"]:
+                    host, port = None, None
+                else:
+                    host, port = response[3].decode().split(":")
```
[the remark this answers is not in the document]
petyaslavova: done

redis/maintenance_events.py
```diff
@@ -636,12 +668,35 @@ def handle_node_moving_event(self, event: NodeMovingEvent):
...
+            # mark them for reconnect after they complete the current command
+            self.pool.update_active_connections_for_reconnect(
+                moving_address_src=moving_address_src,
+            )
+            # take care for the inactive connections in the pool
+            # delete them and create new ones
+            self.pool.disconnect_free_connections(
+                moving_address_src=moving_address_src,
+            )
+
```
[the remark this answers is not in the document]
petyaslavova: done.

redis/_parsers/base.py
petyaslavova: This is a special case. We should continue with the execution, and instead of immediately handling the reconnects, we should do it after half of the ttl time, also instead of reconnecting to a new host address(which is not provided in this case), we reconnect to the original one.

redis/_parsers/base.py
petyaslavova: Done.
@vladvildanov I was thinking about those msg_type comparisons... Does it make sense to decode the msg_type - it is usually a small string and change the constants to contain only strings and compare against strings only from the very beginning?

vladvildanov: Yeah, I think strings is better, this is maintenance information that is never exposed to the end user, so we might treat it the way it works better to us

PR #3754, opened 2025-08-27
Extract additional interfaces and abstract classes

vladvildanov:
### Pull Request check-list

### Description of change

Additional base classes will be further re-used in async implementation

### DIFF
--- a/redis/multidb/circuit.py
+++ b/redis/multidb/circuit.py
@@ -45,48 +45,11 @@ class CircuitBreaker(ABC):
         """Set database associated with this circuit."""
         pass
 
-    @abstractmethod
-    def on_state_changed(self, cb: Callable[["CircuitBreaker", State, State], None]):
-        """Callback called when the state of the circuit changes."""
-        pass
-
-class PBListener(pybreaker.CircuitBreakerListener):
-    """Wrapper for callback to be compatible with pybreaker implementation."""
-    def __init__(
-            self,
-            cb: Callable[[CircuitBreaker, State, State], None],
-            database,
-    ):
-        """
-        Initialize a PBListener instance.
-
-        Args:
-            cb: Callback function that will be called when the circuit breaker state changes.
-            database: Database instance associated with this circuit breaker.
-        """
-
-        self._cb = cb
-        self._database = database
-
-    def state_change(self, cb, old_state, new_state):
-        cb = PBCircuitBreakerAdapter(cb)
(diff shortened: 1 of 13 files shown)


redis/multidb/circuit.py
```diff
@@ -75,8 +116,7 @@ def state_change(self, cb, old_state, new_state):
         new_state = State(value=new_state.name)
         self._cb(cb, old_state, new_state)
 
-
-class PBCircuitBreakerAdapter(CircuitBreaker):
+class PBCircuitBreakerAdapter(SyncCircuitBreaker, BaseCircuitBreaker):
```
petyaslavova, line 119: Is there a specific reason to use multiple inheritance here? If `SyncCircuitBreaker` extends `BaseCircuitBreaker`, you could inherit from just one class, which would make it easier to follow.

redis/multidb/database.py
```diff
@@ -5,65 +5,92 @@
...
-        """Set the health check URL associated with the current database."""
+    def circuit(self, circuit: SyncCircuitBreaker):
+        """Set the circuit breaker for the current database."""
         pass
 
-Databases = WeightedList[tuple[AbstractDatabase, Number]]
+Databases = WeightedList[tuple[SyncDatabase, Number]]
 
-class Database(AbstractDatabase):
+class Database(BaseDatabase, SyncDatabase):
```
petyaslavova, line 89: Same question here. If you use single class inheritance, it is easier to follow the initialization of the object and you can add an `__init__` method in the `SyncDatabase` class that covers the initialization of the client and circuit.

redis/multidb/circuit.py
vladvildanov: Well, `SyncCircuitBreaker` is an interface and `BaseCircuitBreaker` is an abstract class that provides methods implementation for `CircuitBreaker` interface.

So the idea is to split the interfaces, so `BaseCircuitBreaker` could be used by async implementation as well. THis way we can re-use everything that could be re-used and what's not will be specified by `SyncCircuitBreaker` and `AsyncCircuitBreaker` interfaces

petyaslavova (approved):

redis/multidb/database.py
vladvildanov: Here's the same as circuit breaker, since Database for async implementation using async client under the hood, the interfaces has been separated and base class contain all re-usable code

PR #3768, opened 2025-09-11
Added support for Lag-Aware Healthcheck and OSS Cluster API

vladvildanov:
### Pull Request check-list

### Description of change


redis/asyncio/http/http_client.py
```diff
@@ -0,0 +1,216 @@
...
+        body: Optional[Union[bytes, str]] = None,
+        timeout: Optional[float] = None,
+    ) -> HttpResponse:
+        """
+        Invoke HTTP request with given method."""
+        pass
+
+class AsyncHTTPClientWrapper(AsyncHTTPClient):
+    """
+    An async wrapper around sync HTTP client with thread pool execution.
```
petyaslavova, line 99: It will be more efficient if you use an async HTTP library like aiohttp or httpx.

redis/asyncio/multidb/healthcheck.py
```diff
@@ -67,9 +71,107 @@ async def _returns_echoed_message(self, database) -> bool:
             # For a cluster checks if all nodes are healthy.
             all_nodes = database.client.get_nodes()
             for node in all_nodes:
-                actual_message = await node.redis_connection.execute_command("ECHO", "healthcheck")
+                actual_message = await node.execute_command("ECHO" ,"healthcheck")
```
petyaslavova: the space should be after the comma

redis/asyncio/http/http_client.py
vladvildanov: I agree, but the question is do we really want to introduce new dependency now to achieve that?

For now the only purpose for this HTTP client is to be used by Lag-aware health check. It doesn't requires a performance as it sends requests once per few seconds. So in my opinion this tiny wrapper works well.

I defined an interface `AsyncHttpClient` whenever we would need a real HTTP client we can implement it and change it easily without a breaking changes.

petyaslavova (approved):
LGTM.

PR #3772, opened 2025-09-18
Adding ssl_verify_flags_config argument for ssl connection configuration

### LINKED ISSUE #3685: Asking for an option to deactivate x509 strict mode for TLS connections
Hello,

We are using Google Cloud for our python jobs using Redis. When we tried to upgrade some of our code to Python 3.13 we noticed that our jobs could not access our hosted Redis instance.

This is related to the Python 3.13 change: https://docs.python.org/3/whatsnew/3.13.html#ssl

The `VERIFY_X509_STRICT` is now used by default by Python 3.13 and unfortunately the certificates generated by the cloud provider don't have that TLS x509 extension.

Currently:

- I can't recreate the server certificate which is handled by the cloud provider
- I can't find a parameter on the redis-py client to deactivate the x509 strict mode
- I'm stuck with Python 3.12

One solution would be to completely deactivate the certificate checking but this is something we can't do for security purpose.

Would it be possible to have a way to deactivate x509 strict mode inside Redis-py ?

I looked at existing parameters and didn't see something helpful.

The error message we get:

`redis.exceptions.ConnectionError: Error 1 connecting to redis-production:6378. [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Missing Authority Key Identifier (_ssl.c:1028).`

Best regards

petyaslavova:
### Pull Request check-list

### Description of change

Fixes #3685


redis/asyncio/connection.py
```diff
@@ -1021,6 +1039,34 @@ def parse_url(url: str) -> ConnectKwargs:
 
         if parsed.scheme == "rediss":
             kwargs["connection_class"] = SSLConnection
+
+            if "ssl_verify_flags_config" in kwargs:
+                # flags are passed in as a string representation of a list,
+                # e.g. [(VERIFY_X509_STRICT, False), (VERIFY_X509_PARTIAL_CHAIN, True)]
+                # To parse it sucessfully, we need transform the flags to strings with quotes.
```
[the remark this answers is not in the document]
petyaslavova: Fixed.

redis/asyncio/connection.py
```diff
@@ -1021,6 +1039,34 @@ def parse_url(url: str) -> ConnectKwargs:
...
+                verify_flags_config_str = kwargs.pop("ssl_verify_flags_config")
+                # First wrap any VERIFY_* name in quotes
+                verify_flags_config_str = re.sub(
+                    r"\b(VERIFY_[A-Z0-9_]+)\b", r'"\1"', verify_flags_config_str
+                )
+
+                # transform the string to a list of tuples - the first element of each tuple is a string containing the name of the flag,
+                # and the second is a boolean that indicates if the flad should be enabled or disabled
```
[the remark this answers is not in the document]
petyaslavova: Fixed.

redis/connection.py
```diff
@@ -1376,7 +1382,19 @@ def __init__(
...
+            ssl_verify_flags_config: A list with flags configuration to be set on the SSLContext. Defaults to None.
+                              Valid format is as follows:
+                                [
+                                    (config_flag, enabled/disabled),
+                                    ...
+                                ]
+                              Example:
+                                [
+                                    (ssl.VERIFY_X509_STRICT, False),       # disable strict
+                                    (ssl.VERIFY_X509_PARTIAL_CHAIN, True), # ensure partial chain is enabled
```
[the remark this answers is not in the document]
petyaslavova: Done.

redis/connection.py
```diff
@@ -1632,6 +1657,34 @@ def parse_url(url):
         if url.scheme == "rediss":
             kwargs["connection_class"] = SSLConnection
 
+            if "ssl_verify_flags_config" in kwargs:
+                # flags are passed in as a string representation of a list,
+                # e.g. [(VERIFY_X509_STRICT, False), (VERIFY_X509_PARTIAL_CHAIN, True)]
+                # To parse it sucessfully, we need transform the flags to strings with quotes.
```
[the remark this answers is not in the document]
petyaslavova: Done.

redis/connection.py
```diff
@@ -1632,6 +1657,34 @@ def parse_url(url):
...
+                verify_flags_config_str = kwargs.pop("ssl_verify_flags_config")
+                # First wrap any VERIFY_* name in quotes
+                verify_flags_config_str = re.sub(
+                    r"\b(VERIFY_[A-Z0-9_]+)\b", r'"\1"', verify_flags_config_str
+                )
+
+                # transform the string to a list of tuples - the first element of each tuple is a string containing the name of the flag,
+                # and the second is a boolean that indicates if the flad should be enabled or disabled
```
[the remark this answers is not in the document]
petyaslavova: Done.

redis/asyncio/connection.py
```diff
@@ -906,6 +918,12 @@ def get(self) -> SSLContext:
             context = ssl.create_default_context()
             context.check_hostname = self.check_hostname
             context.verify_mode = self.cert_reqs
+            if self.verify_flags_config:
+                for flag, enabled in self.verify_flags_config:
+                    if enabled:
+                        context.options |= flag
```
vladvildanov: Should we use `context.verify_flags` instead?

https://docs.python.org/3/library/ssl.html#context-creation

redis/asyncio/cluster.py
```diff
@@ -299,6 +300,7 @@ def __init__(
         ssl_ca_certs: Optional[str] = None,
         ssl_ca_data: Optional[str] = None,
         ssl_cert_reqs: Union[str, VerifyMode] = "required",
+        ssl_verify_flags_config: Optional[List[Tuple[VerifyFlags, bool]]] = None,
```
vladvildanov: I think it makes sense to support it as two separate parameters, something like `include_verify_flags` and `exclude_verify_flags` with  type `Optional[List[VerifyFlags]]`. URL specification allows to pass parameters as arrays, which will be converted to a list, but tuples is something that beyond the query parameters specification. So it will make it more useful if Redis URL is used.

redis/asyncio/connection.py
```diff
@@ -1021,6 +1039,34 @@ def parse_url(url: str) -> ConnectKwargs:
...
         if parsed.scheme == "rediss":
             kwargs["connection_class"] = SSLConnection
+
+            if "ssl_verify_flags_config" in kwargs:
+                # flags are passed in as a string representation of a list,
+                # e.g. [(VERIFY_X509_STRICT, False), (VERIFY_X509_PARTIAL_CHAIN, True)]
+                # To parse it successfully, we need to transform the flags to strings with quotes.
+                verify_flags_config_str = kwargs.pop("ssl_verify_flags_config")
+                # First wrap any VERIFY_* name in quotes
+                verify_flags_config_str = re.sub(
```
vladvildanov: You won't need this if you would follow my recommendation above

vladvildanov (changes requested):

redis/asyncio/connection.py
petyaslavova: You are right :) I fixed it. I really don't know how I have ended up using the options...

redis/asyncio/cluster.py
petyaslavova: Done.

redis/asyncio/connection.py
petyaslavova: Fixed.

redis/asyncio/connection.py
```diff
@@ -953,6 +978,20 @@ def to_bool(value) -> Optional[bool]:
     return bool(value)
 
 
+def parse_ssl_verify_flags(value):
+    # flags are passed in as a string representation of a list,
+    # e.g. VERIFY_X509_STRICT, VERIFY_X509_PARTIAL_CHAIN
+    verify_flags_str = value.replace("[", "").replace("]", "")
```
vladvildanov, line 984: I think even this is not needed. According to a URL query parameters spec it is possible to pass array like this:

`?foo[]=bar&foo[]=baz` and I expect that url parsing function will return you a list `foo => [bar, baz]`. Could you please check it?

vladvildanov (changes requested):

redis/asyncio/connection.py
petyaslavova: @vladvildanov this almost works. The syntax is a little different: `?foo=bar&foo=baz`, but then in our code we break the received list and get just the first element. This should be fixed, but I don't want to mix this new functionality with the parsing fix.

PR #3779, opened 2025-09-26
Adding 'auto' option to MaintNotificationsConfig.enabled

petyaslavova:
### Pull Request check-list

### Description of change

disabled: The client doesn’t send the CLIENT MAINT_NOTIFICATIONS ON ... command as part of the handshake. The feature isn’t used by the client. 

enabled: The client forcefully sends the CLIENT MAINT_NOTIFICATIONS ON ... command to the server. If the server responds with an error, the connection is interrupted and an error is raised.

auto: The client tries to send the CLIENT MAINT_NOTIFICATIONS ON ... command to the server. If the server returns an error, the client should continue without the push notifications.


redis/connection.py
```diff
@@ -709,15 +713,21 @@ def on_connect_check_health(self, check_health: bool = True):
...
+                ):
+                    # Log warning but don't fail the connection
+                    import logging
 
-                logger = logging.getLogger(__name__)
-                logger.warning(f"Failed to enable maintenance notifications: {e}")
+                    logger = logging.getLogger(__name__)
+                    logger.warning(f"Failed to enable maintenance notifications: {e}")
+                else:
+                    raise
```
vladvildanov, line 732: Previously the exception wasn't raised, is there's a reason why it should be raised now?

redis/maint_notifications.py
```diff
@@ -447,7 +447,7 @@ class MaintNotificationsConfig:
 
     def __init__(
         self,
-        enabled: bool = True,
+        enabled: Union[bool, Literal["auto"]] = "auto",
```
vladvildanov, line 450: I would suggest to use Enum instead. If there's 3 possible variants now, most likely 4th would appear somehow. Defining Enum now prevent us from type changing in a future

petyaslavova: I started doing it with Enum, but.. I have submitted the final version of the docs example for client initialization and configuration and the `auto` there is a string and the property is `enabled`.

redis/connection.py
petyaslavova: The current behavior is aligned with the requirements outlined in the specification. I think the initial handling have been done before this exact detail has changed or there has been some misunderstanding about redis-py being excluded from this requirement.

redis/maint_notifications.py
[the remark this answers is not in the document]
petyaslavova: Fixed.

vladvildanov (approved):

PR #3785, opened 2025-10-02
Reorganizing the code related to maintenance push notifications.

petyaslavova:
### Pull Request check-list

### Description of change

Extracting maintenance push notifications logic into separate abstract classes. Using multiple inheritance for `Connection` and `ConnectionPool` classes to expose the maintenance push notifications for the existing connections.

### DIFF
--- a/redis/commands/search/__init__.py
+++ b/redis/commands/search/__init__.py
@@ -1,4 +1,4 @@
-import redis
+from redis.client import Pipeline as RedisPipeline
 
 from ...asyncio.client import Pipeline as AsyncioPipeline
 from .commands import (
@@ -181,7 +181,7 @@ class AsyncSearch(Search, AsyncSearchCommands):
         return p
 
 
-class Pipeline(SearchCommands, redis.client.Pipeline):
+class Pipeline(SearchCommands, RedisPipeline):
     """Pipeline for the module."""
 
 
--- a/redis/commands/search/aggregation.py
+++ b/redis/commands/search/aggregation.py
@@ -1,4 +1,4 @@
-from typing import List, Union
+from typing import List, Optional, Tuple, Union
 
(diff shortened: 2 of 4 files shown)


redis/connection.py
```diff
@@ -1328,6 +1513,119 @@ def pack_commands(self, commands):
...
+            raise NotImplementedError(
+                "Maintenance notifications are not supported by this connection type"
+            )
+
+    @property
+    def maintenance_state(self) -> MaintenanceState:
+        if isinstance(self._conn, MaintNotificationsAbstractConnection):
+            return self._conn.maintenance_state
+        else:
+            raise NotImplementedError(
```
vladvildanov, line 1585: Let's extract this part as it used in multiple places

redis/maint_notifications.py
```diff
@@ -501,7 +500,7 @@ def is_relaxed_timeouts_enabled(self) -> bool:
         return self.relaxed_timeout != -1
 
     def get_endpoint_type(
-        self, host: str, connection: "ConnectionInterface"
+        self, host: str, connection: "MaintNotificationsAbstractConnection"
```
vladvildanov, line 503: Needs to make sure to release as a part of 7.0.0, otherwise breaking change

vladvildanov (dismissed):

redis/maint_notifications.py
petyaslavova: Yeah, I know - that's why I've been working on this with the highest priority the last days :)

redis/connection.py
petyaslavova: Done. Extracted those in a helper method that will raise the error if the connection is not a MaintNotificationsAbstractConnection and will return the provided object in case it is - I'm doing this so that I can somehow "show" the static code analysis tools that this object is an instance of the correct type and those attributes are available and errors should not be reported.

PR #3788, opened 2025-10-03
Fix automation scenarion maint notification

kiryazovi-redis:
The current scenario tests have 3 issues which we have to fix for our nightly CI to work.
1. they would not work so well with a cluster that has multiple dbs, so I fixed the logic to choose: 
    Master node of our target db
    Any node without shards of the target db
2. They require more logs (This part of the PR can be removed, depending on comments)
3. They take too long due to clean-up. The clean-up can be mostly removed besides the failover one. Because migrate and bind are operations that should take the cluster into usable for next scenario tests. The only exception is migrate only tests. 
(a notification is sent only when we did mgirate + bind, because the proxy is pointing towards "empty" node)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,12 @@ filterwarnings = [
     # Ignore a coverage warning when COVERAGE_CORE=sysmon for Pythons < 3.12.
     "ignore:sys.monitoring isn't available:coverage.exceptions.CoverageWarning",
 ]
+log_cli_level = "DEBUG"
+log_cli_date_format = "%H:%M:%S:%f"
+log_cli = true
+log_cli_format = "%(asctime)s %(levelname)s %(threadName)s: %(message)s"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -80,6 +80,12 @@ filterwarnings = [
     # Ignore a coverage warning when COVERAGE_CORE=sysmon for Pythons < 3.12.
     "ignore:sys.monitoring isn't available:coverage.exceptions.CoverageWarning",
 ]
+log_cli_level = "DEBUG"
```
petyaslavova: This config will affect all other tests as well and the log from the standard cicd actions will become unreadable. 
The default level should be something like `INFO` - usually debug is enabled only for troubleshooting runs.
And the default value of log_cli should be set to false - you can enable it just for the pytest command used to start the scenario tests using `--log-cli`.

PR #3789, opened 2025-10-03
Update ping command docstrings and method return type hint.

### LINKED ISSUE #3714: Documentation of ping() is not clear
*The issue refers to: https://redis.readthedocs.io/en/stable/commands.html#redis.commands.core.CoreCommands.ping*
***
The documentation of the ping() method is unclear or confusing. According to the documentation, the return value is:
```python
Union[Awaitable, Any]
```

In practice, however, the callback logic always returns a bool by default. See:
- https://github.com/redis/redis-py/blob/64f291f7797b5b3e5652da8941947b8c2db517e9/redis/_parsers/helpers.py#L734

I suggest changing the return-type in the documentation to:
```python
Union[Awaitable, bool]
```

It would also be helpful if the mode of operation were also mentioned in the description. Because the current text only refers to the redis.io PING documentation, which says that the return value is a string with the value “PONG”:

> Ping the Redis server
>
> For more information see https://redis.io/commands/ping

I would suggest something like:

> Ping the Redis server
> 
> Send `PING`<sup>1</sup> command to server and returns boolean `True` if the server's response equals "PONG"<sup>2</sup>.
>
> 
> <sup>1</sup>For more information on the underlying ping command see https://redis.io/commands/ping
> <sup>2</sup>This default behaviour can be overwritten by [setting a custom callback](https://redis.readthedocs.io/en/stable/connections.html#redis.Redis.set_response_callback).

**Note:** *This is a follow-up of https://github.com/redis/redis-doc/issues/2717 which wasn't resolved yet, but may not be handled any more as the repository is marked as obsolete.*

petyaslavova:
### Pull Request check-list

### Description of change

Fixes #3714

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -1161,9 +1161,12 @@ class PubSub:
             return await self.handle_message(response, ignore_subscribe_messages)
         return None
 
-    def ping(self, message=None) -> Awaitable:
+    def ping(self, message=None) -> Awaitable[bool]:
         """
-        Ping the Redis server
+        Ping the Redis server to test connectivity.
+
+        Sends a PING command to the Redis server and returns True if the server
+        responds with "PONG".
         """
         args = ["PING", message] if message is not None else ["PING"]
         return self.execute_command(*args)
--- a/redis/client.py
+++ b/redis/client.py
@@ -1186,7 +1186,10 @@ class PubSub:
 
     def ping(self, message: Union[str, None] = None) -> bool:
         """
-        Ping the Redis server
+        Ping the Redis server to test connectivity.
+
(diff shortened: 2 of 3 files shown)


zealcc:
@petyaslavova Incompatible types in "await" (actual type "Awaitable[bool] | bool", expected type "Awaitable[Any]")mypy[misc](https://mypy.readthedocs.io/en/stable/_refs.html#code-misc)

```
from redis.asyncio import ConnectionPool, Redis


async def main():
    pool = ConnectionPool.from_url("redis://localhost/1")
    redis = Redis(connection_pool=pool)
    await redis.ping()
```

PR #3790, opened 2025-10-03
Add redis 8.4-M01-pre to tested redis versions.

petyaslavova:
### Pull Request check-list

### Description of change

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -74,7 +74,7 @@ jobs:
       max-parallel: 15
       fail-fast: false
       matrix:
-        redis-version: ['8.2.1-pre', '${{ needs.redis_version.outputs.CURRENT }}', '8.0.2' ,'7.4.4', '7.2.9']
+        redis-version: ['8.2.2-pre', '${{ needs.redis_version.outputs.CURRENT }}', '8.0.2' ,'7.4.4', '7.2.9']
         python-version: ['3.9', '3.13']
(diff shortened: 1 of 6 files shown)


.github/workflows/integration.yaml
```diff
@@ -74,7 +76,7 @@ jobs:
       max-parallel: 15
       fail-fast: false
       matrix:
-        redis-version: ['8.2.1-pre', '${{ needs.redis_version.outputs.CURRENT }}', '8.0.2' ,'7.4.4', '7.2.9']
+        redis-version: ['8.4-M01-pre', '${{ needs.redis_version.outputs.CURRENT }}', '8.0.2' ,'7.4.4', '7.2.9']
```
vladvildanov, line 79: I think we shouldn't change 8.2 to 8.4 and instead support both. 8.4 is another version to support so instead we have to remove 7.4 and 7.2 as soon as they EOL

petyaslavova: I found that into the image version that is marked as CURRENT, once a patch is released, the base version image gets updated/rebuilt. For example, when 8.2.1 has been released, the 8.2 image gets rebuilt to contain Redis server with version 8.2.1. So now when 8.2.2 is release the used CURRENT(which is 8.2) will automatically start using it for the tests.
I was wondering if it is time to remove at least 7.2, but decided to add this removal  in an additional PR after discussing this internally ...

PR #3799, opened 2025-10-16
fix: improve Lock.extend() and Lock.do_extend() return type to Literal[True]

### LINKED ISSUE #3792: Typing: `Lock.extend` and `Lock.do_extend` could return `Literal[True]`
Hi there! I ran into this recently: both the sync and async `Lock` APIs have `extend` and `do_extend` typed as returning `bool`, but in practice only `True` is returned.

For example, here:

https://github.com/redis/redis-py/blob/b96d29c40ffa96e75330a933023f2f8eacd5ce64/redis/lock.py#L287-L317

Instead of `False`, these APIs use exceptions to communicate failure states.

The current return type can lead to developer confusion, e.g. developers mistakenly doing this:

```python
if not lock.extend(...):
    # handle lock extension failure
```

...which doesn't work as expected, since lock failure causes a `raise LockError` instead.

## Proposed solution

I think the simplest solution here would be to refine the return types on these APIs: instead of `-> bool` these APIs could be annotated with `-> typing.Literal[True]` to communicate clearly to developers that there's no meaningful conditional check on the return value.

## Alternatives

An alternative to the proposed solution above would be to document in `redis-py`'s docs that the return type of `bool` is really just the `True` value, and that users must handle errors solely through exception handling. I think this would be fine, but that it would be a lot clearer to have this encoded at the type layer additionally, per above 🙂

enioxt:
## Description

Fixes #3792

This PR improves the type safety of  and  methods by changing their return type from  to .

## Problem

Currently, both methods are typed as returning , but in practice they only return  or raise a  exception. This can lead to developer confusion, as developers might write code like:

```python
if not lock.extend(...):
    # This code will never run\!
    # Lock extension failures raise LockError instead
```

## Solution

Changed return types to  to clearly communicate that:
- These methods only return  on success
- Failures are communicated via  exception
- There's no meaningful conditional check on the return value

## Changes

**Sync version** (`redis/lock.py`):
- `extend()`: `bool` → `Literal[True]`
- `do_extend()`: `bool` → `Literal[True]`

**Async version** (`redis/asyncio/lock.py`):
- `extend()`: `Awaitable[bool]` → `Awaitable[Literal[True]]`
- `do_extend()`: `bool` → `Literal[True]`

## Testing

- Type-only change, no runtime behavior modified
- Existing tests continue to pass
- Improves type checking for users of these APIs

## Benefits

- **Better type safety**: IDEs and type checkers can warn about pointless conditionals
- **Clearer API**: Developers immediately understand the exception-based error model
- **Prevents bugs**: Reduces likelihood of incorrect error handling code

**Sacred Code**: 000.111.369.963.1618

### DIFF
--- a/redis/asyncio/lock.py
+++ b/redis/asyncio/lock.py
@@ -3,7 +3,7 @@ import logging
 import threading
 import uuid
 from types import SimpleNamespace
-from typing import TYPE_CHECKING, Awaitable, Optional, Union
+from typing import TYPE_CHECKING, Awaitable, Literal, Optional, Union
 
 from redis.exceptions import LockError, LockNotOwnedError
 from redis.typing import Number
@@ -284,7 +284,7 @@ class Lock:
 
     def extend(
         self, additional_time: Number, replace_ttl: bool = False
-    ) -> Awaitable[bool]:
+    ) -> Awaitable[Literal[True]]:
         """
         Adds more time to an already acquired lock.
 
@@ -301,7 +301,7 @@ class Lock:
             raise LockError("Cannot extend a lock with no timeout")
         return self.do_extend(additional_time, replace_ttl)
 
-    async def do_extend(self, additional_time, replace_ttl) -> bool:
+    async def do_extend(self, additional_time, replace_ttl) -> Literal[True]:
         additional_time = int(additional_time * 1000)
         if not bool(
             await self.lua_extend(
--- a/redis/lock.py
+++ b/redis/lock.py
@@ -3,7 +3,7 @@ import threading
 import time as mod_time
 import uuid
 from types import SimpleNamespace, TracebackType
-from typing import Optional, Type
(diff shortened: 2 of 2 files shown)


petyaslavova:
Hi @enioxt, thank you for your contribution! We will review your changes soon.

petyaslavova:
Hi @enioxt, the changes look good! 
I have just one request - to add in the docstrings info about the behavior. It will be a lot easier for the developers to gain orientation for the methods behavior.

       ```
        Returns:
            Literal[True]
                Short description when this value is returned....

        Raises:
            <Error Type>
                Short description when this error is raised
```

petyaslavova:
Hey @enioxt, would you be willing to reintroduce this change? We’re already planning to include breaking changes in the upcoming major release, so this might be a good opportunity to bring it back in.

PR #3803, opened 2025-10-17
Improve NodesManager locking

praboud:
### Pull Request check-list

### Description of change
`NodesManager` can mutate state (ie: nodes cache + slots cache + startup nodes + default node) from multiple client threads, through both re-initializing from `CLUSTER SLOTS`, and from following a `MOVED`/`ASK` redirect. Right now, there isn't proper synchronization of state across multiple threads, which can result in the state getting corrupted, or the NodesManager otherwise behaving weirdly:
1. `update_moved_exception` just sets an exception on the `NodesManager`, which we expect to trigger an update to the state the next time we fetch a node from the `NodesManager`. But `_moved_exception` isn't synchronized. Suppose two threads A & B sequence like: A calls update_moved_exception, B calls update_moved_exception, A calls get_node_from_slot, B calls get_node_from_slot. A's update to `_moved_exception` gets lost, and when A calls `get_node_from_slot`, it doesn't actually follow the redirect. To avoid this problem, I've changed the slot move logic to immediately apply the update to the slot state, rather than queueing it up for later.
2. `_get_or_create_cluster_node` can mutate the `role` of a `ClusterNode`, but the node is referenced from the `slots_cache`. Because we expect `slots_cache[node][0]` to always be the primary, this can cause strange behavior for readers of `slots_cache` between the time `_get_or_create_cluster_node` is called, and when `initialize` resets `slots_cache` at the end of the update.
3. `initialize` & `_update_moved_slots` both mutate `slots_cache`, and aren't synchronized with each other. This can cause the slots cache to get into a weird state, where eg: nodes are deleted from the slots cache, or duplicated.
4. `initialize` allows multiple callers to initialize concurrently, which is both extra load on the cluster, and can cause strange behavior in corner cases.

To fix all of these:
1. I've ensured that we hold `self._lock` around all places where any mutation happens in `NodesManager`.
2. I've replaced `update_moved_exception` & `_update_moved_slots` with just `move_slot`, to avoid racing multiple slot updates.
3. I've added `_initialize_lock` to serialize / deduplicate calls to `initialize`.

I've added some tests to try to exercise the situations above, and verified that they fail before / pass after this change.

This PR mainly focuses on the sync Redis client, but I've tried to update the asyncio one as well. It doesn't (as far as I can tell) suffer from most of these issues, other than issue 4, so the changes there are a bit lighter. This PR is pretty hefty already - I'm happy to split the asyncio changes out into a separate PR if that's easier for review.

Also, in general, this PR is a lot easier to review with "ignore whitespace" turned on in the diff options, because a lot of these changes mean indenting big blocks of code inside a `with self._lock` block.

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -1928,7 +1928,10 @@ class NodesManager:
                     # bother running again
                     return
 
-            for startup_node in tuple(self.startup_nodes.values()):
+            with self._lock:
+                startup_nodes = tuple(self.startup_nodes.values())
+
+            for startup_node in startup_nodes:
                 try:
                     if startup_node.redis_connection:
                         r = startup_node.redis_connection
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -3031,13 +3031,6 @@ class TestNodesManager:
 
             execute_command_mock.side_effect = execute_command
 
-            r = get_mocked_redis_client(
-                host=default_host,
-                port=default_port,
-                cluster_enabled=True,
-            )
-            nm = r.nodes_manager
-
             nm = NodesManager(
                 startup_nodes=[ClusterNode(host=default_host, port=default_port)],
                 from_url=False,
@@ -3109,7 +3102,7 @@ class TestNodesManager:
             cluster_enabled=True,
         )
         nm = r.nodes_manager
-        # Move slots 0-1000 to 127.0.0.1 in concurrent threads
+        # Move slots 0-999 to 127.0.0.1 in concurrent threads
         num_threads = 20
         slots_per_thread = 50  # 1000 slots / 20 threads = 50 slots per thread
         errors: list[Exception] = []
@@ -3140,7 +3133,7 @@ class TestNodesManager:
         # Check that no errors occurred
         assert len(errors) == 0, f"Errors occurred: {errors}"
 
-        # Verify that all slots 0-1000 are moved to 127.0.0.1:7000
+        # Verify that all slots 0-999 are moved to 127.0.0.1:7001
         for slot_id in range(num_threads * slots_per_thread):
             assert slot_id in nm.slots_cache, f"Slot {slot_id} missing"
             slot_nodes = nm.slots_cache[slot_id]


redis/cluster.py
```diff
@@ -1393,14 +1404,6 @@ def __repr__(self):
...
         return isinstance(obj, ClusterNode) and obj.name == self.name
 
-    def __del__(self):
-        try:
-            if self.redis_connection is not None:
-                self.redis_connection.close()
-        except Exception:
-            # Ignore errors when closing the connection
-            pass
-
```
praboud, line 1568: This causes issues with the change I made to `_get_or_create_cluster_node `, to make a new `ClusterNode` (sharing the `redis_connection` of the old node), instead of mutating the old one in place.

`Redis.__del__` already closes the connection pool down when the object is dropped, so I think this code was always unnecessary anyhow.

redis/cluster.py
```diff
@@ -1562,66 +1577,56 @@ def _update_moved_slots(self):
...
         :param server_type: 'primary' or 'replica'
         :return: list of ClusterNode
         """
-        return [
-            node
-            for node in self.nodes_cache.values()
-            if node.server_type == server_type
-        ]
-
-    def populate_startup_nodes(self, nodes):
```
praboud, line 1787: This was unused, and I don't _think_ this is considered part of the public interface? We could also just protect this with `self._lock`, but it's easier to delete it if we don't need it.

redis/cluster.py
```diff
@@ -1713,135 +1728,144 @@ def initialize(self):
...
+                if epoch != self._epoch:
+                    # another thread has already re-initialized the nodes; don't
+                    # bother running again
+                    return
 
-                target_node = self._get_or_create_cluster_node(
-                    host, port, PRIMARY, tmp_nodes_cache
+                startup_nodes = random.sample(
+                    list(self.startup_nodes.values()), k=len(self.startup_nodes)
                 )
```
praboud: I can pull this out to a separate PR if preferred, but another thing that I've observed is that all the redis clients init from the same node, which can hammer that node if you have a lot of clients. Randomizing the order should allow this to scale better.

petyaslavova:
Hi @praboud, thank you for your contribution! We will review it soon.

praboud:
@petyaslavova hey, just checking in on when you'll have a chance to review this PR.

petyaslavova:
> @petyaslavova hey, just checking in on when you'll have a chance to review this PR.

Hi @praboud, my todo list for the next several weeks is quite full, and I'll probably be able to get to your several PRs at the beginning of January.

redis/asyncio/cluster.py
```diff
@@ -1522,135 +1515,147 @@ async def initialize(self) -> None:
...
+                # [[slot_section[from_slot,to_slot,master,replica1,...,replicaN]]]
+                # where each node contains the following list: [IP, port, node_id]
+                # Therefore, cluster_slots[0][2][0] will be the IP address of the
+                # primary node of the first slot section.
+                # If there's only one server in the cluster, its ``host`` is ''
+                # Fix it to the host in startup_nodes
+                if (
+                    len(cluster_slots) == 1
+                    and not cluster_slots[0][2][0]
+                    and len(self.startup_nodes) == 1
```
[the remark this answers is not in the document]
petyaslavova: This is old code and is out of the scope of this PR.

redis/cluster.py
```diff
@@ -1631,25 +1635,33 @@ def __init__(
...
     ):
-        self.nodes_cache: Dict[str, Redis] = {}
-        self.slots_cache = {}
-        self.startup_nodes = {}
-        self.default_node = None
-        self.populate_startup_nodes(startup_nodes)
+        self.nodes_cache: dict[str, ClusterNode] = {}
+        self.slots_cache: dict[int, list[ClusterNode]] = {}
+        self.startup_nodes: dict[str, ClusterNode] = {n.name: n for n in startup_nodes}
+        self.default_node: ClusterNode | None = None
```
petyaslavova: Please use Union and Optional in type hints - it is used everywhere in the code and we better have some alignment in the used syntax.

redis/cluster.py
```diff
@@ -1631,25 +1635,33 @@ def __init__(
...
         self.from_url = from_url
         self._require_full_coverage = require_full_coverage
         self._dynamic_startup_nodes = dynamic_startup_nodes
         self.connection_pool_class = connection_pool_class
         self.address_remap = address_remap
-        self._cache = cache
-        self._cache_config = cache_config
-        self._cache_factory = cache_factory
-        self._moved_exception = None
+        self._cache: CacheInterface | None = None
```
petyaslavova: Optional[CacheInterface]

redis/cluster.py
```diff
@@ -1598,30 +1601,31 @@ def get_server_index(
...
 
 
 class NodesManager:
     def __init__(
         self,
-        startup_nodes,
+        startup_nodes: list[ClusterNode],
         from_url=False,
         require_full_coverage=False,
-        lock=None,
```
petyaslavova, line 1624: This is a public api exposed, so we can't just remove this argument in case somebody is using it. 
Even if we include this change in a new major release where breaking changes are allowed, we still first need to deprecate it first.

redis/cluster.py
```diff
@@ -1631,25 +1635,33 @@ def __init__(
...
-        if lock is None:
-            lock = threading.RLock()
-        self._lock = lock
+
+        # nodes_cache / slots_cache / startup_nodes / default_node are protected by _lock
+        self._lock: threading.RLock = threading.RLock()
+        # initialize holds _initialization_lock to dedup multiple calls to reinitialize;
+        # note that if we hold both _lock and _initialization_lock, we _must_ acquire
+        # _initialization_lock first (ie: to have a consistent order) to avoid deadlock.
+        self._initialization_lock: threading.Lock = threading.Lock()
```
petyaslavova: There is a new feature in development right now, that might lead to the case where we try to get that lock more than once within the same thread(not 100% sure, but from what I've seen while debugging, it is a possible corner case), so this lock should also use RLock, otherwise we might end up in a deadlocked state.

redis/cluster.py
```diff
@@ -1659,7 +1671,12 @@ def __init__(
         )
         self.initialize()
 
-    def get_node(self, host=None, port=None, node_name=None):
+    def get_node(
+        self,
+        host: str | None = None,
```
petyaslavova: Please use Optional

redis/cluster.py
```diff
@@ -1669,53 +1686,50 @@ def get_node(self, host=None, port=None, node_name=None):
...
+            with self._lock:
+                return self.nodes_cache.get(node_name)
         else:
             return None
 
-    def update_moved_exception(self, exception):
-        self._moved_exception = exception
-
-    def _update_moved_slots(self):
+    def move_slot(self, e: AskError | MovedError):
```
petyaslavova: Please use Union

redis/cluster.py
```diff
@@ -1727,66 +1741,56 @@ def _update_moved_slots(self):
     )
     def get_node_from_slot(
         self,
-        slot,
-        read_from_replicas=False,
-        load_balancing_strategy=None,
-        server_type=None,
+        slot: int,
+        read_from_replicas: bool = False,
+        load_balancing_strategy: LoadBalancingStrategy | None = None,
```
petyaslavova: Please use Optional

redis/cluster.py
```diff
@@ -1727,66 +1741,56 @@ def _update_moved_slots(self):
...
-        else:
-            # return a replica
-            # randomly choose one of the replicas
-            node_idx = random.randint(1, len(self.slots_cache[slot]) - 1)
+        with self._lock:
+            if self.slots_cache.get(slot) is None or len(self.slots_cache[slot]) == 0:
+                raise SlotNotCoveredError(
+                    f'Slot "{slot}" not covered by the cluster. '
+                    + f'"require_full_coverage={self._require_full_coverage}"'
+                )
```
petyaslavova, line 1770: Can you please add one empty line between the check that raises an error and the main logic?

redis/cluster.py
petyaslavova: Unless it is named with `_` prefix, it should be considered as part of the public interface, so the strategy with the added lock protection will be better. Also, since we are no longer using it internally, you can add an annotation that the method is deprecated.

redis/cluster.py
petyaslavova: I would prefer to leave that one out of this PR for now - let's add this separately.

petyaslavova: I'm not sure it is needed actually... In this PR you protect the current method to be executed just once, so in the case of many threads - this is already fixed - will run once.

redis/cluster.py
```diff
@@ -2032,15 +2055,17 @@ def remap_host_port(self, host: str, port: int) -> Tuple[str, int]:
             return self.address_remap((host, port))
         return host, port
 
-    def find_connection_owner(self, connection: Connection) -> Optional[Redis]:
+    def find_connection_owner(self, connection: Connection) -> ClusterNode | None:
```
petyaslavova: Please use Optional

redis/cluster.py
```diff
@@ -1883,135 +1897,144 @@ def initialize(self):
...
+            for startup_node in startup_nodes:
+                try:
+                    if startup_node.redis_connection:
+                        r = startup_node.redis_connection
+                    else:
+                        # Create a new Redis connection
+                        r = self.create_redis_node(
+                            startup_node.host, startup_node.port, **kwargs
+                        )
+                        self.startup_nodes[startup_node.name].redis_connection = r
```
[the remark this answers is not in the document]
petyaslavova: For the sync client startup_nodes are updated only inside `initialize` method, which is already protected by a lock

redis/cluster.py
```diff
@@ -1883,135 +1897,144 @@ def initialize(self):
...
-                else:
-                    # Create a new Redis connection
-                    r = self.create_redis_node(
-                        startup_node.host, startup_node.port, **kwargs
-                    )
-                    self.startup_nodes[startup_node.name].redis_connection = r
-                # Make sure cluster mode is enabled on this node
-                try:
-                    cluster_slots = str_if_bytes(r.execute_command("CLUSTER SLOTS"))
-                    r.connection_pool.disconnect()
```
petyaslavova, line 1904: You remove this one because it will be automatically closed during garbage collection, or there is some other reason?

redis/cluster.py
```diff
@@ -1883,135 +1897,144 @@ def initialize(self):
...
+                # [[slot_section[from_slot,to_slot,master,replica1,...,replicaN]]]
+                # where each node contains the following list: [IP, port, node_id]
+                # Therefore, cluster_slots[0][2][0] will be the IP address of the
+                # primary node of the first slot section.
+                # If there's only one server in the cluster, its ``host`` is ''
+                # Fix it to the host in startup_nodes
+                if (
+                    len(cluster_slots) == 1
+                    and len(cluster_slots[0][2][0]) == 0
+                    and len(self.startup_nodes) == 1
```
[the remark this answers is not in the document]
petyaslavova: the same as above.

petyaslavova:
Hi @praboud, this is a great addition to the library—thank you!

I’ve reviewed your changes and have a few minor requests. Once those are addressed, I think this PR will significantly improve performance when the cluster needs to be re-initialized.

Thanks again for the time and effort you’ve put into this work!

redis/cluster.py
petyaslavova: changed.

redis/cluster.py
petyaslavova: This one should stay for now. Will be changed to trigger connections disconnects without directly killing connections that might have in-flight commands in another PR.

redis/cluster.py
```diff
@@ -1886,135 +1919,139 @@ def initialize(self):
...
+            for startup_node in tuple(self.startup_nodes.values()):
+                try:
+                    if startup_node.redis_connection:
+                        r = startup_node.redis_connection
+                    else:
+                        # Create a new Redis connection
+                        r = self.create_redis_node(
+                            startup_node.host, startup_node.port, **kwargs
+                        )
+                        self.startup_nodes[startup_node.name].redis_connection = r
```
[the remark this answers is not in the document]
petyaslavova: This review comment is already posted and replied once.

redis/cluster.py
```diff
@@ -1886,135 +1919,139 @@ def initialize(self):
...
+                    # another thread has already re-initialized the nodes; don't
+                    # bother running again
+                    return
 
-                target_node = self._get_or_create_cluster_node(
-                    host, port, PRIMARY, tmp_nodes_cache
-                )
-                nodes_for_slot.append(target_node)
+            for startup_node in tuple(self.startup_nodes.values()):
+                try:
```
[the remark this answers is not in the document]
petyaslavova: done

tests/test_cluster.py
```diff
@@ -3009,6 +3009,223 @@ def test_allow_custom_queue_class(self, queue_class):
...
+        ):
+
+            def execute_command(*_args, **_kwargs):
+                if _args[0] == "CLUSTER SLOTS":
+                    # Track how many times we actually fetch cluster slots
+                    initialization_count["count"] += 1
+                    return default_cluster_slots
+                else:
+                    return execute_command_mock(*_args, **_kwargs)
+
```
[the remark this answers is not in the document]
petyaslavova: the test is working without issues, so I'll leave it this way

tests/test_cluster.py
```diff
@@ -3009,6 +3009,223 @@ def test_allow_custom_queue_class(self, queue_class):
...
+                cluster_enabled=True,
+            )
+            nm = r.nodes_manager
+
+            def execute_command(*_args, **_kwargs):
+                if _args[0] == "CLUSTER SLOTS":
+                    return default_cluster_slots
+                else:
+                    return execute_command_mock(*_args, **_kwargs)
+
```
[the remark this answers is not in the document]
petyaslavova: same applies as for the similar comment above - won't change this one.

PR #3804, opened 2025-10-18
Fix issues with ClusterPipeline connection management 

praboud:
### Pull Request check-list

### Description of change
ClusterPipeline doesn't correctly handle returning connections to the connection pools.
1. It's possible to leak connections if an error other than ConnectionError or TimeoutError is thrown when establishing connections. We only catch those specific errors (and the try/catch doesn't wrap the entire area where we're getting the connections); if others are thrown, the exception is simply raised to the caller, and the connections are never returned to their respective pools.
2. More problematically, it's possible to return dirty connections to the pool if an error is thrown after at least one of the connections has been written to, and before all of the connections were read from. This can cause pretty bad correctness issues, where the client gets a response intended for a different request.

This diff addresses these issues by wrapping the entire area in a try/catch with a finally that ensures the connections are released, but which closes the connections if we have read but not written to the connection.

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -3511,6 +3511,7 @@ class PipelineStrategy(AbstractStrategy):
                     try:
                         connection = get_connection(redis_node)
                     except (ConnectionError, TimeoutError):
+                        # Release any connections we've already acquired before clearing nodes
                         for n in nodes.values():
                             n.connection_pool.release(n.connection)
                         # Connection retries are being handled in the node's
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -3906,14 +3906,21 @@ class TestClusterPipeline:
 
         # Patch Connection.disconnect so we can assert that at least one
         # connection was disconnected when the write error occurred.
-        with patch.object(Connection, "disconnect", wraps=Connection.disconnect) as mock_disconnect:
+        original_disconnect = Connection.disconnect
+        disconnect_called = []
+
+        def track_disconnect(self, *args):
+            disconnect_called.append(True)
+            return original_disconnect(self, *args)
+
+        with patch.object(Connection, "disconnect", track_disconnect):
             with patch.object(redis.cluster.NodeCommands, "write", mock_write):
                 with pytest.raises(RuntimeError):
                     r.pipeline().get("a").get("b").execute()
 
             # Ensure that at least one connection was disconnected as part of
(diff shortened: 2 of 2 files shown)


redis/cluster.py
```diff
@@ -2772,13 +2774,15 @@ def _send_cluster_commands(
...
-        nodes = {}
+        nodes: dict[str, NodeCommands] = {}
+        nodes_written = 0
+        nodes_read = 0
 
-        # as we move through each command that still needs to be processed,
-        # we figure out the slot number that command maps to, then from
-        # the slot determine the node.
-        for c in attempt:
-            while True:
```
praboud, line 3434: Unless I've completely lost the plot, this used to be a `while True` loop which always `break`s on the first iteration, so this does ~nothing.

petyaslavova:
Hi @praboud, thank you for your contribution! We will review your changes soon.

praboud:
@petyaslavova hey, just checking in on when you'll have a chance to review this PR.

praboud:
@petyaslavova this PR addresses a fairly serious correctness issue with the Pipeline implementation, which can result in eg: one request getting the data from another request. If those two requests are eg: handling different users' data, this could be a really problematic security or privacy issue. I'm currently using a forked version of redis-py with this fix patched in, but I would love to get this upstreamed so that others don't run into the same problem. Do you have a sense of when it would be possible to get this reviewed? (I know there's a merge conflict - if you're +1 on the general approach in this PR, I can do a pass to resolve the conflict.)

praboud:
Sorry, I just saw your comment on https://github.com/redis/redis-py/pull/3803#issuecomment-3615767994 re: timing. Thanks for the update there, though I am a little worried that other folks using this library are unknowingly having correctness problems with the bug that this PR addresses .

redis/cluster.py
```diff
@@ -3064,50 +3068,48 @@ def _send_cluster_commands(
                         self._nodes_manager.initialize()
                         if is_default_node:
                             self._pipe.replace_default_node()
+                        nodes = {}
```
petyaslavova, line 3522: Shouldn't this be done in the finally block?

redis/cluster.py
petyaslavova, line 3516: I think you can remove the releasing of the connections from here - they will be released in the finally block.

petyaslavova (approved):

PR #3813, opened 2025-10-24
Adding support for HYBRID search.

petyaslavova:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/commands/search/hybrid_query.py
+++ b/redis/commands/search/hybrid_query.py
@@ -228,18 +228,21 @@ class HybridPostProcessingConfig:
         """
         Create a new hybrid post processing configuration object.
         """
-        self._load_fields = []
-        self._groupby = []
-        self._apply = []
+        self._load_statements = []
+        self._apply_statements = []
+        self._groupby_statements = []
         self._sortby_fields = []
         self._filter = None
         self._limit = None
 
     def load(self, *fields: str) -> Self:
         """
-        Add load parameters to the query.
+        Add load statement parameters to the query.
         """
-        self._load_fields = fields
+        if fields:
+            fields_str = " ".join(fields)
+            fields_list = fields_str.split(" ")
+            self._load_statements.extend(("LOAD", len(fields_list), *fields_list))
         return self
 
     def group_by(self, fields: List[str], *reducers: Reducer) -> Self:
@@ -262,7 +265,7 @@ class HybridPostProcessingConfig:
             if reducer._alias is not None:
                 ret.extend(("AS", reducer._alias))
 
-        self._groupby.extend(ret)
+        self._groupby_statements.extend(ret)
         return self
 
     def apply(self, **kwexpr) -> Self:
@@ -274,11 +277,14 @@ class HybridPostProcessingConfig:
                 the alias for the projection, and the value is the projection
                 expression itself, for example `apply(square_root="sqrt(@foo)")`.
         """
+        apply_args = []
         for alias, expr in kwexpr.items():
             ret = ["APPLY", expr]
             if alias is not None:
                 ret.extend(("AS", alias))
-            self._apply.extend(ret)
+            apply_args.extend(ret)
+
+        self._apply_statements.extend(apply_args)
 
         return self
 
@@ -310,14 +316,12 @@ class HybridPostProcessingConfig:
 
     def build_args(self) -> List[str]:
         args = []
-        if self._load_fields:
-            fields_str = " ".join(self._load_fields)
-            fields = fields_str.split(" ")
-            args.extend(("LOAD", len(fields), *fields))
-        if self._groupby:
-            args.extend(self._groupby)
-        if self._apply:
-            args.extend(self._apply)
+        if self._load_statements:
+            args.extend(self._load_statements)
+        if self._groupby_statements:
+            args.extend(self._groupby_statements)
+        if self._apply_statements:
+            args.extend(self._apply_statements)
         if self._sortby_fields:
             sortby_args = []
             for f in self._sortby_fields:
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -4097,32 +4097,6 @@ class TestHybridSearch(SearchTestsBase):
 
     @pytest.mark.redismod
     @skip_if_server_version_lt("8.3.224")
-    def test_review_feedback_hybrid_search(self, client):
-        # Create index and add data
-        self._create_hybrid_search_index(client)
-        self._add_data_for_hybrid_search(client, items_sets=5)
-
-        # set search query
-        search_query = HybridSearchQuery("@color:{red} @color:{green}")
-        search_query.scorer("TFIDF")
-
-        vsim_query = HybridVsimQuery(
-            vector_field_name="@embedding",
-            vector_data=np.array([-100, -200, -200, -300], dtype=np.float32).tobytes(),
-        )
-
-        hybrid_query = HybridQuery(search_query, vsim_query)
-
-        res = client.ft().hybrid_search(query=hybrid_query)
-        if is_resp2_connection(client):
-            assert len(res.results) > 0
-            assert res.warnings == []
-        else:
-            assert len(res["results"]) > 0
-            assert res["warnings"] == []
-
-    @pytest.mark.redismod
-    @skip_if_server_version_lt("8.3.224")
     def test_basic_hybrid_search(self, client):
         # Create index and add data
         self._create_hybrid_search_index(client)
@@ -5289,3 +5263,56 @@ class TestHybridSearch(SearchTestsBase):
             assert len(vsim_res_from_cursor.rows) == 5
         else:
             assert len(vsim_res_from_cursor[0]["results"]) == 5
+
+    @pytest.mark.redismod
+    @skip_if_server_version_lt("8.3.224")
+    def test_hybrid_search_query_with_multiple_loads_and_applies(self, client):
+        # Create index and add data
+        self._create_hybrid_search_index(client)
+        self._add_data_for_hybrid_search(client, items_sets=1)
+
+        # set search query
+        search_query = HybridSearchQuery("@color:{red|green}")
+
+        vsim_query = HybridVsimQuery(
+            vector_field_name="@embedding",
+            vector_data=np.array([1, 2, 7, 6], dtype=np.float32).tobytes(),
+        )
+
+        hybrid_query = HybridQuery(search_query, vsim_query)
+
+        postprocessing_config = HybridPostProcessingConfig()
+        postprocessing_config.load("@color", "@price")
+        postprocessing_config.load("@description")
+        postprocessing_config.apply(discount_10_percents="@price - (@price * 0.1)")
+        postprocessing_config.apply(
+            additional_discount="@discount_10_percents - (@discount_10_percents * 0.1)"
+        )
+        postprocessing_config.filter(HybridFilter('@price=="15"'))
+        postprocessing_config.load("@description")
+        postprocessing_config.sort_by(
+            SortbyField("@discount_10_percents", asc=False),
+            SortbyField("@color", asc=True),
+        )
+        postprocessing_config.limit(0, 5)
+
+        res = client.ft().hybrid_search(
+            query=hybrid_query, post_processing=postprocessing_config, timeout=10
+        )
+        print(res)
+        if is_resp2_connection(client):
+            assert len(res.results) == 2
+            for item in res.results:
+                assert item["color"] is not None
+                assert item["price"] is not None
+                assert item["description"] is not None
+                assert item["discount_10_percents"] is not None
+                assert item["additional_discount"] is not None
+        else:
+            assert len(res["results"]) == 2
+            for item in res["results"]:
+                assert item["color"] is not None
+                assert item["price"] is not None
+                assert item["description"] is not None
+                assert item["discount_10_percents"] is not None
+                assert item["additional_discount"] is not None
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 ---
 # image tag 8.0-RC2-pre is the one matching the 8.0 GA release
 x-client-libs-stack-image: &client-libs-stack-image
-  image: "redislabs/client-libs-test:${CLIENT_LIBS_TEST_STACK_IMAGE_TAG:-8.2}"
+  image: "redislabs/client-libs-test:${CLIENT_LIBS_TEST_STACK_IMAGE_TAG:-8.4-RC1-pre.2}"
 
 x-client-libs-image: &client-libs-image
   image: "redislabs/client-libs-test:${CLIENT_LIBS_TEST_IMAGE_TAG:-8.4-RC1-pre.2}"


tests/test_asyncio/test_search.py
```diff
@@ -41,1972 +52,3166 @@
...
+        #     assert res.warnings == []
+        #     assert res.execution_time > 0
+        # else:
+        #     assert res["total_results"] >= 3
+        #     assert len(res["results"]) == 3
+        #     assert res["results"] == self._convert_dict_values_to_str(expected_results)
+        #     assert res["warnings"] == []
+        #     assert res["execution_time"] > 0
+
+        # combine with RRF, not all possible params provided
```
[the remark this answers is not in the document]
petyaslavova: done

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,279 @@
+from typing import List, Literal, Optional, Union
+
+try:
+    from typing import Self  # Py 3.11+
+except ImportError:
+    from typing_extensions import Self
+
+from redis.commands.search.aggregation import Limit, Reducer
+from redis.commands.search.query import Filter, SortbyField
```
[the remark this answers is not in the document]
petyaslavova: it is used.

tests/test_search.py
```diff
@@ -45,3824 +61,4957 @@
...
+    @staticmethod
+    def getClient(client):
+        """
+        Gets a client client attached to an index name which is ready to be
+        created
+        """
+        return client
+
+    @staticmethod
+    def createIndex(client, num_docs=100, definition=None):
```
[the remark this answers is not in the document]
petyaslavova: Old method - changing it is out of scope

tests/test_asyncio/test_search.py
```diff
@@ -41,1972 +52,3166 @@
...
+    def getClient(decoded_r: redis.Redis):
+        """
+        Gets a client client attached to an index name which is ready to be
+        created
+        """
+        return decoded_r
 
-        q = Query('"king henry"').paging(0, 1).summarize().highlight()
+    @staticmethod
+    async def createIndex(decoded_r, num_docs=100, definition=None):
```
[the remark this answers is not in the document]
petyaslavova: Old method - changing it is out of scope

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+        return self
+
+    def build_args(self) -> List[str]:
+        args = []
+        if self._combine:
+            args.extend(("COMBINE", *self._combine))
+        if self._load_fields:
+            fields_str = " ".join(self._load_fields)
+            fields = fields_str.split(" ")
+            args.extend(("LOAD", len(fields), *fields))
```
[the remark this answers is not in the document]
petyaslavova: This is done to be able to support calls like the following q.load("val1", "val2", "val3 AS the_value")

redis/commands/search/commands.py
```diff
@@ -470,7 +504,7 @@ def get_params_args(
             return []
         args = []
         if len(query_params) > 0:
-            args.append("params")
+            args.append("PARAMS")
```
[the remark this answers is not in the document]
petyaslavova: In the official docs of all commands where it is used it is with capital letters.

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+        self._yield_score_as = yield_score_as
+
+    def query_string(self) -> str:
+        """Return the query string of this query object."""
+        return self._query_string
+
+    def scorer(self, scorer: str) -> "HybridSearchQuery":
+        """
+        Scoring algorithm for text search query.
+        Allowed values are "TFIDF" or  "BM25"
```
vladvildanov: Allowed values are `type?: 'BM25' | 'TFIDF' | 'DISMAX' | 'DOCSCORE'` according to the design. Also maybe it makes sense to keep value as Enum so we don't need to add validation around it?

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+        """
+        self._query_string = query_string
+        self._scorer = scorer
+        self._yield_score_as = yield_score_as
+
+    def query_string(self) -> str:
+        """Return the query string of this query object."""
+        return self._query_string
+
+    def scorer(self, scorer: str) -> "HybridSearchQuery":
```
vladvildanov, line 39: Apart of scorer algorithm, scorer also supports named parameters

`SCORER algorithm params...`

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+    def vsim_method_params(
+        self,
+        method: str,
+        **kwargs,
+    ) -> "HybridVsimQuery":
+        """
+        Add search method parameters to the query.
+
+        Args:
+            method: Vector search method name. Supported values are "KNN" or "RANGE".
```
vladvildanov, line 129: Enum value could be used

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+            kwargs: Search method parameters. Use the param names for keys and the
+                values for the values. Example: {"K": 10, "EF_RUNTIME": 100}.
```
vladvildanov, line 131: K is a required argument, whereas EF_RUNTIME is optional. Makes sense to add some validation around it

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+        conditions: str,
+    ) -> None:
+        """
+        Create a new hybrid filter object.
+
+        Args:
+            conditions: Filter conditions.
+        """
+        args = [conditions]
+        Filter.__init__(self, "FILTER", *args)
```
vladvildanov, line 351: Filters are way more complex then this. 

 ```
 [FILTER] "<filter-expression>"
        [POLICY [ADHOC/BATCHES/ACORN]]
            [BATCHES BATCH_SIZE <batch-size-value>]
```

File: redis/commands/search/hybrid_query.py [hunk 12]
```diff
@@ -0,0 +1,326 @@
...
+    ) -> "HybridVsimQuery":
+        """
+        Add search method parameters to the query.
+
+        Args:
+            method: Vector search method name. Supported values are "KNN" or "RANGE".
+            kwargs: Search method parameters. Use the param names for keys and the
+                values for the values. Example: {"K": 10, "EF_RUNTIME": 100}.
+        """
+        vsim_method_params: List[Union[str, int]] = [method]
```
vladvildanov: It's not really clear which parameters are corresponding to which method and which one is optional, which one is required. I would suggest to add more clear structures around it for better user experience

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+        args.extend(self._vector_similarity_query.get_args())
+        return args
+
+
+class HybridPostProcessingConfig:
+    def __init__(self) -> None:
+        """
+        Create a new hybrid post processing configuration object.
+        """
+        self._combine = []
```
vladvildanov: Combine isn't really a post processing configuration. AFAIU, this is fully pre-processing configuration to specify how does documents and vectors should be combined

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+        self._load_fields = []
+        self._groupby = []
+        self._apply = []
+        self._sortby_fields = []
+        self._filter = None
+        self._limit = None
+
+    def combine(
+        self,
+        method: Literal["RRF", "LINEAR"],
```
vladvildanov: Design also specifies third one: `FUNCTION`. PRD also mentioned it

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+        **kwargs,
+    ) -> Self:
+        """
+        Add combine parameters to the query.
+
+        Args:
+            method: The combine method to use - RRF or LINEAR.
+            kwargs: Additional combine parameters.
```
vladvildanov, line 198: Same problem here, isn't really clear which parameters could be used with which method

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+                of strings. In both cases, the field should be specified as `@field`.
+            reducers: One or more reducers. Reducers may be found in the
+                `aggregation` module.
+        """
+
+        fields = [fields] if isinstance(fields, str) else fields
+
+        ret = ["GROUPBY", str(len(fields)), *fields]
+        for reducer in reducers:
+            ret += ["REDUCE", reducer.NAME, str(len(reducer.args))]
```
vladvildanov: So it needs to add `REDUCE` for each reducer? From PRD it seems that it accepts count argument and single `REDUCE` keyword. Also, it doesn't says about aliases. Correct me if I'm wrong

`[GROUPBY count field... REDUCE function count...]`

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+    def apply(self, **kwexpr) -> Self:
+        """
+        Specify one or more projection expressions to add to each result.
+
+        Args:
+            kwexpr: One or more key-value pairs for a projection. The key is
+                the alias for the projection, and the value is the projection
+                expression itself, for example `apply(square_root="sqrt(@foo)")`.
+        """
+        for alias, expr in kwexpr.items():
```
vladvildanov, line 281: PRD also doesn't specifies an option to have multiple APPLY keywords

redis/commands/search/hybrid_query.py
```diff
@@ -0,0 +1,326 @@
...
+            ret = ["APPLY", expr]
+            if alias is not None:
+                ret += ["AS", alias]
+            self._apply.extend(ret)
+
+        return self
+
+    def sort_by(self, *sortby: "SortbyField") -> Self:
```
vladvildanov, line 291: Looks like `SortByField` is missing `WITHCOUNT` option

redis/commands/search/hybrid_query.py
petyaslavova: Not for this version - currently only the scorer name can be provided.

redis/commands/search/hybrid_query.py
petyaslavova: Tested the other options - they also work + found some other that are not listed in the spec and are supported. I would prefer to leave it as string and even not be too detailed in the docstring - I am not sure what are all of the actually supported values...

redis/commands/search/hybrid_query.py
petyaslavova: There are more details in the docstrings + a few examples in the tests that will also be useful.

redis/commands/search/hybrid_query.py
petyaslavova: It is not supported in this version.

redis/commands/search/hybrid_query.py
petyaslavova: Done.

redis/commands/search/hybrid_query.py
petyaslavova: I would prefer not to add validation of the inputs, and this way allow more flexibility for future server api changes, but I have added more details in the docstrings

redis/commands/search/hybrid_query.py
petyaslavova: I can't find such an option in the spec.

redis/commands/search/hybrid_query.py
petyaslavova: But it works... And there is no other way to apply more than one field transformation except to have the apply twice or more. There  is a test covering that case.

redis/commands/search/hybrid_query.py
petyaslavova: Moved it in separate class config with enum for the methods.

redis/commands/search/hybrid_query.py
petyaslavova: This is an example of a working query: 
`FT.HYBRID idx SEARCH @color:{red|green} VSIM @embedding testtesttesttest LOAD 4 @color @price @size @item_type GROUPBY 3 @item_type @size @price REDUCE COUNT_DISTINCT 1 @color AS colors_count REDUCE AVG 1 @price LIMIT 0 4 TIMEOUT 10`
There are some examples in the spec that are not working, that are with just one REDUCE keyword and count of reducers.
Its true that aliases are not mentioned, but they are accepted and the response contains the alias provided for column name.

redis/commands/search/hybrid_query.py
petyaslavova: When POLICY and BATCHES are added server returns an error "Unknown argument"

`FT.HYBRID idx SEARCH @color:{missing} VSIM @embedding abcd1234efgh5678 FILTER "@price:[15 16] @size:[10 11]" POLICY ADHOC LOAD 2 @price @size TIMEOUT 10 `
`(error) POLICY: Unknown argument`

`FT.HYBRID idx SEARCH @color:{missing} VSIM @embedding abcd1234efgh5678 FILTER "@price:[15 16] @size:[10 11]" BATCHES BATCH_SIZE 3 LOAD 2 @price @size TIMEOUT 10 `
`(error) BATCHES: Unknown argument`

redis/commands/search/hybrid_query.py
vladvildanov: @petyaslavova Well even if there's only few that you're sure about, I would make it Enum. This way user knows which values could be used, if even we dunno what work and what's not, for user it would be even harder to find what is the actual values

redis/commands/search/hybrid_query.py
vladvildanov: Got you, not supported

redis/commands/search/hybrid_query.py
vladvildanov: This is in the design doc. I think it based on FT.SEARCH spec

https://redis.io/docs/latest/commands/ft.search/

petyaslavova: This one is for the ft.search command. For ft.hybrid we still have only the confluence page. It doesn't support everything that is supported for ft.search.

redis/commands/search/hybrid_query.py
petyaslavova: Fixed.

redis/commands/search/hybrid_query.py
petyaslavova: @vladvildanov I checked what all the supported scorers are currently and found that the module allows/supports custom scorers to be added - https://redis.io/docs/latest/develop/ai/search-and-query/administration/extensions/ 
In such a case we will restrict the client to be used only with the default ones and exclude the custom additions, if we choose to change the scorer input type to enum.
One more argument to leave this one as string is that for the other commands it is also a string( the same argument is available for the ft.aggregate command). 
To make things easier for the end user, I can provide a link to documentation that explains the supported values.

vladvildanov (approved):

PR #3818, opened 2025-10-27
Apply policies for normal and pipeline mode

vladvildanov:
### Pull Request check-list

### Description of change

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -727,7 +727,7 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands):
 
         self._policies_callback_mapping: dict[Union[RequestPolicy, ResponsePolicy], Callable] = {
             RequestPolicy.DEFAULT_KEYLESS: lambda command_name: [self.get_random_primary_or_all_nodes(command_name)],
-            RequestPolicy.DEFAULT_KEYED: lambda command, *args: self.get_node_from_slot(command, *args),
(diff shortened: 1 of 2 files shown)


redis/cluster.py
```diff
@@ -804,6 +843,74 @@ def get_default_node(self):
         """
         return self.nodes_manager.default_node
 
+    def get_node_from_slot(self, command: str, *args):
```
petyaslavova: Since a list is returned here, `get_nodes_from_slot` would be a better name for the method

redis/cluster.py
```diff
@@ -804,6 +843,74 @@ def get_default_node(self):
...
+            })
+
+        return commands
+
+    def get_special_nodes(self) -> Optional[list["ClusterNode"]]:
+        """
+        Returns a list of nodes for commands with a special policy.
+        """
+        if not self._aggregate_nodes:
+            raise RedisClusterException('Cannot execute FT.CURSOR commands without FT.AGGREGATE')
```
petyaslavova, line 879: I guess this exception reports relation with FT.CURSOR and FT.AGGREGATE, because the Special policy is related just to those two, but if this type of policy is extended to cover additional commands, this error message might become confusing.

vladvildanov: For special cases we would anyway need something like separate mapping or strategies for each special command when they would appear. But for now, we don't know if there will be any soon, so I didn't implement it

petyaslavova (approved):

PR #3863, opened 2025-12-03
Adding retries for the overall connect - socket connect + handshake. Fix for pubsub reconnect issues.

### LINKED ISSUE #3203: Bug in retry-handling in the PubSub-client

### LINKED ISSUE #3555: Retry Mechanism Fails When Redis Container is Paused

petyaslavova:
With this PR, retries now cover both the socket connection and the initial connection handshake, rather than retrying the socket connect alone.

This improves retry behavior in several scenarios, including:

- The socket connects successfully, but the server is still booting and not yet fully functional.
- Pub/Sub connection issues (Fixes #3203).
- Retry handling when the Redis container is paused (Fixes #3555).

A huge thank-you to @ManelCoutinhoSensei — his excellent work on PR #3596 and his deep investigations into these issues revealed that our retry logic was incomplete. That insight was key to driving this improvement.
Really appreciate the effort and the quality of his analysis!

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -299,7 +299,9 @@ class AbstractConnection:
         # try once the socket connect with the handshake, retry the whole
         # connect/handshake flow based on retry policy
         await self.retry.call_with_retry(
-            lambda: self.connect_check_health(check_health=True, retry_socket_connect=False),
+            lambda: self.connect_check_health(
+                check_health=True, retry_socket_connect=False
+            ),
             lambda error: self.disconnect(),
         )
 


ManelCoutinhoSensei:
@petyaslavova I was testing this solution yesterday and while it does hide the bug that was previously there, I believe this solution is not 100% correct.
Essentially what this is doing is adding a new retry outside the normal one and that's why the logs now, when you configure the connection to retry 3 times for example, are something like:
```logs
WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.
WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.
WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.
**WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.**
WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.
WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.
WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.
**WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.**
WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.
WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.
WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.
**WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.**
WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.
WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.
WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.
```
Instead of only doing it once

petyaslavova:
> @petyaslavova I was testing this solution yesterday and while it does hide the bug that was previously there, I believe this solution is not 100% correct. Essentially what this is doing is adding a new retry outside the normal one and that's why the logs now, when you configure the connection to retry 3 times for example, are something like:
> 
> ```
> WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.
> WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.
> WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.
> **WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.**
> WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.
> WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.
> WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.
> **WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.**
> WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.
> WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.
> WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.
> **WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.**
> WARNING:root:Attempt 1 / 3. Retrying in 0.50 seconds.
> WARNING:root:Attempt 2 / 3. Retrying in 1.00 seconds.
> WARNING:root:Attempt 3 / 3. Retrying in 2.00 seconds.
> ```
> 
> Instead of only doing it once

@ManelCoutinhoSensei, please provide more info... Some simple code example, what is the Redis server state?

ManelCoutinhoSensei:
@petyaslavova the case that I was talking about in #3555 for example.
the simplest way to do it is just start the redis-standalone instance (from your docker-compose) and pause it
then just try to do a random command
Code example:
```python
    redis_conn = Redis(
        connection_pool=ConnectionPool.from_url(
            "redis://localhost:6379",
            encoding="utf-8",
            decode_responses=False,
            socket_keepalive=True,
            socket_timeout=1,
            socket_connect_timeout=30,
            health_check_interval=10,
            retry_on_error=(ConnectionError, TimeoutError, socket.timeout),
            retry=Retry(ExponentialBackoff(5.0, 0.25), retries=3),
        )
    )

    for i in range(n_messages):
        print("Trying to add")
        res = redis_conn.xadd(topic, {"value": str(i)}, maxlen=10000, approximate=False)
        print(res)
        time.sleep(0.1)
```

you will have to put logs in the retry class to see the logs

it does make sense for this to be happening since the solution was just to put another retry wrapper around the connection

ManelCoutinhoSensei:
@petyaslavova Did you have a chance to verify this?

petyaslavova:
> @petyaslavova Did you have a chance to verify this?

Hey @ManelCoutinhoSensei, unfortunately, not yet. Hopefully, in the coming weeks, I’ll have more time to focus on the open PRs and issues 🤞

ManelCoutinhoSensei:
Hey @petyaslavova , did you have any time yet?

PR #3864, opened 2025-12-04
Added intrastructure and integration point with OTel

vladvildanov:
### Description of change

### Pull Request check-list

### DIFF
--- a/redis/observability/attributes.py
+++ b/redis/observability/attributes.py
@@ -155,7 +155,7 @@ class AttributeBuilder:
         return attrs
 
     @staticmethod
-    def build_connection_pool_attributes(
+    def build_connection_attributes(
             pool_name: str,
             connection_state: Optional[ConnectionState] = None,
             is_pubsub: Optional[bool] = None,
--- a/redis/observability/config.py
+++ b/redis/observability/config.py
@@ -82,10 +82,10 @@ class OTelConfig:
     def __init__(
             self,
             # Core enablement
-            enabled_telemetry: List[TelemetryOption] = None,
+            enabled_telemetry: Optional[List[TelemetryOption]] = None,
             # Metrics-specific
             metrics_sample_percentage: float = 100.0,
-            metric_groups: List[MetricGroup] = None,
+            metric_groups: Optional[List[MetricGroup]] = None,
             # Redis-specific telemetry controls
             include_commands: Optional[List[str]] = None,
             exclude_commands: Optional[List[str]] = None,
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -30,3 +30,7 @@ numpy>=1.24.0,<2.0 ; platform_python_implementation == "PyPy"
 
 redis-entraid==1.0.0
 pybreaker>=1.4.0
+
+opentelemetry-api>=1.18.0
+opentelemetry-sdk>=1.18.0
+opentelemetry-exporter-otlp-http>=1.18.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,11 @@ jwt = [
 circuit_breaker = [
     "pybreaker>=1.4.0"
 ]
+otel = [
+    "opentelemetry-api>=1.18.0",
+    "opentelemetry-sdk>=1.18.0",
+    "opentelemetry-exporter-otlp-http>=1.18.0",
+]
 
 [project.urls]
 Changes = "https://github.com/redis/redis-py/releases"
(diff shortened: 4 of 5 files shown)


redis/observability/attributes.py
```diff
@@ -0,0 +1,296 @@
...
+        if retry_attempts is not None and retry_attempts > 0:
+            attrs[REDIS_CLIENT_OPERATION_RETRY_ATTEMPTS] = retry_attempts
+
+        if is_blocking is not None:
+            attrs[REDIS_CLIENT_OPERATION_BLOCKING] = is_blocking
+
+        return attrs
+
+    @staticmethod
+    def build_connection_pool_attributes(
```
petyaslavova: Having in mind that this is for the connection pool, not for a concrete connection, connection_state and is_pubsub seem strange - they are specific per connection.

redis/observability/config.py
```diff
@@ -0,0 +1,169 @@
...
+            otel.init(OTelConfig())
+    """
+
+    DEFAULT_TELEMETRY = TelemetryOption.METRICS
+    DEFAULT_METRIC_GROUPS = MetricGroup.COMMAND | MetricGroup.CONNECTION_BASIC | MetricGroup.RESILIENCY
+
+    def __init__(
+            self,
+            # Core enablement
+            enabled_telemetry: List[TelemetryOption] = None,
```
petyaslavova: The List is Optional

redis/observability/config.py
```diff
@@ -0,0 +1,169 @@
...
+            # Metrics-specific
+            metrics_sample_percentage: float = 100.0,
+            metric_groups: List[MetricGroup] = None,
```
petyaslavova: Optional

redis/observability/providers.py
```diff
@@ -0,0 +1,333 @@
...
+        if not self.config.is_enabled():
+            return None
+
+        # Lazy import - only import OTel when metrics are enabled
+        try:
+            from opentelemetry import metrics
+            from opentelemetry.metrics import NoOpMeterProvider
+        except ImportError:
+            raise ImportError(
+                "OpenTelemetry is not installed. Install it with:\n"
```
petyaslavova, line 73: Those libs can be added in the toml file as optional dependencies

redis/observability/config.py
```diff
@@ -0,0 +1,169 @@
...
+            # Redis-specific telemetry controls
+            include_commands: Optional[List[str]] = None,
+            exclude_commands: Optional[List[str]] = None,
+    ):
+        # Core enablement
+        if enabled_telemetry is None:
+            self.enabled_telemetry = self.DEFAULT_TELEMETRY
+        else:
+            self.enabled_telemetry = TelemetryOption(0)
```
petyaslavova, line 97: We have to somehow use the input values. This way the user might configure a list of options and to expect that they should be working. I believe here the provided list should be checked and if invalid ot not yet supported options are provided - an error should be either logged or raised.

tests/test_observability/test_recorder.py
```diff
@@ -0,0 +1,956 @@
...
+These tests verify that recorder functions correctly pass arguments through
+to the underlying OTel Meter instruments (Counter, Histogram, UpDownCounter).
+The MeterProvider is mocked to verify the actual integration point where
+metrics are exported to OTel.
+"""
+
+import pytest
+from unittest.mock import MagicMock, patch
+
+from redis.observability import recorder
```
petyaslavova, line 13: I don't see the new otel dependencies added to the dev-dependencies list - aren't they needed for the successful test execution?

redis/observability/config.py
vladvildanov: But we use the input values, I used `IntFlag` approach to store the result of multiple Enum values. This line is just instantiate an object with no value and then here, we apply all configuration given. And the same as Enums it will be automatically verified that the type is correct.

https://github.com/redis/redis-py/pull/3864/files#diff-f2fe2a40b49fd19e25110a4b17f58b3a6353f79c43a67e8439ed53a02f5873dbR98

redis/observability/attributes.py
vladvildanov: fixed

redis/observability/config.py
vladvildanov: fixed

redis/observability/config.py
vladvildanov: fixed

redis/observability/providers.py
vladvildanov: Added

tests/test_observability/test_recorder.py
vladvildanov: You're right, they're needed, added them

redis/observability/config.py
petyaslavova: Yeah, just below :) Thanks :)

PR #3866, opened 2025-12-04
Idle connection management

praboud:
### Description of change
Adds idle connection management to `ConnectionPool`. This is disabled by default, but clients can configure a `idle_connection_timeout`. If set, connections that aren't used for longer than the configured timeout are closed automatically by a background thread. (There's a singleton `IdleConnectionCleanupManager` which schedules these checks for all connection pools, to avoid needing to allocate a separate thread for each pool.)

Right now, because connections are never closed once opened, clients have a "high water mark" property - the client will hold open as many connections as it needed during the busiest period that client has experienced. This can cause far more connections to be used than are really needed, as well as making the number of connections used difficult to reason about. I've deployed this patch to production, and seen a roughly 2x reduction in connections used in practice, on a large deployment of clients.

### Pull Request check-list


petyaslavova:
Hi @praboud, thank you for your contribution! We will take a look at it soon.

PR #3868, opened 2025-12-05
Applying the changes in the SMIGRATED notification format

petyaslavova:
In this PR the changed format of the SMIGRATED notification is applied.
The Proxy Mock helper needed some changes to generate the resp format of the messages.
Parser was changed, as well as the OSSNodeMigratedNotification fields - now the object contains a map with nodes host and port and slots.
Unit tests to related changes are also updated.

This is the format of the notification:
`>3
+SMIGRATED
:SeqID
*<num_entries>
  *2
    +<host:port>
    +<slots-or-ranges>
`

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -319,6 +319,8 @@ class MaintNotificationsAbstractConnection:
             oss_cluster_maint_notifications_handler,
             parser,
         )
+        self._processed_start_maint_notifications = set()
+        self._skipped_end_maint_notifications = set()
 
     @abstractmethod
     def _get_parser(self) -> Union[_HiredisParser, _RESP3Parser]:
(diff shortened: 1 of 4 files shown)


tests/maint_notifications/proxy_server_helpers.py
```diff
@@ -11,37 +11,51 @@ class RespTranslator:
...
     @staticmethod
-    def str_or_list_to_resp(txt: str) -> str:
-        """
-        Convert specific string or list to RESP format.
-        """
-        if re.match(r"^<.*>$", txt):
-            items = txt[1:-1].split(",")
-            return f"*{len(items)}\r\n" + "\r\n".join(
-                f"${len(x)}\r\n{x}" for x in items
+    def oss_maint_notification_to_resp(txt: str) -> str:
```
vladvildanov, line 14: I'm not in a context, but why does client needs to convert plain text notification into RESP? As I understand, server sends it as a plain text, so we do it for some kind internal purposes?

petyaslavova: We have a proxy server that we use for testing. We send the resp that needs to be pushed to the connections to test how clients handle the push messages.

vladvildanov: Got you

PR #3869, opened 2025-12-05
Add handling for parallel slot migrations with maintenance notifications flow for OSS Cluster API

petyaslavova:
### Description of change

### Pull Request check-list

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -669,7 +669,7 @@ class MaintNotificationsAbstractConnection:
     def maintenance_state(self, state: "MaintenanceState"):
         self._maintenance_state = state
 
-    def add_miant_start_notification(self, id: int):
+    def add_maint_start_notification(self, id: int):
         self._processed_start_maint_notifications.add(id)
 
     def get_processed_start_notifications(self) -> set:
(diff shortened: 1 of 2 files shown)


redis/connection.py
```diff
@@ -667,6 +669,22 @@ def maintenance_state(self) -> MaintenanceState:
     def maintenance_state(self, state: "MaintenanceState"):
         self._maintenance_state = state
 
+    def add_miant_start_notification(self, id: int):
```
vladvildanov: miant -> maint

redis/maint_notifications.py
```diff
@@ -1024,15 +1034,26 @@ def handle_oss_maintenance_completed_notification(
             else:
                 if self.config.is_relaxed_timeouts_enabled():
                     # reset the timeouts for the node to which the connection is connected
-                    # TODO: add check if other maintenance ops are in progress for the same node - CAE-1038
-                    # and if so, don't reset the timeouts
+                    # Perform check if other maintenance ops are in progress for the same node
+                    # and if so, don't reset the timeouts and wait for the last maintenance
+                    # to complete
                     for conn in (
                         *current_node.redis_connection.connection_pool._get_in_use_connections(),
```
vladvildanov: I think we need to somehow control access to connections, in multi-threaded environment in use and free connections are frequently accessed and could be modified while we're loop over them

redis/connection.py
petyaslavova: fixed.

redis/maint_notifications.py
petyaslavova: Good catch :) I have totally missed that - added a pool lock while iterating over the connections in the connection pool.

Issue #3874, opened 2025-12-09
SentinelConnectionPool does not retrieve updated master address from Sentinel on connection retries


SamuelGD:
**Description:**
When the Redis master address changes (e.g., after a Sentinel failover), the Redis client using a SentinelConnectionPool does not fetch the updated master address during connection retries.

Currently, only the `connect` method in SentinelManagedConnection retrieves the master address from Sentinel https://github.com/redis/redis-py/blob/v7.1.0/redis/asyncio/sentinel.py#L53.

However, on  retries, `connect_check_health` call the internal `_connect` method, which does not update the master address: https://github.com/redis/redis-py/blob/v7.1.0/redis/asyncio/connection.py#L306-L311.

As a result, the client may continue attempting to connect to the old master.

**Expected behavior:**
Retries should retrieve the latest master address from Sentinel, ensuring the client reconnects to the correct master after a failover.

petyaslavova:
Hi @SamuelGD, thank you for reporting this! We will have a look at  it.

SamuelGD:
Thank you @petyaslavova, I also opened a PR for this https://github.com/redis/redis-py/pull/3875

SamuelGD:
The bug may have been introduced in this PR https://github.com/redis/redis-py/pull/3557/files.

In send_packed_command, self.connect was swapped with self.connect_check_health but SentinelConnectionPool only overrides connect.

jbadson:
@SamuelGD that PR may be contributing to the problem, but I don't think it's the only cause. I've had the same issue since at least version 5.2.1, which was released in Dec 2024.  The Sentinel client automatically reconnects to the new sentinel after a failover, but the master connection is not updated.

I assumed this was the intended functionality since last time I checked the docs didn't explicitly say the client would automatically reconnect, but if it is a bug then it would be convenient to have it fixed.

For now I've worked around it by adding a wrapper that catches ConnectionError on Redis client calls and recreates the client by calling Sentinel.master_for() again.

ManelCoutinhoSensei:
I ran into the same issue. For now, I’m working around it on my side with the following override:
```python
class SentinelAwareManagedConnection(SentinelManagedConnection):
    def connect_check_health(self, check_health: bool = True, retry_socket_connect: bool = True) -> None:
        """
        Override to re-establish the connection using sentinel master re-discovery when the socket is dead.

        Args:
            check_health: whether to perform a PING/PONG health check after connecting.
[...]

PR #3875, opened 2025-12-09
SentinelConnectionPool should retrieve updated master address from Sentinel on connection retries

SamuelGD:
### Description of change

https://github.com/redis/redis-py/issues/3874

### Pull Request check-list

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -37,29 +37,21 @@ class SentinelManagedConnection(Connection):
             s = s % host_info
         return s
 
-    def connect_to(self, address):
-        self.host, self.port = address
-
-        self.connect_check_health(
-            check_health=self.connection_pool.check_connection,
-            retry_socket_connect=False,
-        )
-
-    def _connect_retry(self):
+    def _connect(self):
         if self._sock:
-            return  # already connected
+            super()._connect()
+            return None # already connected
         if self.connection_pool.is_master:
-            self.connect_to(self.connection_pool.get_master_address())
-        else:
-            for slave in self.connection_pool.rotate_slaves():
-                try:
-                    return self.connect_to(slave)
-                except ConnectionError:
-                    continue
-            raise SlaveNotFoundError  # Never be here
-
-    def connect(self):
-        return self.retry.call_with_retry(self._connect_retry, lambda error: None)
+            self.host, self.port = self.connection_pool.get_master_address()
+            super()._connect()
+            return None
+        for slave in self.connection_pool.rotate_slaves():
+            try:
+                self.host, self.port = slave
+                return super()._connect()
+            except ConnectionError:
+                continue
+        raise SlaveNotFoundError  # Never be here
 
     def read_response(
         self,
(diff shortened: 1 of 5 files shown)


petyaslavova:
Hi @SamuelGD, than you for your contribution! We will review it soon.

redis/sentinel.py
```diff
@@ -37,29 +37,19 @@ def __repr__(self):
...
-        self.connect_check_health(
-            check_health=self.connection_pool.check_connection,
-            retry_socket_connect=False,
-        )
-
-    def _connect_retry(self):
+    def _connect(self):
         if self._sock:
-            return  # already connected
+            return super()._connect() # already connected
```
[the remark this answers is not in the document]
SamuelGD: If the connection is already established, _connect is never called https://github.com/redis/redis-py/blob/v7.1.0/redis/connection.py#L852

SamuelGD:
hi @petyaslavova, any chance someone from your team could take a look this week?

SamuelGD:
Hi @petyaslavova, we've been running a patch on our side for a few weeks now, and it has fixed all the issues we had during failover. I think everyone would benefit from this fix, would someone on your side be able to take a look?

SamuelGD:
hi @petyaslavova, it's been a month and a half, any chance someone could take a look?

redis/sentinel.py
```diff
@@ -37,29 +37,19 @@ def __repr__(self):
...
-        self.connect_check_health(
-            check_health=self.connection_pool.check_connection,
-            retry_socket_connect=False,
-        )
-
-    def _connect_retry(self):
+    def _connect(self):
         if self._sock:
-            return  # already connected
+            return super()._connect()  # already connected
```
ManelCoutinhoSensei, line 42: if it is already connected do we really need to create a new socket?

redis/sentinel.py
```diff
@@ -37,29 +37,19 @@ def __repr__(self):
             s = s % host_info
         return s
 
-    def connect_to(self, address):
-        self.host, self.port = address
-
-        self.connect_check_health(
-            check_health=self.connection_pool.check_connection,
```
ManelCoutinhoSensei, line 44: this setting is being silently ignored/dropped with this PR

redis/sentinel.py
ManelCoutinhoSensei: i know that the `_connect` also has the guard inside but still...

eeshsaxena (commented):
Nice approach moving the master address lookup into `_connect` so each retry through the base `connect()` re-resolves it from Sentinel rather than reusing a stale address.

One small thing on the already-connected guard:

```python
def _connect(self):
    if self._sock:
        return super()._connect()  # already connected
```

The comment says "already connected", but `super()._connect()` builds a fresh socket rather than being a no-op. So if this branch is ever reached on a live connection it reconnects and replaces `self._sock` (and on the async side `self._reader`/`self._writer`) without closing the previous one. The old code returned early here.

In the normal flow `_connect` is only entered when not connected, since `connect_check_health` guards it with `if self.is_connected: return`, so today this is effectively defensive. But returning early (as the old code did) would match the comment and avoid the reconnect if that invariant ever changes. Same note for the async `if self.is_connected: await super()._connect()`.

Issue #3123, opened 2024-01-26
Wrong typing for client parameters in Script


abbasyadollahi:
## Version

5.0.1

## Platform

Python 3.12.1
macOS 14.3

## Description

The arguments referring to a client in the `Script` class are type annotated to the `Redis` class, although instances of `RedisCluster` can also end up being the client. This also applies to the [ScriptCommands.register_script](https://github.com/redis/redis-py/blob/master/redis/commands/core.py#L5389C32-L5389C37) function, where it expects `self` to be a `Redis` instance, even though I can run this successfully with a `RedisCluster` client.

Causes the following type error:
```
Cannot access member "register_script" for type "RedisCluster"
  Could not bind method "register_script" because "RedisCluster" is not assignable to parameter "self"
    "RedisCluster" is incompatible with "Redis" Pylance (reportGeneralTypeIssues)
```

## Code

```python
from redis import Redis
from redis.cluster import RedisCluster

client = Redis("localhost")
script = client.register_script("""return redis.call("SET", KEYS[1], ARGV[1])""")
script(["mykey"], ["myvalue"])

client = RedisCluster("localhost")
script = client.register_script("""return redis.call("SET", KEYS[1], ARGV[1])""")
script(["mykey"], ["myvalue"]) # Type error
```

bcmills:
Ah, staleness bots: because bugs just go away on their own if you ignore them! 🙉

petyaslavova:
Fixed with #3876

arlosnow1231-stack:
I confirm this is fixed

PR #3876, opened 2025-12-10
Fix type hints for register_script to support RedisCluster

### LINKED ISSUE #3712: mypy error: Invalid self argument "RedisCluster" to attribute function "register_script"
When our code calls `register_script` on a `RedisCluster` script, mypy flags the call as an error because the `register_script` method is defined like this in `ScriptComamnds`:
```
    def register_script(self: "redis.client.Redis", script: ScriptTextT) -> Script:
        # ... lines omitted ...
        return Script(self, script)
```

I believe this was done because the `Script` class expects an object of type `redis.client.Redis`, but I'd imagine this could be solved by making `Script` accept `Union[Redis, RedisCluster]` instead.

## Steps to Repro
1. Create a file with these contents:
    ```python
    from typing import Union

    import redis
    from redis.commands.core import Script


    def load_script(redis_connection: Union[redis.Redis, redis.RedisCluster], script_filename: str) -> Script:
        with open(script_filename) as lua_script_file:
            lua_script_contents = lua_script_file.read()

            return redis_connection.register_script(lua_script_contents)
    ```
2. Run mypy on the code.

## Expected Results
No error from mypy.

## Actual Results
mypy prints this error:
> example.py:11: Invalid self argument "RedisCluster" to attribute function "register_script" with type "Callable[[Redis, Union[bytes, str, memoryview[int]]], Script]

majiayu000:
## Summary
Fixes #3712 #3123

Update type annotations in `Script`, `AsyncScript`, and their corresponding `register_script` methods to accept both `Redis` and `RedisCluster` types.

## Changes
- Added `redis.cluster` and `redis.asyncio.cluster` to TYPE_CHECKING imports
- Updated `Script.__init__` to accept `Union["redis.client.Redis", "redis.cluster.RedisCluster"]`
- Updated `Script.__call__` client parameter similarly
- Updated `ScriptCommands.register_script` self type
- Made equivalent changes for async versions (`AsyncScript`, `AsyncScriptCommands`)

## Test
Before fix, mypy showed:
```
Invalid self argument "RedisCluster" to attribute function "register_script" with type "Callable[[Redis, Union[bytes, str, memoryview[int]]], Script]"
```
After fix, no such error.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to type hints plus small cluster-only tests, with no functional logic changes beyond widened accepted client types.
> 
> **Overview**
> Fixes typing for Lua scripting helpers so `register_script`, `Script`, and `AsyncScript` accept both standalone `Redis` and `RedisCluster` clients (including the optional `client` override when calling a script).
> 
> Adds *cluster-only* tests (sync + asyncio) that register a script on a cluster client and assert it executes successfully.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 63a259d177b353cad5c3eb0481dd2f3898be5dda. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


petyaslavova:
Hi @majiayu000, thank you for your contribution! We will take a look at it soon.

majiayu000:
I've proactively added tests for the type hints fix:

- `test_script_with_cluster_client` in `tests/test_scripting.py`: Tests that sync `Script` class accepts `RedisCluster` as registered_client
- `test_async_script_with_cluster_client` in `tests/test_asyncio/test_scripting.py`: Tests that `AsyncScript` class accepts `RedisCluster` as registered_client

These tests verify that the type hints fix works correctly for both sync and async implementations.

tests/test_asyncio/test_scripting.py
```diff
@@ -150,3 +152,27 @@ async def test_eval_msgpack_pipeline_error_in_lua(self, r):
...
+
+    @pytest.mark.asyncio()
+    async def test_async_script_with_cluster_client(self):
+        """Test that AsyncScript class accepts RedisCluster as registered_client.
+
+        This verifies the type hints fix for register_script to support RedisCluster.
+        We use a mock-like approach since we don't need actual cluster connection.
+        Using bytes script to avoid encoder dependency in mock.
+        """
+        from unittest.mock import MagicMock
```
petyaslavova: Import should be added at the top of the file.

tests/test_scripting.py
```diff
@@ -53,6 +54,24 @@ def test_encoder(self, r, script_bytes):
...
+        Using bytes script to avoid encoder dependency in mock.
+        """
+        from unittest.mock import MagicMock
+
+        # Create a mock RedisCluster instance
+        mock_cluster = MagicMock(spec=RedisCluster)
+
+        # Script should accept RedisCluster without type errors
+        # Using bytes script to bypass encoder.encode() call
+        script = Script(mock_cluster, script_bytes)
```
petyaslavova: It would be better to validate that you can actually register the script using the RedisCluster client. With unit tests you won't be able to validate that we don't have a static code analyses errors, but what we can do is to validate that by providing the RedisCluster as valid input type - the client actually can be used to execute a register_script call.

majiayu000:
Thanks! I moved the import to the top and added onlycluster tests that register and execute scripts using RedisCluster (sync + asyncio). Please take another look.

PR #3880, opened 2025-12-10
Add DriverInfo class for upstream driver tracking

vchomakov:
### Description of change

This PR adds a DriverInfo class to enable upstream drivers (like django-redis, celery, rq, etc.) to identify themselves when using redis-py as their underlying client.

Key features:
- DriverInfo dataclass that tracks upstream driver name and version
- Formatted output follows the pattern: redis-py(driver1_vVersion1;driver2_vVersion2)
- Validation ensures driver names follow Python package naming conventions (lowercase, start with letter, only letters/digits/hyphens/underscores)
- Validation ensures driver versions contain only printable ASCII without spaces or braces
- Integration with both Redis and AsyncRedis clients via optional `driver_info` parameter
- Pool propagation: `driver_info` is automatically propagated from client to connection pool to individual connections
- Deprecation: `lib_name` and `lib_version` parameters are now deprecated in favor of `driver_info`
- Fallback behavior: if `driver_info` is not provided, it's created from `lib_name/lib_version` (if present)
- Most recently added driver appears first in the formatted name

Example usage:
```
from redis import Redis, DriverInfo

# Create DriverInfo with upstream driver
info = DriverInfo().add_upstream_driver("django-redis", "5.4.0")
client = Redis(driver_info=info)
# Client will identify as: redis-py(django-redis_v5.4.0)

# Legacy parameters (deprecated)
client = Redis(lib_name="django-redis", lib_version="5.4.0")  # Still works but shows deprecation warning
```

### Pull Request check-list

  - All 17 tests passing (11 DriverInfo tests + 6 client setinfo tests)
  - ruff check passes
  - ruff format --check passes
  - Will enable and monitor
  - Comprehensive test suite covering:
    - Default behavior (no upstream drivers)
    - Single upstream driver
    - Multiple upstream drivers (latest first)
    - Invalid driver name validation (4 test cases)
    - Invalid driver version validation (3 test cases)
    - Client setinfo with driver_info parameter (sync and async)
    - Pool propagation (async with single connection and pool)
    - Deprecation warnings for `lib_name/lib_version`
  - Docstrings added to all public methods and classes
  - Examples included in docstrings
  - Deprecation notices added to `lib_name` and `lib_version` parameters
  - Not applicable - this is a utility class for library integrators, not end users

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1220,9 +1220,6 @@ class ConnectionPool:
         if self._event_dispatcher is None:
             self._event_dispatcher = EventDispatcher()
 
-        # Store driver_info for propagation to connections
-        self.driver_info = self.connection_kwargs.get("driver_info", None)
-
     def __repr__(self):
         conn_kwargs = ",".join([f"{k}={v}" for k, v in self.connection_kwargs.items()])
         return (
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -2502,9 +2502,6 @@ class ConnectionPool(MaintNotificationsAbstractConnectionPool, ConnectionPoolInt
         if self._event_dispatcher is None:
             self._event_dispatcher = EventDispatcher()
 
-        # Store driver_info for propagation to connections
-        self.driver_info = self._connection_kwargs.get("driver_info", None)
-
         # a lock to protect the critical section in _checkpid().
         # this lock is acquired when the process id changes, such as
         # after a fork. during this time, multiple threads in the child


redis/client.py
```diff
@@ -280,6 +282,9 @@ def __init__(
         decode_responses:
             if `True`, the response will be decoded to utf-8.
             Argument is ignored when connection_pool is provided.
+        driver_info:
+            Optional DriverInfo object to identify upstream libraries.
```
petyaslavova, line 290: The current implementation causes to have lib_name to be ignored - it will be valuable addition to the details in the docstrings.

vchomakov: Documented the `lib_name` / `lib_version` deprecation over `driver_info`.

vchomakov:
> I think we should unify driver metadata handling:
> 
> 1. Store the redis-py client version in DriverInfo.
>    If a DriverInfo object is provided, ignore lib_name and lib_version - it should be clearly stated in the docstrings.
> 2. Propagate DriverInfo through pools and connections.
>    Pools should hold a driver_info attribute and pass it down so each connection reads driver fields only from that object.
> 3. Fallback behavior:
>    If no DriverInfo is provided, initialize one using the existing lib_name / lib_version.
> 4. Deprecate lib_name and lib_version.
>    Mark them as deprecated now and plan to remove them after a few releases.(we have an annotation that can be used to deprecate concrete arguments)
> 
> This will keep the driver info consistent, will remove duplicated parameters, and simplify the API going forward.

Thanks, @petyaslavova, for the detailed suggestions. I've addressed them in the last commit.

redis/asyncio/client.py
```diff
@@ -306,6 +313,17 @@ def __init__(
             # Create internal connection pool, expected to be closed by Redis instance
             if not retry_on_error:
                 retry_on_error = []
+
+            # Handle driver_info: if provided, use it; otherwise create from lib_name/lib_version
+            if driver_info is not None:
```
bobymicroby: The same fallback logic is repeated in 4 places: 
client.py ,  asyncio/client.py, connection.py, asyncio/connection.py 

```
  if driver_info is not None:
      computed_driver_info = driver_info
  else:
      name = lib_name if lib_name is not None else "redis-py"
      version = lib_version if lib_version is not None else get_lib_version()
      computed_driver_info = DriverInfo(name=name, lib_version=version)
```
You could add a helper in `driver_info.py`  - then each call site becomes a one-liner.

bobymicroby (approved):
@vchomakov LGTM, maybe consider adding a helper to do the fallback logic as I mentioned above.

redis/asyncio/connection.py
```diff
@@ -1190,6 +1227,9 @@ def __init__(
         if self._event_dispatcher is None:
             self._event_dispatcher = EventDispatcher()
 
+        # Store driver_info for propagation to connections
```
petyaslavova: Where do you use that? When a new connection is created only the connection_kwargs are provided as init arguments.

redis/connection.py
```diff
@@ -2482,6 +2510,9 @@ def __init__(
         if self._event_dispatcher is None:
             self._event_dispatcher = EventDispatcher()
 
+        # Store driver_info for propagation to connections
```
petyaslavova: Same as for the async pool - I think we don't need to have this property stored for the pool object.

PR #3881, opened 2025-12-11
Added export of operation duration metric 

vladvildanov:
### Description of change

### Pull Request check-list

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -84,13 +84,9 @@ class TestRedisClientEventEmission:
             '_get_or_create_collector',
             return_value=collector
         ):
-            # Create event dispatcher (real one, to test the full chain)
-            event_dispatcher = EventDispatcher()
-
             # Create Redis client with mocked connection pool
             client = redis.Redis(
                 connection_pool=mock_connection_pool,
-                event_dispatcher=event_dispatcher,
             )
 
             yield client, self.operation_duration
@@ -145,11 +141,8 @@ class TestRedisClientEventEmission:
         with mock.patch.object(
             recorder, '_get_or_create_collector', return_value=collector
         ):
-            event_dispatcher = EventDispatcher()
-
             client = redis.Redis(
                 connection_pool=mock_connection_pool,
-                event_dispatcher=event_dispatcher,
             )
 
             client._send_command_parse_response = mock.MagicMock(return_value=b'value1')
@@ -181,11 +174,8 @@ class TestRedisClientEventEmission:
         with mock.patch.object(
             recorder, '_get_or_create_collector', return_value=collector
         ):
-            event_dispatcher = EventDispatcher()
-
             client = redis.Redis(
                 connection_pool=mock_connection_pool,
-                event_dispatcher=event_dispatcher,
             )
 
             # Make command raise an exception
@@ -241,11 +231,8 @@ class TestRedisClientEventEmission:
         with mock.patch.object(
             recorder, '_get_or_create_collector', return_value=collector
         ):
-            event_dispatcher = EventDispatcher()
-
             client = redis.Redis(
                 connection_pool=mock_connection_pool,
-                event_dispatcher=event_dispatcher,
             )
 
             client._send_command_parse_response = mock.MagicMock(return_value=True)
@@ -292,11 +279,8 @@ class TestRedisClientEventEmission:
         with mock.patch.object(
             recorder, '_get_or_create_collector', return_value=collector
         ):
-            event_dispatcher = EventDispatcher()
-
             client = redis.Redis(
                 connection_pool=mock_connection_pool,
-                event_dispatcher=event_dispatcher,
             )
 
             client._send_command_parse_response = mock.MagicMock(return_value=True)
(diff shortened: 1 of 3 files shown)


redis/event.py
```diff
@@ -295,6 +298,20 @@ def commands(self) -> tuple:
     def exception(self) -> Exception:
         return self._exception
 
+@dataclass
+class AfterCommandExecutionEvent:
```
petyaslavova, line 302: It would be easier to understand what this event handles if the name suggests it is related to metrics collection or if it might be used for other things as well - at least in the docstrings, we can add a mention of the metrics collection.

tests/test_client.py
```diff
@@ -0,0 +1,347 @@
...
+            collector = RedisMetricsCollector(mock_meter, config)
+
+        # Patch the recorder to use our collector
+        with mock.patch.object(
+            recorder,
+            '_get_or_create_collector',
+            return_value=collector
+        ):
+            # Create event dispatcher (real one, to test the full chain)
+            event_dispatcher = EventDispatcher()
```
petyaslavova: It is not mandatory - in case it is not provided as input argument a default EventDispatcher instances will be created.

tests/test_client.py
```diff
@@ -0,0 +1,347 @@
...
+        recorder.reset_collector()
+        config = OTelConfig(metric_groups=[MetricGroup.COMMAND])
+
+        with mock.patch('redis.observability.metrics.OTEL_AVAILABLE', True):
+            collector = RedisMetricsCollector(mock_meter, config)
+
+        with mock.patch.object(
+            recorder, '_get_or_create_collector', return_value=collector
+        ):
+            event_dispatcher = EventDispatcher()
```
petyaslavova: I suggest to have some of the test (we can also look at them as examples how some features can be used or need to be configured) without pre-creating the event dispatcher and leave the init method to create the default instance - in this case the end result will be the same - instance of type EventDispatcher

tests/test_pipeline.py
```diff
@@ -454,3 +459,392 @@ def test_pipeline_with_msetex(self, r):
...
+
+        # Verify attributes
+        attrs = call_args[1]['attributes']
+        assert attrs['db.operation.name'] == 'MULTI'
+        assert attrs['db.operation.batch.size'] == 3
+        assert attrs['server.address'] == 'localhost'
+        assert attrs['server.port'] == 6379
+        assert attrs['db.namespace'] == '0'
+
+    def test_pipeline_transaction_emits_multi_command_name(
```
petyaslavova: Isn't this covered in the test above? The difference between those two tests is that the first one sends several commands as transaction and this one sends just one command

tests/test_pipeline.py
```diff
@@ -454,3 +459,392 @@ def test_pipeline_with_msetex(self, r):
...
+
+        recorder.reset_collector()
+
+    def test_empty_pipeline_does_not_emit_event(
+        self, mock_connection_pool, mock_connection, mock_meter
+    ):
+        """
+        Test that an empty pipeline (no commands) does not emit an event.
+        """
+        from redis.client import Pipeline
```
petyaslavova: I think all those imports can be removed - they are already added to the top level imports in the beginning of the file.

tests/test_cluster_transaction.py
```diff
@@ -396,3 +400,258 @@ def test_transaction_discard(self, r):
...
+
+        # Patch the recorder to use our collector
+        with patch.object(
+            recorder,
+            '_get_or_create_collector',
+            return_value=collector
+        ):
+            # Create a new event dispatcher and attach it to the cluster
+            event_dispatcher = EventDispatcher()
+            r._event_dispatcher = event_dispatcher
```
petyaslavova: Is this needed? It should already be initialized. Or it needs to be initialized after OTelConfig?

redis/event.py
vladvildanov: I totally disagree, event name represent an action occurred and event listener represent a reaction on this action that's why I named event listener in more descriptive way to what's actually handled

https://github.com/redis/redis-py/pull/3881/files#diff-8340d4c3344d848477836a04aef890f8bdc064d6be55f938d01432fa97a1b737R487

Event to listener relation is 1:m and multiple listeners could be assigned to single event, that's why event should be fully decoupled from any of the listeners.

I could add a docstring, but in my opinion event name is quite self-descriptive

tests/test_client.py
vladvildanov: You're right, removed it

tests/test_client.py
vladvildanov: I removed all the dispatchers, so now it's relies on constructor instead

tests/test_pipeline.py
vladvildanov: Yeah, nice catch, removed this test

tests/test_pipeline.py
vladvildanov: Done

tests/test_cluster_transaction.py
vladvildanov: It's actually not, the same as it was with standalone client it initialised in constructor

PR #3888, opened 2025-12-16
Added export of resiliency metrics

vladvildanov:
### Description of change

### Pull Request check-list

### DIFF
--- a/redis/event.py
+++ b/redis/event.py
@@ -331,7 +331,7 @@ class OnMaintenanceNotificationEvent:
     Event fired whenever a maintenance notification is received.
     """
     notification: "MaintenanceNotification"
-    connection: Union["MaintNotificationsAbstractConnection"]
+    connection: "MaintNotificationsAbstractConnection"
 
 class AsyncOnCommandsFailEvent(OnCommandsFailEvent):
     pass


redis/observability/recorder.py
```diff
@@ -457,6 +454,45 @@ def record_streaming_lag(
...
+        >>> record_maint_notification_count('localhost', 6379, 'localhost', 6379, 'MOVING')
+    """
+    global _metrics_collector
+
+    if _metrics_collector is None:
+        _metrics_collector = _get_or_create_collector()
+        if _metrics_collector is None:
+            return
+
+    # try:
```
petyaslavova, line 484: If commented try/except block is not needed, it should be removed

redis/retry.py
```diff
@@ -100,7 +100,7 @@ def __eq__(self, other: Any) -> bool:
     def call_with_retry(
         self,
         do: Callable[[], T],
-        fail: Callable[[Exception], Any],
```
petyaslavova, line 103: Would this be considered a breaking change?

redis/client.py
```diff
@@ -974,13 +1016,42 @@ def clean_health_check_responses(self) -> None:
...
+                        duration_seconds=time.monotonic() - start_time,
+                        server_address=conn.host,
+                        server_port=conn.port,
+                        db_namespace=str(conn.db),
+                        error=error,
+                        retry_attempts=failure_count,
+                    )
+                )
+
+        self._event_dispatcher.dispatch(
```
petyaslavova: OnErrorEvent will be registered even if we are not providing error - like in the case when we have marked the connections for reconnect as part of maintenance notification handling and we are just calling reconnect without an error input

redis/connection.py
```diff
@@ -51,7 +51,7 @@
     MaintenanceState,
     MaintNotificationsConfig,
     MaintNotificationsConnectionHandler,
-    MaintNotificationsPoolHandler,
+    MaintNotificationsPoolHandler, MaintenanceNotification,
```
petyaslavova, line 54: Won't this cause linter error? Usually every imported type is on a new line...

redis/connection.py
petyaslavova, line 290: I think that the event_dispatcher should also be added as an argument in the __init__ function, not only to the _configure_maintenance_notifications

redis/connection.py
```diff
@@ -853,14 +858,34 @@ def connect_check_health(
...
         except socket.timeout:
-            raise TimeoutError("Timeout connecting to server")
+            e = TimeoutError("Timeout connecting to server")
+            self._event_dispatcher.dispatch(
+                OnErrorEvent(
+                    error=e,
+                    server_address=self.host,
+                    server_port=self.port,
+                    is_internal=False,
+                    retry_attempts=self.retry.get_retries(),
```
petyaslavova: You can't be sure that this number will be same as the final failure count. If users completely override the retryable exceptions, Timeout errors might not be retried.

redis/connection.py
petyaslavova: Another change that probably will be good to be added is to provide the already initialized event dispatcher when this _init_ is called - this way if custom event dispatcher is used, it will be provided to Mmaint notifications enabled connections as well. It is used when initializing the AbstractConnection and CacheProxyConnection.

redis/observability/recorder.py
vladvildanov: @petyaslavova It's actually needed, we need to handle every exception related to otel metrics export and skip it to not break application flow. I'm keeping it unhandled for now to ensure that it works correctly and I do not miss any exceptions during development, so I'll uncomment it's later

redis/retry.py
vladvildanov: I also have doubts about it, I'll better go safe way and will support both types of fail functions. Otherwise if someone uses `Retry` object outside of redis-py it will end up with exception

redis/client.py
vladvildanov: Ah, missed a tab, thanks

redis/connection.py
vladvildanov: Yeah, maybe, I didn't run linters yet, it's still on development branch

redis/connection.py
vladvildanov: Yeah, you're right, I'll remove it completely. The real retry attempts will pop up in metrics from retry callbacks, we don't need a final one with overall retries

redis/retry.py
vladvildanov: Updated

redis/connection.py
vladvildanov: Done

redis/event.py
```diff
@@ -332,8 +330,8 @@ class OnMaintenanceNotificationEvent:
     """
     Event fired whenever a maintenance notification is received.
     """
-    notification: MaintenanceNotification
-    connection: Union[AbstractConnection, MaintNotificationsAbstractConnection]
+    notification: "MaintenanceNotification"
+    connection: Union["MaintNotificationsAbstractConnection"]
```
petyaslavova: The Union here seems to be redundant

vladvildanov: Removed

PR #3891, opened 2025-12-19
Added export of connection basic metrics

vladvildanov:
### Description of change

### Pull Request check-list

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -2098,8 +2098,12 @@ class MaintNotificationsAbstractConnectionPool:
                     "Maintenance notifications handlers on connection are only supported with RESP version 3"
                 )
 
+            self._event_dispatcher = kwargs.get("event_dispatcher", None)
+            if self._event_dispatcher is None:
+                self._event_dispatcher = EventDispatcher()
+
             self._maint_notifications_pool_handler = MaintNotificationsPoolHandler(
-                self, maint_notifications_config
+                self, maint_notifications_config, self._event_dispatcher
             )
 
             self._update_connection_kwargs_for_maint_notifications(
@@ -2167,7 +2171,7 @@ class MaintNotificationsAbstractConnectionPool:
         # first update pool settings
         if not self._maint_notifications_pool_handler:
             self._maint_notifications_pool_handler = MaintNotificationsPoolHandler(
-                self, maint_notifications_config
+                self, maint_notifications_config, self._event_dispatcher
             )
         else:
             self._maint_notifications_pool_handler.config = maint_notifications_config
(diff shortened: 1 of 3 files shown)


redis/connection.py
```diff
@@ -2060,6 +2063,13 @@ def set_retry(self, retry: Retry):
     def re_auth_callback(self, token: TokenInterface):
         pass
 
+    @abstractmethod
+    def get_connection_count(self) -> list[tuple[int, dict]]:
```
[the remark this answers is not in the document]
vladvildanov: Supported version >= 3.9

redis/connection.py
```diff
@@ -2785,6 +2809,21 @@ async def _mock(self, error: RedisError):
         """
         pass
 
+    def get_connection_count(self) -> list[tuple[int, dict]]:
```
petyaslavova: The BlockingConnectionPool has a different internal structure for free and used connections, so this method should either be implemented there as well, or you should use the exposed methods that will return the lists `_get_free_connections` and `_get_in_use_connections`.

redis/event.py
```diff
@@ -333,6 +345,30 @@ class OnMaintenanceNotificationEvent:
...
+@dataclass
+class AfterConnectionCreatedEvent:
+    """
+    Event fired after connection is created in pool.
+    """
+    connection_pool: "ConnectionPoolInterface"
+    duration_seconds: float
+
+@dataclass
+class AfterConnectionTimeoutRelaxedEvent:
```
petyaslavova: Shouldn't we also have a metric for the unrelaxing of the timeouts?

redis/maint_notifications.py
```diff
@@ -560,13 +561,19 @@ def __init__(
         self,
         pool: "MaintNotificationsAbstractConnectionPool",
         config: MaintNotificationsConfig,
+        event_dispatcher: Optional[EventDispatcherInterface] = None,
```
petyaslavova, line 564: You should add the event_listener argument in the methods where this class is instantiated - in case we want to use the same event_dispatcher object instance.

redis/maint_notifications.py
```diff
@@ -816,3 +838,12 @@ def handle_maintenance_completed_notification(self):
         # timeouts by providing -1 as the relaxed timeout
         self.connection.update_current_socket_timeout(-1)
         self.connection.maintenance_state = MaintenanceState.NONE
+
+        if kwargs.get('notification'):
+            self.connection.event_dispatcher.dispatch(
+                AfterConnectionTimeoutRelaxedEvent(
```
petyaslavova: I think this one should be different event. With the one above we relax the timeouts, with this one we remove the timeout relaxation and revert to default timeouts.

redis/event.py
vladvildanov: Nope, we don't have it

redis/connection.py
vladvildanov: fixed

redis/maint_notifications.py
vladvildanov: Done

redis/maint_notifications.py
vladvildanov: Clarified in the discussion

PR #3899, opened 2025-12-29
Fix handling of circular MOVED redirects in cluster slot mapping

barshaul:
### Description of change

Circular MOVED responses were incorrectly treated as failover events, causing the
primary node to be demoted and leaving shards with replicas only.

This can occur in DNS-based environments where a MOVED response intentionally
points back to the same address (e.g., MOVED + disconnect to force DNS re-resolution).

This change detects circular MOVED redirects and avoids updating the slot map
when the redirected node is already the shard primary.
### Pull Request check-list

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1456,30 +1456,30 @@ class NodesManager:
                 e.host, e.port, PRIMARY, **self.connection_kwargs
             )
             self.set_nodes(self.nodes_cache, {redirected_node.name: redirected_node})
-        if redirected_node in self.slots_cache[e.slot_id]:
-            if redirected_node == self.slots_cache[e.slot_id][0]:
-                # The redirected node is already the primary of this shard. Do nothing.
-                return
-            # The MOVED error resulted from a failover, and the new slot owner
-            # had previously been a replica.
-            old_primary = self.slots_cache[e.slot_id][0]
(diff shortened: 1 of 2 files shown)


petyaslavova:
Hi @barshaul, thank you for your contribution. We will review it soon.

barshaul:
> Hi @barshaul, thank you for your contribution. We will review it soon.

Thanks

redis/asyncio/cluster.py
```diff
@@ -1457,6 +1457,9 @@ def _update_moved_slots(self) -> None:
             )
             self.set_nodes(self.nodes_cache, {redirected_node.name: redirected_node})
         if redirected_node in self.slots_cache[e.slot_id]:
+            if redirected_node == self.slots_cache[e.slot_id][0]:
```
petyaslavova: @barshaul Shouldn't you embrace this in an if-else block and just skip the rest of the handling in case we have same address? 
When you directly return you don't reset the  `self._moved_exception` property.

barshaul: good point, I missed the resetting logic. fixing

barshaul:
@petyaslavova ready for another round

barshaul:
Hey @petyaslavova 
Could this fix be backported to older redis-py versions?

PR #3905, opened 2026-01-06
Fix unrecoverable connection state

### LINKED ISSUE #3904: Unrecoverable connection state where parser raises `RedisError("Buffer is closed.")`
We've observed an issue where an async Redis connection is in a state where:

- `connection.is_connected` returns `True`
- But `connection._parser._connected` is `False`

In this state, every subsequent operation fails with `RedisError: Buffer is closed.`, and the connection never recovers.

### Root cause

1. `ConnectionPool.ensure_connection()` calls `connection.can_read_destructive()` to verify the connection is healthy, which calls `self._parser.can_read_destructive()`.
2. `_AsyncRESPBase.can_read_destructive()` raises `RedisError("Buffer is closed.")` when `parser._connected` is `False`.
3. `ensure_connection()` only catches `ConnectionError`, `TimeoutError`, and `OSError`, but not `RedisError`.
4. Because the exception isn't caught, no recovery is attempted. The broken connection is returned to the pool.
5. On the next request, `ensure_connection()` calls `connection.connect()`, which checks `is_connected` (based on whether `_reader`/`_writer` are set). Since they are still set, `connect()` returns early **without calling `on_connect()`**, so `parser._connected` remains `False`.
6. `can_read_destructive()` raises `RedisError` again. The cycle repeats forever.

### Inconsistency

`_AsyncHiredisParser.can_read_destructive()` raises `ConnectionError` for the same condition, which **would** be caught and recovered. Only `_AsyncRESPBase` raises `RedisError`.

### How the state occurs

The connection must have `_reader`/`_writer` set but `parser._connected` be `False`. I think this might happen if a previous connection was established, then disconnected (setting `_connected = False` and clearing `_reader`/`_writer`), and during reconnection `_connect()` set `_reader`/`_writer` but `on_connect()` was never called due to an interruption (e.g., `CancelledError` between `_connect()` completing and `on_connect_check_health()` being called).

With Sentinel connections, `connect_check_health()` is called with `retry_socket_connect=False`, meaning exceptions during `_connect()` don't trigger cleanup, increasing the likelihood of this inconsistent state.

itssimon:
### Description of change

Fixes #3904.

### Pull Request check-list

### DIFF
--- a/redis/asyncio/multidb/client.py
+++ b/redis/asyncio/multidb/client.py
@@ -1,18 +1,29 @@
 import asyncio
 import logging
-from typing import Any, Awaitable, Callable, Coroutine, List, Optional, Union
+from typing import Any, Awaitable, Callable, List, Optional, Union
 
 from redis.asyncio.client import PubSubHandler
 from redis.asyncio.multidb.command_executor import DefaultCommandExecutor
-from redis.asyncio.multidb.config import DEFAULT_GRACE_PERIOD, MultiDbConfig
-from redis.asyncio.multidb.database import AsyncDatabase, Databases
+from redis.asyncio.multidb.config import (
+    DEFAULT_GRACE_PERIOD,
+    DatabaseConfig,
+    InitialHealthCheck,
+    MultiDbConfig,
+)
+from redis.asyncio.multidb.database import AsyncDatabase, Database, Databases
 from redis.asyncio.multidb.failure_detector import AsyncFailureDetector
 from redis.asyncio.multidb.healthcheck import HealthCheck, HealthCheckPolicy
+from redis.asyncio.retry import Retry
 from redis.background import BackgroundScheduler
+from redis.backoff import NoBackoff
 from redis.commands import AsyncCoreCommands, AsyncRedisModuleCommands
 from redis.multidb.circuit import CircuitBreaker
 from redis.multidb.circuit import State as CBState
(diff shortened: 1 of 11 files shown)


petyaslavova:
Hi @itssimon, thank you for your contribution! We will review it soon.

redis/_parsers/base.py
```diff
@@ -426,7 +425,7 @@ def on_disconnect(self):
 
     async def can_read_destructive(self) -> bool:
         if not self._connected:
-            raise RedisError("Buffer is closed.")
+            raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
```
petyaslavova: I think that raising an OSError would be better here. This method is called in the AbstractConnection's can_read_destructive where OSError is caught and a ConnectionError is raised with additional info about the host and port.

itssimon: Makes sense. Updated.
Note that `_AsyncHiredisParser` raises `ConnectionError` for the same condition. Might want to update that too for consistency. Happy to do that in this PR as well.

petyaslavova: @itssimon yeah, it will be great, thank you! 
Just use this error only for the `can_read_destructive` method.
`on_connect` should still raise ConnectionError as it is used in a different type of context.

itssimon: All done!

Issue #3867, opened 2025-12-05
sentinel client should listen for switch-master


tonyay163:
After a failover, we rely on a retry to pick up the new master. It seems like other client libraries will listen for the switch-master and automatically point to it. This will speed up failover recovery

See https://github.com/redis/go-redis/blob/16b55f9519d46711228a1b81792e971542557ca6/sentinel.go#L1110-L1118

petyaslavova:
Hi @tonyay163, thank you for bringing this up! We will take a look.

ManelCoutinhoSensei:
Hey @petyaslavova , I implemented a draft solution for this with #3915 .
Let me know if this is a valid path and I should implement the async version and tests as well or if we don't want to take this path.
This PR would also be the 1st step to solve #3543

Issue #3543, opened 2025-03-05
Missing keyspace notifications after master failover


ManelCoutinhoSensei:
### Expected behavior
When a Redis master fails over in a Sentinel-managed cluster, the client should continue receiving keyspace notifications from the new master without requiring any additional configuration.

### Actual behavior
Keyspace notifications from the new master are lost after master fails over.

### Reproduce behavior

1. Create a Redis Sentinel cluster using something like the following `docker-compose.yml`:
```yaml
services:

  # Redis Node #1 (initial master) + Sentinel
  redis-node-1:
    container_name: redis_node_1
    image: bitnami/redis:7.2.4
    environment:
      - ALLOW_EMPTY_PASSWORD=yes
      - REDIS_AOF_ENABLED=no
    ports:
      - 6380:6379
    networks:
      - nw

  redis-node-1-sentinel:
    container_name: redis-node-1-sentinel
    image: bitnami/redis-sentinel:7.2.4
    depends_on:
      - redis-node-1
    environment:
      - REDIS_MASTER_HOST=redis-node-1
      - REDIS_SENTINEL_MASTER_NAME=mymaster
      - REDIS_SENTINEL_DOWN_AFTER_MILLISECONDS=5000
      - REDIS_SENTINEL_FAILOVER_TIMEOUT=10000
      - REDIS_SENTINEL_QUORUM=2
    ports:
      - 36380:26379
    networks:
      - nw

  # Redis Node #2  + Sentinel  
  redis-node-2:
    container_name: redis_node_2
    image: bitnami/redis:7.2.4
    environment:
      - ALLOW_EMPTY_PASSWORD=yes
      - REDIS_REPLICATION_MODE=slave
      - REDIS_MASTER_HOST=redis-node-1
      - REDIS_AOF_ENABLED=no
    ports:
      - 6381:6379
    networks:
      - nw

  redis-node-2-sentinel:
    container_name: redis_node_2_sentinel
    image: bitnami/redis-sentinel:7.2.4
    depends_on:
      - redis-node-2
    environment:
      - REDIS_MASTER_HOST=redis-node-1
      - REDIS_SENTINEL_MASTER_NAME=mymaster
      - REDIS_SENTINEL_DOWN_AFTER_MILLISECONDS=5000
      - REDIS_SENTINEL_FAILOVER_TIMEOUT=10000
      - REDIS_SENTINEL_QUORUM=2
    ports:
      - 36381:26379
    networks:
      - nw

[...]

PR #3915, opened 2026-01-19
Add Sentinel +switch-master linstener option to monitor failovers.

ManelCoutinhoSensei:
### Description of change

Fix for #3867 inspired by the [redis-go](https://github.com/redis/go-redis/blob/16b55f9519d46711228a1b81792e971542557ca6/sentinel.go#L1110-L1118) library.
This solution adds a thread that cycles through valid sentinel clients to listen for `+switch-master` events whenever a failover happens and the user decided to monitor it.

It also provides a first step to implement a solution for #3543.

### Pull Request check-list

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -1,5 +1,8 @@
 import random
+import time
 import weakref
+from collections import defaultdict
+from threading import Event, Lock, Thread
 from typing import Optional
 
 from redis.client import Redis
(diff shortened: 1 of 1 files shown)


petyaslavova:
Hi @ManelCoutinhoSensei, thank you for your contribution! We will have a look at it.

eeshsaxena (commented):
Read through the failover listener and it is put together well. Things I checked:

- The `+switch-master` parsing matches the message format (`<name> <old-ip> <old-port> <new-ip> <new-port>`), so `parts[0]`/`parts[3]`/`parts[4]` pick the service name and new address, with the `len(parts) < 5` guard and an `int()` guard on the port.
- `_on_switch_master` snapshots the WeakSet with `list(proxies)` and uses `.get(service_name)` so it does not create empty defaultdict entries, and the run loop re-subscribes on `ConnectionError`/`TimeoutError` with pubsub cleanup in `finally`.
- `update_master_address` does the compare-and-set under `self._lock` and disconnects outside the lock, which is the right ordering.

Two smaller things worth a look:

1. The listener passes the raw new IP from the `+switch-master` payload straight to `proxy.update_master_address`, which skips the `force_master_ip` normalization that `discover_master` applies (around L354). With `force_master_ip` set, a failover would set `master_address` to the real new IP, then the next `get_master_address` re-applies the forced IP and disconnects again, so it self-corrects but with an extra disconnect and a transient non-forced address. Routing the failover update through the same forced-ip logic (or just re-running `discover_master` for that service instead of trusting the payload IP) would keep the two paths in agreement.

2. Tiny doc typo: the `monitor_failover` docstring has a stray combining grave accent on the first letter, so it renders as `m̀onitor_failover`.

PR #3925, opened 2026-01-29
Refactoring the SMIGRATED flow - the notification is changed to contain the src node address for each slot range movement.

petyaslavova:
### Description of change

### Pull Request check-list

### DIFF
--- a/redis/_parsers/base.py
+++ b/redis/_parsers/base.py
@@ -11,7 +11,6 @@ from redis.maint_notifications import (
     NodeMigratedNotification,
     NodeMigratingNotification,
     NodeMovingNotification,
-    NodesToSlotsMapping,
     OSSNodeMigratedNotification,
     OSSNodeMigratingNotification,
 )
@@ -193,22 +192,26 @@ class MaintenanceNotificationsParser:
     @staticmethod
     def parse_oss_maintenance_completed_msg(response):
         # Expected message format is:
-        # SMIGRATED <seq_number> [<host:port> <slot, range1-range2,...>, ...]
+        # SMIGRATED <seq_number> [[<src_host:port> <dest_host:port> <slot_range>], ...]
         id = response[1]
         nodes_to_slots_mapping_data = response[2]
-        nodes_to_slots_mapping = []
-        for src_node, node, slots in nodes_to_slots_mapping_data:
-            # Parse the node address to extract host and port
+        # Build the nodes_to_slots_mapping dict structure:
+        # {
+        #     "src_host:port": [
+        #         {"dest_host:port": "slot_range"},
+        #         ...
+        #     ],
+        #     ...
+        # }
+        nodes_to_slots_mapping = {}
+        for src_node, dest_node, slots in nodes_to_slots_mapping_data:
             src_node_str = safe_str(src_node)
-            node_str = safe_str(node)
+            dest_node_str = safe_str(dest_node)
             slots_str = safe_str(slots)
 
-            mapping = NodesToSlotsMapping(
-                src_node_address=src_node_str,
-                dest_node_address=node_str,
-                slots=slots_str,
-            )
-            nodes_to_slots_mapping.append(mapping)
+            if src_node_str not in nodes_to_slots_mapping:
+                nodes_to_slots_mapping[src_node_str] = []
+            nodes_to_slots_mapping[src_node_str].append({dest_node_str: slots_str})
 
         return OSSNodeMigratedNotification(id, nodes_to_slots_mapping)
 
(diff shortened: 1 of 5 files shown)


redis/_parsers/base.py
```diff
@@ -341,17 +353,20 @@ def handle_push_response(self, response, **kwargs):
...
                     msg_type
                 ][1]
                 notification = parser_function(response)
 
                 if notification is not None:
-                    return self.oss_cluster_maint_push_handler_func(notification)
+                    if self.maintenance_push_handler_func:
+                        self.maintenance_push_handler_func(notification)
+                    if self.oss_cluster_maint_push_handler_func:
+                        self.oss_cluster_maint_push_handler_func(notification)
```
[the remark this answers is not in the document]
petyaslavova: the order is as intended.

redis/maint_notifications.py
```diff
@@ -463,15 +471,15 @@ class OSSNodeMigratedNotification(MaintenanceNotification):
 
     Args:
         id (int): Unique identifier for this notification
-        nodes_to_slots_mapping (Dict[str, str]): Mapping of node addresses to slots
+        nodes_to_slots_mapping (List[NodesToSlotsMapping]): List of node-to-slots mappings
     """
 
-    DEFAULT_TTL = 30
+    DEFAULT_TTL = 120
```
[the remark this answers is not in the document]
petyaslavova: It is intentional - some connections might read the notifications with delay, so we can consider the notification valid for longer in order not to update the cluster topology without needed, because it is an expensive operation.

redis/maint_notifications.py
```diff
@@ -1011,77 +1029,58 @@ def handle_oss_maintenance_completed_notification(
...
+                            conn.mark_for_reconnect()
+
+                    if (
+                        current_node
+                        not in self.cluster_client.nodes_manager.nodes_cache.values()
+                    ):
+                        # disconnect all free connections to the node - this node will be dropped
+                        # from the cluster, so we don't need to revert the timeouts
+                        for conn in current_node.redis_connection.connection_pool._get_free_connections():
+                            conn.disconnect()
```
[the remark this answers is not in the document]
petyaslavova: The unrelaxing of the timeouts is now handled on connection level. This part needed to be redesigned.

redis/maint_notifications.py
```diff
@@ -454,6 +455,13 @@ def __hash__(self) -> int:
         return hash((self.__class__.__name__, int(self.id)))
 
 
+@dataclass
+class NodesToSlotsMapping:
```
vladvildanov: Does it really requires separate class for this mapping? Maybe it makes sense to use more primitive type like  `Mapping[...]`

redis/maint_notifications.py
```diff
@@ -463,15 +471,15 @@ class OSSNodeMigratedNotification(MaintenanceNotification):
 
     Args:
         id (int): Unique identifier for this notification
-        nodes_to_slots_mapping (Dict[str, str]): Mapping of node addresses to slots
+        nodes_to_slots_mapping (List[NodesToSlotsMapping]): List of node-to-slots mappings
```
vladvildanov: This shouldn't be a list, we can keep all mappings in the single mapping and access it by key

vladvildanov (changes requested):

redis/maint_notifications.py
petyaslavova: @vladvildanov In the notification we have a list with multiple slots/slot ranges that are being moved - in this list we can have the src multiple times, as well as the destination. 
For example `[ [src1 dest1 slot-range-1] [src2 dest 1 slot-range-2] [src1 dest3 slot-range3] ]` - with this structure using just map I should have the src as key and a list of dicts for the values and append for each element containing the same source. 
```
{
   "src1":[
       {"dest1":"slot-range-1"}, 
       {"dest3":"slot-range-3"}
   ],
   "src2": ...
}
```
I believe with the additional object it is easier to understand what data we operate with and is less error prone.

vladvildanov: So this map is exactly what I'm talking about, isn't it more natural to group it in a map so it will be easier to access specific slot ranges? The naming itself is confusing,`nodes_to_slots_mapping ` makes me feel that It needs to be a map and not list

redis/maint_notifications.py
petyaslavova: Removed.

redis/maint_notifications.py
petyaslavova: Changed.

vladvildanov (approved):

PR #3938, opened 2026-02-05
Disabled SCH in MultiDBClient underlying clients

vladvildanov:
### Description of change

### Pull Request check-list

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -78,7 +78,22 @@ runs:
         echo "::endgroup::"
 
         echo "::group::Starting Redis servers"
-        redis_major_version=$(echo "$REDIS_VERSION" | grep -oP '^\d+')
+        # Check if REDIS_VERSION is in the custom map
+        mapped_version=""
+        if [[ -n "${REDIS_VERSION_CUSTOM_MAP:-}" ]]; then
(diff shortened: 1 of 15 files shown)


vladvildanov:
augment review

redis/multidb/config.py
```diff
@@ -164,8 +165,15 @@ def databases(self) -> Databases:
         for database_config in self.databases_config:
             # The retry object is not used in the lower level clients, so we can safely remove it.
             # We rely on command_retry in terms of global retries.
+            #
+            # Maintenance notifications are temporarily disabled in underlying clients.
             database_config.client_kwargs.update(
-                {"retry": Retry(retries=0, backoff=NoBackoff())}
+                {
+                    "retry": Retry(retries=0, backoff=NoBackoff()),
+                    "maint_notifications_config": MaintNotificationsConfig(
```
petyaslavova: This would replace the configuration with disabled, even when customers want it to be enabled. 
This morning at the hitless meeting, we discussed a different approach: the config should be disabled by default, but still allowed to be provided and used. 
If the customer is ok with timeouts relaxation - he/she should be still able to enabled it and use the feature.

vladvildanov: Okay, I had an assumption that we don't want to expose it at all

PR #3940, opened 2026-02-06
Test/aa hitless combined

kiryazovi-redis:
## Summary

Two mirror tests proving that SCH timeout relaxation is the decisive factor in preventing Circuit Breaker trips during network latency on an OSS Active-Active cluster.

- **Test 1** (`relaxed_timeout=5`): During SMIGRATING, the effective timeout (5s) exceeds the 2s network latency, so commands succeed and the CB stays closed. No AA failover.
- **Test 2** (`relaxed_timeout=-1`): The timeout stays at 1s < 2s latency. Health checks fail, the CB trips, and AA failover occurs.

### What changed

- Added `TestActiveActiveWithHitless` class with two parametrized tests in `test_active_active.py`
- Added `r_multi_db_with_hitless` fixture in `conftest.py` (MultiDBClient with MaintNotificationsConfig + CB thresholds)
- Added `NETWORK_LATENCY` action type and `trigger_network_latency()` to `REFaultInjector` in `fault_injector_client.py`
- Added `trigger_name="migrate"` to `ClusterOperations.trigger_effect` calls for SLOT_SHUFFLE

### Test parameters

| Parameter | Value |
|---|---|
| `socket_timeout` | 1s |
| `network_latency` | 2000ms / 60s duration |
| `data_preload` | 100 keys x 100KB (~10MB) |
| `min_num_failures` | 20 |
| `failure_rate_threshold` | 0.90 |
| `relaxed_timeout` | 5 (Test 1) / -1 (Test 2) |

### Prerequisites

- OSS AA database (`re-active-active-oss-cluster`) with `shards_count: 6`, `shards_placement: sparse` across 3 nodes
- Fault Injector with `network_latency` and `slot_migrate` actions available

## Test plan

- [x] Test 1 passes: CB does not trip with relaxation enabled
- [x] Test 2 passes: CB trips with relaxation disabled
- [ ] Run both tests back-to-back on a fresh CRDB deployment
- [x] Verify no leftover `tc netem` rules on cluster nodes after tests complete

---

### DIFF
--- a/tests/test_scenario/conftest.py
+++ b/tests/test_scenario/conftest.py
@@ -35,6 +35,62 @@ DEFAULT_ENDPOINT_NAME = "m-standard"
 DEFAULT_OSS_API_ENDPOINT_NAME = "maint-notifications-oss-api"
 
 
+def _wait_for_cluster_healthy_in_fixture(
+    host, port, username, password, timeout=120, ping_threshold_ms=1500, required_streak=3, use_ssl=False
+):
+    """
+    Poll cluster until PINGs respond within a reasonable time.
+    Used by fixtures to ensure cluster is healthy before creating clients.
+    """
+    from time import sleep, time
+
+    logging.info("Waiting for cluster to become healthy (timeout=%ds, ssl=%s)", timeout, use_ssl)
+    start = time()
+    healthy_streak = 0
+
+    while time() - start < timeout:
+        try:
+            kwargs = {
+                "host": host,
+                "port": port,
+                "username": username,
+                "password": password,
+                "socket_timeout": 5,
+                "protocol": 3,
+            }
+            if use_ssl:
+                kwargs["ssl"] = True
+                kwargs["ssl_cert_reqs"] = "none"
+                kwargs["ssl_check_hostname"] = False
+            cluster = RedisCluster(**kwargs)
+            ping_start = time()
+            cluster.ping()
+            ping_ms = (time() - ping_start) * 1000
+            cluster.close()
+
+            if ping_ms < ping_threshold_ms:
+                healthy_streak += 1
+                logging.info(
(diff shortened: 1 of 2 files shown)


kiryazovi-redis:
This is not meant to be merged, I am just putting it cus I worked briefly (~day) to test if AA and SCH work in OSS mode.
It only scratches the surface of possible interactions between the two features, so in on way does it represent "both work together in all cases.
Examples of issues encountered during writing it:


1. 3000ms latency broke the FI's ability to reach the cluster
The network_latency action applies tc netem delay on ALL egress traffic from ALL cluster nodes. With a single RQ worker, network_latency runs first, then slot_migrate tries the REST API and fails with Name or service not known DNS errors. The 3000ms blanket latency destabilized the cluster's internal communication (node gossip, health checks), likely causing health-check-based DNS to return NXDOMAIN. Fixed by reducing latency to 2000ms and socket_timeout to 1s, which kept the cluster healthy enough while still exceeding the client timeout.
2. CB tripped in Test 1 after migration completed
The latency (60s duration) outlasted the SMIGRATING window (~20-30s). After SMIGRATED reverted the timeout to 1s, commands during the remaining latency all timed out, pushing the failure rate above 0.90. Fixed by adding a threading.Event (migration_complete) so the command loop in Test 1 stops as soon as slot_shuffle finishes -- only measuring behavior during the SMIGRATING window.
3. assert not errors.empty() failed in Test 2
The CB tripped via background health check probes (not via command-level exceptions caught by the test loop). The MultiDBClient's internal health check thread detected the database as unreachable and tripped the CB before commands raised visible exceptions. "Fixed" by removing the errors.empty() assertion -- is_changed_flag == True is the meaningful check.
This for example should be investigated with another test.

Anyway due to:
https://github.com/redis/redis-py/pull/3938
This option to use both will be disabled by default with opt-in by client

PR #3959, opened 2026-02-17
Add port feature for unix socket for OTel instrumentation compatibility

### LINKED ISSUE #3957: Regression from #3954
I found a regression with #3954. When using a unix domain socket via a URL scheme (specifcially `unix:///run/redis/redis.sock?db=0`), there is an excpetion raised, since there is no `port` for unix socket connections.

```
   File "/opt/netbox/venv/lib/python3.13/site-packages/redis/commands/core.py", line 2069, in get
     return self.execute_command("GET", name, keys=[name])
            ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/opt/netbox/venv/lib/python3.13/site-packages/redis/client.py", line 716, in execute_command
     return self._execute_command(*args, **options)
            ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
   File "/opt/netbox/venv/lib/python3.13/site-packages/redis/client.py", line 755, in _execute_command
     server_port=conn.port,
                 ^^^^^^^^^
 AttributeError: 'UnixDomainSocketConnection' object has no attribute 'port'
```

This was discovered when using `django-redis` for caching in `netbox`.
Downgrading to `7.1.1` solves this issue.

### LINKED ISSUE #3958: AttributeError when using Unix sockets with OTel metrics (v7.2.0)
After upgrading to redis-py 7.2.0 (released Feb 16, 2026), any Redis operations using Unix domain socket connections crash with:

```
AttributeError: 'UnixDomainSocketConnection' object has no attribute 'port'
```

Seems like a regression introduced by the new OpenTelemetry native metrics support added in #3954. The OTel instrumentation added accesses `conn.host` and `conn.port` unconditionally when recording metrics:

```python
# redis/client.py - multiple locations
record_operation_duration(
    command_name=command_name,
    duration_seconds=time.monotonic() - start_time,
    server_address=conn.host,
    server_port=conn.port,  # <-- AttributeError for UnixDomainSocketConnection - no port attribute
    db_namespace=str(conn.db),
)
```

### Environment

1. Redis server listening on a Unix socket (not TCP)
2. redis-py 7.2.0
3. Execute any Redis command (e.g., `PING`)

### Reproduction Script

Here's the Python script to reproduce the issue:

```python
"""
Requirements:
- Redis server running with: redis-server --unixsocket /tmp/redis-test.sock --unixsocketperm 777 --port 0 --daemonize yes
- redis-py 7.2.0: pip install redis==7.2.0
"""
import redis
def main():
    r = redis.Redis(
        unix_socket_path='/tmp/redis-test.sock',
        encoding='utf-8',
        decode_responses=True
    )
    
    print(f'redis-py version: {redis.__version__}')
    print(f'Connection class: {r.connection_pool.connection_class.__name__}')
    print()
    print('Executing PING command...')
    
    # This triggers _execute_command() which crashes at line 748:
    #   server_port=conn.port,  # AttributeError!
    result = r.ping()
    print(f'Result: {result}')
if __name__ == '__main__':
    main()
```

Bojun-Feng:
### Description of change

Fix #3957 
Fix #3958 

Add `host` and `port` properties to `UnixDomainSocketConnection` to fix `AttributeError: 'UnixDomainSocketConnection' object has no attribute 'port'` when using Unix sockets with the new OTel metrics introduced in v7.2.0.

### Pull Request check-list

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1016,6 +1016,17 @@ class UnixDomainSocketConnection(AbstractConnection):
         self.path = path
         super().__init__(**kwargs)
 
+    # host and port properties are required for OTel instrumentation compatibility.
+    # TCP Connection has these as attributes; Unix sockets use path instead.
+
+    @property
+    def host(self):
+        return self.path
+
+    @property
+    def port(self):
+        return None
+
     def repr_pieces(self) -> Iterable[Tuple[str, Union[str, int]]]:
         pieces = [("path", self.path), ("db", self.db)]
         if self.client_name:
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -2090,6 +2090,17 @@ class UnixDomainSocketConnection(AbstractConnection):
         self.path = path
         self.socket_timeout = socket_timeout
 
+    # host and port properties are required for OTel instrumentation compatibility.
+    # TCP Connection has these as attributes; Unix sockets use path instead.
+
+    @property
+    def host(self):
+        return self.path
+
+    @property
+    def port(self):
+        return None
+
     def repr_pieces(self):
         pieces = [("path", self.path), ("db", self.db)]
         if self.client_name:


Bojun-Feng:
Ran the connection tests (tests/test_connection.py) and observability tests (tests/test_observability/) with a local Redis server. All 47 connection tests passed (1 skipped due to missing `HiredisParser`), and all 143 observability tests passed. Also verified manually that both sync and async `UnixDomainSocketConnection` now correctly return path for host and None for port. End-to-end verification confirmed the fix resolves the AttributeError when using Unix sockets.

Cornelicorn:
Also fixes #3957

petyaslavova:
Hi @Bojun-Feng, thank you for your contribution!

I’ve also been working on a few related improvements around attribute access when extracting metrics. We still need to verify whether the host and port attributes exist, since someone might implement either `ConnectionInterface` or `AbstractConnection` and configure a Redis client with a `connection_class` that doesn’t define `host` or `port`. I’m addressing that in a separate PR.

Given that, it won’t be necessary to provide both properties here. Also, since port is not relevant when using a Unix socket path, I suggest we skip it for now. You can keep only the host property for both sync and async implementations.

Bojun-Feng:
Hi @petyaslavova , thank you for the response.

I see https://github.com/redis/redis-py/pull/3964 is a better fix. Closing this down.

petyaslavova:
Hi @Bojun-Feng, my PR isn't covering the Unix socket connection's path attribute to be provided as host and this way we are losing those connectionsd to have their data reported. 
So, your solution was also a good addition, and the only change it needed was to leave only the host property. 
If you are willing to add it again, it will be great :)

PR #3965, opened 2026-02-20
Fix protocol validation: replace finally with else and store parsed int

bysiber:
Two issues in the protocol validation logic during connection initialization:

**1. `finally` block causes `UnboundLocalError` when `ValueError` is raised**

When `protocol` is a non-numeric string like `"abc"`, `int(protocol)` raises `ValueError`. The `except ValueError` clause re-raises as `ConnectionError`. But the `finally` block always executes — and `p` was never assigned, so `p < 2` raises `UnboundLocalError`, replacing the meaningful error message with a confusing one.

Changed `finally` to `else` so the range check only runs after successful parsing.

**2. Async connection stores raw `protocol` instead of parsed `p`**

The sync version correctly does `self.protocol = p` (the validated integer), but the async version had `self.protocol = protocol` (the raw input, which could be a string). This forces workarounds throughout the async codebase (`self.protocol not in [2, "2"]`, `int(self.protocol)`, etc.). Fixed to store the parsed integer `p`, consistent with the sync version.

### DIFF
--- a/redis/_parsers/resp3.py
+++ b/redis/_parsers/resp3.py
@@ -29,17 +29,23 @@ class _RESP3Parser(_RESPBase, PushNotificationsParser):
         return response
 
     def read_response(self, disable_decoding=False, push_request=False):
-        pos = self._buffer.get_pos() if self._buffer else None
+        pos = self._buffer.get_pos() if self._buffer is not None else None
         try:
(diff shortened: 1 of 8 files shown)


petyaslavova (approved):
LGTM.

PR #3954, opened 2026-02-12
Added OTel instrumentation and metrics export for sync client

vladvildanov:
### Description of change

This PR adds optional OpenTelemetry (OTel) instrumentation to redis-py, enabling automatic metrics collection and export for Redis operations. The implementation follows a singleton pattern where observability is initialised once globally, and all Redis clients automatically collect metrics without additional configuration.

It provides:
- Automatic metrics collection for commands, connections, client-side caching, pub/sub, and streaming operations
- Configurable metric groups to enable/disable specific categories
- Privacy controls to hide sensitive information (channel names, stream names)
- Zero overhead when disabled - no performance impact if OTel dependencies aren't installed
- Uses global MeterProvider - integrates with your existing OTel setup

### How to Enable

Install OTel dependencies:

```
pip install redis[otel]
```

Set up your OTel exporter and initialize redis-py observability:

```python
from opentelemetry import metrics
from opentelemetry.sdk.metrics import MeterProvider
from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
from opentelemetry.exporter.otlp.proto.http.metric_exporter import OTLPMetricExporter
```

##### Configure your OTel exporter
```python
exporter = OTLPMetricExporter(endpoint='http://localhost:4318/v1/metrics')
reader = PeriodicExportingMetricReader(exporter, export_interval_millis=10000)
provider = MeterProvider(metric_readers=[reader])
metrics.set_meter_provider(provider)
```
##### Initialize redis-py observability (once at app startup)

```python
from redis.observability.providers import get_observability_instance
from redis.observability.config import OTelConfig, MetricGroup

otel = get_observability_instance()
otel.init(OTelConfig())
```

##### All Redis clients now automatically collect metrics

```python
import redis
r = redis.Redis(host='localhost', port=6379)
r.set('key', 'value')  # Metrics collected automatically
```

##### Optional: Customise metric groups:

```python
from redis.observability.config import MetricGroup

otel = get_observability_instance()
otel.init(OTelConfig(
    metric_groups=[
        MetricGroup.COMMAND,
        MetricGroup.PUBSUB,
        MetricGroup.STREAMING,
        MetricGroup.CSC,
        MetricGroup.CONNECTION_BASIC,
        MetricGroup.CONNECTION_ADVANCED,
        MetricGroup.RESILIENCY,
    ]
))
```

### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Do tests and lints pass with this change?
[...]

PR #3977, opened 2026-02-24
Added OTel instrumentation and metrics export for async client

vladvildanov:
### Description of change

This PR provides recording of metrics for async clients. In terms of public API nothing has been changed so you can use #3954 description or [official docs](https://redis.readthedocs.io/en/stable/opentelemetry.html#native-opentelemetry-integration-recommended) to understand the  OTel configuration flow.

We only bring new async recorder functions that is used across async clients to collect the same metrics (where it's possible, CSC and SCH are not available via async clients) 

### Pull Request check-list

---

> [!NOTE]
> **Medium Risk**
> Touches async command execution, retry callbacks, and connection lifecycle paths to emit metrics, which could impact error handling and performance under load. Changes are mostly additive/instrumentation but are in high-traffic code paths (client/cluster/connection/pipeline).
> 
> **Overview**
> Adds a new async observability layer (`redis.asyncio.observability.recorder`) and wires it into asyncio `Redis`, `PubSub`, cluster, pipelines/transactions, connection pools, and connection close/health-check paths to record **operation durations**, **errors/retries**, **pubsub receive events**, **streaming lag**, **connection create/wait/close metrics**, and **geo failover metrics**.
> 
> Updates async retry to optionally pass a failure count (`with_failure_count`) and uses it throughout async code to report retry attempts, while also extending sync streaming commands to export lag metrics for both sync and async (wrapping coroutine responses). Adds an async mode to `benchmarks/otel_benchmark.py` and new/updated asyncio tests validating metric emission and required connection metadata.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0e7d587c060dd6f904e0f832fd28de1a279551fc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -738,7 +738,11 @@ class Redis(
         After we disconnect the connection, it will try to reconnect and
         do a health check as part of the send_command logic(on connection level).
         """
-        if error and failure_count is not None and failure_count <= conn.retry.get_retries():
+        if (
+            error
+            and failure_count is not None
+            and failure_count <= conn.retry.get_retries()
+        ):
             await record_operation_duration(
                 command_name=command_name,
                 duration_seconds=time.monotonic() - start_time,
@@ -1101,7 +1105,11 @@ class PubSub:
 
         In this error handler we are trying to reconnect to the server.
         """
-        if error and failure_count is not None and failure_count <= conn.retry.get_retries():
+        if (
+            error
+            and failure_count is not None
+            and failure_count <= conn.retry.get_retries()
+        ):
             if command_name:
                 await record_operation_duration(
                     command_name=command_name,
@@ -1591,7 +1599,11 @@ class Pipeline(Redis):  # lgtm [py/init-calls-subclass]
         After we disconnect the connection, it will try to reconnect and
         do a health check as part of the send_command logic(on connection level).
         """
-        if error and failure_count is not None and failure_count <= conn.retry.get_retries():
+        if (
+            error
+            and failure_count is not None
+            and failure_count <= conn.retry.get_retries()
+        ):
             await record_operation_duration(
                 command_name=command_name,
                 duration_seconds=time.monotonic() - start_time,
@@ -1836,7 +1848,11 @@ class Pipeline(Redis):  # lgtm [py/init-calls-subclass]
         After we disconnect the connection, it will try to reconnect and
         do a health check as part of the send_command logic(on connection level).
         """
-        if error and failure_count is not None and failure_count <= conn.retry.get_retries():
+        if (
+            error
+            and failure_count is not None
+            and failure_count <= conn.retry.get_retries()
+        ):
             await record_operation_duration(
                 command_name=command_name,
                 duration_seconds=time.monotonic() - start_time,


redis/asyncio/observability/recorder.py
```diff
@@ -0,0 +1,597 @@
...
+    from redis.asyncio.multidb.database import AsyncDatabase
+    from redis.observability.config import OTelConfig
+
+# Global metrics collector instance (lazy-initialized)
+_async_metrics_collector: Optional[RedisMetricsCollector] = None
+
+CONNECTION_COUNT_REGISTRY_KEY = "connection_count"
+
+
+async def _get_or_create_collector() -> Optional[RedisMetricsCollector]:
```
petyaslavova: Do we need this to be async? I don't see any async method used inside.

redis/asyncio/observability/recorder.py
```diff
@@ -0,0 +1,597 @@
...
+    try:
+        manager = get_observability_instance().get_provider_manager()
+        if manager is None:
+            return None
+        return manager.config
+    except Exception:
+        return None
+
+
+async def record_operation_duration(
```
petyaslavova, line 95: This function is defined as async, but in practice there is currently no opportunity for a context switch during its execution.
Although it awaits _get_or_create_collector, that function does not perform any asynchronous operations at the moment. As a result, it executes fully without yielding control, and the code then continues synchronously with recording the data.

The only reason I can see for keeping these functions async is to preserve an interface that allows us to introduce asynchronous calls in the future without causing breaking changes.

redis/asyncio/client.py
```diff
@@ -714,7 +721,14 @@ async def _send_command_parse_response(self, conn, command_name, *args, **option
         await conn.send_command(*args)
         return await self.parse_response(conn, command_name, **options)
 
-    async def _close_connection(self, conn: Connection):
+    async def _close_connection(
+        self,
+        conn: Connection,
```
petyaslavova, line 726: Based on the provided initialization parameters, we may end up creating a UnixDomainSocketConnection, which derives from AbstractConnection.

redis/asyncio/client.py
```diff
@@ -724,7 +738,18 @@ async def _close_connection(self, conn: Connection):
         After we disconnect the connection, it will try to reconnect and
         do a health check as part of the send_command logic(on connection level).
         """
-        await conn.disconnect()
+        if error and failure_count <= conn.retry.get_retries():
+            await record_operation_duration(
+                command_name=command_name,
+                duration_seconds=time.monotonic() - start_time,
+                server_address=conn.host,
```
petyaslavova: Should be read conditionally - the connection might not have host and port.
There are several other places this should be done as well.

redis/asyncio/client.py
```diff
@@ -1239,6 +1355,13 @@ async def handle_message(self, response, ignore_subscribe_messages=False):
                 "data": response[2],
             }
 
+        if message_type in ["message", "pmessage"]:
+            channel = str_if_bytes(message["channel"])
```
petyaslavova, line 1369: The channel name might not be in a valid-UTF-8 format. There was a PR merged recently fixing resubscribe issue related exactly to this type of channel names(#3944). 
Maybe some error handling can be added here for such cases.

redis/asyncio/cluster.py
```diff
@@ -817,6 +821,52 @@ def _parse_target_nodes(self, target_nodes: Any) -> List["ClusterNode"]:
             )
         return nodes
 
+    async def _record_error_metric(
+        self,
+        error: Exception,
+        connection: Union[Connection, "ClusterNode"],
```
petyaslavova, line 827: Just a note here: the cluster init interface basically doesn't allow connections of type unix socket connections - in this class we are safe to use directly `con.host` and `con.port`

redis/asyncio/connection.py
```diff
@@ -325,27 +340,59 @@ async def connect(self):
...
+            with_failure_count=True,
         )
 
     async def connect_check_health(
         self, check_health: bool = True, retry_socket_connect: bool = True
     ):
         if self.is_connected:
             return
+        # Track actual retry attempts for error reporting
+        actual_retry_attempts = [0]
```
petyaslavova: Can't this just be an int? Now it looks a little strange - first we have a property initialised as list containing just one element - number 0. Then we use it as dict with one key - number 0 and assign the actual retry attempts count to it..

redis/asyncio/observability/recorder.py
vladvildanov: Exactly, async recorders exists just for interface separation. For example, some of the recorders accept ConnectionPool interface and it's much clearer to keep it separate to ensure that async functions operate on async pools only and etc.

redis/asyncio/client.py
vladvildanov: str_if_bytes method apply error handling by setting error="replace"

```
def str_if_bytes(value: Union[str, bytes]) -> str:
    return (
        value.decode("utf-8", errors="replace") if isinstance(value, bytes) else value
    )
```

redis/asyncio/observability/recorder.py
vladvildanov: Fixed

redis/asyncio/client.py
vladvildanov: Fixed

redis/asyncio/client.py
vladvildanov: Fixed

redis/asyncio/connection.py
vladvildanov: Fixed

petyaslavova (approved):
LGTM.
Just check if the cursor comment is valid, and I think it is ready for merging

PR #3983, opened 2026-03-01
Normalize scalar prefixes in IndexDefinition

### LINKED ISSUE #3890: IndexDefinition Prefix Footgun
Discovered a very confusing footgun during a recent refactor with the peculiar syntax of RediSearch `IndexDefinition` and insufficient type checking.
Consider the following, sensible-looking code:

```
my_prefix = "MY:INDEXABLE:KEY:"
index_key = my_prefix + "index"

defn = IndexDefinition(prefix=my_prefix, index_type=IndexType.HASH)
redis_client.ft(index_key).create_index(my_fields, definition=defn)
```

What would we expect this to do?
Create an index of hashes prefixed by `my_prefix` ? No.
Error, because the arg `prefix=` expects an iterable? No..
Index every single hash prefixed by `{"M", "Y", ":", "I", "N", "D", "E", "X", "A", "B", "L", "K"}` ? It seems this is actually what happens.

Might I suggest `IndexDefinition` takes a keyword argument of `prefixes=` to indicate that it expects an iterable?

Or, in `index_definition.py`:
```
    def _append_prefix(self, prefix):
        """Append PREFIX."""
        if len(prefix) > 0:
            self.args.append("PREFIX")
            self.args.append(len(prefix))
            for p in prefix:
                self.args.append(p)
```
A quick `raise RuntimeError` if `prefix` is a string

Thanks for considering!

Br1an67:
### Description of change

Passing a string to the `prefix` parameter of `IndexDefinition` causes it to be iterated character-by-character, resulting in each character being treated as a separate prefix. For example:

```python
defn = IndexDefinition(prefix="MY:KEY:", index_type=IndexType.HASH)
# Actually indexes hashes prefixed by 'M', 'Y', ':', 'K', 'E', ...
```

The likely user intent is for scalar prefix values to represent a single prefix:
```python
defn = IndexDefinition(prefix="MY:KEY:", index_type=IndexType.HASH)
# Now produces PREFIX 1 "MY:KEY:"
```

This updates _append_prefix to normalize prefixes with list_or_args, so str, bytes, bytearray, and memoryview are treated as single-prefix values, while existing list/tuple inputs continue to work as before. 
Unit tests are added.

Fixes #3890

---

> [!NOTE]
> **Low Risk**
> Low risk: small input-normalization change limited to `IndexDefinition` argument building, with added tests covering new scalar prefix types.
> 
> **Overview**
> Fixes `IndexDefinition(prefix=...)` so *scalar* prefixes are treated as a single value (instead of being iterated character-by-character) by normalizing via `list_or_args` and requiring `prefix` to be non-`None`.
> 
> Extends `list_or_args` to treat `bytearray` and `memoryview` as scalar inputs, and adds unit tests validating scalar and list/tuple prefix handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1adbc5345f89b82483f7a7526166aa92ec874e7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


petyaslavova:
Hi @Br1an67, thank you for your contribution! We will have a look at it.

redis/commands/search/index_definition.py
```diff
@@ -42,6 +42,11 @@ def _append_index_type(self, index_type):
 
     def _append_prefix(self, prefix):
         """Append PREFIX."""
+        if isinstance(prefix, str):
```
petyaslavova: Instead of raising for str, I think we should normalize scalar prefix values to a single-prefix list. 
`prefix` is currently untyped, and the likely user intent for prefix="my:" is PREFIX 1 "my:", not an exception and not character-by-character expansion. The same applies to bytes-like prefixes, e.g. prefix=b"my:" currently expands to byte values.

Could you update _append_prefix to use the existing list_or_args normalization, or equivalent logic, so str/bytes/bytearray/memoryview become one-element prefix lists while existing list/tuple inputs continue to work?

Also, the regression test can be a pure unit test without the Redis fixture/redismod marker, since it only checks IndexDefinition.args.

petyaslavova (changes requested):

redis/commands/search/index_definition.py
```diff
@@ -42,6 +42,11 @@ def _append_index_type(self, index_type):
...
+            raise TypeError(
+                "prefix must be a list of strings, not a single string. "
+                "Did you mean prefix=['{}']?".format(prefix)
+            )
```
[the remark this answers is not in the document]
petyaslavova: not relevant anymore

redis/commands/search/index_definition.py
petyaslavova: done

petyaslavova (approved):

PR #3994, opened 2026-03-09
Refactored health check logic for MultiDBClient

vladvildanov:
### Description of change

Refactoring this PR brings:

1. Configuration on Health check level. Currently, the configuration for health check probes, delay, timeout etc. is done on the level of Health check policy which eliminates configuration of each specific health check.
2. Add health check timeout configuration, to control how long the health check operation may take (including all probes).
3. Concurrent health check execution. Currently, health checks per database are executed sequentially, whereas the idea is to refactor it to make them concurrently executable and independent of each other.
4. Timeout exception handling refactor. Currently, if health check exceeds timeout we propagate `TimeoutError`
to the end user, but it makes more sense to continue operating and mark the specific database as unhealthy instead.
5. Only async health checks are used across both clients, to have an alignment for health check logic.
6. Health checks now use dedicated connections instead of using shared pool

### Pull Request check-list

---

> [!NOTE]
> **Medium Risk**
> Refactors core MultiDBClient health-check execution (sync + asyncio), including new timeout/concurrency behavior and dedicated connection management; regressions could affect failover/circuit state transitions under load. Changes are well-covered by updated/added tests but touch critical availability logic.
> 
> **Overview**
> MultiDBClient health checks are reworked to be **async-only across both sync and asyncio clients**, with per-health-check configuration (`health_check_probes`, `health_check_delay`, new `health_check_timeout`) moved onto `AbstractHealthCheck` and propagated via config defaults.
> 
> Health checks now run **concurrently** (across checks and databases) with per-check `asyncio.wait_for` timeouts; timeouts/exceptions are treated as an *unhealthy database* (`UnhealthyDatabaseException`) rather than bubbling interval timeouts to callers. Policies now manage **dedicated cached health-check clients per database** and expose `close()`; both clients ensure these pools are closed (`MultiDBClient.close()` / async client `aclose()`), and the sync client uses an enhanced `BackgroundScheduler` with a shared background event loop to run async health checks safely.
> 
> Docs and tests are updated to reflect the new async custom health-check interface/signature, new pytest marker (`no_mock_connections`), and expanded coverage for timeout behavior, concurrency, and client lifecycle.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit acde14693f83f13563ededddef6a7e27c366bb52. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redis/asyncio/multidb/healthcheck.py
+++ b/redis/asyncio/multidb/healthcheck.py
@@ -189,6 +189,10 @@ class AbstractHealthCheckPolicy(HealthCheckPolicy):
                     client = AsyncRedisCluster(
                         host=first_node.host,
                         port=first_node.port,
+                        dynamic_startup_nodes=database.client.nodes_manager._dynamic_startup_nodes,
+                        address_remap=database.client.nodes_manager.address_remap,
+                        require_full_coverage=database.client.nodes_manager._require_full_coverage,
+                        retry=database.client.retry,
                         **filtered_kwargs,
                     )
                 else:
--- a/docs/geographic_failover.rst
+++ b/docs/geographic_failover.rst
@@ -342,26 +342,17 @@ asyncio event loop, so please ensure that `check_health` method is async:
 
 .. code-block:: python
 
-    from redis.asyncio.multidb.healthcheck import AbstractHealthCheck
-    from redis.asyncio import Connection
+    from redis.asyncio.multidb.healthcheck import AbstractHealthCheck, AsyncRedisClientT
 
-    class PingHealthCheck(AbstractHealthCheck):
-        def __init__(
-            self,
-            health_check_probes: int = 10,
-            health_check_delay: float = 0.1,
-            health_check_timeout: float = 2,
-        ):
-            super().__init__(
-                health_check_probes=health_check_probes,
-                health_check_delay=health_check_delay,
-                health_check_timeout=health_check_timeout,
-            )
+    class EchoHealthCheck(AbstractHealthCheck):
+        """
+        Health check based on ECHO command.
+        """
 
-        async def check_health(self, database, connection: Connection) -> bool:
-            await connection.send_command("PING")
+        async def check_health(self, database, hc_client: AsyncRedisClientT) -> bool:
+            await connection.send_command("ECHO", "healthcheck")
             response = await connection.read_response()
-            return response in (b"PONG", "PONG")
+            return response in (b"healthcheck", "healthcheck")
 
 Failure Detection (Reactive Monitoring)
 -----------------


redis/multidb/healthcheck.py
```diff
@@ -42,51 +37,83 @@ def health_check_delay(self) -> float:
...
+    Abstract health check policy.
     """
 
-    def __init__(self, health_check_probes: int, health_check_delay: float):
-        super().__init__(health_check_probes, health_check_delay)
-
     def execute(self, health_checks: List[HealthCheck], database) -> bool:
-        for health_check in health_checks:
-            for attempt in range(self.health_check_probes):
+        with ThreadPoolExecutor(max_workers=len(health_checks)) as executor:
```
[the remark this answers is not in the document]
vladvildanov: By default there's always at least one health check

redis/multidb/healthcheck.py
```diff
@@ -42,51 +37,128 @@ def health_check_delay(self) -> float:
...
     def execute(self, health_checks: List[HealthCheck], database) -> bool:
+        """Execute health checks and return database health status."""
         pass
 
+    @abstractmethod
+    def _execute(self, health_check: HealthCheck, database) -> bool:
+        """
+        Executes health check against given database.
+        """
+        pass
```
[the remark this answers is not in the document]
vladvildanov: For sure he does, cause _execute() method is called within public execute() method

redis/multidb/healthcheck.py
```diff
@@ -42,51 +37,128 @@ def health_check_delay(self) -> float:
...
                 try:
-                    if not health_check.check_health(database):
+                    # Calculate remaining time for this future
+                    remaining = deadline - monotonic()
+                    if remaining <= 0:
+                        raise TimeoutError(
+                            f"Health check timed out after {max_timeout}s"
+                        )
+
+                    result = future.result(timeout=max(0, remaining))
```
petyaslavova: Why do you need a timeout here? When you are extracting the result, the future should have already completed , so the result should basically be just ready to consume

redis/multidb/healthcheck.py
```diff
@@ -42,51 +37,128 @@ def health_check_delay(self) -> float:
...
+            }
+
+            # Use as_completed to process results as they finish
+            # This allows early return on first failure
+            for future in as_completed(future_to_hc, timeout=max_timeout):
                 try:
-                    if not health_check.check_health(database):
+                    # Calculate remaining time for this future
+                    remaining = deadline - monotonic()
+                    if remaining <= 0:
```
petyaslavova: Wouldn't this be ensured by the as_completed timeout?

redis/multidb/healthcheck.py
```diff
@@ -42,51 +37,128 @@ def health_check_delay(self) -> float:
...
+        try:
+            # Submit all health checks and track their individual timeouts
+            future_to_hc = {
+                executor.submit(self._execute, health_check, database): health_check
+                for health_check in health_checks
+            }
+
+            # Use as_completed to process results as they finish
+            # This allows early return on first failure
+            for future in as_completed(future_to_hc, timeout=max_timeout):
```
petyaslavova: This will enforce each/all health checks to complete for the max timeout, but won't respect the timeout config for each separate health check. 
It seems that for a sync multi-threaded execution, the most reliable option is to make _execute apply the timeout and raise the TimeoutError in case some health check exceeds its timeout.

redis/multidb/healthcheck.py
vladvildanov: You're right

redis/multidb/healthcheck.py
vladvildanov: It's no longer needed since we removed the timeout above

redis/multidb/healthcheck.py
vladvildanov: Fixed

redis/multidb/healthcheck.py
vladvildanov: Removed

redis/multidb/healthcheck.py
vladvildanov: After the discussion, we decided to leave it as it is and revisit when command_timeout will be supported

redis/multidb/healthcheck.py
```diff
@@ -42,51 +37,118 @@ def health_check_delay(self) -> float:
...
-        for health_check in health_checks:
-            for attempt in range(self.health_check_probes):
+        if not health_checks:
+            return True
+
+        max_timeout = max(hc.health_check_timeout for hc in health_checks)
+        executor = ThreadPoolExecutor(max_workers=len(health_checks))
+
+        try:
+            future_to_hc = {
```
uglide: I think even after this change we're not addressing the main problem: Health checks that stuck on I/O. Here is the AI generated test case that showcases the problem:
```
@pytest.mark.onlynoncluster
class TestThreadLeaksOnStuckHealthChecks:
    """
    BUG: When health checks get stuck on I/O, threads accumulate
    because executor.shutdown(wait=False) doesn't terminate them.

    These tests FAIL to show the bug exists.
    """

    def test_threads_should_not_accumulate_on_timeouts(self):
        """
        BUG: Repeated health check timeouts cause thread accumulation.
        This test FAILS because stuck threads are not cleaned up.

        EXPECTED: No thread leaks after execute() returns
        ACTUAL: Each timed-out health check leaves a running thread
        """
        stuck_threads = []
        thread_barrier = threading.Barrier(2)  # For synchronization

        def stuck_health_check(database):
            stuck_threads.append(threading.current_thread())
            try:
                thread_barrier.wait(timeout=0.05)
            except threading.BrokenBarrierError:
                pass
            time.sleep(2.0)  # Simulate stuck I/O
            return True

        mock_db = Mock(spec=Database)

        # Run multiple health checks that will all timeout
        for i in range(3):
            mock_hc = _configure_mock_health_check(
                Mock(spec=HealthCheck),
                probes=1,
                timeout=0.05,
            )
            mock_hc.check_health.side_effect = stuck_health_check

            policy = HealthyAllPolicy()
            try:
                policy.execute([mock_hc], mock_db)
            except UnhealthyDatabaseException:
                pass  # Expected

        # Give a small window for proper cleanup
        time.sleep(0.2)

        # BUG: This assertion FAILS - all 3 threads are still running
        alive_threads = [t for t in stuck_threads if t.is_alive()]
        assert len(alive_threads) == 0, (
            f"BUG: {len(alive_threads)} threads still running after execute() returned! "
            "Each health check timeout leaves a stuck thread. "
            "In production, this causes unbounded thread accumulation and resource exhaustion."
        )

        # Cleanup: wait for threads to complete
        time.sleep(2.5)

    def test_executor_shutdown_should_terminate_blocked_threads(self):
        """
        BUG: executor.shutdown(wait=False, cancel_futures=True) does NOT
        terminate threads blocked on I/O. This test FAILS.

        EXPECTED: shutdown() should terminate all threads
        ACTUAL: Blocked threads continue running indefinitely
        """
        thread_running = threading.Event()
        thread_should_exit = threading.Event()
        thread_ref = []

        def blocking_task():
            thread_ref.append(threading.current_thread())
            thread_running.set()
            while not thread_should_exit.is_set():
                time.sleep(0.1)
            return "done"

        executor = ThreadPoolExecutor(max_workers=1)
        future = executor.submit(blocking_task)

        # Wait for thread to start
        thread_running.wait(timeout=1.0)

        # Shutdown with cancel_futures=True
        executor.shutdown(wait=False, cancel_futures=True)

        # Give time for shutdown to take effect
        time.sleep(0.2)

        # BUG: This assertion FAILS - thread is still alive
        assert not thread_ref[0].is_alive(), (
            "BUG: Thread is still running after executor.shutdown()! "
            "Python's ThreadPoolExecutor cannot forcibly terminate running threads. "
            "This is a fundamental limitation that affects health check cancellation."
        )

        # Clean up
        thread_should_exit.set()
        time.sleep(0.2)
```

vladvildanov: @uglide Thanks for raising this. I didn't notice that canceling futures doesn't affect the one that already running and works only for new one. So there's no proactive monitoring on health check threads which is a deal breaker here.

I think switch to async based approach is reasonable, until we don't have a per command timeouts

redis/asyncio/multidb/healthcheck.py
```diff
@@ -43,29 +37,79 @@ def health_check_delay(self) -> float:
...
+        # Check results - handle exceptions and failures
+        for result in results:
+            if isinstance(result, Exception):
+                # Any exception (including TimeoutError) makes the database unhealthy
+                raise UnhealthyDatabaseException("Unhealthy database", database, result)
+            elif not result:
+                # Health check returned False
+                return False
+
+        return True
```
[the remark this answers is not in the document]
vladvildanov: Async is the only option for now

redis/multidb/client.py
```diff
@@ -78,20 +77,29 @@ def __init__(self, config: MultiDbConfig):
...
         """
 
         # Initial databases check to define initial state
-        self._perform_initial_health_check()
+        asyncio.run(self._perform_initial_health_check())
 
         # Starts recurring health checks on the background.
-        self._bg_scheduler.run_recurring(
+        # Uses run_recurring_coro to run async health checks in a background thread
+        self._bg_scheduler.run_recurring_coro(
```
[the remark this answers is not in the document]
vladvildanov: Fixed, now we have a single event loop

redis/background.py
```diff
@@ -90,7 +194,11 @@ async def run_recurring_async(
             if self._stopped:
                 return
 
-        loop = asyncio.get_running_loop()
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
+
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/multidb/client.py
```diff
@@ -382,7 +393,11 @@ def _on_circuit_state_change_callback(
         self, circuit: CircuitBreaker, old_state: CBState, new_state: CBState
     ):
         if new_state == CBState.HALF_OPEN:
-            self._check_db_health(circuit.database)
+            # Schedule health check - use fire-and-forget with proper exception handling
+            _schedule_coro_fire_and_forget(
+                self._check_db_health(circuit.database),
+                "HALF_OPEN health check",
+            )
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/background.py
```diff
@@ -79,6 +83,185 @@ def run_recurring(self, interval: float, callback: Callable, *args):
...
+            # Start the loop in a background thread
+            self._health_check_thread = threading.Thread(
+                target=self._run_health_check_loop,
+                daemon=True,
+            )
+            self._health_check_thread.start()
+
+            # Wait for loop to be running
+            while not self._health_check_loop.is_running():
+                pass
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/multidb/client.py
```diff
@@ -406,6 +419,26 @@ def close(self):
         if self.command_executor.active_database:
             self.command_executor.active_database.client.close()
 
+        # Close health check policy - must wait for cleanup to complete
+        _run_coro_and_wait(self._health_check_policy.close())
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/multidb/client.py
```diff
@@ -317,48 +329,47 @@ def _check_db_health(self, database: SyncDatabase) -> bool:
...
+                logger.debug(
+                    "Health check failed, due to exception",
+                    exc_info=result.original_exception,
                 )
-        return results
 
-    def _perform_initial_health_check(self):
+                db_results[unhealthy_db] = False
+            elif isinstance(result, Exception):
+                db_results[database] = False
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/multidb/client.py
```diff
@@ -78,20 +77,32 @@ def __init__(self, config: MultiDbConfig):
             auto_fallback_interval=self._auto_fallback_interval,
         )
         self.initialized = False
-        self._hc_lock = threading.RLock()
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/background.py
```diff
@@ -79,6 +85,230 @@ def run_recurring(self, interval: float, callback: Callable, *args):
...
+                self._execute_recurring_coro,
+                loop,
+                interval,
+                coro,
+                *args,
+            )
+
+        try:
+            # Schedule the coroutine to run in the event loop
+            task = asyncio.ensure_future(coro(*args), loop=loop)
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/multidb/client.py
```diff
@@ -263,8 +273,7 @@ def add_health_check(self, healthcheck: HealthCheck):
         """
         Adds a new health check to the database.
         """
-        with self._hc_lock:
-            self._health_checks.append(healthcheck)
+        self._health_checks.append(healthcheck)
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/background.py
```diff
@@ -79,6 +85,229 @@ def run_recurring(self, interval: float, callback: Callable, *args):
...
+
+            # Start the loop in a background thread
+            self._health_check_thread = threading.Thread(
+                target=self._run_health_check_loop,
+                daemon=True,
+            )
+            self._health_check_thread.start()
+
+        # Wait for loop to be running (outside the lock to avoid blocking other threads)
+        self._health_check_loop_ready.wait()
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/background.py
```diff
@@ -91,7 +320,8 @@ async def run_recurring_async(
                 return
 
         loop = asyncio.get_running_loop()
-        wrapped = _async_to_sync_wrapper(loop, coro, *args)
+
+        wrapped = _async_to_sync_wrapper(coro, *args)
```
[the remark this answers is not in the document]
vladvildanov: Fixed

redis/multidb/client.py
```diff
@@ -317,48 +332,56 @@ def _check_db_health(self, database: SyncDatabase) -> bool:
...
+                logger.debug(
+                    "Health check failed, due to exception",
+                    exc_info=result,
                 )
-        return results
 
-    def _perform_initial_health_check(self):
+                db_results[database] = False
+
+        return db_results
```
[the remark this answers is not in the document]
vladvildanov: We won't to keep it per client

redis/asyncio/multidb/client.py
```diff
@@ -352,6 +363,15 @@ async def _check_databases_health(self) -> dict[Database, bool]:
 
                 db_results[unhealthy_db] = False
             elif isinstance(result, Exception):
+                # Generic exceptions also indicate unhealthy database -
```
petyaslavova: Do we need to handle the exceptions separately? Almost the whole code in if and else blocks is the same - the only significant difference is the way we provide exception info in the debug log.

redis/asyncio/multidb/healthcheck.py
```diff
@@ -105,6 +120,39 @@ async def execute_with_timeout(health_check: HealthCheck):
...
+                    conns.append(ConnectionPool(**conn_kwargs))
+            self._connections[db_id] = conns
+
+        return conns
+
+    async def close(self) -> None:
+        disc_tasks = []
+
+        for db_id in self._connections:
+            for pool in self._connections[db_id]:
```
petyaslavova: Naming is a little confusing here. First we have a local var `_connections` that holds a list of ConnectionPools, in the method above we iterate over the `_connections` using var name `conns` and here we have pools. 
Renaming the `_connections` field to `_pools` or possibly `_redis_connection_pools` might improve code readability and better reflect what the field represents.

redis/asyncio/multidb/healthcheck.py
```diff
@@ -113,16 +161,42 @@ async def _execute(self, health_check: HealthCheck, database) -> bool:
         pass
 
 
+async def _check_health(
+    database, pool: ConnectionPool, health_check: HealthCheck
+) -> bool:
+    conn = await pool.get_connection()
```
petyaslavova: How will this behave with maintenance notifications? This additional pool doesn't have the handlers to update the settings in case of MOVING and SMIGRATED. One more question - if we have a cluster and it goes through an upgrade, the nodes will change. Currently, I don't see how the pools list will refresh and use the new topology.

redis/asyncio/multidb/client.py
vladvildanov: This is really redundant handling, since we already wrap any exceptions during health checks into UnhealthyDatabaseException, so I'll just remove it

redis/asyncio/multidb/healthcheck.py
vladvildanov: Yeah, pools might be not good enough in this situation, using underlying client is much more reliable. But the problem that we have no maintenance notifications support for async clients

redis/asyncio/multidb/client.py
vladvildanov: Removed

redis/asyncio/multidb/healthcheck.py
vladvildanov: Fixed naming issue

redis/asyncio/multidb/healthcheck.py
vladvildanov: Refactored to use clients instead, will perform additional e2e testing to understand the behaviour

redis/asyncio/multidb/client.py
```diff
@@ -360,8 +362,6 @@ async def _check_databases_health(self) -> dict[Database, bool]:
                 )
 
                 db_results[unhealthy_db] = False
-            elif isinstance(result, Exception):
-                db_results[database] = False
```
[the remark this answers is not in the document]
vladvildanov: We already handle underlying exceptions and wrap them into UnhealthyDatabaseException

redis/asyncio/multidb/healthcheck.py
```diff
@@ -120,38 +160,57 @@ async def execute_with_timeout(health_check: HealthCheck):
...
             # Check for both sync and async standalone Redis clients
             if isinstance(database.client, (AsyncRedis, SyncRedis)):
                 conn_kwargs = database.client.get_connection_kwargs()
-                conns = [ConnectionPool(**conn_kwargs)]
+                filtered_kwargs = _filter_kwargs(conn_kwargs, AsyncRedis)
+                client = AsyncRedis(**filtered_kwargs)
+            elif isinstance(database.client, (AsyncRedisCluster, SyncRedisCluster)):
+                # Cluster client - create a single cluster client that handles
+                # topology changes internally
+                conn_kwargs = database.client.get_connection_kwargs().copy()
```
petyaslavova, line 183: There are other configs that might be worth copying, for example the retry config, dynamic_startup_nodes, address_remap and maybe require_full_coverage - these are used by the cluster client instance, but not added to the connection_kwargs

vladvildanov: Updated

PR #3995, opened 2026-03-09
Align deprecation of lib_name/lib_version in async cluster

alisaifee:
### Description of change

The non cluster client marks lib_name & lib_version as deprecated however the cluster client had a default value which was getting passed to the connection which was raising a deprecation warning.

This change aligns the constructor arguments for the async cluster client with the non cluster client.

### Pull Request check-list

---

> [!NOTE]
> **Low Risk**
> Low risk: small constructor-parameter and connection-metadata plumbing change, with added deprecation warnings and a targeted test to confirm expected `CLIENT INFO` fields.
> 
> **Overview**
> Aligns async cluster client metadata handling with the rest of the library by **deprecating `lib_name`/`lib_version`** on `redis.asyncio.RedisCluster` and introducing a `driver_info` parameter.
> 
> The constructor now resolves and passes a single computed `DriverInfo` (`driver_info` or derived from `lib_name`/`lib_version`) into connection kwargs instead of defaulting and forwarding `lib_name`/`lib_version`. Tests add coverage to ensure passing the deprecated args emits warnings while still propagating `lib-name`/`lib-ver` via `client_info()`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 53b625a12c39d335c067cc11aa83e350fd10a14c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/test_asyncio/test_cluster.py
+++ b/tests/test_asyncio/test_cluster.py
@@ -477,12 +477,12 @@ class TestRedisClusterObj:
             ),
         ):
             startup_nodes = [ClusterNode("127.0.0.1", 16379)]
-            cluster = RedisCluster(
+            async with RedisCluster(
                 startup_nodes=startup_nodes, lib_name="test2", lib_version="1234"
-            )
-            info = await cluster.client_info()
-            assert info["lib-ver"] == "1234"
-            assert info["lib-name"] == "test2"
+            ) as cluster:
+                info = await cluster.client_info()
+                assert info["lib-ver"] == "1234"
+                assert info["lib-name"] == "test2"
 
     async def test_empty_startup_nodes(self) -> None:
         """


petyaslavova:
Hi @alisaifee, thank you for the contribution! Could you please apply the same pattern to the sync cluster client as well? It would also be great to include a few unit tests covering these changes.

alisaifee:
> Could you please apply the same pattern to the sync cluster client as well

The sync cluster client doesn't actually have this problem since it doesn't have default values for `lib_name`, `lib_version` etc. These parameters are captured by the extra `**kwargs` :https://github.com/redis/redis-py/blob/53e6d8af1e0dde9658d9bf49d927681f1224f1cb/redis/cluster.py#L613

I could ofcourse still add the `deprecated_args` decorator to the constructor but IMO it would be a bit confusing/weird since that argument is not explicitly being captured by that constructor.

tests/test_asyncio/test_cluster.py
```diff
@@ -465,6 +465,25 @@ async def test_cluster_retry_object(self, request: FixtureRequest) -> None:
...
+            pytest.warns(
+                DeprecationWarning,
+                match="deprecated usage of input argument/s 'lib_name'",
+            ),
+            pytest.warns(
+                DeprecationWarning,
+                match="deprecated usage of input argument/s 'lib_version'",
+            ),
+        ):
+            startup_nodes = [ClusterNode("127.0.0.1", 7000)]
```
petyaslavova: The standard setup used in the pipeline doesn't have this port exposed. 
Cluster ports are 16379-16381. 
Also, since this test is meaningful only for the cluster, it will be better to be annotated with `@pytest.mark.onlycluster` - this will skip its execution when standalone tests are ran.

tests/test_asyncio/test_cluster.py
```diff
@@ -465,6 +465,25 @@ async def test_cluster_retry_object(self, request: FixtureRequest) -> None:
...
+                match="deprecated usage of input argument/s 'lib_version'",
+            ),
+        ):
+            startup_nodes = [ClusterNode("127.0.0.1", 16379)]
+            cluster = RedisCluster(
+                startup_nodes=startup_nodes, lib_name="test2", lib_version="1234"
+            )
+            info = await cluster.client_info()
+            assert info["lib-ver"] == "1234"
+            assert info["lib-name"] == "test2"
```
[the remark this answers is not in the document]
alisaifee: Resolved in 53b625a12c39d335c067cc11aa83e350fd10a14c

tests/test_asyncio/test_cluster.py
alisaifee: Thanks. I've fixed the incorrect port (a7ef4e4dedc151e77b8be7d622ff43e6781e1a38) and rebased against master.

Regarding the `onlycluster` mark, it is already present at the test module level so it is implictly applied to all test cases in the file ([ref](https://github.com/redis/redis-py/blob/7e8ea8842a87e1c65c37b60c3b65e8ea4fe172c9/tests/test_asyncio/test_cluster.py#L61)).

PR #4029, opened 2026-04-09
Reorganize unit tests grouping - extracting the tests that define their hardcoded client configuration separately

petyaslavova:
Reorganize unit tests grouping - extracting the tests that define their hardcoded client configuration separately so that they won't be executed twice - once with the standalone tests and once with cluster tests.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to CI/test execution and pytest markers, with no runtime library behavior changes. Main risk is CI misconfiguration causing tests to be skipped or run in the wrong matrix combinations.
> 
> **Overview**
> **Reorganizes CI test execution to avoid running certain unit tests twice.** A new pytest marker, `fixed_client`, is introduced and applied broadly to tests that hardcode their client configuration.
> 
> CI is updated to run these `fixed_client` tests as a separate job/config (`fixed-clients`), while the existing standalone/cluster suites now explicitly exclude them. The GitHub Action input is changed from `protocol` to `test-config`, and artifacts are renamed accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10c37e2b98b1d74635b7b26bc49f19219a15c080. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -161,21 +161,24 @@ runs:
           echo "::group::RESP${protocol} standalone tests"
           echo "REDIS_MOD_URL=${REDIS_MOD_URL}"
 
-          if (( $REDIS_MAJOR_VERSION < 7 )) && [ "$protocol" == "3" ]; then
-            echo "Skipping module tests: Modules doesn't support RESP3 for Redis versions < 7"
-            invoke standalone-tests --redis-mod-url=${REDIS_MOD_URL} $eventloop --protocol="${protocol}" --extra-markers="not redismod and not cp_integration"
+          if [ "$protocol" == "None" ]; then
+            echo "::group::RESP${protocol} fixed client tests"
+            echo "Running fixed client tests - in those tests client instance and configuration is fixed and not changed by pytest configuration"
+            invoke fixed-client-tests $eventloop
+            echo "::endgroup::"
           else
+            echo "::group::RESP${protocol} standalone tests"
             invoke standalone-tests --redis-mod-url=${REDIS_MOD_URL} $eventloop --protocol="${protocol}"
+            echo "::endgroup::"
+
+            echo "::group::RESP${protocol} cluster tests"
+            invoke cluster-tests $eventloop --protocol=${protocol}
+            echo "::endgroup::"
           fi
-
-          echo "::endgroup::"
-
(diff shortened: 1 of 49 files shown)


tests/test_connection.py
```diff
@@ -82,6 +82,7 @@ def inner():
     # assert mod.get('fookey') == d
 
 
+@pytest.mark.fixed_client
```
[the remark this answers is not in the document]
petyaslavova: false positive - there are no such tests in the class - all tests that don't extract the client config from request are eligible for this annotation.

petyaslavova:
@sera bypass

petyaslavova:
@sera bypass

.github/actions/run-tests/action.yml
```diff
@@ -22,8 +22,8 @@ inputs:
     description: 'Event loop to use'
     required: false
     default: 'asyncio'
-  protocol:
-    description: 'RESP protocol version to use'
+  test-config:
+    description: 'Test configuration to run. Can be "all", "fixed-clients", "2-standalone", "2-cluster", "3-standalone" or "3-cluster"'
```
vladvildanov: Could you please specify that `2-standalone` means `protocol-client_type`? Do we also need the same for sentinel?

petyaslavova: 2-standalone means resp2 + standalone client tests
Sentinel tests depend on the config related to protocol version, but they can be extracted to fixed-clients with protocol provided as a resource with values 2 and 3. With the current code, all sentinel tests run twice for each protocol - once with standalone tests and once with cluster tests.
Good catch :)

petyaslavova: So the latest state is the following - a few tests were migrated to run as part of the fixed-client tests, but the ones that run for the two protocol versions will run as part of the standalone client tests - all of them already had an annotation to be skipped for cluster client tests and we would not benefit much to move them to the fixed client tests - they take just a few seconds. After this reorg of the tests the cluster tests are the ones that take much time, all other actions take less than 10 minutes.

petyaslavova:
@sera bypass

vladvildanov (approved):

petyaslavova:
@sera bypass

petyaslavova:
@sera bypass

PR #4043, opened 2026-04-22
fix(cluster): enable health check in ClusterNode command execution

### LINKED ISSUE #4028: `ClusterNode.execute_command` and `execute_pipeline` bypass connection health checks, causing  retry penalty on stale connections
ClusterNode.execute_command and ClusterNode.execute_pipeline both hardcode check_health=False when calling connection.send_packed_command():

https://github.com/redis/redis-py/blob/7021617890d46a08e39bfd73caea226573c27958/redis/asyncio/cluster.py#L1597-L1625

This means the health_check_interval parameter passed to RedisCluster(...) has no effect on cluster operations. Connections that have gone stale (silently closed by the server, a firewall, or NAT timeout during idle periods) are never detected before use.
NOTE: we have timeout set to 0 for our valkey(redis) cluster

We are most probably observing the same in our application, when a cluster node connection goes stale after an idle period:

1. execute_pipeline / execute_command acquires the dead connection from the pool
2. is_connected returns True (it only checks if _reader is not None, not actual TCP state)
3. send_packed_command(..., check_health=False) skips the health check entirely
4. The write goes into the kernel buffer, drain() or read_response() eventually fails with ConnectionError
5. The error bubbles up to PipelineStrategy.execute() which catches ERRORS_ALLOW_RETRY
6. The retry path runs: [aclose() -> asyncio.sleep(0.25) -> loop re-enters -> initialize() (full topology rediscovery)](https://github.com/redis/redis-py/blob/7021617890d46a08e39bfd73caea226573c27958/redis/asyncio/cluster.py#L2425-L2433)
7. That asyncio.sleep(0.25) alone is 250ms. Add topology rediscovery (~150-250ms for CLUSTER SLOTS + reconnecting to all nodes) and the total penalty is ~400 ms per occurrence.

What would be the cons to passing True other than we'll pay for health check in each pipeline execution?
```python
# ClusterNode.execute_command
await connection.send_packed_command(connection.pack_command(*args), True)
# ClusterNode.execute_pipeline  
await connection.send_packed_command(
    connection.pack_commands(cmd.args for cmd in commands), True
)
```

@nadiatazi262-tech  @petyaslavova  Could you please have a look if this analysis is correct? I'm planning to monkey patch the client to send `True` as a workaround.

armorbreak001:
## Summary

ClusterNode.execute_command and execute_pipeline hardcode check_health=False when calling send_packed_command(), which completely bypasses connection health validation. This means stale connections (silently closed by server, firewall, or NAT timeout during idle periods) are never detected before use, causing ConnectionError that triggers expensive retry logic with full topology rediscovery (~400ms penalty per occurrence).

## Changes

- Change check_health=False to check_health=True in ClusterNode.execute_command()
- Change check_health=False to check_health=True in ClusterNode.execute_pipeline()

This lets the connection health check run as intended, detecting stale connections early before they cause command failures.

Fixes #4028

---

> [!NOTE]
> **Medium Risk**
> Changes `ClusterNode.execute_command`/`execute_pipeline` to run `Connection.send_packed_command` with its default `check_health=True`, which can alter runtime behavior/latency and potentially surface new connection errors if health checks fail. Dependency tweaks for PyPy `cryptography` are low risk but may affect dev environment resolution.
> 
> **Overview**
> Re-enables connection health validation for asyncio cluster node command execution by no longer forcing `check_health=False` when calling `Connection.send_packed_command()` in `ClusterNode.execute_command` and `ClusterNode.execute_pipeline`.
> 
> Updates `dev_requirements.txt` to pin `cryptography<46` for PyPy 3.10 using `platform_python_implementation` markers (in addition to the existing `implementation_name` marker).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed04c6e667a9bbf69c635f08a5c2d43c9226aa63. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redis/asyncio/cluster.py
```diff
@@ -1603,7 +1603,7 @@ async def execute_command(self, *args: Any, **kwargs: Any) -> Any:
         await self.disconnect_if_needed(connection)
 
         # Execute command
-        await connection.send_packed_command(connection.pack_command(*args), False)
+        await connection.send_packed_command(connection.pack_command(*args), True)
```
petyaslavova: Since the default value is True - you can just omit setting this argument.

redis/asyncio/cluster.py
```diff
@@ -1621,7 +1621,7 @@ async def execute_pipeline(self, commands: List["PipelineCommand"]) -> bool:
 
         # Execute command
         await connection.send_packed_command(
-            connection.pack_commands(cmd.args for cmd in commands), False
+            connection.pack_commands(cmd.args for cmd in commands), True
```
petyaslavova: Same applies here - no need to provide True - this is the default value.

petyaslavova (changes requested):
Hi @armorbreak001, thank you for your contribution!

armorbreak001:
Hi @petyaslavova, thanks for the review! Good catch — I've removed both redundant `True` arguments in 1d101c3. The calls now just omit the parameter since it defaults to `True`.

petyaslavova (approved):
LGTM.

petyaslavova:
Hey @armorbreak001 , your tests with pypy 3.10 are failing because of a new cryptography version that is not compatible. Can you please add the following restriction in the dev dependencies `cryptography<46; implementation_name == "pypy" and python_version == "3.10"`?

armorbreak001:
Thanks @petyaslavova ! I've added `cryptography<46` with a pypy 3.10 environment marker in pyproject.toml. Pushed to this branch — CI should pick it up.

petyaslavova:
> Thanks @petyaslavova ! I've added `cryptography<46` with a pypy 3.10 environment marker in pyproject.toml. Pushed to this branch — CI should pick it up.

Hey @armorbreak001, the mentioned line should be added in the dev-requirements file, not the .toml .
So now we have two options :)
1. You move the restriction to the proper file
2. The change in the .toml file is just reverted and this PR is merged after #4044

armorbreak001:
Thanks @petyaslavova ! Done — reverted the pyproject.toml change and added `cryptography<46; platform_python_implementation == "PyPy" and python_version == "3.10"` to dev_requirements.txt instead. Pushed.

dev_requirements.txt
```diff
@@ -31,6 +31,8 @@ pybreaker>=1.4.0
 
 xxhash==3.6.0
 
+cryptography<46; platform_python_implementation == "PyPy" and python_version == "3.10"
```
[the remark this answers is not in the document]
petyaslavova: Will be fixed later.

PR #4045, opened 2026-04-22
fix: Type Lock.extend and Lock.reacquire as Literal[True]

### LINKED ISSUE #3792: Typing: `Lock.extend` and `Lock.do_extend` could return `Literal[True]`

Brumbelow:
### Description of change

`Lock.extend`, `Lock.do_extend`, `Lock.reacquire`, and `Lock.do_reacquire` (both sync and async variants) only ever return `True` on success; every failure path raises `LockError` or `LockNotOwnedError`. Typing them as `bool` suggests callers can meaningfully branch on the return value:
```
    if not lock.extend(10):   # never triggers — a failure raises instead
        handle_failure()
```

Retyping as `Literal[True]` makes the intent explicit at the type layer and preserves full backward compatibility at runtime. The issue calls out `extend`/`do_extend`; `reacquire`/`do_reacquire` have identical semantics and are included for consistency.

### Pull Request check-list

  - [x] Do the CI tests pass with this change?
  - [ ] Is a documentation update included?
  - [ ] Is there an example added to the examples folder?

Fixes redis/redis-py#3792

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 12f4daca3f4b51ad8c76feda027e2fe4582a50c3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/lock.py
+++ b/redis/asyncio/lock.py
@@ -3,7 +3,7 @@ import logging
 import threading
 import uuid
 from types import SimpleNamespace
-from typing import TYPE_CHECKING, Awaitable, Optional, Union
+from typing import TYPE_CHECKING, Awaitable, Literal, Optional, Union
 
 from redis.exceptions import LockError, LockNotOwnedError
 from redis.typing import Number
@@ -295,7 +295,7 @@ class Lock:
 
     def extend(
         self, additional_time: Number, replace_ttl: bool = False
-    ) -> Awaitable[bool]:
+    ) -> Awaitable[Literal[True]]:
         """
         Adds more time to an already acquired lock.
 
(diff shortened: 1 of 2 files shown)


petyaslavova (approved):
Hi @Brumbelow, thank you for your contribution! The changes LGTM.

PR #4055, opened 2026-05-07
Added support for new array commands

vladvildanov:
### Description of change

Adds Python client support for the new **Redis Array** data type, introduced in Redis 8.8.0 (preview).
                                                                                                                                                                                                               
#### Commands implemented                                   

All 18 array commands available under `Redis()` and `redis.asyncio.Redis()`:                                                                                                                                   
 
| Group | Commands |                                                                                                                                                                                           
|---|---|                                                 
| Read | `arget`, `armget`, `argetrange`, `arscan`, `arcount`, `arlen`, `arnext`, `arinfo` |
| Write | `arset`, `armset`, `arinsert`, `arring`, `arseek` |                                                                                                                                                  
| Delete | `ardel`, `ardelrange` |                                                                                                                                                                             
| Search | `argrep` |                                                                                                                                                                                          
| Aggregate | `arop` |                                                                                                                                                                                         
| History | `arlastitems` |                               
                                                                                                                                                                                                               
#### New public enums (re-exported from `redis.commands.core`)                                                                                                                                                  
                                                                                                                                                                                                               
- `ArrayAggregateOperations` — operation selector for `arop` (`SUM`, `MIN`, `MAX`, `AND`, `OR`, `XOR`, `MATCH`, `USED`)                                                                                        
- `ArrayPredicateType` — predicate kind for `argrep` (`EXACT`, `MATCH`, `GLOB`, `RE`)
- `ArrayPredicateCombinator` — predicate combiner for `argrep` (`AND`, `OR`)                                                                                                                                   
                                                                                                                                                                                                               
#### RESP2 / RESP3 compatibility                                                                                                                                                                                 
                                                                                                                                                                                                               
17 of the 18 commands return identical Python values across protocols and need no callback. **ARINFO** is the only command with protocol-divergent server output (RESP2 returns a flat key/value list; RESP3   
returns a Map).
                                                                                                                                                                                                               
A `parse_arinfo` callback is registered in `_RedisCallbacks` (`redis/_parsers/response_callbacks.py`) so that **all five** documented modes — `(protocol=2|3|None) × (legacy_responses=True|False)` — produce a
 consistent `dict[str, int]` shape. This matches the `XINFO STREAM` pattern.

### Pull Request check-list

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Adds a new public command surface for the preview Redis Array type and introduces a new response parser (`ARINFO`) that affects RESP2/RESP3 callback selection; behavior depends on server version and response-shape normalization.
> 
> **Overview**
> Adds first-class Python client support for the preview **Redis Array** data type by introducing `ArrayCommands` (mixed into `DataAccessCommands` and async variants) with methods for read/write/delete/search/aggregate operations (e.g. `ARSET`, `ARMSET`, `ARGET*`, `ARGREP`, `AROP`, etc.) plus new public enums (`ArrayAggregateOperations`, `ArrayPredicateType`, `ArrayPredicateCombinator`).
> 
> Normalizes `ARINFO` responses across RESP2/RESP3 by adding `parse_arinfo` and registering it in the shared response-callback map, and adds broad sync+async test coverage for the new commands gated by `skip_if_server_version_lt("8.7.2")`. CI dependency-audit config is updated to ignore `PYSEC-2025-183`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e349b1873e3ba3bca2f9f5e74d0a67a39005c26. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


redis/commands/core.py
```diff
@@ -5304,6 +5304,547 @@ def sort_ro(
...
+        self,
+        name: KeyT,
+        start: int,
+        end: int,
+        limit: int | None = None,
+    ) -> list[int | bytes | str] | Awaitable[list[int | bytes | str]]:
+        """
+        Iterate populated elements of the array stored at ``name`` in the
+        inclusive range [``start``, ``end``] and return alternating
+        index/value pairs as a flat list: ``[idx1, val1, idx2, val2, ...]``.
```
petyaslavova: Wouldn't it be better to add a callback here as well? It is more intuitive to return a dict object.

tests/test_asyncio/test_commands.py
```diff
@@ -2486,6 +2493,545 @@ async def test_lset(self, r: redis.Redis):
         assert await r.lset("a", 1, "4")
         assert await r.lrange("a", 0, 2) == [b"1", b"4", b"3"]
 
+    @skip_if_server_version_lt("8.8.0")
```
petyaslavova: The skip_if_version should be changed to 8.7.x when we have a build that includes the changes.

redis/commands/core.py
vladvildanov: Good catch, but I'm wondering why it doesn't solve on protocol level. I left a comment regarding this, let's see if it would be fixed server-side

tests/test_asyncio/test_commands.py
vladvildanov: Fixed

redis/commands/core.py
vladvildanov: @petyaslavova So ARSCAN command is compatible with other SCAN-like commands and it returns index-value pairs as flat array regardless of protocol. Redis-py does convert HSCAN reply into a dict, but in my opinion arrays are different as indexes might be not sequential and could be even in reverse order. Ordering is more a property of list rather than dict

petyaslavova: Ok, let's leave it this way then - the ordering might really be important and if we return a dict, we will lose it.

petyaslavova (approved):
LGTM.

PR #4060, opened 2026-05-12
Randomize cluster startup node order during topology refresh

### LINKED ISSUE #4049: CLUSTER SLOTS re-initialization cascade: deterministic startup_nodes ordering causes network saturation on first-slot node in large clusters
**Version**: redis-py 5.3.0 (also confirmed in latest `master`, the core issue is unchanged)

**Redis Engine**: 7.1 (ElastiCache), but applies to any engine 6.2+ where `CLUSTER SLOTS` returns slots in sorted order

**Platform**: Python 3.10, ~10,000 ECS tasks connecting to a 130-shard Redis Cluster (ElastiCache, m7g.2xlarge nodes)

## Description

When a `ConnectionError` or `TimeoutError` occurs, redis-py calls `NodesManager.initialize()` which iterates `startup_nodes` to find a node for `CLUSTER SLOTS`. With `dynamic_startup_nodes=True` (default), `startup_nodes` is replaced with `nodes_cache` after each initialization, and `nodes_cache` is always in slot order from the `CLUSTER SLOTS` response. This means the node owning slot 0 is always tried first on every re-initialization across all clients.

In a large cluster with many clients, this creates a feedback loop:

1. Some perturbation causes `TimeoutError`s across the fleet
2. Each timeout triggers `initialize()`, which sends `CLUSTER SLOTS` to the slot-0 node
3. The `CLUSTER SLOTS` response is large (130 shards x 2 replicas = 260 entries), which saturates the slot-0 node's network bandwidth
4. The slot-0 node becomes slow, causing more timeouts, which trigger more `initialize()` calls, which send more `CLUSTER SLOTS` to the same node
5. When the slot-0 node becomes unreachable, `initialize()` falls through to the next node in dict order (the node owning slot 10), cascading the same problem
6. The cascade continues in exact ascending slot order until it stabilizes (we observed 4 nodes deep)

This is related to #3478 (connection imbalance on shard 0001), but the root cause is not `default_node`. It's the deterministic iteration order of `startup_nodes` during re-initialization. `default_node` is only used for admin commands and changing it does not fix the issue, as the original reporter in #3478 also noted.

## Root Cause

The relevant code path:

1. `_execute_command` catches `ConnectionError`/`TimeoutError` and calls `self.nodes_manager.initialize()`:
   https://github.com/redis/redis-py/blob/v5.3.0/redis/cluster.py#L1235

2. `initialize()` iterates `self.startup_nodes.values()` to find a node for `CLUSTER SLOTS`:
   https://github.com/redis/redis-py/blob/v5.3.0/redis/cluster.py#L1649

3. After success, `startup_nodes` is replaced with `tmp_nodes_cache` (slot-ordered):
   https://github.com/redis/redis-py/blob/v5.3.0/redis/cluster.py#L1768

4. `tmp_nodes_cache` is built by iterating the `CLUSTER SLOTS` response in slot order, so the node owning slot 0 is always the first dict entry:
   https://github.com/redis/redis-py/blob/v5.3.0/redis/cluster.py#L1697-L1705

The `_initialization_lock` added in #3803 serializes `initialize()` within a single process but does not fix the cross-process deterministic ordering.

### Option 1: Shuffle startup_nodes before iterating

```python
def initialize(self):
    self.reset()
    # ...
    startup_nodes = list(self.startup_nodes.values())
    random.shuffle(startup_nodes)
    for startup_node in startup_nodes:
        # ... existing logic
```

This distributes `CLUSTER SLOTS` calls across all nodes instead of always targeting the slot-0 node. The fallback behavior is preserved: if the randomly chosen first node fails, it tries the next random node, and so on.

### Option 2: Expose a target node strategy for topology refresh

Add a constructor parameter (similar to `target_nodes` for user commands) that controls which node `initialize()` sends `CLUSTER SLOTS` to:

```python
RedisCluster(
    startup_nodes=[ClusterNode(host=endpoint, port=port)],
    topology_refresh_target="random",  # or "first", or a specific node
)
```

Currently there is no way to control this. `initialize()` always iterates `startup_nodes` in dict order and uses the first reachable node. A `topology_refresh_target="random"` option would let users opt into randomized node selection for `CLUSTER SLOTS` without changing the default behavior.

## Affected Versions

This bug requires two conditions:

1. **redis-py >= 4.1.0**, when `RedisCluster` was added to redis-py (replacing the separate `redis-py-cluster` package). The deterministic `startup_nodes` ordering from `nodes_cache` has been present since the initial implementation.

2. **Redis engine >= 6.2**, when `CLUSTER SLOTS` started returning slots in sorted order. Prior to 6.2, each node returned `CLUSTER SLOTS` in a different pseudo-random order, so `nodes_cache` dict insertion order was effectively random and no single node was consistently first.

With both conditions met, the node owning slot 0 is always the first entry in `startup_nodes` after every `initialize()` call.

The `_initialization_lock` added in #3803 (merged Jan 2026) serializes `initialize()` within a single process, reducing amplification from concurrent threads. However, it does not fix the cross-process deterministic ordering. All processes still independently target the same slot-0 node.

**Confirmed affected**: 5.3.0
**Confirmed still present in**: latest `master` as of April 2026

## Additional Note

[`CLUSTER SLOTS`](https://redis.io/docs/latest/commands/cluster-slots/) is deprecated as of Redis 7.0 in favor of [`CLUSTER SHARDS`](https://redis.io/docs/latest/commands/cluster-shards/). However, the underlying issue is not specific to which command is used for topology discovery. The problem is that `initialize()` always sends the topology discovery command to the first node in a deterministically ordered `startup_nodes` dict. If `CLUSTER SHARDS` also returns results in a deterministic order (or if `startup_nodes` remains slot-ordered regardless), the same cascade would occur.

The fix should ensure that topology discovery is distributed across nodes, regardless of which command is used.

## Related Issues

- #3478: High connection count on shard 0001 (same symptom, misattributed to `default_node`)
- #3693: startup_nodes depletion after ConnectionError/TimeoutError
- #3697: PR to fix #3693 (not merged)
- #3803: NodesManager locking (helps within one process, doesn't fix cross-process ordering)

petyaslavova:
Randomizes the startup node iteration order during cluster topology initialization for both sync and async clients. This prevents many clients from consistently querying the same first startup node when reinitializing cluster state.

The implementation copies `startup_nodes` to a list, shuffles it when multiple nodes are available, and then proceeds with the existing initialization flow. Sync behavior still includes any additional startup nodes after the shuffled startup node list, preserving the existing MOVED refresh path behavior.

Adds sync and async cluster tests that use the real cluster fixture and mock only `random.shuffle` to make the order deterministic. The tests verify that initialization queries the node that becomes first after shuffling.

Fixes #4049 

---

> [!NOTE]
> **Medium Risk**
> Changes cluster topology refresh/initialization ordering (sync + asyncio) and retry behavior by deferring the last failed node, which could affect failover/reconnect paths in production clusters. Test updates reduce flakiness but new shuffle/defer logic may surface edge cases with dynamic startup nodes and additional startup nodes.
> 
> **Overview**
> **Randomizes Redis Cluster topology refresh startup node selection** for both sync and asyncio clients by shuffling the `startup_nodes` iteration order when multiple nodes exist, reducing the chance that many clients hammer the same first node.
> 
> Propagates a `last_failed_node_name` hint through retry/refresh paths so the node that just errored is tried *after* other startup and `additional_startup_nodes` during reinitialization (sync `redis/cluster.py`, async `redis/asyncio/cluster.py`).
> 
> Adds/adjusts tests to deterministically validate the shuffle behavior (mocking `random.shuffle`), updates maintenance-notification tests to pin startup-node ordering, and makes lock blocking-timeout tests deterministic by monkeypatching time/asyncio sleep.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit caba14df88883cc57bf07de0beb5e5fc387d21dd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redis/asyncio/cluster.py
```diff
@@ -524,15 +524,22 @@ def __init__(
...
             if not self._lock:
                 self._lock = asyncio.Lock()
             async with self._lock:
                 if self._initialize:
                     try:
-                        await self.nodes_manager.initialize()
+                        await self.nodes_manager.initialize(
+                            additional_startup_nodes_info=additional_startup_nodes_info,
+                            last_failed_node_name=last_failed_node_name,
+                        )
```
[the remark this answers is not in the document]
petyaslavova: it will be used in recent future

redis/asyncio/cluster.py
```diff
@@ -974,10 +981,12 @@ async def execute_command(self, *args: EncodableT, **kwargs: Any) -> Any:
 
         # Start timing for observability
         start_time = time.monotonic()
+        last_failed_node_name = None
 
         for _ in range(execute_attempts):
             if self._initialize:
-                await self.initialize()
+                await self.initialize(last_failed_node_name=last_failed_node_name)
```
vladvildanov, line 988: So async initialisation never passes `additional_startup_nodes_info`?

petyaslavova: @vladvildanov  Currently, no - it was added for the maintenance notifications handling in sync client. I'm adding this now, so that later, when the maintenance notifications are added to the sync client we will have the same order of the arguments.

vladvildanov (approved):

PR #4063, opened 2026-05-13
Avoid zero-timeout async reads in hiredis connections readiness checks and replace async's can_read_destructive with non-destructive can_read.

### LINKED ISSUE #2912: Missing lines of coverage when using redis
I left an inquiry in the python coverage project (https://github.com/nedbat/coveragepy/issues/1671).  I'll repeat much of what I've said over there:

**Describe the bug**
When using asyncio with redis.asyncio, some lines that are definitely hit remain mysteriously uncovered, even when using a combination of different concurrency settings

**To Reproduce**
1. Python version 3.11.4, with coverage 7.3.0 and redis 5.0.0 (OS is Ubuntu 22.04)
2. My .coveragerc file:

```
[run]
source = .
parallel = True
concurrency = 
    thread
    gevent
    multiprocessing
```

(I've tried various combinations of concurrency, but I most often leave it blank.  I suspect it's not even appropriate for my application, though I'm not certain)

My minimal example (coverage_test.py):

```python
import asyncio

import redis.asyncio as redis


async def main():
    client = redis.Redis()
    result = await client.get('key')
    print('running main')

if __name__ == '__main__':
    asyncio.run(main())
    print('finished main')
```

I issue the coverage command with a bash script (coverage.sh) as follows:

```
coverage erase
coverage run coverage_test.py
coverage combine
coverage html
xdg-open htmlcov/index.html
```

The output of the script is:

```
running main
finished main
```

And the coverage of the file looks like this:

image

I would expect the last print statement (which is executed, according to the output) to be marked green.  I see there are some similar issues (most notably https://github.com/nedbat/coveragepy/issues/1082 and https://github.com/nedbat/coveragepy/issues/1598), but as I mentioned I haven't found a concurrency setting that mitigates this.

I was wondering if any of the maintainers of redis-py had any particular insights about what might cause this.  I haven't been able to reproduce it so far with other libraries.

** Update **

I've been experimenting a bit with copying code from redis-py and asyncio to illustrate exactly where this problem happens.  Here is my longer example:

```python
import asyncio
from asyncio import timeout as async_timeout

import redis.asyncio as redis


# These functions are lifted off the asyncio library.  I've done
# this to illustrate as much detail as I can about the condition
# where this abberant behavior in coverage manifests.  I've 
# called out the critical line with a comment below
async def stream_wait_for_data(stream, func_name):
    if stream._waiter is not None:
        raise RuntimeError(
            f'{func_name}() called while another coroutine is '
            f'already waiting for incoming data')

    assert not stream._eof, '_wait_for_data after EOF'

    if stream._paused:
        stream._paused = False
        stream._transport.resume_reading()

    stream._waiter = stream._loop.create_future()
    try:
        ###
        # This "return" is the key.  If it's here, the last
        # line of the file will be covered.  If it is missing, the last
        # line of the file will be uncovered.  In both conditions, it
        # prints to stdout, though
        ###
        return
        await stream._waiter
    finally:
        stream._waiter = None

async def stream_read(stream, n=-1):
    if stream._exception is not None:
        raise stream._exception

    if n == 0:
        return b''

    if n < 0:
        blocks = []
        while True:
            block = await stream_read(stream, stream._limit)
            if not block:
                break
            blocks.append(block)
        return b''.join(blocks)

    if not stream._buffer and not stream._eof:
        await stream_wait_for_data(stream, 'read')

    data = bytes(stream._buffer[:n])
    del stream._buffer[:n]

    stream._maybe_resume_transport()
    return data


# This function is lifted from the redis-py base parser
async def can_read_destructive(parser):
    try:
        stream = parser._stream
        async with async_timeout(0):
            return await stream_read(stream)
    except TimeoutError:
        return False


async def main():
    print('running main')
    client = redis.Redis()

    # Much of this logic below is just replicated from redis-py.  It's
    # the code that gets executed when you call "get" on the client
    await client.initialize()
    pool = client.connection_pool
    pool._checkpid()
    async with pool._lock:
        try:
            connection = pool._available_connections.pop()
        except IndexError:
            connection = pool.make_connection()
        pool._in_use_connections.add(connection)

    # ensure this connection is connected to Redis
    await connection.connect()
    try:
        parser = connection._parser
        return await can_read_destructive(parser) 

    except OSError as e:
        await connection.disconnect(nowait=True)
        host_error = connection._host_error()
        raise ConnectionError(f"Error while reading from {host_error}: {e.args}")

if __name__ == '__main__':
    asyncio.run(main())

    # This is the line missing coverage in certain conditions.  It is always
    # executed, though:
    print('finished main')
```

This is what coverage looks like with the return (the one in the example above, called out in the comment, in the "stream_wait_for_data" function):

image

And without:

image

petyaslavova:
## Summary

This PR replaces the async hiredis readiness probe with a non-destructive `can_read()` check that only inspects already-buffered parser/stream state or EOF. The previous implementation used `async_timeout(0)` around `StreamReader.read()`, which could create and immediately cancel a read waiter when no data was available.

That zero-timeout read path can interfere with tracing tools such as coverage.py: code after the readiness check may execute normally but still be reported as uncovered. 
The pure-Python async parser already avoided this shape; this brings the async hiredis parser in line with that behavior.

Since the logic inside `can_read_destructive()` is not destructive anymore it is deprecated in favor of `can_read()`, and connection-pool readiness checks now use the new method. 
The deprecated wrapper remains for compatibility.

Tests were added for the async hiredis parser to verify that `can_read()` detects hiredis-buffered responses, asyncio stream-buffered bytes, and EOF without calling `StreamReader.read()`.

**Important !!!** Users with custom async parser classes must implement can_read() instead of relying on can_read_destructive(), because connection readiness checks now use the non-destructive parser API directly.

Fixes #2912 

---

> [!NOTE]
> **Medium Risk**
> Touches connection pool health checks, hiredis parsing readiness, and client-cache invalidation paths; custom async parsers must add `can_read()`.
> 
> **Overview**
> Replaces async connection readiness checks with a **non-destructive** `can_read()` that only inspects hiredis/parser buffers, `StreamReader` backlog, or EOF—no zero-timeout `StreamReader.read()`, which could skew coverage tools and differ from the pure-Python async parser. **`can_read_destructive()` is deprecated (8.0)** and async pools/connections call `can_read()` instead; custom async parsers must implement it.
> 
> Sync **hiredis** readiness no longer caches partial parses in `can_read()`; it uses `Reader.has_data()`, **`select`** (plus SSL `pending()`), and aligns zero-timeout **would-block** behavior with **`TimeoutError`**. Client-tracking invalidation drains use **`read_response(..., timeout=0, disconnect_on_error=False)`** and stop on timeout. Tests cover buffer/EOF detection without consuming the stream.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 06d09db4d9783f16419c4eb6898c18cbc20cff3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/_parsers/base.py
+++ b/redis/_parsers/base.py
@@ -1,7 +1,6 @@
 import logging
-import sys
 from abc import ABC
-from asyncio import IncompleteReadError, StreamReader, TimeoutError
+from asyncio import IncompleteReadError, StreamReader
 from typing import Awaitable, Callable, List, Optional, Protocol, Union
 
 from redis.maint_notifications import (
@@ -14,12 +13,7 @@ from redis.maint_notifications import (
     OSSNodeMigratedNotification,
     OSSNodeMigratingNotification,
 )
-from redis.utils import safe_str
-
-if sys.version_info.major >= 3 and sys.version_info.minor >= 11:
-    from asyncio import timeout as async_timeout
-else:
-    from async_timeout import timeout as async_timeout
+from redis.utils import deprecated_function, safe_str
 
 from ..exceptions import (
     AskError,
@@ -169,7 +163,13 @@ class AsyncBaseParser(BaseParser):
         self._stream: Optional[StreamReader] = None
         self._read_size = socket_read_size
 
+    @deprecated_function(
+        version="8.0.0", reason="Use can_read() instead", name="can_read_destructive"
+    )
     async def can_read_destructive(self) -> bool:
+        raise NotImplementedError()
+
+    async def can_read(self) -> bool:
         raise NotImplementedError()
 
     async def read_response(
@@ -515,16 +515,20 @@ class _AsyncRESPBase(AsyncBaseParser):
         """Called when the stream disconnects"""
         self._connected = False
 
+    @deprecated_function(
+        version="8.0.0",
+        reason="Use can_read() instead",
+        name="can_read_destructive",
+    )
     async def can_read_destructive(self) -> bool:
+        return await self.can_read()
+
+    async def can_read(self) -> bool:
         if not self._connected:
             raise OSError("Buffer is closed.")
         if self._buffer:
             return True
-        try:
-            async with async_timeout(0):
-                return self._stream.at_eof()
-        except TimeoutError:
-            return False
+        return self._stream.at_eof()
 
     async def _read(self, length: int) -> bytes:
         """
--- a/redis/_parsers/base.py
+++ b/redis/_parsers/base.py
@@ -120,12 +120,6 @@ class BaseParser(ABC):
     def on_connect(self, connection):
         pass
 
-    @abstractmethod
-    def can_read(self, timeout: float = 0) -> bool:
-        # TODO: Rename this API; it detects pending data or dirty/closed
-        # connection state, not only whether application data can be read.
-        pass
-
 
 class _RESPBase(BaseParser):
     """Base class for sync-based resp parsing"""
@@ -179,11 +173,11 @@ class AsyncBaseParser(BaseParser):
         version="8.0.0", reason="Use can_read() instead", name="can_read_destructive"
     )
     @abstractmethod
-    async def can_read_destructive(self, timeout: float = 0) -> bool:
+    async def can_read_destructive(self) -> bool:
         pass
 
     @abstractmethod
-    async def can_read(self, timeout: float = 0) -> bool:
+    async def can_read(self) -> bool:
         # TODO: Rename this API; it detects pending data or dirty/closed
         # connection state, not only whether application data can be read.
         pass
@@ -536,10 +530,10 @@ class _AsyncRESPBase(AsyncBaseParser):
         reason="Use can_read() instead",
         name="can_read_destructive",
     )
-    async def can_read_destructive(self, timeout: float = 0) -> bool:
-        return await self.can_read(timeout=timeout)
+    async def can_read_destructive(self) -> bool:
+        return await self.can_read()
 
-    async def can_read(self, timeout: float = 0) -> bool:
+    async def can_read(self) -> bool:
         # TODO: Rename this API; it detects pending data or dirty/closed
         # connection state, not only whether application data can be read.
         if not self._connected:
--- a/redis/_parsers/base.py
+++ b/redis/_parsers/base.py
@@ -1,5 +1,5 @@
 import logging
-from abc import ABC
+from abc import ABC, abstractmethod
 from asyncio import IncompleteReadError, StreamReader
 from typing import Awaitable, Callable, List, Optional, Protocol, Union
 
@@ -112,11 +112,19 @@ class BaseParser(ABC):
             return exception_class(response, status_code=error_code)
         return ResponseError(response)
 
+    @abstractmethod
     def on_disconnect(self):
-        raise NotImplementedError()
+        pass
 
+    @abstractmethod
     def on_connect(self, connection):
-        raise NotImplementedError()
+        pass
+
+    @abstractmethod
+    def can_read(self, timeout: float = 0) -> bool:
+        # TODO: Rename this API; it detects pending data or dirty/closed
+        # connection state, not only whether application data can be read.
+        pass
 
 
 class _RESPBase(BaseParser):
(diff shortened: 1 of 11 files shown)


redis/_parsers/base.py
```diff
@@ -515,16 +515,20 @@ def on_disconnect(self):
...
 
+    @deprecated_function(
+        version="8.0.0",
+        reason="Use can_read() instead",
+        name="can_read_destructive",
+    )
     async def can_read_destructive(self) -> bool:
+        return await self.can_read()
+
+    async def can_read(self) -> bool:
```
vladvildanov, line 536: Just for clarification, the `can_read` method identify "if you can perform read operation", but not "if there's actual data to read"?

petyaslavova: @vladvildanov yes  exactly - it only checks if the connection is still usable, but it doesn't actually consume any data.

petyaslavova: @vladvildanov update on the answer to the question: after digging a little bit more, this method should also check for data waiting in the connection, this is the expectation for a non breaking check - it is not supposed to consume this data, but just check if there is something waiting to be processed - in this concrete method this is represented by the self._buffer check. In hiredis the stream's underlying _buffer is checked if it contains data. The name is still a little confusing because if self._stream.at_eof() is True, this doesn't mean we have data(in the meaning of some unconsumed response or push message) in the socket, this means that we have been notified that the server has closed the connection, but I think that such a rename should be out of scope for this bug fix.

vladvildanov: @petyaslavova Yeah, it also confuses me how it is possible to read if stream at_eof(), maybe just leave a comment or TODO to be aware that naming isn't reflecting what actually happens

petyaslavova: done.

redis/_parsers/hiredis.py
```diff
@@ -90,17 +90,16 @@ def on_connect(self, connection, **kwargs):
...
             raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
 
-        if self._next_response is NOT_ENOUGH_DATA:
-            self._next_response = self._reader.gets()
-            if self._next_response is NOT_ENOUGH_DATA:
-                return self.read_from_socket(timeout=timeout, raise_on_timeout=False)
-        return True
+        if self._reader.has_data():
+            return True
+        return _socket_can_read(self._sock, timeout)
```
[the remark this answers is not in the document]
petyaslavova: fixed

redis/_parsers/hiredis.py
```diff
@@ -243,16 +225,24 @@ def on_connect(self, connection):
...
             return True
-        try:
-            async with async_timeout(0):
-                return await self.read_from_socket()
-        except asyncio.TimeoutError:
-            return False
+        # asyncio.StreamReader has no public non-destructive API for checking
+        # buffered bytes. Preserve dirty-connection detection for hiredis and
+        # fail loudly if the private buffer API changes.
+        return bool(self._stream._buffer)
```
ndyakov, line 246: let's make sure this is detected in the ci instead of failing runtime for the user.

redis/_parsers/hiredis.py
```diff
@@ -141,26 +143,6 @@ def read_response(
...
-                    return response
-
-                return self.read_response(
-                    disable_decoding=disable_decoding,
-                    push_request=push_request,
-                    timeout=timeout,
-                )
-            return response
-
         if disable_decoding:
```
ndyakov, line 146: since this was removed do we have a test that verifies handling when first parsed message is a push notification on a sync hiredis connection?

ndyakov (approved):
looks valid, but have two concerns that can be addressed with adding couple of tests.

redis/_parsers/hiredis.py
petyaslavova: Done. Test covering the property availability is added.

redis/_parsers/hiredis.py
petyaslavova: New tests covering this corner case are added.

PR #4076, opened 2026-05-21
test + fix: add sharded pub/sub e2e coverage and fix the client defects it exposed

petyaslavova:
## Change summary

This pull request adds end-to-end sharded pub/sub coverage for Redis Cluster and fixes the client defects that coverage exposed. Shard-channel reconciliation was purely event-driven and had no defence against a node that stops answering: a node that leaves the deployment replies ECONNREFUSED rather than MOVED, so the channels pinned to it never moved to their new owner, and because one reader serves every per-node pubsub, a single sick node could stall delivery cluster-wide. The read path also had no lock between the poller and the writers of the same socket, so a reconnect from reconciliation could land underneath a reader parked in `parse_response`.

The core of the change is on the read path of `ClusterPubSub` in `redis/cluster.py` and `redis/asyncio/cluster.py`. The per-node poll is split out of `PubSub.get_message` into `_poll_node_pubsub`, which holds a new per-node wire I/O lock across the read and the subscription bookkeeping but dispatches user handlers outside it, so a handler is free to call `ssubscribe`/`sunsubscribe` without deadlocking against the reconciliation worker. The round robin now keeps going when one node fails: it records the first error and re-raises only if the whole pass failed, cools the node off for a second, marks it unreachable so a migration skips a `SUNSUBSCRIBE` that cannot arrive, and asks for a throttled `CLUSTER SLOTS` refresh so a departed node's channels get a new owner. A `MOVED` raised out of a poll is now handled rather than propagated: the affected channels are detached and re-routed, the redirect is applied, and reconciliation is scheduled. `_pubsubs_generator` yields `None` instead of returning on a momentarily empty mapping, and retiring an empty per-node pubsub is centralized in `_drop_node_pubsub`.

Supporting fixes, all found while running the new tests: `execute_command` delegates `SSUBSCRIBE`/`SUNSUBSCRIBE` to the tracked methods instead of dispatching raw and recording nothing; `reset()`/`aclose()` tear down per-node pubsubs under the I/O lock, and `reset()` swaps in a fresh reconcile executor so the scheduling sites can stay lock-free and out of a three-thread deadlock cycle with the maintenance-notifications handler lock; `handle_message` discards instead of removing so a racing detach cannot raise `KeyError` into a poll; `PubSub.get_message` no longer raises `TypeError` for `timeout=None`; `NodesManager.initialize` skips a re-entrant refresh triggered by an SMIGRATED push handled inline on the refreshing thread, and the sync maintenance handler now takes `_initialization_lock` before its own `_lock`. The parsers were hardened so a concurrent disconnect surfaces as a retryable `ConnectionError` rather than `ValueError`/`TypeError`/`AttributeError` (`_parsers/socket.py`, `hiredis.py`, `base.py`), and the RESP3 parser forwards the caller's `timeout` across a push continuation.

No public signatures, defaults, or return types change; every new member is private. Sync and async stay mirrored, with the deliberate divergences documented in place: the async `PubSub` has no `subscribed_event` to wait on, the async maintenance handler needs no lock-ordering rule because handling never runs inline inside `initialize()`, async gains `_ensure_cluster_initialized` and keeps topology-repair tasks out of the set `aclose()` cancels, and the async `ClusterNode` pool gains the maintenance-recycle debug logging the sync pool already had.

## Test coverage

New integration coverage in `tests/test_pubsub.py` and `tests/test_asyncio/test_pubsub.py` exercises sharded delivery to multiple subscribers, duplicate `ssubscribe`, silence after `sunsubscribe`, lazy and per-shard-shared connections, subscriber-mode protocol errors, and server-side `SHARDNUMSUB` cleanup on close. Around forty mirrored unit tests cover each new behavior directly: poll cool-off and its blocking/zero-timeout variants, the throttled topology repair, `MOVED` repair on both the round-robin and target-node paths, the unreachable-node skip, which half of `handle_message` runs under the I/O lock, handler re-entrancy and the reconciler deadlock, and collection of per-node pubsubs a detach empties. `tests/test_connection.py` and its async mirror cover the parser guards; `tests/test_cluster.py` and its mirror cover the re-entrant `initialize` skip and the recycle logging.

A new scenario suite, `tests/test_scenario/test_pubsub.py`, drives shard migration and node/shard/proxy failure against a real Redis Enterprise deployment through the fault injector, asserting a delivery ratio after recovery for both the sync and async clients; it needed new `migrate` and `wait_for_database_active` fault-injector actions. The Active-Active fixtures now poll the endpoints until they answer PING before building the client, replacing a blind ten-second teardown sleep that was letting a previous test's injected fault fail the next test's initial health check.

---

> [!NOTE]
> **High Risk**
> Large concurrent changes to cluster sharded pub/sub, lock ordering, and connection/parser teardown paths; behavior of errors and topology recovery changes for production cluster users.
> 
> **Overview**
> Adds end-to-end sharded pub/sub coverage and fixes the cluster client behavior that testing exposed. **Sharded pub/sub** in sync and async `ClusterPubSub` now keeps shard channels on the slot owner, re-routes on `MOVED`, cools off and throttles `CLUSTER SLOTS` refresh when a node is unreachable, and only surfaces connection errors from `get_sharded_message()` when every polled node fails (except when `target_node` is set). Polling goes through **`_poll_node_pubsub`** with a per-node wire I/O lock so reads do not race reconnects; user handlers run outside the lock to avoid deadlocks with reconciliation.
> 
> Supporting changes: **`SSUBSCRIBE`/`SUNSUBSCRIBE`** via `execute_command` delegate to tracked methods; unsubscribe bookkeeping uses **`discard`**; **`PubSub.get_message`** no longer breaks on `timeout=None`; **`NodesManager.initialize`** skips re-entrant refresh from inline SMIGRATED handling; sync maintenance notifications take **`_initialization_lock`** before the handler lock; reconcile scheduling avoids lock cycles on reset/close.
> 
> **Parsers** (`socket`, hiredis, async base, RESP3) turn concurrent disconnects into retryable **`ConnectionError`** and forward read **timeouts** across RESP3 push continuations. Docs and **CONTRIBUTING** describe new `get_sharded_message()` error semantics and security reporting; **pytest** log formats gain sub-second timestamps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4484456e1c4174fdfa3b19c873f8fd5bddcfb58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_scenario/test_active_active.py
```diff
@@ -44,7 +44,7 @@ def trigger_network_failure_action(
     logger.info(f"Action completed. Status: {status_result['status']}")
 
 
-@pytest.mark.skip(reason="Temporarily disabled")
+# pytest.mark.skip(reason="Temporarily disabled")
```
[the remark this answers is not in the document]
petyaslavova: On purpose...

redis/cluster.py
```diff
@@ -3055,6 +3106,23 @@ def listen(self, event: object) -> None:
...
+# siblings hold undelivered messages.
+SHARD_POLL_COOL_OFF_SECONDS = 1.0
+
+# How often a failed sharded-pubsub poll may trigger a slots-cache refresh.
+# Reconciliation is otherwise purely event-driven, and a node that has left the
+# deployment answers ECONNREFUSED rather than MOVED - so without this the reader
+# would cool off against the departed node forever and the shard channels pinned
+# to it would never move to their new owner. Throttled because the refresh costs
+# a CLUSTER SLOTS round trip and a failing node fails every poll.
+SHARD_TOPOLOGY_REPAIR_INTERVAL_SECONDS = 5.0
```
vladvildanov, line 3131: The reconciliation seems to me like a behavioural changes. The current users are probably handling an exception from `get_sharded_message()` as a trigger for manual reconciliation and now we handle it client-side. So it won't break an existing users, it just makes their code redundant. Since, it's not a breaking behaviour worth to document so users could remove the exception handling

petyaslavova: Added this info to the docs.

vladvildanov (approved):

Issue #2220, opened 2022-06-07
Reduce default max_connections number


laukhin:
**Version**: 4.3.3

**Platform**: Python 3.10 (actually any python)

**Description**: By default, there is a 2**31 max connection [pool size](https://github.com/redis/redis-py/blob/c54dfa49dda6a7b3389dc230726293af3ffc68a3/redis/connection.py#L1282). Such a large number seems unreasonable and unsafe to me for several reasons:

1. By default Redis instance can handle [10000 connections](https://redis.io/docs/reference/clients/#maximum-concurrent-connected-clients) at a time which is far less than our default pool size.
1. Such an enormous pool size could lead to a DDoS attack on the Redis instance if used without proper care. Actually, it happened with our production not so long ago - we had some problems with a load and it led to our connections pool growing infinitely. Our Redis instances just couldn't recover from a massive amount of new connections. With a stricter connections size policy we could avoid that.
1. I made some research and other Redis clients have a much stricter pool policy. For example, the Java client [has 8 connections by default](https://github.com/apache/commons-pool/blob/master/src/main/java/org/apache/commons/pool2/impl/GenericObjectPoolConfig.java#L37). Unofficial go client implementation has 10 connections per CPU [by default](https://github.com/go-redis/redis/blob/4ddd7d18036f93ac0753e1905c7dbd17cde3ca58/options.go#L88).

My suggestion is to reduce the default max_connections number (both in sync and async versions) to provide a safe configuration by default and let the end-user override it to fine-tune the performance for its needs. The number is discussable, I'd stay for about 100 default connections. I can handle this issue if you will decide to fix that :)

Fogapod:
I started getting `ConnectionError: Too many connections` after setting `max_connections=50`. Looks like connections aren't released properly and 2**31 hides that fact?

relevant line:
https://github.com/redis/redis-py/blob/c54dfa49dda6a7b3389dc230726293af3ffc68a3/redis/asyncio/connection.py#L1532

Why would pool ever raise Too many connections, it should just wait instead

Fogapod:
this is not solved..

anushkmittal:
@Fogapod any workaround to this? do u just not set max_connections

Fogapod:
> @Fogapod any workaround to this? do u just not set max_connections

I use blocking pool: https://github.com/redis/redis-py/issues/2517#issuecomment-1998680722

mohd-akram:
[...]

PR #4082, opened 2026-05-27
Updates in default connection and retry settings

### LINKED ISSUE #2243: process hangs due to default socket_timeout=None
**Version**: master (https://github.com/redis/redis-py/tree/bea72995fd39b01e2f0a1682b16b6c7690933f36)

**Platform**: Python 3.10 on Ubuntu 20.04

**Description**: 

There is no default **socket_timeout** set for connections, which could lead to outages. (only tested `RedisCluster`, but the connection object seems to be shared with the single-node implementation)

### steps to reproduce

1. Example code:
```
from redis.cluster import RedisCluster as Redis
from time import sleep

rc = Redis(host="172.20.1.1", port=6379)

for x in range(10000):
    print(f"{x} : {rc.cluster_keyslot(x)}")
    rc.get(x)
    sleep(2)
```

2. isolate the node hosting the keyslot, ensure there is a TCP timeout instead of closing the connection, for example:
    - add an `iptables` DROP rule, or
    - `docker network disconnect NETWORK_ID CONTAINER_ID` (stopping/killing the container does NOT work, as the host might immediately generate a `no route to host` response)

### expected behaviour

redis-py throws a timeout

### actual behaviour:
the process hangs

### relevant Stack trace (gdb)

The stack trace shows `socket_timeout=None`

```
#12 Frame 0x558935950720, for file /workspace/redis-cluster-issue/redis-py/redis/connection.py, line 192, in _read_from_socket (self=<SocketBuffer(_sock=<socket at remote 0x7f01866520e0>, socket_read_size=65536, socket_timeout=None, _buffer=<_io.BytesIO at remote 0x7f0186bd4f90>, bytes_written=0, bytes_read=0) at remote 0x7f01863382e0>, length=None, timeout=<object at remote 0x7f0187458df0>, raise_on_timeout=True, sock=<...>, socket_read_size=65536, buf=<_io.BytesIO at remote 0x7f0186bd4f90>, marker=0, custom_timeout=False)
    data = self._sock.recv(socket_read_size)

#32 Frame 0x7f018667d970, for file /workspace/redis-cluster-issue/redis-py/redis/connection.py, line 824, in read_response (self=<Connection(pid=77375, host='172.20.2.2', port=6379, db=0, username=None, client_name=None, password=None, socket_timeout=None, socket_connect_timeout=None, socket_keepalive=None, socket_keepalive_options={}, socket_type=0, retry_on_timeout=False, retry_on_error=[], retry=<Retry(_backoff=<NoBackoff(_backoff=0) at remote 0x7f018662b700>, _retries=0, _supported_errors=(<type at remote 0x558935865b00>, <type at remote 0x558935866680>, <type at remote 0x5589343d5380>)) at remote 0x7f018662b7c0>, health_check_interval=0, next_health_check=0, redis_connect_func=<method at remote 0x7f018736ca00>, encoder=<Encoder(encoding='utf-8', encoding_errors='strict', decode_responses=False) at remote 0x7f0186338160>, _sock=<socket at remote 0x7f01866520e0>, _socket_read_size=65536, _parser=<ClusterParser(socket_read_size=65536, encoder=<...>, _sock=<...>, _buffer=<SocketBuffer(_sock=<...>, socket_read_si...(truncated)
    response = self._parser.read_response(disable_decoding=disable_decoding)
```

### suggested fixes

- enable keepalive per default
- add a sane default timeout (might interfere with `BRPOP` / `BLPOP` -> increase the socket timeout dynamically if those are used?)
- improve documentation

### additional info

https://docs.python.org/3/library/socket.html#socket.socket.settimeout

(There is a related issue (#722) outlining some workarounds, I submitted this one to discuss changing the defaults to more sensible values)

petyaslavova:
## Connection configs

This PR updates the default connection behavior for redis-py clients.
`TCP keepalive` is now enabled by default, and unspecified keepalive options resolve to platform-supported defaults: `30s idle`, `5s interval`, and `3 probes`. The default socket timeout and socket connect timeout are now both 5 seconds.

The default `socket read buffer size` is reduced to `32kB`, and the default `connection pool limit` is reduced from the previous effectively-unbounded value to `100 connections`. 
These defaults are applied across sync and async connection paths, with async cluster connection limits aligned as well.

## Retry configs

The default retry policy now uses `ExponentialWithJitterBackoff(base=0.01, cap=1)` with `10 retries`. 
This gives standalone retry paths a 10ms base backoff, a 1000ms cap, and more retry attempts before surfacing retryable failures.

Cluster retry defaults are aligned to 10 attempts as well. For regular cluster commands, the retry object is still primarily used as a retry-count container, allowing more topology recovery attempts; cluster transactions use the full retry object and therefore honor the jittered backoff configuration.

Fixes #2243 and #2220

---

> [!NOTE]
> **High Risk**
> Default socket timeouts, keepalive, pool size, read buffer, and retry counts/backoff change runtime behavior for all callers that did not override these options, which can affect latency, connection churn, and high-concurrency workloads.
> 
> **Overview**
> Introduces **`redis/_defaults.py`** as the single source for shared connection and retry defaults, then wires those constants through sync/async **`Redis`**, **`Connection`**, **`ConnectionPool`**, cluster, multidb, and **`HttpClient`**.
> 
> **Connection behavior** now defaults to **5s** socket and connect timeouts, **32KB** read buffer (down from 64KB), **TCP keepalive on** with platform-specific idle/interval/probe options when `socket_keepalive_options` is omitted, and **`max_connections` capped at 100** (replacing the prior effectively unbounded pool default). Clients expose **`socket_read_size`** in kwargs/URL parsing; docs describe keepalive option semantics.
> 
> **Retry behavior** shifts to **`ExponentialWithJitterBackoff(base=0.01, cap=1)`** with **10 attempts** (standalone, multidb command retry, cluster `cluster_error_retry_attempts`, and related paths). Cluster docs/comments reflect the new defaults.
> 
> **`SENTINEL`** for “argument omitted” is consolidated in **`redis.utils`** (parsers stop defining their own). Tests and maintenance-notification expectations are updated for non-`None` default timeouts; **`tests/test_defaults.py`** asserts the new signatures and values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cfe8b9a9ecb3888e3d71623001970b40db2f8064. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redis/connection.py
```diff
@@ -2853,7 +2871,7 @@ def __init__(
         maint_notifications_config: Optional[MaintNotificationsConfig] = None,
         **connection_kwargs,
     ):
-        max_connections = max_connections or 2**31
```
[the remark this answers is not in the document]
petyaslavova: fixed.

redis/asyncio/cluster.py
```diff
@@ -1463,7 +1473,7 @@ def __init__(
         port: Union[str, int],
         server_type: Optional[str] = None,
         *,
-        max_connections: int = 2**31,
+        max_connections: int = 100,
```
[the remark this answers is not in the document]
petyaslavova: Sync cluster nodes connect to each node using standalone Redis client instance with default pool config - which is already updated to have 100 max_connections as default

redis/_defaults.py
```diff
@@ -0,0 +1,35 @@
...
+
+    # Not every platform exposes interval/probe tuning constants.
+    tcp_keepcnt = getattr(socket, "TCP_KEEPCNT", None)
+    if tcp_keepcnt is not None:
+        options[tcp_keepcnt] = 3
+
+    return options
+
+
+# Retry defaults
```
elena-kolevska, line 36: Seems like this wasn't finished. We have the default retry object instantiated in too many palces with magic number. Let's centralise it with some kind of a `get_default_retry()` factory.

redis/_defaults.py
```diff
@@ -0,0 +1,35 @@
+"""Internal default helper functions shared across redis-py modules."""
+
+import socket
+
+# Connection defaults
+
+DEFAULT_SOCKET_TIMEOUT = 5
```
elena-kolevska: Will this affect blocking commands? Do we dynamically change/disable the socket  timeout on those connections?

redis/asyncio/connection.py
```diff
@@ -168,30 +163,30 @@ class AbstractConnection:
...
+        socket_connect_timeout: float | None = DEFAULT_SOCKET_CONNECT_TIMEOUT,
         retry_on_timeout: bool = False,
-        retry_on_error: Union[list, _Sentinel] = SENTINEL,
+        retry_on_error: list | object = SENTINEL,
         encoding: str = "utf-8",
         encoding_errors: str = "strict",
         decode_responses: bool = False,
         parser_class: Type[BaseParser] = DefaultParser,
-        socket_read_size: int = 65536,
+        socket_read_size: int = 32768,
```
elena-kolevska: I feel like this should be configurable on the Client level, because it can be problematic for users with huge keys.

elena-kolevska (changes requested):

redis/_defaults.py
petyaslavova: It will. We don't currently have special handling for blocking commands, and there is an open issue for that, so I consider this fix out of scope for the current change.

redis/_defaults.py
petyaslavova: I left most of the values directly set in the args because this way when you create the object and hover over the constructor call, you can see all the available args + the configured default values. If I replace all those with a factory method call, the users of the library will lose that visibility over the configuration and will have to inspect the code to follow the internal logic on the retry config that will be applied. There were just two config changes that were not good for this approach and those are the keep alive config dict + the socket timeouts(those needed to be also used in the maintenance notification handling). 
For the retries I can add the default values for retry count, base and cap as constants and use those. 
The factory method is more appropriate for initialization of the object inside the method when it is not provided as default value.

redis/asyncio/connection.py
petyaslavova: Done

redis/_defaults.py
petyaslavova: As agreed offline - the configs are extracted as constants and reused.

redis/_defaults.py
elena-kolevska: Ok, so let's make it very clear in the release notes.

elena-kolevska (approved):

PR #4085, opened 2026-05-28
security: add optional key parameter to JWToken for signature verification

dfgvaetyj3456356-hash:
This PR fixes a JWT signature verification bypass in JWToken.

**CWE:** CWE-345 (Insufficient Verification of Cryptographic Signature)
**File:** redis/auth/token.py

JWToken.__init__() previously decoded JWTs with verify_signature=False, meaning any forged token would be accepted regardless of the signing key. An attacker could craft arbitrary JWT payloads and have them accepted by downstream authorization decisions that rely on is_expired(), ttl(), and try_get().

**Fix:**
- Added an optional key parameter to JWToken.__init__()
- When key is provided, the JWT signature is verified using standard PyJWT verification
- When key is None, existing behavior is preserved for backward compatibility
- Updated existing tests to pass the signing key
- Added new test verifying forged tokens are rejected with wrong key

---

> [!NOTE]
> **Medium Risk**
> Fixes a real signature-bypass (CWE-345) in auth code, but callers who still use `JWToken(token)` alone remain unverified until they pass `key` and `algorithms`.
> 
> **Overview**
> **`JWToken`** now accepts optional **`key`** and **`algorithms`**. When **`key`** is set, PyJWT verifies the signature (and requires **`algorithms`**); **`aud`** / **`nbf`** checks stay off. With no **`key`**, behavior is unchanged: decode without signature verification for backward compatibility.
> 
> Tests and mocks were updated so the default mock identity provider uses verified tokens; a **`legacy`** mock path and dedicated signature tests cover the old init and forged-token rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ee2ff9f2c7169808eff578973d8764e4e9d8f14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/auth/token.py
+++ b/redis/auth/token.py
@@ -78,7 +78,7 @@ class SimpleToken(TokenInterface):
 class JWToken(TokenInterface):
     REQUIRED_FIELDS = {"exp"}
 
-    def __init__(self, token: str):
+    def __init__(self, token: str, key: str = None, algorithms: list = None):
         try:
             import jwt
         except ImportError as ie:
@@ -86,11 +86,24 @@ class JWToken(TokenInterface):
                 f"The PyJWT library is required for {self.__class__.__name__}.",
             ) from ie
         self._value = token
-        self._decoded = jwt.decode(
-            self._value,
-            options={"verify_signature": False},
-            algorithms=[jwt.get_unverified_header(self._value).get("alg")],
-        )
+        if key is not None:
+            if algorithms is None:
+                algorithms = ["HS256"]
+            self._decoded = jwt.decode(
+                self._value,
+                key,
+                algorithms=algorithms,
+                options={"verify_exp": False},
+            )
+        else:
+            if algorithms is None:
+                header = jwt.get_unverified_header(self._value)
+                algorithms = [header["alg"]]
+            self._decoded = jwt.decode(
+                self._value,
+                options={"verify_signature": False, "verify_exp": False},
+                algorithms=algorithms,
+            )
         self._validate_token()
 
     def is_expired(self) -> bool:
--- a/redis/auth/token.py
+++ b/redis/auth/token.py
@@ -93,7 +93,7 @@ class JWToken(TokenInterface):
                 self._value,
                 key,
                 algorithms=algorithms,
-                options={"verify_exp": False},
+                options={"verify_exp": False, "verify_aud": False, "verify_nbf": False},
             )
         else:
             if algorithms is None:
--- a/redis/auth/token.py
+++ b/redis/auth/token.py
@@ -78,7 +78,28 @@ class SimpleToken(TokenInterface):
 class JWToken(TokenInterface):
     REQUIRED_FIELDS = {"exp"}
 
-    def __init__(self, token: str, key: str = None, algorithms: list = None):
+    def __init__(
+        self, token: str, key: str | None = None, algorithms: list[str] | None = None
+    ):
+        """Initialize a JWT token wrapper.
+
+        Args:
+            token: The encoded JWT string.
+            key: The secret key to verify the token signature. If None,
+                signature verification is skipped (backward compatibility).
+            algorithms: A list of allowed algorithms. Required when key is
+                provided.
+
+        Raises:
+            ImportError: If the PyJWT library is not installed.
+            ValueError: If key is provided but algorithms is None.
+            InvalidTokenSchemaErr: If required fields are missing.
+
+        Note:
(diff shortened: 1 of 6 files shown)


redis/auth/token.py
```diff
@@ -78,19 +78,32 @@ def get_received_at_ms(self) -> float:
 class JWToken(TokenInterface):
     REQUIRED_FIELDS = {"exp"}
 
-    def __init__(self, token: str):
+    def __init__(self, token: str, key: str = None, algorithms: list = None):
```
petyaslavova: Please follow PEP 604 for the type hints

redis/auth/token.py
```diff
@@ -78,19 +78,32 @@ def get_received_at_ms(self) -> float:
...
             ) from ie
         self._value = token
-        self._decoded = jwt.decode(
-            self._value,
-            options={"verify_signature": False},
-            algorithms=[jwt.get_unverified_header(self._value).get("alg")],
-        )
+        if key is not None:
+            if algorithms is None:
+                algorithms = ["HS256"]
```
petyaslavova: When `key` is provided but `algorithms` is omitted, the code silently uses ["HS256"]. 
That default is risky and likely wrong for real IdP tokens; for example, Entra ID / Azure AD commonly uses RS256.
So in this case, it will be better to require a value for `algorithms` arg when `key` is provided.
A test for key provided without algorithms would make this behavior explicit and prevent the silent HS256 fallback from coming back later.
Please also add a docstring describing the behaviour and the arguments expectations.
Add a clear statement that this change doesn't completely resolve the security issue - using `key` and `algorithms` enables a safer path, but the old unsafe usage pattern is still available for compatibility.

redis/auth/token.py
```diff
@@ -78,19 +78,32 @@ def get_received_at_ms(self) -> float:
...
+            self._decoded = jwt.decode(
+                self._value,
+                key,
+                algorithms=algorithms,
+                options={"verify_exp": False, "verify_aud": False, "verify_nbf": False},
```
petyaslavova: Please don’t add `verify_exp=False` here. This disables expiration validation, so expired tokens can still decode successfully. 
Expiration validation should remain enforced by default; otherwise, we introduce a new security problem by allowing expired tokens to be accepted.

redis/auth/token.py
```diff
@@ -78,19 +78,32 @@ def get_received_at_ms(self) -> float:
...
+            )
+        else:
+            if algorithms is None:
+                header = jwt.get_unverified_header(self._value)
+                algorithms = [header["alg"]]
+            self._decoded = jwt.decode(
+                self._value,
+                options={"verify_signature": False, "verify_exp": False},
```
petyaslavova: ditto

tests/conftest.py
```diff
@@ -750,7 +750,7 @@ def mock_identity_provider() -> IdentityProviderInterface:
     mock_provider = Mock(spec=IdentityProviderInterface)
     token = {"exp": datetime.now(timezone.utc).timestamp() + 3600, "oid": "username"}
     encoded = jwt.encode(token, "secret", algorithm="HS256")
-    jwt_token = JWToken(encoded)
+    jwt_token = JWToken(encoded, key="secret", algorithms=["HS256"])
```
petyaslavova, line 753: We still need to have tests with the old jwt_token style, so probably it will be good to introduce another mock and have tests covering both types of token initialization.

petyaslavova (changes requested):
Hey @dfgvaetyj3456356-hash, thank you for your contribution! 
Please take a look at the requested changes and additionally please fix linters and ensure all code paths are covered with tests. 
With the current change, the old, still supported, code path is completely skipped in the tests. 
We should have coverage for the old initialization, when only token is provided, as well as when we have just algorithms and key is None, and when both key and algorithm are provided. 
Also the case when a key is provided and algorithms is None - then we should validate that an error is raised.

Ersa-tech:
Opened #4093 as the replacement/update from a branch I can push to with the current GitHub account. It carries the maintainer-requested coverage for the legacy token-only path, algorithms-without-key path, key+algorithms path, key-without-algorithms error, and removes the optional redis_entraid integration test path that was only added for extra coverage. Local verification on the replacement branch: focused JWT tests, Ruff, and git diff --check all pass.

petyaslavova:
Closing this PR since all the changes are added in #4093

PR #4090, opened 2026-05-29
Fall back to RESP2 when HELLO is not supported by server/proxy

### LINKED ISSUE #4089: RESP3 default in 8.0 breaks connections to servers/proxies that don't support HELLO (no fallback)
## Problem

redis-py 8.0 changed `DEFAULT_RESP_VERSION` from 2 to 3 (PR #4031). When connecting to a Redis server < 6.0 or a proxy that doesn't support the `HELLO` command, the connection fails immediately with `ResponseError: unknown command` — there is no automatic fallback to RESP2.

## Root Cause

`on_connect_check_health()` in `connection.py:1111-1114`:

```python
self.send_command("HELLO", self.protocol, "AUTH", *auth_args, check_health=False)
self.handshake_metadata = self.read_response()  # no try-except, raises directly
```

Unlike Lettuce, Jedis, and go-redis which catch `HELLO` failures and fall back to legacy `AUTH`, redis-py raises the error and the connection is never established.

## Impact

Users upgrading from 7.x to 8.0 with **zero code changes** will experience immediate connection failures if their server/proxy doesn't support `HELLO`. This includes:

- Redis < 6.0
- Certain managed Redis proxies 
- Third-party Redis-compatible services that haven't implemented HELLO

## Suggested Fix

Catch `ResponseError` from the `HELLO` command and fall back to RESP2 + standalone `AUTH`, similar to what other clients do:

```python
try:
    self.send_command("HELLO", self.protocol, "AUTH", *auth_args, check_health=False)
    self.handshake_metadata = self.read_response()
except ResponseError:
    # Server doesn't support HELLO, fall back to RESP2 AUTH
    if isinstance(self._parser, _RESP3Parser):
        self.set_parser(_RESP2Parser)
        self._parser.on_connect(self)
    self.protocol = 2
    self.send_command("AUTH", *auth_args, check_health=False)
    auth_response = self.read_response()
    if str_if_bytes(auth_response) != "OK":
        raise AuthenticationError("Invalid Username or Password")
```

## Workaround

```python
redis.Redis(host=host, port=port, password=password, protocol=2)
```

Explicitly setting `protocol=2` avoids the `HELLO` path entirely.

yangbodong22011:
## Summary

Fixes #4089

redis-py 8.0 defaults to RESP3, which sends a `HELLO` command during connection handshake. If the server (Redis < 6.0) or proxy doesn't support `HELLO`, the connection fails with `ResponseError: unknown command` — there is no fallback.

This PR adds a `try/except ResponseError` around the `HELLO` response in both sync and async connection paths. On failure, it gracefully falls back to RESP2 with standalone `AUTH`, matching the behavior of Lettuce, Jedis, and go-redis.

## Changes

- `redis/connection.py` — `on_connect_check_health()`: catch `ResponseError` after `HELLO`, revert parser to RESP2, issue standalone `AUTH`
- `redis/asyncio/connection.py` — same fix for the async path

Both branch 1 (RESP3 + auth) and branch 3 (RESP3 + no auth) are covered.

## Testing

Manually verified against a proxy that rejects `HELLO` with `ERR unknown command`. Connection now succeeds with automatic RESP2 fallback instead of raising.

---

> [!NOTE]
> **Medium Risk**
> Changes connection handshake and authentication for all default-RESP3 clients; behavior is gated on error text and explicit protocol/maint settings, but misclassified errors could still affect connect success.
> 
> **Overview**
> When the default RESP3 connect path sends **HELLO** and the server or proxy rejects it (unknown command / NOPROTO), sync and async connections now **downgrade to RESP2** instead of failing outright. Shared helpers **`_fallback_to_resp2`** and **`_do_resp2_auth`** centralize parser/protocol downgrade, optional **standalone AUTH** with fresh credentials (avoiding HELLO’s injected `default` username), and a **warning** unless the user set **`protocol=3`** explicitly or **maintenance notifications** are enabled.
> 
> Successful HELLO responses are stored in **`handshake_metadata`** on both paths; after fallback it is set to **`{}`** so downstream code does not see **`None`**. New tests in **`tests/test_hello_fallback.py`** cover auth branches, NOPROTO, no-fallback cases, and logging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ce49533884f01ef309a10f3c9c42b78b43a870b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -270,6 +270,7 @@ class AbstractConnection:
             if p < 2 or p > 3:
                 raise ConnectionError("protocol must be either 2 or 3")
         self.protocol = p
+        self.handshake_metadata = None
         self.legacy_responses = legacy_responses
         if parser_class != _AsyncHiredisParser:
             # The Python parsers are protocol-specific; hiredis supports both.
@@ -565,8 +566,10 @@ class AbstractConnection:
                 "HELLO", self.protocol, "AUTH", *auth_args, check_health=False
             )
             try:
-                response = await self.read_response()
-                if response.get(b"proto") != int(self.protocol) and response.get(
+                self.handshake_metadata = await self.read_response()
+                if self.handshake_metadata.get(
+                    b"proto"
+                ) != int(self.protocol) and self.handshake_metadata.get(
                     "proto"
                 ) != int(self.protocol):
                     raise ConnectionError("Invalid RESP version")
@@ -586,7 +589,12 @@ class AbstractConnection:
                 self._parser.on_connect(self)
             await self.send_command("HELLO", self.protocol, check_health=check_health)
             try:
-                response = await self.read_response()
+                self.handshake_metadata = await self.read_response()
+                if (
+                    self.handshake_metadata.get(b"proto") != self.protocol
+                    and self.handshake_metadata.get("proto") != self.protocol
+                ):
+                    raise ConnectionError("Invalid RESP version")
             except ResponseError as e:
                 await self._fallback_to_resp2(e, parser)
 
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1,6 +1,7 @@
 import asyncio
 import copy
 import inspect
+import logging
 import socket
 import sys
 import time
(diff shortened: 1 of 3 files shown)


petyaslavova:
Hey @yangbodong22011, thank you for the contribution! I’ll review it shortly.

Meanwhile, can you please check Cursor’s comment and the test failures?

There are also a few things we need to be careful about when changing the protocol configuration at this point in the flow: what happens if the parser is `HiredisParser`, which callbacks would be assigned later for command response post-processing, and how this interacts with maintenance notifications and client-side caching configuration.

Having dedicated unit tests for those cases would be very helpful.

PR #4093, opened 2026-05-31
Add optional signature verification args to JWToken and perform verification when they are provided.

Ersa-tech:
## Change summary

This pull request adds opt-in JWT signature verification to `JWToken` (`redis/auth/token.py`) without changing the behaviour of the existing initialization path. 
Until now, `JWToken(token)` was decoded with `verify_signature=False`, so callers had no supported way to have the signature checked before the wrapper's claims were used. 
Two new optional parameters change that: `key` turns verification on, and `algorithms` supplies the allowed algorithm list, which is required whenever a key is given (`ValueError` otherwise) so the allow-list is always caller-configured rather than derived from the token's own header — deriving it from the header would enable algorithm-confusion attacks (RFC 8725 §2.1). `key` is typed to accept anything PyJWT accepts (shared secret, PEM, public-key object, or `PyJWK`) via `TYPE_CHECKING` imports, since PyJWT remains an optional dependency imported lazily inside `__init__`.

On the verified path, claim validation is **pinned explicitly** rather than left to PyJWT's defaults. `_PINNED_DECODE_OPTIONS` names every claim check PyJWT offers (`exp`, `nbf`, `iat`, `aud`, `iss`, `sub`, `jti`) and disables each one, because PyJWT merges caller options over its own defaults and those defaults have changed between releases — `verify_sub` and `verify_jti` arrived in 2.10 both defaulting to on, which would have made a PyJWT upgrade silently reject validly-signed tokens that the unverified path accepts. Pinning keeps redis-py's accepted-token set independent of the installed PyJWT version, and keeps claim semantics where this class already owns them: expiration is decided by `is_expired()`/`ttl()`, including the `exp == -1` never-expiring sentinel.

A third parameter, `decode_options`, lets callers re-enable individual checks on top of that baseline, so an identity provider that wants stricter validation can ask for it without waiting on a redis-py release. `FORBIDDEN_DECODE_OPTIONS` rejects `verify_signature` (decided by whether `key` was passed) and `verify_exp` (owned by this class); `decode_options` without `key` also raises, rather than being silently inert. The merged dict is rebuilt per call because PyJWT writes into the mapping it is handed. The docstring documents the pinning rationale and warns that `verify_aud`/`verify_iss` cannot work through `decode_options` alone, since PyJWT needs an `audience`/`issuer` value this class does not forward.

The unverified path is unchanged in behaviour and slightly simplified: it no longer reads the token header at all. Previously, it passed `[header.get("alg")]` to PyJWT, which is inert because PyJWT only consults the algorithm list when `verify_signature` is true; an intermediate revision indexed `header["alg"]` and raised a bare `KeyError` for a header without `alg`, which is now fixed and covered by a regression test. **Public API and compatibility:** all three parameters are keyword-optional with defaults preserving current behaviour, so `JWToken(token)` is unaffected — the only observable change on that path is the removal of the `KeyError`. **Sync/async parity:** `redis/auth/` is shared between both stacks (there is no `redis/asyncio/auth`), so no mirrored change applies. `whitelist.py` gains two vulture entries for the `TYPE_CHECKING`-only imports, following the existing `AsyncConnectionPool`/`TargetNodesT` entries.

## Test coverage

All JWToken tests are consolidated into `tests/test_auth/test_token.py::TestToken` (23 tests); the separate `tests/test_auth/test_jwt_signature.py` is removed, as it duplicated the audience, sentinel, and valid-key cases already covered there. New coverage: forged tokens rejected with a wrong key; the algorithm allow-list enforced (`InvalidAlgorithmError` when the token's `alg` is outside the list); `key` without `algorithms` raising; `InvalidTokenSchemaErr` on both decode paths; the exact options dict forwarded to PyJWT; non-string `sub`/`jti` accepted by default, which is what proves the pinning works; `decode_options` re-enabling a check, rejecting both forbidden keys, requiring `key`, and leaving the caller's dict unmutated; and a subset guard asserting PyJWT offers no claim check the pinned set omits — that one fails when a future PyJWT adds one, so the addition is triaged deliberately instead of arriving silently. Branch coverage of `redis/auth/token.py` is complete for the new code.

Time-dependent assertions in this file were rewritten for stability on slow CI runners. Eight `pytest.approx(value, N)` calls passed `N` positionally as `rel`, i.e. a 100–1000% tolerance; one of them was asserting a value 900 ms off the token it was built from and passing anyway. Values the test supplies are now compared exactly (they are stored fields or JSON round-trips), the two methods that genuinely read the clock when called are bracketed between two clock reads so the bound is exact at any runner speed, and one token's expiry window was widened from 1 s to 1 h so a stalled runner cannot expire it inside pytest's 30 s timeout. Verified with a 5 s stall injected at each point a runner could stall, and 15 consecutive clean runs.

---

> [!NOTE]
> **High Risk**
> Changes JWT decode and signature-verification behavior in auth code; incorrect caller usage or regressions on the unverified path could weaken token trust, though defaults preserve existing unverified construction.
> 
> **Overview**
> Adds an **opt-in verified** path to `JWToken` via `key`, `algorithms`, and optional `decode_options`, while keeping `JWToken(encoded)` as an unverified decode for backward compatibility.
> 
> On the verified path, PyJWT checks the signature with caller-supplied algorithms (wrong key or disallowed `alg` fails). Claim checks are **explicitly pinned off** (`verify_exp`, `verify_nbf`, `aud`, etc.) so PyJWT upgrades cannot change acceptance; expiration stays in `is_expired()` / `ttl()` including `exp == -1`. Callers cannot override `verify_signature` or `verify_exp`, and `decode_options` without `key` is rejected.
> 
> The legacy unverified path no longer reads `alg` from the JWT header or passes `algorithms` to PyJWT (RFC 8725 algorithm-confusion mitigation). `algorithms` without `key` is ignored and does not imply verification.
> 
> Tests expand coverage for verification, pinning, forbidden options, and CI-stable time assertions; `whitelist.py` whitelists TYPE_CHECKING JWT imports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dad8623315a3e099d6ed4d26fe8935bec79b9f8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


petyaslavova:
Hey @Ersa-tech, I'll take a look at this later this week.

redis/auth/token.py
```diff
@@ -78,19 +78,63 @@ def get_received_at_ms(self) -> float:
...
+            if algorithms is None:
+                raise ValueError(
+                    "algorithms must be provided when key is specified"
+                )
+            unverified_claims = jwt.decode(
+                self._value,
+                options={"verify_signature": False},
+                algorithms=algorithms,
+            )
+            options = {"verify_aud": False, "verify_nbf": False}
```
petyaslavova: I think you should add here also `verify_iat: False`.

With the current code a correctly signed token with a slightly future `iat` can now fail construction due to clock skew, even though the legacy path accepts it.

If we want verified mode to preserve the existing redis-py token semantics, I think this options dict should include `verify_iat: False` too, plus a regression test with a future `iat`.

redis/auth/token.py
```diff
@@ -78,19 +78,63 @@ def get_received_at_ms(self) -> float:
...
+            if unverified_claims.get("exp") == -1:
```
petyaslavova: Sorry, I think my earlier comment about removing `verify_exp=False` was off. I dug further into the surrounding `JWToken` logic and noticed that expiration is still validated later by redis-py via `is_expired()` / `ttl()`.
Given that, can we avoid the double decode here and set `verify_exp=False` on the verified decode directly? 
The current flow first decodes without signature verification only to inspect `exp == -1`, then decodes again with verification. Since redis-py already owns the expiration semantics, including the `exp == -1` sentinel, it seems cleaner and cheaper to verify the signature once and let the existing token methods handle expiration afterward.

tests/test_auth/test_token.py
```diff
@@ -74,4 +74,38 @@ def test_jwt_token(self):
...
+        assert jwt_token.try_get("key") == "value"
+        assert jwt_token.is_expired() is False
+
+    def test_jwt_token_with_nbf(self):
+        jwt = pytest.importorskip("jwt")
+
+        token = {
+            "exp": datetime.now(timezone.utc).timestamp() + 100,
+            "iat": datetime.now(timezone.utc).timestamp(),
+            "nbf": datetime.now(timezone.utc).timestamp() - 10,
```
petyaslavova: In order to validate you actually skip the `nbf` validation, you need to set the `nbf` with time in the future, not in the past.

petyaslavova (changes requested):

eeshsaxena (commented):
Looked at this from the bypass angle and it seems sound:

- In the verified path the real `jwt.decode` always runs with the caller-provided `algorithms`, so a token with `alg: none` or an algorithm other than what the caller expects is rejected by PyJWT rather than silently accepted. No algorithm-confusion opening as long as the caller passes a sane list, which is now required.
- The `exp == -1` handling reads the unverified claims only to decide whether to disable the expiry check, but the signature is still verified in the second decode, so an attacker cannot forge a never-expiring token without the key.
- Requiring `algorithms` when a key is provided closes the accidental "key set but verification effectively skipped" case.

One small behavior change worth noting: the no-key path moved from `header.get("alg")` to `header["alg"]`, so a token with no `alg` header now raises `KeyError` instead of proceeding with `[None]`. Probably fine and arguably better, just flagging it in case any caller relied on the old leniency.

petyaslavova:
Thanks for raising this and for the work on the original proposal — an optional key with an explicit algorithm allow-list, and a forged-token test, was a reasonable shape to put forward, and reviewing it was a useful prompt to look at the area properly.

Before deciding, I went through the design of the whole token-based authentication flow rather than just this diff, and I've concluded that redis-py is not the right layer for this validation. There are three roles here: the identity provider *issues* the token, redis-py *presents* it, and the Redis server *verifies* it. redis-py is never the relying party. `JWToken` is a thin wrapper whose only job is to expose `exp` so `TokenManager` can schedule renewal, and `try_get()` so a credentials provider can pick the AUTH username — no authorization decision is made from these claims on our side. A client verifying the signature of a token it just obtained from its own identity provider over an authenticated channel is checking its own work, and it does not change what the server accepts.

Where verification does belong is the `IdentityProviderInterface` implementation, which is caller-owned and already able to do it — and to do it more completely than a key/algorithms pair on this class can. `jwt.decode(encoded, key, algorithms=[...], audience=..., issuer=...)` followed by `JWToken(encoded)` is a couple of lines, and it can check `aud` and `iss`, which this class does not forward. A JWKS-backed provider can additionally use `PyJWKClient` for `kid` lookup. Adding a signature-only option to the core wrapper would hand users the weaker half of the job while implying the whole job was done.

So I'm going to close this PR. If you run into a concrete case where the identity provider cannot verify the token before wrapping it, please open an issue with that scenario — it would be a different design discussion, one that would need `audience`/`issuer` and JWKS support rather than this shape.

PR #4095, opened 2026-06-01
Fix maintenance notifications guard for Unix socket connections

### LINKED ISSUE #4086: Redis-py 8.0.0 - Cannot enable maintenance notifications for connection object that doesn't have a host attribute.
Greetings all, 

In hopes this isn't a redundant issue where I've failed to do a proper amount of research first, I've installed `8.0.0` with the use of `unix_socket_path`. This threw `ValueError: Cannot enable maintenance notifications for connection object that doesn't have a host attribute.`, thus I added `maint_notifications_config=MaintNotificationsConfig(enabled=False)` to the Redis connection object's initialization. It seems like this may be ignored downstream, as it continues to throw the ValueError. 

It seems like the ValueError's raised by PUBSUB, stemming from the `subscribe()` function in `redis/client.py`, which calls `connect_check_health()` downstream, which calls `activate_maint_notifications_handling_if_enabled(check_health=check_health)`, leading to `self._enable_maintenance_notifications()` throwing the ValueError. 

I'm running with `decode_responses` and `retry_on_timeout` enabled, and `legacy_responses` disabled. Have I missed something here, or is this a bug? In my mind, I've missed something, but figured I'd ask here as I prepare to dig further. 

I've also reviewed the 8.0.0 changelog, though didn't see anything that specifically stood out related to this. It _is_ possible I've overlooked something. 

Thank you.

sapirbaruch:
Fixes #4086.

## What's wrong

`activate_maint_notifications_handling_if_enabled` uses `hasattr(self, "host")` to check whether a connection has a host before attempting the maintenance notifications handshake. For `UnixDomainSocketConnection` this check always passes because the `host` abstract property is defined on the parent class — `hasattr` returns `True` even though the actual value is `None`.

The check was meant to skip Unix socket connections (which have no host), but it doesn't. `_enable_maintenance_notifications` then calls `getattr(self, "host", None)`, gets `None`, and raises:

```
ValueError: Cannot enable maintenance notifications for connection object that doesn't have a host attribute.
```

This happens even when the user explicitly passes `maint_notifications_config=MaintNotificationsConfig(enabled=False)` because the config check earlier in the guard may still evaluate truthy depending on how the pool sets it up.

## Fix

Replace `hasattr(self, "host")` with `getattr(self, "host", None) is not None`. This correctly skips any connection where `host` is `None` rather than relying on attribute existence.

```python
# before
and hasattr(self, "host")

# after
and getattr(self, "host", None) is not None
```

One-line change, no new dependencies, no API surface changes.

---

> [!NOTE]
> **Low Risk**
> Single guard change in connect-time maintenance notification activation; no API or behavior change for TCP connections with a real host.
> 
> **Overview**
> Fixes a guard in `activate_maint_notifications_handling_if_enabled` so Unix domain socket connections no longer enter the maintenance-notifications handshake path.
> 
> The condition now requires a **non-`None` host** (`getattr(self, "host", None) is not None`) instead of only checking that a `host` attribute exists. That matches the intent to skip connections without a TCP hostname (UDS uses `path` only) and avoids a `ValueError` during connect when RESP3 maintenance handling is configured on the pool.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4731d9d0b971a4191801c0a412dffb48f0a1a238. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -604,7 +604,7 @@ class MaintNotificationsAbstractConnection:
             and self.maint_notifications_config
             and self.maint_notifications_config.enabled
             and self._maint_notifications_connection_handler
-            and hasattr(self, "host")
+            and getattr(self, "host", None) is not None
         ):
             self._enable_maintenance_notifications(
                 maint_notifications_config=self.maint_notifications_config,


petyaslavova:
Hey @sapirbaruch, thank you for your contribution! I actually worked on the same issue yesterday but forgot to publish my PR. It's now available as #4097.

The fix in #4097 includes a few additional checks and more test coverage, so I think it would be safer to merge that one instead. Thanks again for taking the time to work on this!

petyaslavova:
Closing this PR for the reasons mentioned above.

PR #4052, opened 2026-04-30
Changing the default protocol used to RESP3 keeping the responses for default config compatible with current RESP2 shape. Adding possibility to opt-in for unified responses for both protocol.

petyaslavova:
## Summary

This PR changes redis-py 8.0 to use RESP3 on the wire by default while preserving legacy RESP2-compatible Python response shapes unless users explicitly opt into unified responses with `legacy_responses=False`. It adds the protocol/response-shape routing needed for the full matrix: default protocol(when it is not provided empty string is used to identify the case when we should use resp3 wire communication and return the responses in the legacy resp2 shape), `protocol=2`, and `protocol=3`, each with legacy and unified response modes.

The unified response implementation is expanded across core commands, cluster commands, JSON, Search, probabilistic commands, TimeSeries, and VectorSet. Response callbacks now normalize RESP2 and RESP3 wire replies into the approved unified shapes, while keeping legacy RESP2 and explicit native RESP3 behavior intact. Type hints and parser behavior were updated where unified responses introduce new structures.

Tests were updated to assert expected shapes using shared helpers for RESP2 legacy, RESP3 legacy, and unified responses. The test tasks now accept protocol and legacy-response combinations, emit distinct coverage/JUnit artifact names, and include matrix support for running all protocol/legacy combinations. This also includes fixes for cluster/default response configuration issues and a timing-sensitive `XREADGROUP ... CLAIM` test expectation.

Docs were refreshed for end users: README and RESP3 docs now describe the redis-py 8.0 defaults, and a new unified responses migration guide explains how to enable unified responses and what changes to expect. The old RESP unification docs/specs were replaced with unified-response-focused migration material.

| `protocol` setting | `legacy_responses` setting | Wire protocol | Python response shape |
| --- | --- | --- | --- |
| unset / default | unset / default (`True`) | RESP3 | Legacy RESP2-compatible shape |
| unset / default | `False` | RESP3 | Unified response shape |
| `protocol=2` | unset / default (`True`) | RESP2 | Legacy RESP2 shape |
| `protocol=2` | `False` | RESP2 | Unified response shape |
| `protocol=3` | unset / default (`True`) | RESP3 | Native RESP3 shape |
| `protocol=3` | `False` | RESP3 | Unified response shape |

---

> [!NOTE]
> **High Risk**
> High risk because it changes default wire protocol behavior and introduces new callback routing that can affect response shapes across many core/module commands, plus expands CI coverage matrix that may surface previously untested incompatibilities.
> 
> **Overview**
[...]

PR #4109, opened 2026-06-08
fix(search): parse RESP3 FT.SEARCH responses with bytes-typed keys

### LINKED ISSUE #4107: FT.SEARCH returns Result{0 total} with RESP3 byte-key response when decode_responses=False
redis-py: 8.0.0
Redis: Redis Cloud / Redis 8 Search
Connection: decode_responses=False

client.ft("INDEX").search(Query(...)) returns Result{0 total, docs: []}

But raw FT.SEARCH returns:
{
  b"total_results": 2,
  b"results": [...],
  b"warning": []
}

Result.from_resp3() appears to read string keys like "total_results" and "results",
so byte-key RESP3 mappings are silently parsed as empty.

mokashang:
### Description of change

Fixes #4107.

Since the wire protocol default switched to RESP3 in #4052, the server returns FT.SEARCH responses as RESP3 maps. When a client is opened with `decode_responses=False` (the default for binary callers), those map keys arrive as `bytes`, but `Result.from_resp3` looked them up as plain `str`:

```py
instance.total = res.get("total_results", 0)
for result_item in res.get("results", []):
    ...
```

Because `b"total_results" != "total_results"`, every lookup missed and `client.ft("INDEX").search(...)` returned `Result{0 total, docs: []}` even when the server had matched documents. The reporter confirms the raw response looks like:

```py
{b"total_results": 2, b"results": [...], b"warning": []}
```

### Fix

Normalise the top-level map and each per-result map to string keys before reading them. This mirrors the pattern already established in `_parse_hybrid_search_resp3` (`redis/commands/search/commands.py`), whose docstring calls out: *"Top-level keys are normalised to strings."* The fix is a 5-line change, no value semantics are altered — bytes values are still passed through `decode_field_value` exactly as before.

### Testing

Added `tests/test_search_result.py` (marked `fixed_client` since `Result.from_resp3` is a pure function and needs no Redis):

- `test_str_keys_are_parsed` — baseline, str-keyed map still works.
- `test_bytes_keys_are_parsed` — regression test for #4107, fails on master.
- `test_mixed_str_and_bytes_keys` — outer bytes / inner str (different parser layers can produce mixed maps), fails on master.
- `test_empty_or_none_response` — `None` and empty-dict edge cases.

Locally on Python 3.13:

```
$ pytest tests/test_search_result.py -v
tests/test_search_result.py::TestResultFromResp3::test_str_keys_are_parsed PASSED
tests/test_search_result.py::TestResultFromResp3::test_bytes_keys_are_parsed PASSED
tests/test_search_result.py::TestResultFromResp3::test_mixed_str_and_bytes_keys PASSED
tests/test_search_result.py::TestResultFromResp3::test_empty_or_none_response PASSED
========================= 4 passed in 0.02s =========================
```

`ruff check` and `ruff format --check` are clean for both files.

The `fixed_client` suite passes on this branch with the same set of unrelated pre-existing failures (`test_connection.py::test_connect_with_retries` and friends — confirmed they reproduce on master without this change).

### Scope note

The same RESP3-bytes-keys pattern likely affects `_parse_aggregate_resp3` (FT.AGGREGATE) and possibly other RESP3 callbacks reached from `decode_responses=False` clients. I've deliberately kept this PR scoped to the reported FT.SEARCH path; happy to open a follow-up that audits the rest if useful.

### Pull Request check-list

- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? — N/A, behavior fix only.
- [x] Is there an example added to the examples folder (if applicable)? — N/A

---

> [!NOTE]
> **Low Risk**
> Targeted parser normalisation for structural keys only; document field values and RESP2 paths are unchanged, with broad unit and integration coverage.
> 
> **Overview**
> Fixes **#4107**: with default RESP3 and `decode_responses=False`, RediSearch map keys are `bytes`, so legacy parsers that looked up `"total_results"` / `"results"` missed and returned empty search, aggregate, and spellcheck results.
> 
> **Parser changes** use `str_if_bytes` on structural map keys (and per-hit maps) in `Result.from_resp3`, `_parse_aggregate_resp3`, and `_parse_spellcheck_resp3`, matching the existing hybrid-search RESP3 path. `_parse_spellcheck_resp3` no longer coerces suggestion text to `str`, so term/suggestion types stay aligned with RESP2 when decoding is off.
> 
> **Tests** add `tests/test_search_result.py` (no server), parametrised integration cases for search/aggregate/spellcheck under RESP2 vs default RESP3, and harden `waitForIndex` for raw `FT.INFO` with `str` or `bytes` keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f20d858cd90035b8abc1a59f4d62633dd1f69ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/commands/search/commands.py
+++ b/redis/commands/search/commands.py
@@ -666,8 +666,11 @@ class SearchCommands:
                     score_str = str(score)
                     if score_str.endswith(".0"):
                         score_str = score_str[:-2]
+                    # Preserve ``suggestion`` as-is so it keeps the
+                    # ``decode_responses`` shape RESP2 would produce
+                    # (``str`` when decoded, ``bytes`` otherwise).
                     term_corrections.append(
-                        {"score": score_str, "suggestion": str(suggestion)}
+                        {"score": score_str, "suggestion": suggestion}
                     )
             if term_corrections:
                 corrections[term] = term_corrections
--- a/redis/commands/search/commands.py
+++ b/redis/commands/search/commands.py
@@ -494,11 +494,19 @@ class SearchCommands:
         else:
             data = res
 
+        if data is None:
+            data = {}
+        # On RESP3 connections with decode_responses=False the server's map
+        # keys arrive as bytes, so normalise structural keys to strings
+        # before lookup.  Mirrors ``Result.from_resp3``.
+        data = {str_if_bytes(k): v for k, v in data.items()}
+
         warnings = [str_if_bytes(w) for w in data.get("warning", [])]
         total = data.get("total_results", 0)
 
         rows = []
         for result_item in data.get("results", []):
+            result_item = {str_if_bytes(k): v for k, v in result_item.items()}
             extra_attrs = result_item.get("extra_attributes", {})
             # Convert dict to flat list [key, value, key, value, ...]
             # to match RESP2 row format consumers expect.
@@ -640,6 +648,10 @@ class SearchCommands:
         """
         if not isinstance(res, dict):
             return self._parse_spellcheck(res, **kwargs)
+        # On RESP3 connections with decode_responses=False the server's map
+        # keys arrive as bytes, so normalise the structural ``results`` key
+        # to a string before lookup.  Mirrors ``Result.from_resp3``.
+        res = {str_if_bytes(k): v for k, v in res.items()}
         corrections = {}
         results = res.get("results", {})
         for term, suggestions in results.items():
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -5610,9 +5610,10 @@ class TestHybridSearch(SearchTestsBase):
 # expected legacy output shape with ``decode_responses=False``.  The
 # default protocol (``SENTINEL`` -> not specified) leaves the wire on
 # RESP3 with the ``_RedisCallbacksRESP3toRESP2Legacy`` adapter selected,
-# which is the path the fix actually exercises.  An explicit
-# ``protocol=3`` would route through ``_RedisCallbacksRESP3`` instead
-# and bypass the changed methods.
+# which is where the bytes-key normalisation in ``_parse_search_resp3``,
+# ``_parse_aggregate_resp3`` and ``_parse_spellcheck_resp3`` lives.  An
+# explicit ``protocol=3`` would route through ``_RedisCallbacksRESP3``
+# instead and bypass the methods we want to test.
 _SEARCH_BYTES_PROTOCOLS = [
     pytest.param(2, id="resp2"),
     pytest.param(SENTINEL, id="default-resp3"),
@@ -5644,9 +5645,9 @@ class TestSearchResp3BytesKeys(SearchTestsBase):
     parsers in ``_RedisCallbacksRESP3toRESP2Legacy``).
     """
 
-    @pytest.mark.parametrize("protocol", _SEARCH_BYTES_PROTOCOLS)
     @pytest.mark.redismod
     @pytest.mark.fixed_client
+    @pytest.mark.parametrize("protocol", _SEARCH_BYTES_PROTOCOLS)
     def test_search_resp3_bytes_keys(self, request, stack_url, protocol):
         client = _make_bytes_search_client(request, stack_url, protocol)
         client.ft().create_index((TextField("title"), TextField("body")))
@@ -5665,9 +5666,9 @@ class TestSearchResp3BytesKeys(SearchTestsBase):
         assert res.total == 2
         assert {d.id for d in res.docs} == {"doc1", "doc2"}
 
-    @pytest.mark.parametrize("protocol", _SEARCH_BYTES_PROTOCOLS)
     @pytest.mark.redismod
     @pytest.mark.fixed_client
+    @pytest.mark.parametrize("protocol", _SEARCH_BYTES_PROTOCOLS)
     def test_aggregate_resp3_bytes_keys(self, request, stack_url, protocol):
         client = _make_bytes_search_client(request, stack_url, protocol)
         client.ft().create_index((TextField("title"), TextField("parent")))
@@ -5693,9 +5694,9 @@ class TestSearchResp3BytesKeys(SearchTestsBase):
         assert b"parent" in row
         assert b"redis" in row
 
-    @pytest.mark.parametrize("protocol", _SEARCH_BYTES_PROTOCOLS)
     @pytest.mark.redismod
     @pytest.mark.fixed_client
+    @pytest.mark.parametrize("protocol", _SEARCH_BYTES_PROTOCOLS)
     def test_spellcheck_resp3_bytes_keys(self, request, stack_url, protocol):
         client = _make_bytes_search_client(request, stack_url, protocol)
         client.ft().create_index((TextField("f1"),))
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -5590,3 +5590,95 @@ class TestHybridSearch(SearchTestsBase):
                 assert item["description"] is not None
                 assert item["discount_10_percents"] is not None
                 assert item["additional_discount"] is not None
+
+
+class TestSearchResp3BytesKeys(SearchTestsBase):
+    """Regression tests for #4107.
+
+    With ``protocol=3`` on the wire and ``decode_responses=False`` the
+    server's structural map keys arrive as ``bytes`` (e.g. ``b"results"``
+    rather than ``"results"``).  The RESP3->legacy-RESP2 search callbacks
+    used to look those keys up as plain strings, missed them, and
+    silently produced empty results.  These tests pin a
+    ``decode_responses=False`` client against a RESP3 wire with the
+    library's default legacy responses and exercise the three callbacks
+    that were affected: FT.SEARCH, FT.AGGREGATE and FT.SPELLCHECK.
+    """
+
+    @pytest.fixture
+    def bytes_client(self, request, stack_url):
+        r = _get_client(
+            redis.Redis,
+            request,
+            decode_responses=False,
+            protocol=3,
+            from_url=stack_url,
+        )
+        r.flushdb()
+        return r
+
+    @pytest.mark.redismod
+    @pytest.mark.fixed_client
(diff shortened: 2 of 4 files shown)


petyaslavova (changes requested):
Thanks for the focused fix, this covers the reported `FT.SEARCH` path nicely.

Could you also extend the same RESP3 byte-key handling to the remaining Search RESP3 parsers that still read structural map keys as plain strings? I think there are two left:

- `_parse_aggregate_resp3()` for `FT.AGGREGATE` / `FT.CURSOR READ` / `FT.PROFILE AGGREGATE`
- `_parse_spellcheck_resp3()` for `FT.SPELLCHECK`

For example, with `decode_responses=False`, byte-keyed RESP3 aggregate responses still parse as total `0` with no rows, and byte-keyed spellcheck responses parse as `{}`. A couple of small regression tests for those byte-keyed RESP3 shapes would make this fix feel complete across the affected Search callbacks.
I would also suggest adding the tests using the integration test approach - just providing a client instance with `decode_responses=False` and use actual communication with the server

mokashang:
Thanks for the review @petyaslavova — extended in 47267ebf:

- `_parse_aggregate_resp3` (FT.AGGREGATE / FT.CURSOR READ / FT.PROFILE AGGREGATE): normalise both the top-level map and the per-result-item map. Document data inside `extra_attributes` is intentionally left as-is so the caller still sees bytes when `decode_responses=False`, mirroring the RESP2 `AggregateResult.rows` shape.
- `_parse_spellcheck_resp3` (FT.SPELLCHECK): normalise the outer `results` key. The inner term keys are not normalised; that matches the RESP2 spellcheck shape with `decode_responses=False`, where `corrections[_correction[1]]` keeps the term in its wire form.

On the test approach: I went with both layers — `tests/test_search_result.py` already covered `Result.from_resp3` and now also has parser-level regression tests for the two new fixes (these are the ones that fail on the unpatched code), and `tests/test_search.py` has a new `TestSearchResp3BytesKeys` class with integration tests for FT.SEARCH, FT.AGGREGATE and FT.SPELLCHECK that hit a real RediSearch via a pinned `decode_responses=False` / `protocol=3` client, as you suggested.

tests/test_search.py
```diff
@@ -5590,3 +5590,95 @@ def test_hybrid_search_query_with_multiple_loads_and_applies(self, client):
...
+        return r
+
+    @pytest.mark.redismod
+    @pytest.mark.fixed_client
+    def test_search_resp3_bytes_keys(self, bytes_client):
+        client = bytes_client
+        client.ft().create_index((TextField("title"), TextField("body")))
+        client.hset("doc1", mapping={"title": "hello", "body": "redis world"})
+        client.hset("doc2", mapping={"title": "hello", "body": "search world"})
+        self.waitForIndex(client, getattr(client.ft(), "index_name", "idx"))
```
petyaslavova, line 5656: waitForIndex needs to be refactored a little - currently this one directly calls FT.INFO through  execute_command which doesn't go through callbacks execution and with the current client instance the keys are bytes, while the helper expects strings - this leads to the test failures in the pipeline.

tests/test_search.py
```diff
@@ -5590,3 +5590,95 @@ def test_hybrid_search_query_with_multiple_loads_and_applies(self, client):
...
+    that were affected: FT.SEARCH, FT.AGGREGATE and FT.SPELLCHECK.
+    """
+
+    @pytest.fixture
+    def bytes_client(self, request, stack_url):
+        r = _get_client(
+            redis.Redis,
+            request,
+            decode_responses=False,
+            protocol=3,
```
petyaslavova: By passing `protocol=3`, these tests don’t actually exercise the changed methods. Explicit `protocol=3` + `legacy_responses=True` keeps the native RESP3 shape, while the fix is in the default RESP3-on-wire legacy path. So the affected parsers (`_parse_search_resp3`, `_parse_aggregate_resp3`, `_parse_spellcheck_resp3`) are bypassed.

Could you make the client setup a small helper instead of a fixture, and parameterize each test over the protocol resource? For this regression, `protocol=2` and the default protocol path should produce the same legacy output format, so the assertions can stay the same. The important case is the default protocol path, because that triggers the problematic RESP3-to-RESP2 legacy parser path.

For example, one test could be shaped like:

```py
from redis.utils import SENTINEL

_SEARCH_PROTOCOLS = [
    pytest.param(2, id="resp2"),
    pytest.param(SENTINEL, id="default-resp3"),
]

def _make_bytes_client(request, stack_url, protocol):
    kwargs = {
        "decode_responses": False,
        "from_url": stack_url,
    }
    if protocol is not SENTINEL:
        kwargs["protocol"] = protocol

    client = _get_client(redis.Redis, request, **kwargs)
    client.flushdb()
    return client

@pytest.mark.parametrize("protocol", _SEARCH_PROTOCOLS)
@pytest.mark.redismod
@pytest.mark.fixed_client
def test_search_resp3_bytes_keys(self, request, stack_url, protocol):
    client = _make_bytes_client(request, stack_url, protocol)

    client.ft().create_index((TextField("title"), TextField("body")))
    client.hset("doc1", mapping={"title": "hello", "body": "redis world"})
    client.hset("doc2", mapping={"title": "hello", "body": "search world"})
    self.waitForIndex(client, getattr(client.ft(), "index_name", "idx"))

    res = client.ft().search(Query("hello"))

    assert res.total == 2
    assert {d.id for d in res.docs} == {"doc1", "doc2"}
```

The same pattern should be applied to aggregate and spellcheck. In that shape, the RESP2 case guards output compatibility, and the default protocol case exercises the actual changed parser path.

redis/commands/search/commands.py
petyaslavova: Would you also fix this line - currently it seems incompatible with what protocol=2 would return - it needs to become `{"score": score_str, "suggestion": suggestion}`
Could you also add more strict validation for this case.. For example, sth like `assert suggestions[0]["suggestion"] == b"important"`

petyaslavova (changes requested):

mokashang:
Thanks for the deeper review @petyaslavova — addressed in 36e7bac7:

- **Suggestion type fix** in `_parse_spellcheck_resp3`: dropped the `str(suggestion)` wrap so the value passes through in its wire form (`str` when `decode_responses=True`, `bytes` otherwise), matching what `_parse_spellcheck` returns for RESP2.
- **`waitForIndex` for bytes responses**: `env.execute_command("FT.INFO", idx)` bypasses the search module's callbacks, so with `decode_responses=False` the helper was looking up structural keys with the wrong type and breaking the `fixed-clients` job. It now accepts both `str` and `bytes` for the `indexing` key on both the RESP2 flat-list and RESP3 dict shapes.
- **Test parametrisation**: replaced the `bytes_client` fixture with a `_make_bytes_search_client` helper and parametrised each test over `[pytest.param(2, id="resp2"), pytest.param(SENTINEL, id="default-resp3")]`. The default-protocol leg leaves `protocol` unset so the client routes through `_RedisCallbacksRESP3toRESP2Legacy` (where the fix actually lives); the previous explicit `protocol=3` was routing through `_RedisCallbacksRESP3` and skipping the changed parsers. RESP2 acts as the anchor for the expected legacy shape.
- **Stricter spellcheck assertion**: now pins the term key to `b"impornant"` and `suggestions[0]["suggestion"] == b"important"`, as suggested. Mirrored the same `b"important"` assertion in the `test_search_result.py` unit test.

tests/test_search.py
```diff
@@ -5592,54 +5605,71 @@ def test_hybrid_search_query_with_multiple_loads_and_applies(self, client):
...
 
 
+# Parametrise the bytes-key regression tests over RESP2 and the
+# default protocol.  RESP2 uses ``_RedisCallbacksRESP2`` and anchors the
+# expected legacy output shape with ``decode_responses=False``.  The
+# default protocol (``SENTINEL`` -> not specified) leaves the wire on
+# RESP3 with the ``_RedisCallbacksRESP3toRESP2Legacy`` adapter selected,
+# which is the path the fix actually exercises.  An explicit
+# ``protocol=3`` would route through ``_RedisCallbacksRESP3`` instead
+# and bypass the changed methods.
```
petyaslavova: "bypass the methods we want to test" is more appropriate here - currently we add those tests targeting a fix for specific issue, but in a few months if we look at the code, we won't remember what are those changed methods, why they are changed, etc.

tests/test_search.py
```diff
@@ -5592,54 +5605,71 @@ def test_hybrid_search_query_with_multiple_loads_and_applies(self, client):
...
-            redis.Redis,
-            request,
-            decode_responses=False,
-            protocol=3,
-            from_url=stack_url,
-        )
-        r.flushdb()
-        return r
-
+    @pytest.mark.parametrize("protocol", _SEARCH_BYTES_PROTOCOLS)
```
petyaslavova: If you move this annotation below the other two, I think it would make the code a bit easier to read and keep the related annotations grouped together. This applies to all similar annotations.

petyaslavova (commented):
Hey @mokashang, great work! Thank you for your efforts on this fix!
There are just two more small/mostly cosmetic/ comments and this PR can be merged :)

mokashang:
Thanks @petyaslavova — addressed in e463696:

- **Comment phrasing**: rewrote the block above `_SEARCH_BYTES_PROTOCOLS` so the rationale stands on its own. It now names the parsers being exercised (`_parse_search_resp3` / `_parse_aggregate_resp3` / `_parse_spellcheck_resp3`) and ends with "would route through `_RedisCallbacksRESP3` instead and bypass the methods we want to test", so the *why* stays legible without the PR diff as context.
- **Decorator order**: moved `@pytest.mark.parametrize("protocol", _SEARCH_BYTES_PROTOCOLS)` below `@pytest.mark.redismod` / `@pytest.mark.fixed_client` on all three `TestSearchResp3BytesKeys` methods, keeping the test-scope marks grouped and `parametrize` last. Collection still yields both `[resp2]` and `[default-resp3]` ids.

petyaslavova (approved):
LGTM.

PR #4111, opened 2026-06-09
Fix async cluster node connection release on write errors

### LINKED ISSUE #4110: Async RedisCluster: ClusterNode.execute_command leaks a pool connection on every write-side error (send_packed_command outside try/finally) -> permanent MaxConnectionsError
### Summary

In `redis/asyncio/cluster.py`, `ClusterNode.execute_command` acquires a connection and calls `send_packed_command` (and `disconnect_if_needed`) **outside** the `try/finally` that returns the connection to the pool. Any write-side error — `ConnectionError`/`TimeoutError`/`OSError`, i.e. the routine errors seen during ElastiCache/Valkey failovers and node replacements — therefore skips `release()`, permanently stranding the connection in `self._connections` without ever returning it to `self._free`. This is a monotonic per-node slot leak. Once `len(self._connections)` reaches `max_connections`, `acquire_connection` raises `MaxConnectionsError` on every subsequent call to that node, even though the server is idle. It never self-heals; only recreating the client recovers.

```python
async def execute_command(self, *args, **kwargs):
    connection = self.acquire_connection()        # appended to self._connections
    await self.disconnect_if_needed(connection)   # <-- outside try
    await connection.send_packed_command(...)      # <-- outside try; raises here on a failover write
    try:
        return await self.parse_response(connection, args[0], **kwargs)
    finally:
        self.release(connection)                   # <-- ONLY path back to self._free; skipped on the above
```

`acquire_connection` appends the new connection to `self._connections` before returning it (counting it against `max_connections`), and `release()` in the `finally` is the only path that puts it into `self._free`. When the pre-`try` `send_packed_command`/`disconnect_if_needed` raises, the connection is counted but never freed.

### Why it never recovers

`self._connections` is only emptied in `ClusterNode.__init__`. A slot-map reinitialize does **not** rebuild existing nodes — `NodesManager.set_nodes` preserves the existing `ClusterNode` instances (and their leaked `_connections`) for any node whose name is unchanged. So a topology refresh does not clear the leak; only constructing a new client does. This matches the long-standing reports of the client being stuck until a process/pod restart.

### Why this surfaces broadly in 8.0.0

The per-node `max_connections` default dropped from `2**31` (7.x) to `100` (8.0.0). The leak existed before but was effectively unreachable under the old ceiling; under 100 a long-lived client exhausts a node's pool after roughly `max_connections` accumulated transient write errors (we observed ~3 days of uptime against AWS ElastiCache in cluster mode before exhaustion).

### Observable symptoms (match existing reports)

- `redis.exceptions.MaxConnectionsError` on every request routed to the affected node, while the server itself shows only a handful of live connections (the leaked sockets were closed on the failed write, so the server never counts them).
- Recovery only on client/process restart.
- This appears to be the undiagnosed root cause of #3929 (async client, AWS ElastiCache upgrade, unrecoverable-until-restart, `MaxConnectionsError` + `pop from an empty deque`) and #3135 (async client, `MaxConnectionsError` once max connections is reached). It is distinct from the sync-path thundering-herd in #4074 / #4083, and adjacent to #4028 which touched the same method for a different concern (health-check bypass).

### Proposed fix

Move `send_packed_command` (and the preceding `disconnect_if_needed`) inside the `try`, so `release()` always runs in the `finally` — mirroring how the standalone client guarantees the connection is returned to the pool. `ClusterNode.execute_pipeline` has the same shape and should be fixed identically.

### Environment

- redis-py 8.0.0 (also confirmed present on current `master`)
- Python 3.11
- `redis[cluster]`, async client (`redis.asyncio.cluster.RedisCluster`)
- AWS ElastiCache, cluster mode enabled

petyaslavova:
This fixes async Redis Cluster node connection handling so acquired connections are returned to the per-node free queue even when write-side setup fails. `ClusterNode.execute_command()` and `ClusterNode.execute_pipeline()` now place lazy disconnect and `send_packed_command()` inside the release-guarded `try` block.

The cleanup path also now releases the connection even if the final `disconnect_if_needed()` raises. This prevents a failed disconnect during cleanup from stranding the connection in `_connections` and eventually exhausting `max_connections`.

Regression tests cover both direct command execution and node pipeline execution with `max_connections=1`. The tests simulate write failures and disconnect failures, then verify the same connection is returned to `_free` and can be reused instead of leaking.

Fixes #4110 

---

> [!NOTE]
> **Medium Risk**
> Changes connection lifecycle on error paths in async cluster command/pipeline execution; incorrect cleanup could still leak or double-release connections, but scope is limited to ClusterNode pool handling and is covered by new tests.
> 
> **Overview**
> Fixes **async Redis Cluster** per-node connection pooling so acquired connections are always returned to `_free`, even when work fails before a response is read.
> 
> `ClusterNode.execute_command()` and `execute_pipeline()` now wrap lazy disconnect, `send_packed_command`, and response handling in a **`try`/`finally`** cleanup path. **`release(connection)`** runs in a nested `finally` so it still executes if **`disconnect_if_needed()`** raises during teardown—avoiding stranded connections and **`MaxConnectionsError`** under **`max_connections=1`** (fixes #4110).
> 
> Regression tests simulate **write failures** and **disconnect failures** on both single-command and node-pipeline paths and assert the same connection is reused from `_free`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1b9a0b122b88ebf9386afeada7133fb424ff5f73. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1683,47 +1683,52 @@ class ClusterNode:
     async def execute_command(self, *args: Any, **kwargs: Any) -> Any:
         # Acquire connection
         connection = self.acquire_connection()
-        # Handle lazy disconnect for connections marked for reconnect
-        await self.disconnect_if_needed(connection)
-
-        # Execute command
-        await connection.send_packed_command(connection.pack_command(*args))
-
-        # Read response
         try:
+            # Handle lazy disconnect for connections marked for reconnect
+            await self.disconnect_if_needed(connection)
+
+            # Execute command
+            await connection.send_packed_command(connection.pack_command(*args))
+
+            # Read response
             return await self.parse_response(connection, args[0], **kwargs)
         finally:
-            await self.disconnect_if_needed(connection)
-            # Release connection
-            self.release(connection)
+            try:
+                await self.disconnect_if_needed(connection)
+            finally:
+                # Release connection
+                self.release(connection)
 
     async def execute_pipeline(self, commands: List["PipelineCommand"]) -> bool:
         # Acquire connection
         connection = self.acquire_connection()
-        # Handle lazy disconnect for connections marked for reconnect
-        await self.disconnect_if_needed(connection)
+        try:
+            # Handle lazy disconnect for connections marked for reconnect
+            await self.disconnect_if_needed(connection)
 
-        # Execute command
-        await connection.send_packed_command(
-            connection.pack_commands(cmd.args for cmd in commands)
-        )
+            # Execute command
+            await connection.send_packed_command(
+                connection.pack_commands(cmd.args for cmd in commands)
+            )
 
-        # Read responses
-        ret = False
-        for cmd in commands:
+            # Read responses
+            ret = False
+            for cmd in commands:
+                try:
+                    cmd.result = await self.parse_response(
+                        connection, cmd.args[0], **cmd.kwargs
+                    )
+                except Exception as e:
+                    cmd.result = e
+                    ret = True
+
+            return ret
+        finally:
             try:
-                cmd.result = await self.parse_response(
-                    connection, cmd.args[0], **cmd.kwargs
-                )
-            except Exception as e:
-                cmd.result = e
-                ret = True
-
-        # Release connection
-        await self.disconnect_if_needed(connection)
-        self.release(connection)
-
-        return ret
+                await self.disconnect_if_needed(connection)
+            finally:
+                # Release connection
+                self.release(connection)
 
     async def re_auth_callback(self, token: TokenInterface):
         tmp_queue = collections.deque()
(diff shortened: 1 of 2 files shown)


vladvildanov (approved):

rlehman221:
@petyaslavova when this fix goes in how can we use it since I know releases don't happen often but this is a p0 for our team right now to have fixed

petyaslavova:
> @petyaslavova when this fix goes in how can we use it since I know releases don't happen often but this is a p0 for our team right now to have fixed

I've been collecting bug fixes for a new minor release or a patch. I have a few more to handle, and I'm planning to release either new minor version or a patch in the following days, or next week.

PR #4114, opened 2026-06-11
Adding support in async standalone client for maintenance notifications

petyaslavova:
### Description of change

### Pull Request check-list

---

> [!NOTE]
> **High Risk**
> Changes core async connection pool locking, reconnection, and timeout behavior during server maintenance events; mistakes could cause connection leaks, missed handoffs, or stuck pools.
> 
> **Overview**
> Adds **maintenance notifications** to the async standalone `Redis` client and connection pool, mirroring the sync feature for RESP3 TCP connections: `MaintNotificationsConfig` on `redis.asyncio.Redis`, auto-enable on default RESP3 pools, and guards for Unix sockets, missing hosts, and non-mixin connection classes.
> 
> New **`redis/asyncio/maint_notifications.py`** handles pool- and connection-level push events (MOVING, migrate/failover start/end), schedules TTL cleanup with `asyncio` tasks, and records observability metrics. Async connections gain mixins for the `CLIENT MAINT_NOTIFICATIONS` handshake, parser push handlers, relaxed read timeouts during maintenance (including rescheduling in-flight `read_response` deadlines), and lifecycle cleanup on disconnect.
> 
> The async pool mixin centralizes **atomic MOVING** updates—matching connections, proactive reconnect, and future `connection_kwargs`—under the non-reentrant pool lock; `BlockingConnectionPool` can serialize get/release during maintenance. Related sync paths are aligned via shared helpers in `redis/maint_notifications.py`, `check_protocol_version` (including `SENTINEL`), stricter `can_read` checks when notifications are enabled, and safer single-connection reconnect in `finally` blocks.
> 
> Large new async test suite plus sync/pool test updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit efbe247883442c3965fd14648668e69ac9de11a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -206,6 +206,10 @@ class AsyncMaintNotificationsAbstractConnection:
     ) -> Any:
         pass
 
+    @abstractmethod
+    def getpeername(self) -> str | None:
+        pass
+
     def _configure_maintenance_notifications(
         self,
         maint_notifications_pool_handler: (
@@ -381,7 +385,7 @@ class AsyncMaintNotificationsAbstractConnection:
         # Method 1: Try to get the actual IP from the established stream.
         # This is most accurate as it shows the exact IP being used.
         try:
-            peer_addr = self._get_peer_address()
+            peer_addr = self.getpeername()
             if peer_addr:
                 return peer_addr
         except (AttributeError, OSError):
@@ -433,18 +437,6 @@ class AsyncMaintNotificationsAbstractConnection:
     def reset_received_notifications(self) -> None:
         self._processed_start_maint_notifications.clear()
         self._skipped_end_maint_notifications.clear()
-
-    def getpeername(self) -> str | None:
-        """
-        Returns the peer name of the connection.
-        """
-        writer = getattr(self, "_writer", None)
-        if writer is None:
-            return None
-        peername = writer.get_extra_info("peername")
-        if isinstance(peername, tuple) and peername:
-            return str(peername[0])
-        return None
 
     def update_current_socket_timeout(
         self, relaxed_timeout: float | None = None
@@ -748,6 +740,18 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
 
     def _get_parser(self) -> BaseParser:
         return self._parser
+
+    def getpeername(self) -> str | None:
+        """
+        Returns the peer name of the connection.
+        """
+        writer = self._writer
+        if writer is None:
+            return None
+        peername = writer.get_extra_info("peername")
+        if isinstance(peername, tuple) and peername:
+            return str(peername[0])
+        return None
 
     async def connect(self):
         """Connects to the Redis server if not already connected"""
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -2232,33 +2232,25 @@ class AsyncMaintNotificationsAbstractConnectionPool:
         can move between active/free lists and escape handling.
         """
         async with self._get_pool_lock():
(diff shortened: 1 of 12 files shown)


redis/asyncio/maint_notifications.py
```diff
@@ -0,0 +1,350 @@
...
+
+    def _schedule(
+        self,
+        delay: float,
+        callback: _ScheduledCallback,
+        *args: Any,
+    ) -> None:
+        task = asyncio.create_task(self._run_after(delay, callback, *args))
+        self._scheduled_tasks.add(task)
+        task.add_done_callback(self._scheduled_tasks.discard)
```
vladvildanov, line 218: `_scheduled_tasks` aren't cancelled on pool/client disconnect

tests/maint_notifications/test_maint_notifications_handling.py
```diff
@@ -709,6 +744,47 @@ def test_connection_pool_creation_with_maintenance_notifications(self, pool_clas
...
+            max_connections=max_connections,
+            protocol=3,
+            maint_notifications_config=MaintNotificationsConfig(enabled=enabled),
+        )
+        connection = MagicMock()
+        connection.pid = pool.pid
+        connection.can_read.side_effect = [RedisConnectionError("closed"), True]
+        connection.should_reconnect.return_value = False
+
+        if pool_class == BlockingConnectionPool:
```
vladvildanov, line 765: In sync version you have a logic to set a `BlockingConnectionPool` to a maintenance mode which prevents creation/releasing new connections while MOVING notifications are processed, but async version missing this.

redis/asyncio/connection.py
```diff
@@ -124,7 +137,367 @@ async def __call__(self, connection: "AbstractConnection"): ...
...
+                raise ResponseError(
+                    "The server doesn't support maintenance notifications"
+                )
+        except Exception as e:
+            if (
+                isinstance(e, ResponseError)
+                and maint_notifications_config.enabled == "auto"
+            ):
+                # Log warning but don't fail the connection
+                import logging
```
vladvildanov, line 369: It looks like there's already a module-level logger, so it should be used instead

vladvildanov (changes requested):

redis/asyncio/connection.py
petyaslavova: No for this file. So added this here to be in sync with the synchronous implementation.

redis/asyncio/maint_notifications.py
petyaslavova: I think we should call the cancel on pool/client close - after that moment the client or pool should not usable and canceled revert of the original config is not needed. If we cancel on disconnect we will leave the configuration for the new and for the existing connections in the tmp state set by MOVING. I added the cancel logic to be called when pool's `aclose` is called.

tests/maint_notifications/test_maint_notifications_handling.py
petyaslavova: Good catch. Fixed it. In my initial pass, the blocking pool was taking _lock on both get_connection and release. I reverted release to always lock — removing the lock around await connection.disconnect() would leave the connection in neither _in_use_connections nor _available_connections while the await is suspended, which is exactly the window where the MOVING handler can iterate both lists and miss it. So release stays as it was. I kept the conditional locking on get_connection only (its critical section has no await so the sync flag pattern is safe there), matching how the sync BlockingConnectionPool does it.

vladvildanov (approved):

Issue #4117, opened 2026-06-12
hiredis parser raises ValueError: filedescriptor out of range in select() — _socket_can_read uses select.select() (FD_SETSIZE limit)


violuke:
**redis-py version:** 8.0.0
**hiredis:** 3.4.0
**Python:** 3.14.5 (also reproduces on earlier 3.x)
**Platform:** Linux

### Describe the bug

When using the hiredis parser, `redis/_parsers/hiredis.py:_socket_can_read()` calls `select.select([sock], [], [], timeout)`. POSIX `select()` cannot handle a file descriptor `>= FD_SETSIZE` (typically 1024) and raises `ValueError: filedescriptor out of range in select()`. This limit is fixed at compile time and is **not** raisable via `ulimit -n` / `RLIMIT_NOFILE`.

`Connection.can_read()` is called on every connection acquisition (`ConnectionPool.get_connection`), so in any long-running process that accumulates file descriptors (web app workers, connection pools, background workers), once a redis socket is assigned an fd `>= 1024`, **every** subsequent command on that process fails with this error until the process is restarted.

The pure-Python parser does not have this problem — `_parsers/socket.py:SocketBuffer.can_read()` uses a non-blocking `recv()` rather than `select()`. Only the hiredis path is affected. This is the same underlying limitation reported long ago in #419 and #1115, which was addressed for the python parser but never for the hiredis parser.

### To Reproduce

```python
import resource, socket, select
_, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
resource.setrlimit(resource.RLIMIT_NOFILE, (hard, hard))

hold = []
while True:
    s = socket.socket()
    if s.fileno() >= 1024:
        break
    hold.append(s)

select.select([s], [], [], 0)   # ValueError: filedescriptor out of range in select()
```

In a real deployment this manifests as the full traceback below, from any cache/throttle/session call:

```
File "redis/connection.py", line 3146, in get_connection
    connection.can_read()
File "redis/connection.py", line 1351, in can_read
    return self._parser.can_read(timeout)
File "redis/_parsers/hiredis.py", line 101, in can_read
    return _socket_can_read(self._sock, timeout)
File "redis/_parsers/hiredis.py", line 31, in _socket_can_read
    return bool(select.select([sock], [], [], timeout)[0])
ValueError: filedescriptor out of range in select()
```

### Expected behavior

Connection readiness checks should work regardless of the socket's fd number, as they already do for the pure-Python parser.

[...]

PR #4118, opened 2026-06-12
Avoid per-check fd allocation in hiredis `_socket_can_read()` — use `poll()` instead of a per-call selector

violuke:
Follow-up to #4115, which fixed the `ValueError: filedescriptor out of range in select()` in the hiredis parser by replacing `select.select()` with `selectors.DefaultSelector()`. This relates also to #4117

The selector approach has a residual failure mode: `DefaultSelector` resolves to `EpollSelector` on Linux and `KqueueSelector` on macOS, both of which **allocate a new file descriptor on every call**. `_socket_can_read()` constructs a fresh selector per readiness check, and `Connection.can_read()` runs on every connection acquisition — so every command now allocates (and frees) an fd just to check readiness.

This matters because fd pressure is exactly the condition that triggers this bug class: a socket only lands on fd >= 1024 in a process holding many descriptors. In a process at or near `RLIMIT_NOFILE`, the readiness check itself now fails with `OSError: [Errno 24] Too many open files` (raised from `epoll_create`/`kqueue` inside `selectors.py`), so a worker that runs out of fds loses the ability to talk to Redis entirely even though its connection socket is healthy.

This PR switches `_socket_can_read()` to `select.poll()`, which:

- has no `FD_SETSIZE` limit (the original bug stays fixed);
- allocates **no** file descriptor — the poll set lives in the syscall arguments, not in a kernel object — so readiness checks keep working under fd exhaustion;
- avoids constructing a selector object per check on the hot path.

Behaviour is preserved: `timeout` seconds is converted to poll's milliseconds, `timeout=0` remains a non-blocking check and `timeout=None` blocks forever; `poll()` always reports `POLLHUP`/`POLLERR`/`POLLNVAL` regardless of the registered mask, so a closed or errored socket still counts as "readable", matching the previous `select()`/selector semantics. On Windows, where `select.poll` is unavailable, the `selectors.DefaultSelector` path from #4115 is retained as the fallback (Windows is not subject to the Unix `FD_SETSIZE` value limit).

The async hiredis parser is unaffected — it reads via asyncio streams and performs no `select`/`poll` calls.

### Changes

- `redis/_parsers/hiredis.py`: `_socket_can_read()` prefers `select.poll()` when available; the selector-based check remains as the Windows fallback.
- `tests/test_connection.py`:
  - `test_hiredis_socket_can_read_uses_poll` (parametrized) — verifies the poll path and the seconds-to-milliseconds timeout conversion for `0`, `0.001`, and `None`.
  - `test_hiredis_socket_can_read_falls_back_to_default_selector` — the existing selector test from #4115, updated to patch `_HAS_POLL = False` so it exercises the fallback path.
  - `test_hiredis_socket_can_read_handles_high_file_descriptor` — kept from #4115; skip condition updated for the new branch order.
  - `test_hiredis_socket_can_read_under_fd_exhaustion` (new) — lowers `RLIMIT_NOFILE`, fills the fd table via `os.dup()` until `EMFILE`, then asserts readiness checks still work for both a readable and an empty socket pair. This test **fails with `OSError: [Errno 24]` against current `master`** and passes with this change.

### How tested

- All hiredis `can_read` unit tests pass (none require a running server).
- Verified the new fd-exhaustion regression test catches the issue: with `redis/_parsers/hiredis.py` reverted to `master`, it fails with `OSError: [Errno 24] Too many open files` raised from `selectors.py`; with this change it passes.
- `ruff check`, `ruff format --check`, and `vulture redis whitelist.py --min-confidence 80` are clean.

---

> [!NOTE]
> **Medium Risk**
> Touches connection hot-path I/O readiness for all sync hiredis users; behavior is intended to match prior semantics but poll vs selector edge cases matter under load and fd pressure.
> 
> **Overview**
> **Hiredis socket readiness** (`_socket_can_read`) now prefers **`select.poll()`** on platforms where it exists, instead of creating a **`selectors.DefaultSelector`** on every check. That keeps the fix for high fds (no `FD_SETSIZE` / `select.select` limit) while avoiding an extra kernel fd per readiness probe—important because `can_read()` runs on the hot path and fd exhaustion was causing **`EMFILE`** when epoll/kqueue selectors were allocated.
> 
> On **Windows** (no `poll`), behavior stays on the **DefaultSelector** fallback. Timeouts are still converted to poll milliseconds; **`timeout=0`** / **`None`** semantics are unchanged.
> 
> **Tests** add poll-path coverage (including timeout conversion), force the selector fallback via `_HAS_POLL`, and a **forked** regression that fills the fd table and asserts readiness checks still work. **`pytest-forked`** was added to dev requirements for that test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54b1c859b18aab2304932f8611188932e8caecd7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1,6 +1,7 @@
 import copy
 import os
 import platform
+import select
 import selectors
 import socket
 import ssl
@@ -114,14 +115,37 @@ def test_hiredis_can_read_checks_socket_readiness_without_reading():
 
 
 @pytest.mark.fixed_client
-def test_hiredis_socket_can_read_uses_default_selector():
+@pytest.mark.parametrize(
+    "timeout,expected_poll_timeout",
+    [(0, 0), (0.001, 1.0), (None, None)],
+)
+def test_hiredis_socket_can_read_uses_poll(timeout, expected_poll_timeout):
+    if not hasattr(select, "poll"):
+        pytest.skip("select.poll not available on this platform")
+
+    sock = Mock(spec=["fileno"])
+    poller = Mock()
+    poller.poll.return_value = [(7, select.POLLIN)]
+
+    with patch("redis._parsers.hiredis.select.poll", return_value=poller):
+        assert _socket_can_read(sock, timeout=timeout) is True
+
+    poller.register.assert_called_once_with(sock, select.POLLIN)
+    poller.poll.assert_called_once_with(expected_poll_timeout)
+
+
+@pytest.mark.fixed_client
+def test_hiredis_socket_can_read_falls_back_to_default_selector():
     sock = Mock(spec=["fileno"])
     selector = MagicMock()
     selector.select.return_value = [(sock, selectors.EVENT_READ)]
     selector.__enter__.return_value = selector
 
-    with patch(
-        "redis._parsers.hiredis.selectors.DefaultSelector", return_value=selector
+    with (
+        patch("redis._parsers.hiredis._HAS_POLL", False),
+        patch(
+            "redis._parsers.hiredis.selectors.DefaultSelector", return_value=selector
+        ),
     ):
         assert _socket_can_read(sock, timeout=0.001) is True
 
@@ -133,9 +157,10 @@ def test_hiredis_socket_can_read_uses_default_selector():
 def test_hiredis_socket_can_read_handles_high_file_descriptor():
     fcntl = pytest.importorskip("fcntl")
 
-    with selectors.DefaultSelector() as selector:
-        if isinstance(selector, selectors.SelectSelector):
-            pytest.skip("Default selector uses select.select")
+    if not hasattr(select, "poll"):
+        with selectors.DefaultSelector() as selector:
+            if isinstance(selector, selectors.SelectSelector):
+                pytest.skip("No poll and default selector uses select.select")
 
     read_fd, write_fd = os.pipe()
     high_read_fd = None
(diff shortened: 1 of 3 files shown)


petyaslavova:
Hi @violuke, thanks for the contribution!

I'll take a closer look on Monday. I had completely missed the fact that `DefaultSelector` actually uses a new file descriptor, so this definitely shouldn't stay as it is. This PR looks like a good first step toward a proper solution.

Thanks for pushing this forward!

tests/test_connection.py
```diff
@@ -114,14 +115,37 @@ def test_hiredis_can_read_checks_socket_readiness_without_reading():
 
 
 @pytest.mark.fixed_client
-def test_hiredis_socket_can_read_uses_default_selector():
+@pytest.mark.parametrize(
+    "timeout,expected_poll_timeout",
+    [(0, 0), (0.001, 1.0), (None, None)],
+)
+def test_hiredis_socket_can_read_uses_poll(timeout, expected_poll_timeout):
+    if not hasattr(select, "poll"):
```
petyaslavova: Can you please move this as decorator:
`@pytest.mark.skipif(not hasattr(select, "poll"), reason="...")`

tests/test_connection.py
```diff
@@ -153,6 +178,42 @@ def test_hiredis_socket_can_read_handles_high_file_descriptor():
...
 
 
+@pytest.mark.fixed_client
+def test_hiredis_socket_can_read_under_fd_exhaustion():
+    # Readiness checks run on every connection acquisition, and fd pressure is
+    # what pushes sockets onto high fds in the first place, so they must not
+    # allocate file descriptors themselves. A per-check epoll/kqueue selector
+    # raises OSError (EMFILE) here; the poll() implementation passes.
+    resource = pytest.importorskip("resource")
+    if not hasattr(select, "poll"):
```
petyaslavova: Please move the skip to be a decorator

tests/test_connection.py
```diff
@@ -114,14 +115,37 @@ def test_hiredis_can_read_checks_socket_readiness_without_reading():
...
 
-    with patch(
-        "redis._parsers.hiredis.selectors.DefaultSelector", return_value=selector
+    with (
+        patch("redis._parsers.hiredis._HAS_POLL", False),
+        patch(
+            "redis._parsers.hiredis.selectors.DefaultSelector", return_value=selector
+        ),
     ):
         assert _socket_can_read(sock, timeout=0.001) is True
```
petyaslavova, line 150: Can you please add validation that  selector.__exit__ was called — so if the `with` block is accidentally removed from the implementation, we would be able to catch this with the test.

tests/test_connection.py
```diff
@@ -133,9 +157,10 @@ def test_hiredis_socket_can_read_uses_default_selector():
 def test_hiredis_socket_can_read_handles_high_file_descriptor():
     fcntl = pytest.importorskip("fcntl")
 
-    with selectors.DefaultSelector() as selector:
-        if isinstance(selector, selectors.SelectSelector):
-            pytest.skip("Default selector uses select.select")
+    if not hasattr(select, "poll"):
+        with selectors.DefaultSelector() as selector:
+            if isinstance(selector, selectors.SelectSelector):
+                pytest.skip("No poll and default selector uses select.select")
```
petyaslavova: This check is redundant — it's always True when select.poll is absent. It can be moved above the method as decorator. 
If `poll` is not available - skip the test.

tests/test_connection.py
```diff
@@ -153,6 +178,42 @@ def test_hiredis_socket_can_read_handles_high_file_descriptor():
...
+    # raises OSError (EMFILE) here; the poll() implementation passes.
+    resource = pytest.importorskip("resource")
+    if not hasattr(select, "poll"):
+        pytest.skip("select.poll not available on this platform")
+
+    readable, writable = socket.socketpair()
+    empty_sock, peer = socket.socketpair()
+    writable.sendall(b"x")
+
+    soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
```
petyaslavova, line 202: `RLIMIT_NOFILE` is process-wide. If pytest runs background threads during this test, they'll hit EMFILE. 
It might be a good idea to run this test in a forked subprocess (pytest-forked).

petyaslavova (changes requested):

violuke:
Thanks for the review @petyaslavova. I've applied the changes you have suggested.

petyaslavova (approved):
LGTM.

PR #4122, opened 2026-06-15
fix(asyncio): prevent Pipeline.reset() connection leak on cancellation

### LINKED ISSUE #4121: `Pipeline.reset()` leaks a pooled connection if cancelled during the UNWATCH read
### Summary

If an `asyncio` `Pipeline` is in a watched transaction and `reset()` is cancelled while awaiting the `UNWATCH` reply, the pooled connection is **never released** back to the pool — it stays in `_in_use_connections` forever. Enough such cancellations exhaust a `BlockingConnectionPool`, after which every acquire fails with `ConnectionError("No connection available.")`.

### What happens

```python
# redis/asyncio/client.py  (Pipeline.reset)
async def reset(self):
    ...
    if self.watching and self.connection:
        try:
            await self.connection.send_command("UNWATCH")
            await self.connection.read_response()      # <-- cancellation lands here
        except ConnectionError:
            if self.connection:
                await self.connection.disconnect()
    ...
    if self.connection:
        await self.connection_pool.release(self.connection)   # <-- skipped on cancel
        self.connection = None
```

A `CancelledError` delivered at `await self.connection.read_response()` is **not** a `ConnectionError`, so the `except ConnectionError` doesn't catch it, and there is **no `finally`**. It propagates out of `reset()` before `connection_pool.release(...)` runs. `release()` is the only thing that removes a connection from `ConnectionPool._in_use_connections` (`disconnect()` only closes the socket and never touches the pool sets), so the slot leaks. `reset()` is reached from `async with pipe:` (`__aexit__`), the `finally` in `execute()`, and explicit calls — and none are shielded at HEAD.

### Reproduction

Self-contained — `pip install redis`, then `python repro.py`. It drives the **real** `pipe.watch()` → `ConnectionPool.get_connection()` path and the **real** `Pipeline.reset()`; only the socket is stubbed (`StubConnection`) so the cancellation can be delivered exactly at the UNWATCH `read_response()`. No live Redis server needed.

```python
import asyncio
import redis
from redis.asyncio.connection import AbstractConnection, ConnectionPool, BlockingConnectionPool
from redis.exceptions import ConnectionError as RedisConnectionError


class StubConnection(AbstractConnection):
    """Real AbstractConnection; I/O faked at the socket boundary. During reset()
    the UNWATCH read suspends on read_gate so a cancel can land there."""
    gate_reads = False
    read_gate = None
    read_started = None

    async def _connect(self): pass
    def _host_error(self): return "stub"
    async def connect(self): pass
    async def disconnect(self, nowait=False): pass
    async def can_read(self, timeout=0): return False
    async def can_read_destructive(self): return False
    async def send_command(self, *args, **kwargs): return None
    async def send_packed_command(self, command, check_health=True): return None

    async def read_response(self, *args, **kwargs):
        if type(self).gate_reads:
            type(self).read_started.set()
            await type(self).read_gate.wait()      # <-- cancellation lands here (UNWATCH)
        return b"OK"


async def cancelled_reset(pool, label):
    client = Redis(connection_pool=pool)
    pipe = client.pipeline(transaction=True)

    StubConnection.gate_reads = False
    await pipe.watch("somekey")                     # real WATCH: acquires a pooled connection
    conn = pipe.connection
    assert conn in pool._in_use_connections

    StubConnection.gate_reads = True
    StubConnection.read_gate = asyncio.Event()
    StubConnection.read_started = asyncio.Event()
    task = asyncio.create_task(pipe.reset())
    await StubConnection.read_started.wait()
    task.cancel()                                   # cancel while reset() awaits the UNWATCH read
    try:
        await task
        pass

    leaked = conn in pool._in_use_connections
    print(f"[{label}] cancelled reset(): in_use={len(pool._in_use_connections)} "
          f"available={len(pool._available_connections)} leaked={leaked}")
    return leaked


async def main():
    print("redis", redis.__version__)
    pool = ConnectionPool(connection_class=StubConnection, max_connections=10)
    assert await cancelled_reset(pool, "ConnectionPool"), "expected leak"

    bpool = BlockingConnectionPool(connection_class=StubConnection, max_connections=1, timeout=1)
    assert await cancelled_reset(bpool, "BlockingPool"), "expected leak"
    try:
        await bpool.get_connection()
        print("UNEXPECTED: acquired a connection")
    except RedisConnectionError as e:
        print(f"[BlockingPool] pool exhausted -> {type(e).__name__}: {e}")


asyncio.run(main())
```

Output (redis 8.0.0):

```
redis 8.0.0
[ConnectionPool] cancelled reset(): in_use=1 available=0 leaked=True
[BlockingPool] cancelled reset(): in_use=1 available=0 leaked=True
[BlockingPool] pool exhausted -> ConnectionError: No connection available.
```

**Control:** if the same stub instead raises `ConnectionError` at the read (rather than being cancelled), the connection is released cleanly and there is no leak — confirming the leak is specific to the cancellation path, not the stub.

### Suggested fix

Release the connection in a `finally`, or shield the UNWATCH read, so cancellation cannot strand the connection:

```python
try:
    if self.watching and self.connection:
        try:
            await self.connection.send_command("UNWATCH")
            await self.connection.read_response()
        except ConnectionError:
            if self.connection:
                await self.connection.disconnect()
    self.watching = False
    self.explicit_transaction = False
finally:
    if self.connection:
        await self.connection_pool.release(self.connection)
        self.connection = None
```

### Relationship to existing issues

Distinct from the CVE-2023-28858 / -28859 cancellation family (those cause a *buffered-response / cross-request data* corruption; their `asyncio.shield` mitigation was later removed, and `reset()`'s read is unshielded at HEAD). Also distinct from the recently-merged cluster-path leak fix #4111 (a *different file* and a write-side error path, not this cancellation-during-`reset()` path). I found no existing report of `reset()`/`UNWATCH` cancellation leaking a pooled connection's slot. (Note: the GitHub search API was intermittently rate-limited during the sweep, so treat "unreported" with mild caution.)

### Environment

- redis 8.0.0 (master `8210f32`), CPython 3.11+

1795771535y-cell:
## Description

Wrap the UNWATCH/reset logic in \Pipeline.reset()\ with try/finally so \connection_pool.release()\ always runs even when the UNWATCH \
ead_response()\ is cancelled.

Without this, a \CancelledError\ delivered at \
ead_response()\ during \
eset()\ would bypass the release, stranding the connection in \_in_use_connections\ forever and eventually exhausting the pool.

Closes #4121

## Changes

- \
edis/asyncio/client.py\: Wrapped \Pipeline.reset()\ cleanup attributes and UNWATCH logic in try/finally, moving the \connection_pool.release()\ into the finally block.

## Testing

Ran the reproduction script from #4121:

\\\
Before:  [ConnectionPool] in_use=1 available=0 leaked=True
         [BlockingPool] in_use=1 available=0 leaked=True
         [BlockingPool] pool exhausted
After:   [ConnectionPool] in_use=0 available=1 leaked=False
         [BlockingPool] in_use=0 available=1 leaked=False
         OK: Pool recovered, acquired connection successfully
\\\

## Notes

The sync \
edis/client.py\ Pipeline reset has the same pattern but is not affected by \CancelledError\, so the fix is scoped to the async path only.

---

> [!NOTE]
> **Medium Risk**
> Touches core async pipeline/pool lifecycle; behavior on cancellation changes but scope is a single method and aligns release with other cleanup patterns.
> 
> **Overview**
> Fixes a **connection pool leak** when async `Pipeline.reset()` is interrupted (e.g. `asyncio.CancelledError` during `read_response()` while unwinding a `WATCH`).
> 
> UNWATCH and state cleanup stay in a `try` block; **`connection_pool.release()`** and clearing `self.connection` move into **`finally`**, so the borrowed connection is always returned even if cleanup is cancelled mid-flight.
> 
> This matters on paths that call `reset()` (`execute()`’s `finally`, `__aexit__`, `aclose()`, and error handlers during `WATCH`), where a stranded connection could leave `_in_use_connections` stuck and eventually exhaust the pool. Change is **async-only**; sync `Pipeline.reset()` is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27957c5ec3ba93838be27e27fd309dcb4d09f3fe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -1680,26 +1680,28 @@ class Pipeline(Redis):  # lgtm [py/init-calls-subclass]
     async def reset(self):
         self.command_stack = []
         self.scripts = set()
-        # make sure to reset the connection state in the event that we were
-        # watching something
-        if self.watching and self.connection:
-            try:
-                # call this manually since our unwatch or
-                # immediate_execute_command methods can call reset()
-                await self.connection.send_command("UNWATCH")
-                await self.connection.read_response()
-            except ConnectionError:
-                # disconnect will also remove any previous WATCHes
-                if self.connection:
-                    await self.connection.disconnect()
-        # clean up the other instance attributes
-        self.watching = False
-        self.explicit_transaction = False
-        # we can safely return the connection to the pool here since we're
-        # sure we're no longer WATCHing anything
-        if self.connection:
-            await self.connection_pool.release(self.connection)
-            self.connection = None
+        try:
+            # make sure to reset the connection state in the event that we were
+            # watching something
+            if self.watching and self.connection:
+                try:
+                    # call this manually since our unwatch or
+                    # immediate_execute_command methods can call reset()
+                    await self.connection.send_command("UNWATCH")
+                    await self.connection.read_response()
+                except ConnectionError:
+                    # disconnect will also remove any previous WATCHes
+                    if self.connection:
+                        await self.connection.disconnect()
+            # clean up the other instance attributes
+            self.watching = False
+            self.explicit_transaction = False
+        finally:
+            # we can safely return the connection to the pool here since we're
+            # sure we're no longer WATCHing anything
+            if self.connection:
+                await self.connection_pool.release(self.connection)
+                self.connection = None
 
     async def aclose(self) -> None:
         """Alias for reset(), a standard method name for cleanup"""


petyaslavova:
Hi @1795771535y-cell , thank you for taking the time to investigate this issue and put together a fix — we really appreciate the effort!

The problem this PR addresses has already been fixed in #4123, which has since been merged. We're going to close this PR to avoid duplication.

Thanks again for contributing to redis-py!

PR #4124, opened 2026-06-17
perf: cache default response callbacks instead of rebuilding on every client init (standalone sync and async clients)

### LINKED ISSUE #3624: Serious performance issue with redis connection pooling due to CaseInsensitiveDict
I was doing performance tests in my app, and noticed that if I serve a large number of requests simultaneously, doing something like this:

```python
async with redis.asyncio.Redis(connection_pool=redis_blocking) as conn:
    await conn.set(...)
```
... the number of requests per second is significantly lower (around 45-50%) compared to the case where I take a connection only once.

Since every request (as the app provides an API) requires taking a connection from the pool, I cannot do it once and then use multiple times - it will be even slower due to serialization (and will actually defeat the purpose of the pool).

Profiling revealed that the slowdown is primarily caused by the use of `CaseInsensitiveDict` in the `Redis.__init__()` method: https://github.com/redis/redis-py/blob/a4df6b272d308e44b3e883bf91a39051eddb91eb/redis/client.py#L382-L387

Especially, `update()` is expensive, since there are many keys to update; the profiler shows that `__setitem__()` alone consumes approximately 50% of the time spent, most of which is attributed to the `str.upper()` method.

Replacing the dictionary with a simple `dict` reduced the time needed to acquire the client (not a connection) by a factor of 12, resulting in a performance boost of approximately 45% in the aforementioned use case, and its performance became nearly comparable to that of acquiring a connection only once.

Now the question - is there a reason to use `CaseInsensitiveDict`? I could not spot anything obvious in the code that requires case-insensitive keys, while the performance gain is significant.

Additionally, moving the initialization of `self.response_callbacks` to the pool (as they are constants anyway) could potentially lead to further performance improvements.

Thank you!

madhavcodez:
Fixes #3624.

`Redis.__init__` (and the async one) builds the whole response-callback table from scratch on every instantiation:

```python
self.response_callbacks = CaseInsensitiveDict(
    get_response_callbacks(
        user_protocol=connection_kwargs.get("protocol"),
        legacy_responses=connection_kwargs.get("legacy_responses", True),
    )
)
```

`get_response_callbacks()` assembles a ~165-key dict, and then `CaseInsensitiveDict.__init__` walks the whole thing again calling `.upper()` on every key, even though those keys are already uppercase. The result only depends on `(protocol, legacy_responses)`, so every client after the first is repeating identical work. In the common "check out a client per request from a shared pool" pattern (FastAPI and friends) this wrap is around 44us, which works out to roughly 80% of the cost of constructing a `Redis(connection_pool=...)`.

## What this changes

Build the canonical table once per `(protocol, legacy_responses)` with `functools.lru_cache`, and hand each client a cheap `CaseInsensitiveDict.copy()` that skips the redundant re-casing:

```python
def copy(self) -> "CaseInsensitiveDict":
    # Keys are already upper-cased; bypass __init__/__setitem__ so we don't
    # re-run str.upper on every entry when copying (see #3624).
    new = CaseInsensitiveDict.__new__(CaseInsensitiveDict)
    dict.update(new, self)
    return new


@lru_cache(maxsize=None)
def _get_default_response_callbacks(user_protocol, legacy_responses):
    return CaseInsensitiveDict(
        get_response_callbacks(
            user_protocol=user_protocol,
            legacy_responses=legacy_responses,
        )
    )
```

Each instance still gets its own mutable dict, so `set_response_callback`, module command registration, and pipelines (which share the client's dict by reference) all behave exactly as before. The cached canonical is never mutated.

## Numbers

Quick local benchmark, 165 callbacks, RESP3/legacy default:

| | before | after |
|---|---|---|
| callback table wrap | ~44us | ~0.7us |
| `Redis(connection_pool=shared)` ctor | ~55us | ~11us |

So roughly 4-5x off client construction in the shared-pool case, and the callback step itself drops by about 40x. The number of `(protocol, legacy_responses)` combinations is tiny and bounded, so the cache stays small.

Repro:

```python
import timeit, redis
pool = redis.ConnectionPool()
print(timeit.timeit(lambda: redis.Redis(connection_pool=pool), number=10000))
```

## Scope and compatibility

Public API and callback contents are unchanged; this is purely about not redoing the same work on every construction. Limited to the base sync and async `Redis` clients. `RedisCluster` assembles its callbacks through a different path and is left alone here, though the same idea could apply as a follow-up.

## Testing

Added `tests/test_response_callbacks_cache.py` covering the invariants that need to hold:

- contents identical to the old `CaseInsensitiveDict(get_response_callbacks(...))` construction, value by identity, across the full `protocol` x `legacy_responses` matrix
- case-insensitive lookups still resolve (`d["geohash"] is d["GEOHASH"]`), including callbacks added later via `set_response_callback`
- mutation isolation: mutating one client's callbacks never leaks into another client with the same combo, and never touches the cached table
- `copy()` returns an independent `CaseInsensitiveDict`
- sync and async stacks produce the same table

Ran the full standalone `protocol` x `legacy_responses` matrix against a dockerized Redis 8 locally, plus `ruff check`, `ruff format --check`, and `vulture`. Everything passes except `test_psync` and `test_ssl_connection_without_ssl`, which fail the same way on an unmodified checkout in my environment (replica/stunnel setup), so they are unrelated to this change.

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -36,7 +36,7 @@ from redis._defaults import (
     DEFAULT_SOCKET_READ_SIZE,
     DEFAULT_SOCKET_TIMEOUT,
 )
-from redis._parsers.helpers import bool_ok, get_response_callbacks
+from redis._parsers.helpers import bool_ok
 from redis.asyncio.connection import (
     Connection,
     ConnectionPool,
@@ -55,7 +55,7 @@ from redis.client import (
     EMPTY_RESPONSE,
     NEVER_DECODE,
     AbstractRedis,
-    CaseInsensitiveDict,
+    _get_default_response_callbacks,
 )
 from redis.commands import (
     AsyncCoreCommands,
@@ -434,12 +434,10 @@ class Redis(
         self.connection: Optional[Connection] = None
 
         connection_kwargs = self.connection_pool.connection_kwargs
-        self.response_callbacks = CaseInsensitiveDict(
-            get_response_callbacks(
-                user_protocol=connection_kwargs.get("protocol"),
-                legacy_responses=connection_kwargs.get("legacy_responses", True),
-            )
-        )
+        self.response_callbacks = _get_default_response_callbacks(
+            connection_kwargs.get("protocol"),
+            connection_kwargs.get("legacy_responses", True),
+        ).copy()
 
         # If using a single connection client, we need to lock creation-of and use-of
         # the client in order to avoid race conditions such as using asyncio.gather
--- a/redis/client.py
+++ b/redis/client.py
@@ -3,6 +3,7 @@ import logging
 import re
 import threading
 import time
+from functools import lru_cache
 from itertools import chain
 from typing import (
     TYPE_CHECKING,
@@ -138,6 +139,34 @@ class CaseInsensitiveDict(dict):
     def update(self, data):
         data = CaseInsensitiveDict(data)
         super().update(data)
+
+    def copy(self) -> "CaseInsensitiveDict":
+        # Keys are already upper-cased; bypass __init__/__setitem__ so we don't
+        # re-run ``str.upper`` on every entry when copying (see #3624).
+        new = CaseInsensitiveDict.__new__(CaseInsensitiveDict)
+        dict.update(new, self)
+        return new
+
+
+@lru_cache(maxsize=None)
+def _get_default_response_callbacks(
+    user_protocol: Optional[Union[int, str]],
+    legacy_responses: bool,
+) -> CaseInsensitiveDict:
+    """Build the canonical response-callback table for a given
+    ``(protocol, legacy_responses)`` combination exactly once.
+
+    The returned mapping is shared and MUST NOT be mutated; callers take a cheap
+    :meth:`CaseInsensitiveDict.copy` so each client owns an independent, mutable
+    mapping. This avoids rebuilding and re-upper-casing the ~165-entry callback
+    table on every client construction (see #3624).
+    """
+    return CaseInsensitiveDict(
+        get_response_callbacks(
+            user_protocol=user_protocol,
+            legacy_responses=legacy_responses,
+        )
+    )
 
 
 class AbstractRedis:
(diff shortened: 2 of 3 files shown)


petyaslavova:
Hi @madhavcodez, thank you for your contribution! I'll take a look at it shortly.

petyaslavova (changes requested):
Thanks for the well-scoped change and the clear benchmark information. 
Memoizing the canonical table per (protocol, legacy_responses) and handing each client an independent copy is a good idea. 
I traced the sharing, and the cached table is never mutated, sync/async parity holds, and the offline tests cover the invariants well.

Three things before we can merge:

1. Rebase onto the latest master and resolve the conflicts, then confirm the full protocol × legacy_responses matrix still passes.

2. Please make the copy guaranteed rather than caller-enforced. Right now every call site has to remember to add .copy(), and forgetting it would mutate the shared canonical and corrupt all clients. Split this into two helpers: a private, lru_cache'd builder that returns the shared canonical (never handed out directly), and a thin wrapper that returns a fresh per-client copy — and have both client stacks call the wrapper only:

   @lru_cache(maxsize=None)
   def _cached_default_response_callbacks(user_protocol, legacy_responses) -> CaseInsensitiveDict:
       ...  # shared canonical

   def default_response_callbacks(user_protocol, legacy_responses) -> CaseInsensitiveDict:
       return _cached_default_response_callbacks(user_protocol, legacy_responses).copy()

   Please add to both methods clear docstrings that state which one clients should use: the wrapper (default_response_callbacks) is the only one clients should call, since it returns an independent, mutable per-client table; the cached builder returns the shared canonical and MUST NOT be mutated or used directly. This keeps the same performance (one cache hit + one cheap copy) while making the "must copy" invariant impossible to get wrong. Please point the tests at the new names and keep a direct copy() test on CaseInsensitiveDict.

3. This is a real but partial improvement of #3624 — it covers the base sync and async Redis clients, while RedisCluster and the broader "initialize callbacks once at the pool level" idea from the issue remain. Please change "Fixes #3624" in the description to a non-closing reference (e.g. "Refs #3624") so the issue stays open for the remaining work after this merges.

Once the rebase is done, the wrapper is in with the docstrings above, and CI is green, this should be ready for another review.

eeshsaxena (commented):
The caching approach is safe as long as every consumer copies, and both standalone clients do. `_get_default_response_callbacks` is `lru_cache`d and returns a shared `CaseInsensitiveDict` that the sync and async `__init__` immediately `.copy()`, so each client owns an independent mapping and later `set_response_callback` mutations never reach the cached table.

The new `CaseInsensitiveDict.copy()` does `dict.update(new, self)` into `CaseInsensitiveDict.__new__(...)`, so it reproduces the subclass and its case handling rather than degrading to a plain `dict` the way the inherited `dict.copy()` would. The `(protocol, legacy_responses)` cache key is hashable and the space is tiny, so the unbounded `maxsize=None` cache stays small in practice. Nice win on the per-client construction cost.

PR #3611, opened 2025-04-21
Multi exec on cluster

robertosantamaria-scopely:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [ ] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?
- [ ] Was the change added to CHANGES file?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

_Please provide a description of the change here._

PR #4130, opened 2026-06-23
fix(cluster): populate ClusterPipeline.command_stack via execution strategy (#3703)

### LINKED ISSUE #3703: self.command_stack isn't set properly for ClusterPipeline

nazarli-shabnam:
### Description of change

Fixes #3703.

After the execution-strategy refactor in #3611, queued commands moved into `ClusterPipeline._execution_strategy.command_queue`. The public `command_stack` attribute on the **sync** `ClusterPipeline` was left initialized to `[]` and never populated again, and the **async** `ClusterPipeline` had no `command_stack` at all.

APM/tracing integrations (e.g. Datadog's `ddtrace`) introspect `command_stack` to build pipeline spans, so after upgrading, cluster pipeline traces silently lost all command information (commands showed up blank). The non-cluster `Pipeline` still populates `command_stack` correctly, so this was a cluster-only regression.

**Change:** expose `command_stack` as a read-only property that delegates to the active execution strategy's `command_queue`, on both the sync and async `ClusterPipeline`. This restores the legacy introspection behavior without altering execution. `len(pipeline)` already delegates to the same queue, so the two stay consistent.

Per the discussion in #3703, no runtime deprecation warning is emitted — tracing tools read this on every pipeline execute, so a per-access warning would be noise. It is documented as legacy in the docstring instead. Both `PipelineStrategy` and `TransactionStrategy` are covered.

This follows the approach proposed by @mathewlee11 and approved by the maintainers in the issue, applied to both the sync and async cluster clients for parity.

### Pull Request check-list

- [x] Do tests and lints pass with this change? (`ruff check`, `ruff format --check`, `vulture` all clean; new tests pass)
- [x] Is the new or changed code fully tested? (added sync + async regression tests in `tests/test_cluster.py` and `tests/test_asyncio/test_cluster.py`)
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? (docstrings on the restored property)
- [ ] Is there an example added to the examples folder (if applicable)? (n/a)

---

> [!NOTE]
> **Low Risk**
> Backward-compatible read-only API restoration with no change to pipeline execution; covered by new sync/async regression tests.
> 
> **Overview**
> **Restores cluster pipeline introspection for APM/tracing** after the execution-strategy refactor left `command_stack` empty (sync) or missing (async), which broke tools like Datadog `ddtrace` that read queued commands on execute.
> 
> On both **sync and async** `ClusterPipeline`, `command_stack` is now a **read-only property** that returns the active strategy’s `command_queue` (same source as `len(pipeline)`). The sync client no longer keeps a separate `command_stack = []` on the pipeline instance.
> 
> `ExecutionStrategy` gains an abstract **`command_queue`**; **`AbstractStrategy`** exposes the internal `_command_queue` through that property. Docstrings mark `command_stack` as legacy; no runtime deprecation warning.
> 
> **Tests** in `tests/test_cluster.py` and `tests/test_asyncio/test_cluster.py` assert queued commands appear in `command_stack` and that both pipeline and transaction strategies delegate correctly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0f88c5f3f3eb5165617b287794e175f18aed3f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


petyaslavova:
Hey @nazarli-shabnam, thank you for your contribution! I'll have a look at it shortly.

nazarli-shabnam:
> Hey @nazarli-shabnam, thank you for your contribution! I'll have a look at it shortly.

@petyaslavova  so, good to go?

eeshsaxena (commented):
Confirmed the regression, and it's a real silent break. On `master`, `ClusterPipeline.__init__` sets `self.command_stack = []` and then nothing ever writes to it again - after the execution-strategy refactor (#3611) queued commands land in `self._execution_strategy._command_queue`, so `pipeline.command_stack` stays `[]` forever. Anything that introspects it (ddtrace's redis integration reads `instance.command_stack` to record the pipelined commands) sees an empty list and records nothing. Replacing the dead attribute with a property that delegates to `self._execution_strategy.command_queue` is the right fix, and returning the live list (so `command_stack[i].args` works) matches what those consumers expect. Nice that both sync and async are covered with tests.

One compatibility angle worth deciding on: the property is **read-only**, whereas on the regular (non-cluster) `Pipeline`, `command_stack` is a plain writable list attribute, and `Pipeline.reset()` reassigns it with `self.command_stack = []`. So code that treats the two pipeline types uniformly and does `pipe.command_stack = []` (or `pipe.command_stack = something`) would now raise `AttributeError: can't set attribute` on a cluster pipeline. In-place mutation (`.clear()`, `.append()`) still works since you hand back the live list, and ddtrace only reads, so this is likely fine in practice - but since the execution strategy already exposes a `command_queue` setter, mirroring it with a `command_stack.setter` that assigns through to `self._execution_strategy.command_queue` would make the cluster pipeline behave identically to the non-cluster one for very little extra code, and remove a subtle behavioural difference between them.

Minor: the docstring says "Deprecated" - if the intent is that this stays as the compatibility surface for tracing integrations, it might read better as "compatibility shim" than "deprecated", since deprecating it would push those same integrations toward `len(pipeline)`, which doesn't give them the command args they actually need.

Deferring to the maintainers - the core fix is correct.

eeshsaxena (commented):
The compat shim reads correctly. `command_stack` is now a read-only property returning `self._execution_strategy.command_queue`, and all the internal queue manipulation (append, reset to `[]`, iteration during send and response parsing) goes through `self._command_queue` on the strategy, not through the pipeline's `command_stack`. So nothing rebinds the property and there is no `AttributeError` from the missing setter.

External introspection (e.g. ddtrace reading `pipeline.command_stack`) gets the live queue back, which is the regression from the execution-strategy refactor being fixed. Sync and async are symmetric. Looks good.

PR #4132, opened 2026-06-24
ci: report test duration metrics to Grafana Cloud (PAWS)

kiryazovi-redis:
## Summary

Adds CI test-performance telemetry to the integration workflow. After each
test run, JUnit XML results are exported as OpenTelemetry metrics (per-test
duration) to Grafana Cloud via the `redis-developer/cae-otel-ci-visibility`
action, enabling performance-regression tracking (PAWS). This mirrors the
setup already running in redis/node-redis.

## Changes

- `tasks.py`: write all JUnit XML into a dedicated `junit-results/` directory.
- `.github/actions/run-tests/action.yml`:
  - new optional `otlp-auth-token` input
  - create `junit-results/` before running tests
  - "Normalize JUnit XML for metrics" step — pytest does not set a `time`
    attribute on the root `<testsuites>` element, which the metrics action
    requires; this step derives it from the child `<testsuite>` times
  - "Self Report Metrics" step — exports the metrics to Grafana Cloud
  - include `junit-results/*-results.xml` in the uploaded artifacts
- `.github/workflows/integration.yaml`: forward the OTLP auth token input to the
  run-tests action for all test jobs.
- `.gitignore`: ignore `junit-results/`.

## Notes

- Metric emitted: `cae_v13_test_duration_seconds` (gauge, seconds), labeled by
  repository, branch, commit, CI run id, and test id.
- Reporting only runs for internal builds and is skipped automatically
  otherwise, so CI behavior is unchanged for external contributors and fork PRs.
- Reporting stays inactive until the required CI configuration is in place;
  until then these steps are skipped and existing CI is unaffected.

## Test plan

- [x] Validated the JUnit-XML normalization against the action's parser locally
      (real pytest output and empty/zero-test suites now parse cleanly).
- [x] Full integration CI on a fork reported successfully and the data appeared
      in Grafana.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI wiring, artifact paths, and optional metrics export with guards for forks and missing secrets; test commands themselves are unchanged aside from JUnit output location.
> 
> **Overview**
> Adds **CI test-performance telemetry** so integration runs can export per-test duration metrics to Grafana Cloud (PAWS), aligned with node-redis.
> 
> **JUnit output** is centralized under `junit-results/`: `tasks.py` ensures that directory exists and writes all `--junit-xml` paths there; the composite test action creates it before tests; `.gitignore` ignores it; uploaded artifacts use `junit-results/*-results.xml`.
> 
> The **run-tests** action gains an optional `otlp-auth-token` input. When a token is set and the run is internal (not a fork PR), two new steps run with `continue-on-error`: a Python step that sets root `<testsuites time="…">` from child suite times (for the metrics action), then `redis-developer/cae-otel-ci-visibility@v2` pushing OTLP metrics to Grafana. **integration.yaml** passes `secrets.SELF_CHECK_OTEL_AUTHORIZATION_TOKEN` into all matrix jobs that use run-tests.
> 
> If the secret is missing or the job is skipped by the guards, test execution and pass/fail behavior stay the same as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19f7607317177daab8316118e27e7af5762b350b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kiryazovi-redis:
I haven't added the secrets, but will once this is approved. Can show as testing the results from my fork.

.github/actions/run-tests/action.yml
```diff
@@ -310,12 +315,39 @@ runs:
         redis-cli -p 16379 CLUSTER NODES
       shell: bash
 
+    - name: Normalize JUnit XML for metrics
```
petyaslavova, line 318: These two steps(Normalize JUnit XML for metrics and Self Report Metrics) run before the artifact-upload and codecov steps and have no `continue-on-error`. In a composite action, a failed step skips later steps that lack an `if:` condition, so a transient Grafana/network error or a malformed JUnit file here would fail an otherwise-green job and skip the test-artifact and coverage uploads. Please add `continue-on-error: true` to both steps so non-critical telemetry can never gate build status or artifact upload.

.github/actions/run-tests/action.yml
```diff
@@ -310,12 +315,39 @@ runs:
...
+      if: always() && inputs.otlp-auth-token != '' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
```
petyaslavova: `always()` also fires when "Setup Test environment" fails before any tests run — in that case `junit-results/` is never created (the `mkdir` lives inside "Run tests"), so `cae-otel-ci-visibility@v2` is invoked against a missing/empty folder. Can we confirm the action no-ops cleanly in that case? If cancelled-run telemetry isn't wanted, consider `!cancelled() && ...` instead of `always() && ...`.

petyaslavova (changes requested):

.github/actions/run-tests/action.yml
kiryazovi-redis: Done in 093403de — added `continue-on-error: true` to both the "Normalize JUnit XML for metrics" and "Self Report Metrics" steps, so a transient Grafana/network error or a malformed JUnit file can no longer fail an otherwise-green job or skip the test-artifact and coverage uploads.

.github/actions/run-tests/action.yml
kiryazovi-redis: Good catch. Switched both steps from `always()` to `!cancelled()` in 093403de, so telemetry no longer runs on cancelled jobs.

For the "Setup Test environment failed before any tests ran" case, it's now harmless either way: the normalize step's `glob("junit-results/*.xml")` simply returns nothing (no error even if the folder is absent), and with `continue-on-error: true` on the `cae-otel-ci-visibility@v2` step, an empty/missing folder can't fail the job or skip the artifact/coverage uploads.

tasks.py
```diff
@@ -179,7 +179,7 @@ def multidb_integration_tests(c, uvloop=False, profile=False):
     run(
         f"pytest {profile_arg} {uvloop_arg} -m multidb_integration "
         f"tests/test_multidb tests/test_asyncio/test_multidb "
-        f"--junit-xml=multidb-integration-results.xml"
+        f"--junit-xml=junit-results/multidb-integration-results.xml"
```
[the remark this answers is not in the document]
kiryazovi-redis: Fixed in 093403de. Added an `_ensure_junit_results_dir()` helper (`os.makedirs("junit-results", exist_ok=True)`) and call it at the start of `fixed_client_tests`, `standalone_tests`, `cluster_tests`, and `multidb_integration_tests`, so local `invoke` runs create the directory before pytest writes the `--junit-xml` report.

.github/actions/run-tests/action.yml
```diff
@@ -310,12 +315,39 @@ runs:
...
+      env:
+        OTEL_EXPORTER_OTLP_PROTOCOL: "http/protobuf"
+
     - name: Upload test results and profiling data
       uses: actions/upload-artifact@v7
       with:
         name: pytest-results-redis_${{inputs.redis-version}}-python_${{inputs.python-version}}-parser_${{env.PARSER_BACKEND}}-el_${{inputs.event-loop}}-config_${{inputs.test-config}}
         path: |
-          *-results.xml
+          junit-results/*-results.xml
```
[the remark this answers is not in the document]
kiryazovi-redis: Fixed in 093403de. Both the "Normalize JUnit XML for metrics" and "Self Report Metrics" steps now have `continue-on-error: true`, so a failure in either can no longer skip the artifact-upload or codecov steps. The steps also use `!cancelled()` instead of `always()`.

petyaslavova (approved):
LGTM.

PR #4137, opened 2026-06-25
fix: detect stale connection via MSG_PEEK in hiredis (#4128)

### LINKED ISSUE #4128: RESP3 + hiredis does not recover from stale pooled connection
Greetings and thanks for maintaining the package!
We encountered a stale connection recovery issue after upgrading to redis-py 8.0.0, where RESP3 became the default.

### Description

With `redis[hiredis]` installed, RESP3 connections do not recover after the server closes a pooled connection. The next command raises:

```txt
```

The same stale-connection scenario succeeds with:

- protocol=2 + hiredis
- protocol=3 + pure Python RESP3 parser
- protocol=3 + hiredis + explicit retry

This affected us after upgrading to redis-py 8.0.0 because RESP3 became the default. We use long-lived pooled Redis connections for a task broker, and idle sockets can be closed by the server/proxy.

### Versions

- redis-py: `8.0.0`
- hiredis: `3.3.1`
- Redis server: `8.2.3`
- Python: `3.14.1`
- Also reproduced against redis-py `main` at 83c2541001fd898190b3617786d5734f8954f179

### How to reproduce

```python
import time

import redis
from redis.connection import _HiredisParser, _RESP3Parser

url = "redis://localhost/15"
admin = redis.Redis.from_url(url)

for parser in (_HiredisParser, _RESP3Parser):
   r = redis.Redis.from_url(
       url,
       protocol=3,
       parser_class=parser,
       health_check_interval=1,
   )

   r.ping()
   client_id = r.client_id()

   # Simulate server/proxy closing an idle pooled connection.
   admin.client_kill_filter(_id=client_id)
   time.sleep(1)

   try:
       print(parser.__name__, r.ping())
   except Exception as exc:
       print(parser.__name__, type(exc).__name__, exc)
```

I am happy to provide more details or try to submit a patch if this confirmed by the maintainers.

Cheers,
Rust

C1-BA-B1-F3:
Fixes #4128

## Problem

When / reports a stale (half-closed) socket as readable due to EOF/POLLHUP,  returned  without checking whether actual data was present. This caused the hiredis  to report the connection had data, but the subsequent  →  returned 0 bytes, raising  instead of allowing the pool to detect and recycle the stale connection.

## Fix

After / says the socket is readable, use a non-destructive   to distinguish real application data from a stale EOF on plain (non-SSL) sockets with . If the peek returns , raise  immediately — the same approach the pure-Python  parser uses.

This fixes the root cause at the  level, benefiting all callers rather than patching individual call sites.

## Testing

- All existing  tests pass (mock-based tests correctly skip MSG_PEEK since mock sockets lack /)
- New test  verifies that a socket with EOF (peer closed, no data) raises 
- All checks passed! passes on both changed files

---

> [!NOTE]
> **Medium Risk**
> Changes connection readiness logic on every hiredis non-blocking check for plain sockets; behavior is scoped to timeout=0 and non-SSL, but it affects pooled connection validation paths.
> 
> **Overview**
> Fixes **#4128** by tightening hiredis `_socket_can_read` so a non-blocking readiness check does not treat a half-closed socket as having data when only EOF is pending.
> 
> After `poll()` / `select()` reports readability, **`timeout=0`** on plain (non-SSL) sockets now uses a non-destructive **`MSG_PEEK`** `recv`: empty peek raises **`ConnectionError`** (server closed), matching the pure-Python socket parser behavior and letting the pool recycle stale connections instead of failing later on zero-byte reads.
> 
> Tests cover EOF when poll reports readable and when poll is mocked to miss EOF while peek still catches it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 165baed83fffbff82ad0a81c5393af7966747343. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/_parsers/hiredis.py
+++ b/redis/_parsers/hiredis.py
@@ -35,6 +35,7 @@ def _socket_can_read(sock, timeout: float) -> bool:
     # timeout=0 must be a non-blocking readiness check only; both branches
     # below are non-destructive and have no FD_SETSIZE limit (select.select
     # raises ValueError for fds >= 1024).
+    socket_readable = False
     if _HAS_POLL:
         # Prefer poll() over selectors.DefaultSelector: epoll/kqueue selectors
         # allocate a file descriptor per check and so fail with EMFILE under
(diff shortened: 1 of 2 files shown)


petyaslavova:
Hey @C1-BA-B1-F3, thank you for your contribution! I'll review it shortly.

petyaslavova:
Hi @C1-BA-B1-F3, thank you for digging into #4128 — you diagnosed the root cause correctly, and all three of your PRs land on the right idea. We're going to consolidate the fix in #4156, which uses a non-destructive poll() check to tell a pending RESP3 push apart from a peer-closed socket. Compared to reading from the socket in can_read() (#4134/#4142), that avoids giving can_read() a side effect for every caller (health checks, pub/sub), and unlike MSG_PEEK (#4137) it also covers SSL connections.

Since these overlap on the same path, I'm going to close this one in favor of #4156. I really appreciate the effort and the multiple approaches you explored!

PR #4142, opened 2026-06-25
fix: RESP3 + hiredis stale connection recovery (#4128)

### LINKED ISSUE #4128: RESP3 + hiredis does not recover from stale pooled connection

C1-BA-B1-F3:
## Problem

When using RESP3 with the hiredis parser, pooled connections closed by the server (stale/EOF) were not detected and recovered. The next command raised  instead of transparently reconnecting.

This affected redis-py 8.0.0+ users because RESP3 became the default protocol.

## Root Cause

 relied solely on / to check socket readability. These OS-level APIs report EOF (server closed connection) as readable, so  returned  for stale connections.

Combined with the connection pool's  bypass (default for RESP3), the stale connection detection in  was completely skipped:

## Fix

When / reports the socket as readable but the hiredis reader has no buffered data, perform an actual non-blocking  to distinguish between real data and EOF:
- Closed socket → 0 bytes → raises  (matching pure-Python parser behavior)
- Data available → feeds to reader → returns 

The same fix is applied to  where  now raises  instead of returning .

## Tests

- Updated  for new behavior
- Added  covering the bug scenario

Fixes #4128

---

> [!NOTE]
> **Medium Risk**
> Changes core connection readiness checks for all hiredis users (sync now may recv during `can_read`), which is intentional but behavior-sensitive; FT.INFO changes are defensive parsing only.
> 
> **Overview**
> **Hiredis stale connections (RESP3 / pooled clients):** When the socket looks readable but the hiredis reader has no buffered data, `can_read` now performs a non-blocking read instead of returning true on poll/select alone. **EOF (0-byte recv)** raises `ConnectionError`, aligning with the pure-Python parser so pools can reconnect; real data is fed into the reader and `can_read` reflects whether the reader has data. **Async hiredis:** `at_eof()` with no reader data now raises `ConnectionError` rather than treating EOF as readable.
> 
> **RediSearch FT.INFO:** `_parse_info`, `_parse_info_unified`, and `_parse_info_resp3_to_legacy` return `{}` for empty/None responses. `_parse_info_resp3_to_legacy` leaves **RESP2 flat lists** unchanged (e.g. cluster) instead of forcing them through the RESP3 dict path.
> 
> Tests cover stale-connection detection, updated socket-readiness behavior, and empty/list FT.INFO cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d7a43717a8e9244a96cbea265559d9a6721c50c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/_parsers/hiredis.py
+++ b/redis/_parsers/hiredis.py
@@ -119,7 +119,19 @@ class _HiredisParser(BaseParser, PushNotificationsParser):
 
         if self._reader.has_data():
             return True
-        return _socket_can_read(self._sock, timeout)
+        if _socket_can_read(self._sock, timeout):
+            # Socket appears readable. This could mean either data is
+            # available or the connection has been closed by the server
+            # (EOF), because poll()/select()/epoll() report EOF as
+            # "readable".  Do an actual non-blocking recv to distinguish
+            # the two cases — a closed socket yields 0 bytes and raises
+            # ConnectionError, matching the pure-Python parser behavior.
+            try:
+                self.read_from_socket(timeout=0, raise_on_timeout=False)
+            except ConnectionError:
+                raise
+            return self._reader.has_data()
+        return False
 
     def read_from_socket(self, timeout=SENTINEL, raise_on_timeout=True):
         sock = self._sock
@@ -257,9 +269,11 @@ class _AsyncHiredisParser(AsyncBaseParser, AsyncPushNotificationsParser):
         # connection state, not only whether application data can be read.
         if not self._connected:
             raise OSError("Buffer is closed.")
-        # EOF means the connection is closed and not safe to reuse.
-        if self._reader.has_data() or self._stream.at_eof():
+        if self._reader.has_data():
             return True
+        # EOF with no reader data means the server closed the connection.
+        if self._stream.at_eof():
+            raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
         # asyncio.StreamReader has no public non-destructive API for checking
         # buffered bytes. Preserve dirty-connection detection for hiredis; tests
         # with a real StreamReader guard this private buffer API in CI.
--- a/redis/commands/search/commands.py
+++ b/redis/commands/search/commands.py
@@ -256,6 +256,8 @@ class SearchCommands:
     # ---- RESP2 legacy parsers ----
 
     def _parse_info(self, res, **kwargs):
+        if not res:
+            return {}
         it = map(str_if_bytes, res)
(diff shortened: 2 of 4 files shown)


petyaslavova:
Please clean up the branch from unrelated changes.

petyaslavova:
Hi @C1-BA-B1-F3, thank you for digging into #4128 — you diagnosed the root cause correctly, and all three of your PRs land on the right idea. We're going to consolidate the fix in #4156, which uses a non-destructive poll() check to tell a pending RESP3 push apart from a peer-closed socket. Compared to reading from the socket in can_read() (#4134/#4142), that avoids giving can_read() a side effect for every caller (health checks, pub/sub), and unlike MSG_PEEK (#4137) it also covers SSL connections.

Since these overlap on the same path, I'm going to close this one in favor of #4156. I really appreciate the effort and the multiple approaches you explored!

PR #4144, opened 2026-06-26
fix: guard RESP parsers against RecursionError on deeply nested replies (#4116)

### LINKED ISSUE #4116: redis-py pure-Python parser hits RecursionError on deeply nested RESP replies
The pure-Python RESP parser path in `redis-py` uses recursive parsing for nested aggregate replies. A malicious Redis server, compromised Redis proxy, or an on-path actor that can inject RESP replies can send sufficiently deep nested aggregates and trigger `RecursionError`, causing client-side denial of service.

The pure-Python RESP parser recursively processes nested aggregate replies without a recursion-depth guard.
Relevant code paths include:

- `redis/_parsers/resp2.py`
- `redis/_parsers/resp3.py`

In the RESP2 parser, nested arrays recurse via `_read_response()` repeatedly until the Python recursion limit is exceeded.

### Proof of concept
Create the following file:
```
import argparse
import sys
from pathlib import Path
sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
from redis._parsers.resp2 import _RESP2Parser
from redis._parsers.socket import SocketBuffer

class FakeSock:
    def __init__(self, payload: bytes):
        self.payload = payload
    def recv(self, n: int) -> bytes:
        if not self.payload:
            return b""
        chunk = self.payload[:n]
        self.payload = self.payload[n:]
        return chunk
    def settimeout(self, _timeout: float) -> None:
        pass

parser = argparse.ArgumentParser()
parser.add_argument("--depth", type=int, default=sys.getrecursionlimit() + 200)
args = parser.parse_args()

payload = (b"*1\r\n" * args.depth) + b"+OK\r\n"
resp = _RESP2Parser(1)
resp._buffer = SocketBuffer(FakeSock(payload), socket_read_size=65536, socket_timeout=1)

print("MODEL=malicious_redis_server_or_proxy")
print(f"PYTHON_RECURSION_LIMIT={sys.getrecursionlimit()}")
print(f"NESTING_DEPTH={args.depth}")
try:
    resp._read_response(disable_decoding=True)
except RecursionError:
    print("RESULT=VULNERABLE_RECURSION_DOS")
else:
    print("RESULT=NO_RECURSION_ERROR_AT_THIS_DEPTH")
```
**Run**
```
python poc.py
python poc.py --depth 1200
```

C1-BA-B1-F3:
## Problem

The pure-Python RESP parser path uses recursive parsing for nested aggregate replies (arrays, sets, maps, push). A malicious Redis server, compromised proxy, or on-path actor can inject deeply nested RESP replies to trigger `RecursionError`, causing client-side denial of service.

**Reproduction** (from #4116):
```python
payload = (b"*1\r\n" * 5000) + b"+OK\r\n"
# Feed to _RESP2Parser → RecursionError
```

## Fix

Add `MAX_NESTING_DEPTH = 100` to all four pure-Python parser classes:
- `_RESP2Parser` / `_AsyncRESP2Parser` (resp2.py)
- `_RESP3Parser` / `_AsyncRESP3Parser` (resp3.py)

When nesting depth exceeds the limit, raise `InvalidResponse` instead of allowing unbounded recursion. The depth limit of 100 is generous for legitimate Redis workloads while preventing stack exhaustion.

All aggregate types are guarded: `*` (array), `~` (set), `%` (map), `>` (push).

## Tests

12 new tests covering:
- Normal nesting (10 levels) parses correctly
- At-limit nesting (100 levels) parses correctly
- Over-limit nesting (101+ levels) raises `InvalidResponse`
- Massively nested (5000 levels) raises `InvalidResponse`, not `RecursionError`
- All four parsers (sync RESP2, async RESP2, sync RESP3, async RESP3)
- RESP3-specific types: set (`~`), map (`%`)

Fixes #4116

---

> [!NOTE]
> **Medium Risk**
> Touches core wire-protocol parsing on every connection; legitimate deeply nested replies above 100 levels would now fail, but the limit is unlikely for normal Redis workloads.
> 
> **Overview**
> Adds a **100-level nesting cap** on pure-Python RESP2/RESP3 parsers (sync and async) so malicious or pathological aggregate replies no longer blow the stack with `RecursionError`.
> 
> `_read_response` now tracks `_depth` and raises **`InvalidResponse`** when depth exceeds `MAX_NESTING_DEPTH`. RESP2 guards nested arrays (`*`); RESP3 also guards sets (`~`), maps (`%`), and push (`>`), with depth passed through recursive child reads.
> 
> New tests in `test_resp_recursion.py` cover normal and at-limit nesting, over-limit and very deep (5000-level) payloads, and all four parser classes plus RESP3 set/map types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4915b52cb9ca1c431b5431fab6253465b0be619. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


petyaslavova:
I think this PR as well is poluted with additional changes meant to be added to other fixes. Please clean up.

C1-BA-B1-F3:
Thanks @petyaslavova! Branch cleaned up — now only contains the RESP RecursionError guard changes. The unrelated commits have been removed.

petyaslavova:
Thanks for cleaning up the branch. The approach is right and the tests exercise the real parser path via SocketBuffer, which is exactly what we want for a security regression. Before merging, please add a non-regression test for a burst of top-level RESP3 push frames so we prove the guard doesn't falsely trip on legitimate push traffic (you can borrow the scenario from #4200).

One thing for a follow-up (not required here): the push continuation in resp3.py recurses without incrementing depth, so a flood of consecutive push frames is a separate recursion vector outside this issue's scope. Once the push-burst test is in and CI (ruff + tests) is green, this should be ready for another review.

redis/_parsers/resp3.py
```diff
@@ -106,41 +109,69 @@ def _read_response(
...
                 resp_dict[key] = self._read_response(
                     disable_decoding=disable_decoding,
                     push_request=push_request,
                     timeout=timeout,
+                    _depth=_depth + 1,
                 )
             response = resp_dict
         # push response
         elif byte == b">":
+            if _depth >= self.MAX_NESTING_DEPTH:
```
Sanjays2402, line 165: the push branch guards the recursion into the elements, but the tail call right below it (the "not a push_request, go read the actual reply" one) doesn't forward _depth, so it restarts at 0. a server/proxy dribbling out-of-band push frames back-to-back gets one stack frame per frame with the counter reset every time, which is the same RecursionError this PR is closing. same in _AsyncRESP3Parser.

petyaslavova:
Hey @C1-BA-B1-F3, will you have some time to work on the requested changes?

eeshsaxena (commented):
Checked the depth guarding across both protocols and it looks complete. RESP3 guards arrays (`*`), sets (`~`), maps (`%`, on both the key and value recursion), and push (`>`), and RESP2 guards arrays. Every recursive branch increments `_depth` and is bounded before recursing, so there is no unguarded nesting path left.

With `MAX_NESTING_DEPTH = 100` and roughly a couple of stack frames per level, this trips `InvalidResponse` well before Python's default recursion limit, so a malicious deeply-nested reply raises a clean protocol error instead of `RecursionError`. The check sitting before the list comprehension means it caps at 100 levels of actual nesting, which seems reasonable.

PR #4145, opened 2026-06-26
fix: make ClusterPipeline.command_stack reflect execution strategy's command queue

### LINKED ISSUE #3703: self.command_stack isn't set properly for ClusterPipeline

C1-BA-B1-F3:
## Problem

Fixes #3703

After the execution strategies refactor (PR #3611), `ClusterPipeline.command_stack` was always an empty list `[]` because commands are now stored in `self._execution_strategy.command_queue` instead. This breaks tracing libraries (e.g. Datadog) and other monitoring tools that inspect `command_stack` to list queued commands.

## Solution

- Remove `self.command_stack = []` from `ClusterPipeline.__init__`
- Add `command_stack` as a property that delegates to `self._execution_strategy.command_queue`

This maintains backward compatibility while keeping the new execution strategy architecture intact.

## Tests

Added two tests:
- `test_pipeline_command_stack_property` - verifies command_stack reflects queued commands
- `test_pipeline_command_stack_empty` - verifies command_stack is empty when no commands queued

## Notes

The async cluster implementation (`redis.asyncio.cluster`) doesn't have this issue as it uses `__slots__` and doesn't expose `command_stack` at all.

cc @mathewlee11 who identified the root cause and proposed the fix approach.

---

> [!NOTE]
> **Medium Risk**
> Changes touch connection readiness, response parsing, and blocking-command timeouts across sync/async paths; incorrect behavior could cause premature disconnects, false timeouts, or break tracing that relied on empty `command_stack`.
> 
> **Overview**
> Beyond restoring **`ClusterPipeline.command_stack`** as a deprecated property over **`_execution_strategy.command_queue`** (sync and async cluster, with tests for Datadog-style introspection), this PR hardens protocol and connection behavior in several places.
> 
> **Hiredis `can_read`:** sync hiredis now does a non-blocking recv when poll/select reports readability so EOF is not mistaken for pending data; async hiredis raises **`ConnectionError`** on EOF when the reader buffer is empty instead of treating the connection as readable.
> 
> **RESP2/RESP3:** nested arrays, sets, maps, and pushes are capped at **`MAX_NESTING_DEPTH` (100)**; deeper replies raise **`InvalidResponse`** instead of **`RecursionError`** (#4116).
> 
> **Blocking list commands** (`BLPOP`, `BRPOP`, `BLMOVE`, etc.) pass **`_blocking_timeout`** through **`parse_response`** into **`read_response`** so socket reads can wait for the command timeout (sync, async, and cluster clients).
> 
> **RediSearch `FT.INFO` parsers** return `{}` for empty/None responses and pass through flat RESP2 lists from cluster when using legacy RESP3 conversion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61f74f4afa072327b20631edfb4a1205bb933cdd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/_parsers/hiredis.py
+++ b/redis/_parsers/hiredis.py
@@ -119,7 +119,19 @@ class _HiredisParser(BaseParser, PushNotificationsParser):
 
         if self._reader.has_data():
             return True
-        return _socket_can_read(self._sock, timeout)
+        if _socket_can_read(self._sock, timeout):
+            # Socket appears readable. This could mean either data is
+            # available or the connection has been closed by the server
+            # (EOF), because poll()/select()/epoll() report EOF as
+            # "readable".  Do an actual non-blocking recv to distinguish
+            # the two cases — a closed socket yields 0 bytes and raises
+            # ConnectionError, matching the pure-Python parser behavior.
+            try:
+                self.read_from_socket(timeout=0, raise_on_timeout=False)
+            except ConnectionError:
+                raise
+            return self._reader.has_data()
+        return False
 
     def read_from_socket(self, timeout=SENTINEL, raise_on_timeout=True):
         sock = self._sock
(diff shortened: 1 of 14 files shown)


petyaslavova:
This PR contains commits for several other fixes. Please don't introduce all your commits for several PRs (create the branch for each PR starting from the current master state) into each one of them. Also, please before adding a PR to fix a concrete issue check if there is already a fix waiting for a review. Closing this one as it duplicates the logic of #4130

PR #4146, opened 2026-06-26
fix: filter None nodes in _determine_nodes during topology changes (#3835)

### LINKED ISSUE #3835: AttributeError during Redis Cluster create-before-destroy process
# Summary

During a GCP Memorystore for Redis Cluster maintenance, nodes were replaced as part of a create-before-destroy process. During this transition, the client’s in-memory cluster topology became temporarily inconsistent with the actual cluster state, resulting in `_determine_nodes()` returning a `None` value for a target node. It resulted in `AttributeError: 'NoneType' object has no attribute 'name'` originating from [this line](https://github.com/redis/redis-py/blob/8f297c2c853b69a63fb1c0ede5a0c507636a5068/redis/asyncio/cluster.py#L1625).

We reproduced this issue by scaling down our Redis cluster.

## Environment
- redis-py version: 5.3.1
- Python version: 3.11
- Deployment environment: Google Cloud Memorystore for Redis Cluster
- Topology: multi-node Redis Cluster
- Client type: redis.asyncio.RedisCluster

## Full stack trace
```
builtins.AttributeError: 'NoneType' object has no attribute 'name'

  File "/usr/local/lib/python3.11/site-packages/ddtrace/_trace/utils_redis.py", line 96, in _instrument_redis_execute_async_cluster_pipeline
    yield span
  File "/usr/local/lib/python3.11/site-packages/ddtrace/contrib/internal/redis/asyncio_patch.py", line 36, in instrumented_async_execute_cluster_pipeline
    return await func(*args, **kwargs)
  File "/usr/local/lib/python3.11/site-packages/redis/asyncio/cluster.py", line 1488, in execute
  File "/usr/local/lib/python3.11/site-packages/redis/asyncio/cluster.py", line 1571, in _execute
AttributeError: 'NoneType' object has no attribute 'name'
```

# Reproduction steps
	1.	Start a Redis Cluster with multiple shards.
	2.	Connect using `redis.asyncio.RedisCluster`.
	3.	Continuously run pipelined commands.
	4.	Scale up/down cluster
	5.	Observe intermittent `AttributeError` exceptions from the client.

# Expected behavior
redis-py should handle temporary None nodes gracefully — either by:
	•	retrying the command after topology refresh, or
	•	raising a more specific error we can use to retry.

# Questions
- I see the `_determine_nodes()` method changed a lot in the newer versions. Could upgrading the version could fix that?
- Is it a known issue or something we should fix in this lib?

Thanks!

C1-BA-B1-F3:
## Problem

During cluster topology changes (scale-down, failover, node replacement), `_determine_nodes()` can return a list containing `None` values when in-memory slot/node caches become temporarily inconsistent.

The existing check `if not target_nodes` only catches empty lists — `[None]` is truthy, so it leaks through to `node.name` and causes:

```
AttributeError: 'NoneType' object has no attribute 'name'
```

This was reported in #3835, where GCP Memorystore cluster maintenance triggered the error during a create-before-destroy node replacement process.

## Fix

Add `None` filtering in `_determine_nodes()` after the policy callbacks return, for all three implementations:

- `redis/asyncio/cluster.py` — async `RedisCluster._determine_nodes()`
- `redis/cluster.py` — sync `RedisCluster._determine_nodes()` 
- `redis/cluster.py` — sync `ClusterPipeline._determine_nodes()`

The filter is placed before any downstream code accesses the node list, so callers (both `execute_command` and pipeline `_execute`) are protected without needing individual changes.

## Tests

Added 4 unit tests (2 sync + 2 async) that verify:

1. `test_determine_nodes_filters_none_nodes` — mixed None/valid nodes are filtered to only valid nodes
2. `test_determine_nodes_returns_empty_when_all_none` — all-None list results in empty list (which callers handle with `RedisClusterException`)

Fixes #3835

---

> [!NOTE]
> **Medium Risk**
> Touches core cluster command routing and retry behavior during topology changes; mis-handling could affect multi-node commands, though the change aligns with existing ClusterDownError retry semantics.
> 
> **Overview**
> Fixes **#3835** where transient `None` values in node lists during maintenance could cause `AttributeError` on `node.name`, or—if filtered—**silent partial execution** on multi-node commands.
> 
> **`_determine_nodes`** in async `RedisCluster`, sync `RedisCluster`, and sync `ClusterPipeline` now detects any `None` in the policy callback result, **triggers topology refresh** (`nodes_manager.initialize()` or `_initialize = True` on async), and raises **`ClusterDownError`** so `execute_command`’s existing **`ERRORS_ALLOW_RETRY`** path retries with a fresh slot map. Return value is normalized with **`nodes or []`** when the list is empty but valid.
> 
> **Tests** (sync + async) assert mixed or all-`None` lists raise `ClusterDownError` and that reinit is signaled (`initialize` called / `_initialize` set).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 025cce3732a8f0816b9e0a24403d3d8e18c238f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -824,16 +824,22 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
         else:
             nodes = policy_callback()
 
-        # Filter out None nodes that can occur during cluster topology
-        # changes (e.g. scale-down, failover) where in-memory slot/node
-        # caches become temporarily inconsistent.
-        if nodes:
-            nodes = [n for n in nodes if n is not None]
+        # None entries can appear during topology changes when the
+        # in-memory slot/node cache is temporarily inconsistent. Silently
+        # dropping them is unsafe for multi-node / broadcast commands
+        # (partial success with no error). Mark the cluster for reinit and
+        # raise a retryable error so execute_command retries cleanly.
+        if nodes and any(n is None for n in nodes):
+            self._initialize = True
+            raise ClusterDownError(
+                "Cluster node mapping returned None during a topology "
+                "change; reinitialized topology for retry"
+            )
 
         if command.lower() == "ft.aggregate":
             self._aggregate_nodes = nodes
 
-        return nodes
+        return nodes or []
 
     async def _determine_slot(self, command: str, *args: Any) -> int:
         if self.command_flags.get(command) == SLOT_ID:
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -824,6 +824,12 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, AsyncRedisClusterCommand
         else:
             nodes = policy_callback()
 
+        # Filter out None nodes that can occur during cluster topology
+        # changes (e.g. scale-down, failover) where in-memory slot/node
+        # caches become temporarily inconsistent.
+        if nodes:
+            nodes = [n for n in nodes if n is not None]
+
         if command.lower() == "ft.aggregate":
             self._aggregate_nodes = nodes
 
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -1347,16 +1347,22 @@ class RedisCluster(
         else:
             nodes = policy_callback()
 
-        # Filter out None nodes that can occur during cluster topology
-        # changes (e.g. scale-down, failover) where in-memory slot/node
-        # caches become temporarily inconsistent.
-        if nodes:
-            nodes = [n for n in nodes if n is not None]
+        # None entries can appear during topology changes when the
+        # in-memory slot/node cache is temporarily inconsistent. Silently
+        # dropping them is unsafe for multi-node / broadcast commands
+        # (partial success with no error). Reinitialize and raise a
+        # retryable error so execute_command retries on a fresh view.
+        if nodes and any(n is None for n in nodes):
+            self.nodes_manager.initialize()
+            raise ClusterDownError(
+                "Cluster node mapping returned None during a topology "
+                "change; reinitialized topology for retry"
+            )
 
         if args[0].lower() == "ft.aggregate":
             self._aggregate_nodes = nodes
 
-        return nodes
+        return nodes or []
 
     def _should_reinitialized(self):
         # To reinitialize the cluster on every MOVED error,
@@ -4364,16 +4370,19 @@ class PipelineStrategy(AbstractStrategy):
         else:
             nodes = policy_callback()
 
-        # Filter out None nodes that can occur during cluster topology
-        # changes (e.g. scale-down, failover) where in-memory slot/node
-        # caches become temporarily inconsistent.
-        if nodes:
-            nodes = [n for n in nodes if n is not None]
+        # See RedisCluster._determine_nodes: do not silently drop None
+        # targets (unsafe for multi-node / broadcast commands).
+        if nodes and any(n is None for n in nodes):
+            self._pipe.nodes_manager.initialize()
+            raise ClusterDownError(
+                "Cluster node mapping returned None during a topology "
+                "change; reinitialized topology for retry"
+            )
 
         if args[0].lower() == "ft.aggregate":
             self._aggregate_nodes = nodes
 
-        return nodes
+        return nodes or []
 
     def multi(self):
         raise RedisClusterException(
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -1347,6 +1347,12 @@ class RedisCluster(
         else:
             nodes = policy_callback()
 
+        # Filter out None nodes that can occur during cluster topology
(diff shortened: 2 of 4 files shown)


petyaslavova:
Please clean up the branch - now it contains tons of unrelated changes.

C1-BA-B1-F3:
Thanks for the review @petyaslavova! I've cleaned up the branch — force pushed with only the relevant commit for filtering None nodes. The unrelated commits from other PRs have been removed.

eeshsaxena (commented):
The crash this prevents is real. I traced the consumer in `redis/cluster.py`:

```python
target_nodes = self._determine_nodes(...)
if not target_nodes:
    raise RedisClusterException(f"No targets were found to execute {args} command on")
...
for node in target_nodes:
    res[node.name] = self._execute_command(node, *args, **kwargs)
```

so a `None` entry reaches `node.name` and blows up with a bare `AttributeError: 'NoneType' object has no attribute 'name'`, which tells the user nothing. And the all-`None` case is already handled sensibly: filtering down to `[]` trips the existing `if not target_nodes` guard and raises the clear `RedisClusterException`. So far so good.

My hesitation is the **partial** case, and I think it's worth thinking through before this lands.

For a broadcast request policy (all-nodes / all-primaries), if the determination yields `[nodeA, None, nodeC]`, this filter turns it into `[nodeA, nodeC]` and the command then runs on two shards and **reports success**. The caller has no way to know a shard was skipped. For something like `CONFIG SET`, `FLUSHALL`, or a keyless broadcast, silently applying to a subset of the cluster and returning OK is arguably a worse failure mode than the `AttributeError` - the exception at least tells you something is wrong, whereas this quietly diverges the cluster.

In other words, the `None` is a symptom of the slot/node cache being stale (which the comment says outright), and dropping it treats the symptom while letting the stale-topology command proceed.

Two directions that would keep the crash fixed without the silent-subset risk:

1. When a `None` is seen, refresh the topology (`nodes_manager.initialize()`) and re-determine, rather than proceeding. The caller already runs inside `for _ in range(execute_attempts)`, so raising a retryable error here would let that loop recover on a re-initialised topology - which is exactly what you'd want during the scale-down/failover window described.
2. Or fix it at the source: whatever slot -> node lookup can return `None` for a live slot is the actual bug, and fixing it there avoids duplicating this guard in three places at all.

Minor points either way:

- The same six-line block is copy-pasted in three locations (async cluster, and `cluster.py` twice). If the filter does stay, a small shared helper would keep them from drifting.
- `if nodes:` is redundant in front of a list comprehension - `[n for n in nodes if n is not None]` already yields `[]` for an empty list, and if `nodes` could ever be `None` (rather than a list) the guard would silently pass `None` through to the caller instead of the empty-list path.

Happy to be wrong about the broadcast case if the affected policies are all single-node in practice - I couldn't convince myself of that from the code. Deferring to the maintainers.

C1-BA-B1-F3:
@eeshsaxena Thanks for the careful review — you are right about the partial-broadcast case.

Silently filtering `[nodeA, None, nodeC]` down to `[nodeA, nodeC]` and reporting success is worse than the original `AttributeError` for multi-node / broadcast commands (`CONFIG SET`, `FLUSHALL`, etc.): the caller has no signal that a shard was skipped.

I updated the approach accordingly:

1. **Do not silently drop `None` targets.** If any entry is `None`, treat it as a stale topology signal.
2. **Reinitialize and raise a retryable error** (`ClusterDownError`, already in `ERRORS_ALLOW_RETRY`):
   - sync / pipeline: `nodes_manager.initialize()` then raise
   - async: set `_initialize = True` then raise (next `execute_command` loop iteration re-inits)
3. That lets the existing `execute_attempts` retry loop re-run policy callbacks on a fresh cluster view instead of proceeding with a subset.
4. Also dropped the redundant `if nodes:` guard in front of the list check, as you noted.

Tests updated to assert reinitialize + `ClusterDownError` for both partial-`None` and all-`None` cases (sync + async).

Happy to adjust further if maintainers prefer a dedicated exception type or a shared helper for the three call sites.

eeshsaxena (commented):
Followed the None-node path through the async retry loop and it is handled correctly. `_determine_nodes` sets `self._initialize = True` and raises `ClusterDownError`, which is caught by the broad `except Exception` in `execute_command`. The retry gate there is `type(e) in self.__class__.ERRORS_ALLOW_RETRY`, and `ClusterDownError` is in that tuple (defined on `AbstractRedisCluster`), so it retries rather than surfacing. On the next iteration `if self._initialize:` reinitializes the topology before `_determine_nodes` runs again, so the retry re-resolves against a fresh view instead of hitting the same stale `None`.

The sync side reinitializes eagerly inside `_determine_nodes` rather than via the flag, but both paths refresh before the retry so the observable behavior matches. `return nodes or []` for the genuinely-empty case is also fine, since `execute_command` raises a clear `RedisClusterException` when there are no targets. Looks correct.

Issue #2967, opened 2023-09-28
Regression in 5.0.1: AttributeError: 'NoneType' object has no attribute 'feed'


PoByBolek:
### Version
- redis-py: 5.0.1
- redis: 7.0.12

### Platform
- Python 3.10.11 on Windows 11

### Description
This is probably a duplicate of #2349, in which case I'd like to re-open that issue, but I got the same error today with redis-py 5.0.1 (worked fine in 5.0.0):

```
Traceback (most recent call last):
  File "C:\Users\PoByBolek\redis-bug\cache.py", line 137, in _get_raw_data
    return redis.hget(self._name, b'data'), None, None,
  File "C:\Users\PoByBolek\redis-bug\venv\lib\site-packages\redis\commands\core.py", line 4946, in hget
    return self.execute_command("HGET", name, key)
  File "C:\Users\PoByBolek\redis-bug\venv\lib\site-packages\redis\client.py", line 536, in execute_command
    return conn.retry.call_with_retry(
  File "C:\Users\PoByBolek\redis-bug\venv\lib\site-packages\redis\retry.py", line 46, in call_with_retry
    return do()
  File "C:\Users\PoByBolek\redis-bug\venv\lib\site-packages\redis\client.py", line 537, in <lambda>
    lambda: self._send_command_parse_response(
  File "C:\Users\PoByBolek\redis-bug\venv\lib\site-packages\redis\client.py", line 513, in _send_command_parse_response
    return self.parse_response(conn, command_name, **options)
  File "C:\Users\PoByBolek\redis-bug\venv\lib\site-packages\redis\client.py", line 553, in parse_response
    response = connection.read_response()
  File "C:\Users\PoByBolek\redis-bug\venv\lib\site-packages\redis\connection.py", line 500, in read_response
    response = self._parser.read_response(disable_decoding=disable_decoding)
  File "C:\Users\PoByBolek\redis-bug\venv\lib\site-packages\redis\_parsers\hiredis.py", line 124, in read_response
    self.read_from_socket()
  File "C:\Users\PoByBolek\redis-bug\venv\lib\site-packages\redis\_parsers\hiredis.py", line 87, in read_from_socket
    self._reader.feed(self._buffer, 0, bufflen)
AttributeError: 'NoneType' object has no attribute 'feed'
```

PoByBolek:
I've added some code that hopefully helps reproducing this issue in case anyone cares.

petyaslavova:
[...]

PR #4147, opened 2026-06-26
fix: check _reader is not None in _HiredisParser.read_from_socket()

C1-BA-B1-F3:
## Problem

Fixes #4003: When using redis-py with the hiredis parser and connection pooling in a multi-threaded environment,  can become  when the socket is closed by another thread, causing:

## Root Cause

The race condition occurs in :

1. Thread A enters  and is about to call 
2. Thread B calls  which sets 
3. Thread A executes  → AttributeError

## Fix

Added a guard at the top of  to check if  is None:

This is consistent with the existing checks in  (line 117) and  (line 175).

## Test

Added  to verify the fix.

---

> [!NOTE]
> **Low Risk**
> Narrow defensive check aligned with existing parser disconnect handling; behavior change only affects the erroneous race path.
> 
> **Overview**
> Fixes a **multi-threaded race** on pooled hiredis connections where one thread can call `on_disconnect()` (setting `_reader` to `None`) while another is inside `_HiredisParser.read_from_socket()`, which previously led to **`AttributeError`** on `_reader.feed()`.
> 
> **`read_from_socket`** now raises **`ConnectionError`** with the same “connection closed” message used elsewhere when `_reader` is missing, matching existing checks in **`can_read`** and **`read_response`**.
> 
> Adds **`test_hiredis_read_from_socket_raises_when_reader_is_none`** to lock in that behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 539975e025a62ada5ba13058adc7134ab1ea172c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/_parsers/hiredis.py
+++ b/redis/_parsers/hiredis.py
@@ -122,6 +122,8 @@ class _HiredisParser(BaseParser, PushNotificationsParser):
         return _socket_can_read(self._sock, timeout)
 
     def read_from_socket(self, timeout=SENTINEL, raise_on_timeout=True):
+        if not self._reader:
+            raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
         sock = self._sock
         custom_timeout = timeout is not SENTINEL
         try:
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -294,6 +294,17 @@ def test_hiredis_read_response_timeout_zero_maps_would_block_to_timeout():
         parser.read_response(timeout=0)
 
 
+def test_hiredis_read_from_socket_raises_when_reader_is_none():
+    """Regression: read_from_socket must raise ConnectionError if _reader is
+    None (e.g. on_disconnect was called by another thread).  Previously this
+    produced AttributeError: 'NoneType' object has no attribute 'feed'."""
+    parser = make_hiredis_parser()
+    parser._reader = None  # simulate on_disconnect clearing the reader
+
+    with pytest.raises(ConnectionError, match="Connection closed by server"):
+        parser.read_from_socket()
+
+
 def test_socket_buffer_timeout_zero_maps_would_block_to_timeout():
     sock = Mock()
     sock.recv.side_effect = BlockingIOError(


redis/_parsers/hiredis.py
```diff
@@ -122,6 +122,8 @@ def can_read(self, timeout: float = 0) -> bool:
         return _socket_can_read(self._sock, timeout)
 
     def read_from_socket(self, timeout=SENTINEL, raise_on_timeout=True):
+        if not self._reader:
```
petyaslavova, line 125: I'm not sure this will help at all. The problem is that the reader gets closed when the client gets closed - when the client gets closed - `self._sock `is gone as well, so this check here is redundant.

petyaslavova:
Thanks for the contribution and for linking the root issue.

Unfortunately this guard doesn't fix the reported problem. read_from_socket() has a single caller, read_response(), which already checks `if not self._reader` before the loop that calls it — so on the normal path _reader is never None here and the new check is redundant. The only way to reach read_from_socket() with _reader is None is the concurrent-close case: on_disconnect() clears _sock and _reader together, and because the added check isn't atomic with the following self._sock.recv_into(...) and self._reader.feed(...), a thread closing the connection right after the check still produces the same AttributeError (or the recv_into/WinError variant). So it doesn't reliably change anything.

The real cause is the same as in the closed #2967: the workers share one client and wrap each operation in `with redis:`. On exit that calls close() → connection_pool.disconnect(), which tears down all pooled connections — including the ones other threads are actively reading from. To be clear, sharing a client/pool across threads is supported; the pool hands each thread its own connection. What isn't supported is closing the client (or disconnecting the pool) from within the worker threads while other threads are still using it.

I'm going to close this for now. The fix on the user side is to not close the shared client per operation — share the client across the threads and close it once, after all workers are done. If you can reproduce #4003 without closing the client/pool from within the worker threads, please share that reproduction and we'll reopen and look at proper synchronization.

PR #4149, opened 2026-06-26
Fix ClusterPipeline.command_stack to reflect queued commands

### LINKED ISSUE #3703: self.command_stack isn't set properly for ClusterPipeline

C1-BA-B1-F3:
## Description

This PR fixes issue #3703 where `self.command_stack` was always `[]` in `ClusterPipeline` even when commands were queued.

### Root Cause

After the changes in PR #3611 which added execution strategies, commands in `ClusterPipeline` are stored in `self._execution_strategy.command_queue` instead of `self.command_stack`. However, `self.command_stack` was still initialized as an empty list in `__init__` and never updated.

This caused tracing libraries (e.g. Datadog) that inspect `command_stack` to report empty command lists, breaking observability.

### Fix

Convert `command_stack` from an instance attribute to a property that delegates to `self._execution_strategy.command_queue`.

### Changes

- **redis/cluster.py**: Remove `self.command_stack = []` from `__init__`, add `command_stack` property
- **tests/test_cluster_transaction.py**: Add tests verifying `command_stack` reflects queued commands in both transactional and non-transactional pipelines

### Testing

- Added unit tests verifying `command_stack` correctly reflects queued commands
- Verified the fix works for both transactional and non-transactional pipelines
- Existing tests continue to pass

Closes #3703

---

> [!NOTE]
> **Low Risk**
> Small backward-compatible API fix delegating to existing queue storage; covered by new cluster pipeline tests.
> 
> **Overview**
> Fixes **ClusterPipeline** observability where `command_stack` stayed empty after queued commands were moved to execution strategies.
> 
> **`command_stack`** is no longer an unused `[]` on init; it is a **property** that returns **`self._execution_strategy.command_queue`**, so tracing integrations that read `command_stack` see the same queued **`PipelineCommand`** entries as transactional and non-transactional pipelines.
> 
> Cluster transaction tests assert stack length and **`args`** for both pipeline modes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c5603b32e9da48fac1a750f853a883fd74704ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -3420,7 +3420,6 @@ class ClusterPipeline(RedisCluster):
         **kwargs,
     ):
         """ """
-        self.command_stack = []
         self.nodes_manager = nodes_manager
         self.commands_parser = commands_parser
         self.refresh_table_asap = False
@@ -3492,6 +3491,10 @@ class ClusterPipeline(RedisCluster):
             self._event_dispatcher = EventDispatcher()
         else:
             self._event_dispatcher = event_dispatcher
+
+    @property
+    def command_stack(self):
+        return self._execution_strategy.command_queue
 
     def __repr__(self):
         """ """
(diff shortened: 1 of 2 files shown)


C1-BA-B1-F3:
Closing as duplicate of #4145 which addresses the same issue (#3703).

PR #4153, opened 2026-06-26
Fix timeout writing to socket after health check (#3741)

### LINKED ISSUE #3741: Timeout writing to socket when using redis-py with configured timeouts and health check
I’m seeing intermittent TimeoutError: **Timeout writing to socket** when using redis-py despite setting custom timeout values and health checks. 

```
import redis
from redis.retry import Retry
from redis.backoff import ExponentialBackoff

client = redis.Redis(
    host="my-redis-host",
    port=6379,
    socket_connect_timeout=15,
    socket_timeout=5,
    socket_keepalive=True,
    retry=Retry(ExponentialBackoff(cap=10, base=1), 25),
    retry_on_error=[ConnectionError, TimeoutError, ConnectionResetError],
    health_check_interval=5,
)

client.set("foo", "bar")  # fails here
```

Observed behavior

Connection succeeds sometimes.

Some requests fail with:

 Timeout writing to socket

Expected behavior
With the above retry/backoff + timeouts, I expect the client to automatically handle transient network issues instead of failing with write socket timeouts.

C1-BA-B1-F3:
## Problem

Users report intermittent `TimeoutError: Timeout writing to socket` when using redis-py with both `socket_timeout` and `health_check_interval` configured. The error occurs even though the operation should complete within the configured timeout.

## Root Cause

Two related issues:

### 1. Socket timeout not re-asserted after health check

In `send_packed_command()`, the health check runs before the `sendall` loop. When the health check triggers a disconnect/reconnect cycle (via `_send_ping` failure → `_ping_failed` → `disconnect` → retry → reconnect), the new socket gets the correct timeout from `_connect()`. However, the code did not explicitly re-assert the socket timeout before `sendall()` after `check_health()` returned. If any concurrent operation (e.g., `can_read()` from another thread) or maintenance notification handler modified the socket timeout between the health check completing and `sendall` executing, the write would use a stale timeout value.

### 2. `update_parser_timeout` didn't handle `_RESP2Parser`

The `update_parser_timeout()` method updated `SocketBuffer.socket_timeout` for `_RESP3Parser` and `_HiredisParser`, but **not** for `_RESP2Parser`. When the connection timeout was changed (e.g., via maintenance notifications), `_RESP2Parser`'s `SocketBuffer` retained the old timeout snapshot. If `_read_from_socket` was later called with a custom timeout, its `finally` block would restore the stale snapshot instead of the current timeout.

## Fix

1. **Re-assert socket timeout in `send_packed_command`**: After `check_health()` returns, explicitly check and reset the socket timeout to `self.socket_timeout` before the `sendall` loop. This is a defensive measure that ensures correctness regardless of what happened during the health check.

2. **Fix `update_parser_timeout` for `_RESP2Parser`**: Add `_RESP2Parser` to the `isinstance` check alongside `_RESP3Parser` so that `SocketBuffer.socket_timeout` is properly updated.

## Tests

- 3 tests verifying socket timeout re-assertion after health check
- 3 tests verifying `update_parser_timeout` handles all parser types (RESP2, RESP3, Hiredis)

Closes #3741

---

> [!NOTE]
> **Medium Risk**
> Touches the hot path for every command write and parser timeout propagation; changes are defensive and well-tested but affect connection I/O behavior under health checks and RESP2 parsing.
> 
> **Overview**
> Fixes intermittent **`TimeoutError: Timeout writing to socket`** when **`socket_timeout`** and **`health_check_interval`** are both set.
> 
> **`send_packed_command`** now resets the socket to **`self.socket_timeout`** before **`sendall`** if **`gettimeout()`** differs—covering stale values left by health-check reconnects, concurrent **`can_read`**, or maintenance-notification timeout tweaks.
> 
> **`update_parser_timeout`** also updates **`SocketBuffer.socket_timeout`** for **`_RESP2Parser`**, not only RESP3/Hiredis, so parser read paths restore the current timeout instead of an old snapshot.
> 
> Regression tests cover write-path timeout re-assertion (with and without health check) and parser timeout updates for RESP2, RESP3, and Hiredis.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5bde3546e989fb0417e083f3dfb906029b6a17cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -750,7 +750,7 @@ class MaintNotificationsAbstractConnection:
     def update_parser_timeout(self, timeout: Optional[float] = None):
         parser = self._get_parser()
         if parser and parser._buffer:
-            if isinstance(parser, _RESP3Parser) and timeout:
+            if isinstance(parser, (_RESP2Parser, _RESP3Parser)) and timeout:
                 parser._buffer.socket_timeout = timeout
             elif isinstance(parser, _HiredisParser):
                 parser._socket_timeout = timeout
@@ -1310,8 +1310,18 @@ class AbstractConnection(MaintNotificationsAbstractConnection, ConnectionInterfa
         try:
             if isinstance(command, str):
                 command = [command]
+            # Re-assert the socket timeout before writing. The health check
+            # (above) may have triggered a disconnect/reconnect cycle, and
+            # concurrent operations (e.g. can_read from another thread) or
+            # maintenance-notification timeout changes can leave the socket
+            # with a stale timeout.  Guarding here ensures the write uses
+            # the configured socket_timeout regardless of what happened
+            # during the health check.
+            sock = self._sock
+            if sock.gettimeout() != self.socket_timeout:
+                sock.settimeout(self.socket_timeout)
             for item in command:
-                self._sock.sendall(item)
+                sock.sendall(item)
         except socket.timeout:
             self.disconnect()
             raise TimeoutError("Timeout writing to socket")
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1508,4 +1508,128 @@ class TestBlockingConnectionPoolGetConnectionCount:
         assert idle_count == 2
         assert used_count == 0
 
-        pool.disconnect()
+
+class TestSendPackedCommandTimeoutReassertion:
+    """Tests that send_packed_command re-asserts socket timeout after health check."""
+
+    def test_send_packed_command_reasserts_timeout_after_health_check(self):
+        """
+        Verify that send_packed_command resets the socket timeout to the
+        configured socket_timeout before calling sendall, even when the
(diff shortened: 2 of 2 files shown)


petyaslavova:
Hey @C1-BA-B1-F3, the described behaviors are not actually reproducible. The connections are guarded against access from another thread, so can_read on the same connection cannot be executed by another thread. This check is therefore redundant.

The only situation where another thread can update the timeout is during maintenance handling, and in this case the socket_timeout is applied.

For the other change, the only situation where we actually update the parser timeout is again related to maintenance notifications handling, which is possible only for RESP3. So we don't need to include the RESP2 option there.

Because of the reasons above, I'm closing this PR.

Issue #4026, opened 2026-04-06
Connection regression since 7.1.0


carl-mastrangelo:
I was trying to update from 7.1.0 to 7.4.0 and noticed that our unit tests stopped passing with the new version.   Somewhere in between the two versions, connecting to a Redis server started taking a very long time, to the point that the tests timed out.  Looking through what changed in a debugger, it appears that some connection retry and backoff was added. 

Our code attempts to connect to a redis server, and if unavailable, logs an exception and continues on.  The unit tests don't start a redis server and thus will never successfully connect.  However, since 7.4.0, it appears there is now a lot more retrying and sleeping, which is not desirable.  We also go out of our way to try and bound the amount of time spent connecting:

```
conn = Redis(port=redis_port(), socket_timeout=2.0, socket_connect_timeout=2.0)
```

I feel like this behavior is a regression, since the behavior was predictable and fast.

nadiatazi262-tech:
Thanks for the report — this does sound like a behavior change on our side.
The added retry/backoff could be causing the delay you’re seeing. We’ll
take a closer look to confirm and see if we can make this more predictable

petyaslavova:
Hey @carl-mastrangelo, thank you for reporting this and for the detailed description of your use case!

The behavior you're observing is actually an intentional change rather than a regression. Previously, the client lacked a retry mechanism for cases when handshake commands (like `AUTH`, `SELECT`, `CLIENT SETINFO`, etc.) failed during connection establishment. This was a bug — when retries were enabled, it was inconsistent that the connection handshake phase wouldn't participate in the retry flow. The fix ensures that the entire command execution pipeline, including the initial connection handshake, has retry capability when retries are configured.

That said, I completely understand this changes the experience for your use case, where you're intentionally connecting to an unavailable server and expecting a fast failure. The default retry configuration (`Retry(ExponentialWithJitterBackoff(base=1, cap=10), retries=3)`) will indeed add noticeable delay when the server is unreachable.

The solution for your tests is to explicitly disable retries by passing a retry object with 0 retries:

```python
from redis.retry import Retry
from redis.backoff import NoBackoff

conn = Redis(
    port=redis_port(),
    socket_timeout=2.0,
    socket_connect_timeout=2.0,
    retry=Retry(NoBackoff(), 0),
)

arlosnow1231-stack:
Thank you for the detailed explanation regarding the change in retry
behavior.

AirRunner:
[...]

PR #4166, opened 2026-07-04
fix(connection): fail fast on unrecoverable connect errors (#4026)

sean-kim05:
## Summary

Addresses the narrow, separable problem I raised in #4026: **definitively-unrecoverable connection errors are retried with full backoff.** The intentional handshake-phase retry behavior (clarified by @petyaslavova in that thread) is left completely untouched.

Since 7.1.0, `Connection.connect()` wraps the entire connect + handshake in the retry policy. During socket establishment every `OSError` is re-wrapped into a generic `ConnectionError`; because the policy's supported errors include `ConnectionError`, these are retried up to `DEFAULT_RETRY_COUNT` (10) times with exponential backoff — even for errors that can never succeed on retry, e.g. a Unix-domain-socket `ENOENT` (the socket path simply doesn't exist). The wrap also discarded the original `errno`, so nothing downstream could distinguish a permanent failure from a transient one.

Net effect: `Redis(unix_socket_path=..., socket_connect_timeout=2)` against a missing socket spends seconds sleeping/retrying, and the common "try UDS, fall back to TCP" pattern becomes very slow.

## Change

Uses the existing `is_retryable` hook in `Retry.call_with_retry` that `connect()` simply wasn't passing:

1. **Preserve the original error** — the `OSError` → `ConnectionError` wrap now uses `raise ... from e`, so the original `errno` survives for classification.
2. **Pass an `is_retryable` predicate** from `connect()` that returns `False` for a conservative set of unrecoverable socket errnos (currently just `ENOENT`). Those fail fast; transient errors such as `ECONNREFUSED` during server startup keep retrying exactly as before.

Applied to **both** the sync (`redis/connection.py`) and async (`redis/asyncio/connection.py`) stacks.

## Tests

- `test_connect_without_retry_on_unrecoverable_oserror` (sync + async): `ENOENT` → single attempt, original `OSError` preserved as `__cause__`.
- `test_connect_retries_on_transient_oserror` (sync + async): `ECONNREFUSED` → still retried (3 attempts with `retries=2`).

## Open questions (carried over from #4026)

Kept deliberately conservative — happy to adjust to your preference:

1. **Scope of "unrecoverable":** just `ENOENT` for now — should it also include `EACCES` / `EAFNOSUPPORT` / `EPROTONOSUPPORT`?
2. **`socket_connect_timeout` as a total-time cap:** left out of scope here (a broader change carrying a deadline across retries); could be a follow-up.
3. **Default vs opt-in:** implemented as the default; happy to gate behind a flag if preferred.

Refs #4026.

---

> [!NOTE]
> **Low Risk**
> Narrow connect-path retry behavior change with conservative errno list and tests; handshake retry logic is unchanged.
> 
> **Overview**
> **Connect failures that cannot succeed on retry (e.g. missing Unix socket, `ENOENT`) now fail immediately** instead of burning the default retry count and exponential backoff.
> 
> Sync and async `Connection.connect()` pass **`is_retryable=self._is_retryable_connect_error`** into the existing `Retry.call_with_retry` hook. The predicate treats errors as non-retryable when the raised `ConnectionError` chains an `OSError` whose `errno` is in `_UNRETRYABLE_CONNECT_ERRNOS` (currently **`ENOENT` only**). **`ECONNREFUSED` and other transient errors keep retrying** as before.
> 
> **`OSError` handling during connect** now uses `raise ConnectionError(...) from e` so the original errno is available on `__cause__` for that classification. Tests cover one-shot failure on `ENOENT` and full retries on `ECONNREFUSED`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3459a64fe9262b3b96a48fff4deda3945ec52b9e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -1,4 +1,5 @@
 import copy
+import errno
 import os
 import socket
 import sys
@@ -1000,6 +1001,25 @@ class AbstractConnection(MaintNotificationsAbstractConnection, ConnectionInterfa
     def _get_parser(self) -> Union[_HiredisParser, _RESP3Parser, _RESP2Parser]:
         return self._parser
 
+    # OS-level errno values for which reconnecting can never succeed, so
+    # there is no point spending the retry/backoff budget on them.
+    _UNRETRYABLE_CONNECT_ERRNOS = frozenset({errno.ENOENT})
+
+    @classmethod
+    def _is_retryable_connect_error(cls, error: BaseException) -> bool:
+        """Whether a failed connection attempt is worth retrying.
+
+        Returns ``False`` for definitively unrecoverable socket errors (e.g. a
+        Unix socket path that does not exist, ``ENOENT``) so that ``connect()``
+        fails fast instead of backing off for every retry. Transient errors
+        such as ``ECONNREFUSED`` during server startup stay retryable.
+        """
+        cause = error.__cause__
+        return not (
+            isinstance(cause, OSError)
+            and cause.errno in cls._UNRETRYABLE_CONNECT_ERRNOS
+        )
+
     def connect(self):
         "Connects to the Redis server if not already connected"
         # try once the socket connect with the handshake, retry the whole
@@ -1009,6 +1029,7 @@ class AbstractConnection(MaintNotificationsAbstractConnection, ConnectionInterfa
                 check_health=True, retry_socket_connect=False
             ),
             lambda error: self.disconnect(error),
+            is_retryable=self._is_retryable_connect_error,
         )
 
     def connect_check_health(
@@ -1044,16 +1065,18 @@ class AbstractConnection(MaintNotificationsAbstractConnection, ConnectionInterfa
             )
             raise e
         except OSError as e:
-            e = ConnectionError(self._error_message(e))
+            conn_error = ConnectionError(self._error_message(e))
             record_error_count(
                 server_address=getattr(self, "host", None),
                 server_port=getattr(self, "port", None),
                 network_peer_address=getattr(self, "host", None),
                 network_peer_port=getattr(self, "port", None),
-                error_type=e,
+                error_type=conn_error,
                 retry_attempts=actual_retry_attempts[0],
             )
-            raise e
+            # Preserve the original OSError (with its errno) as the cause so
+            # the retry policy can tell unrecoverable failures apart.
+            raise conn_error from e
 
         self._sock = sock
         try:
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1,5 +1,6 @@
 import asyncio
 import copy
+import errno
 import inspect
 import math
 import socket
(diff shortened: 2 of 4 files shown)


petyaslavova:
Hi @sean-kim05, thank you for your contribution! I'll review it shortly.

redis/connection.py
```diff
@@ -1000,6 +1001,25 @@ def set_parser(self, parser_class):
     def _get_parser(self) -> Union[_HiredisParser, _RESP3Parser, _RESP2Parser]:
         return self._parser
 
+    # OS-level errno values for which reconnecting can never succeed, so
+    # there is no point spending the retry/backoff budget on them.
+    _UNRETRYABLE_CONNECT_ERRNOS = frozenset({errno.ENOENT})
```
Sanjays2402, line 1006: `ENOENT` is transient for a Unix socket while the server is starting: the path does not exist until the server binds it. This makes an explicitly configured retry stop after one attempt in that common startup race; should this behavior be opt-in rather than overriding every caller’s `Retry` policy?

petyaslavova (changes requested):
Thanks for the careful work on this, and for keeping the handshake-retry behavior intact.

Before we merge, I'd like to apply some changes on the design toward making this configurable on the Retry object rather than hardcoding a non-retryable errno set as default behavior in the connection layer. 

Concretely:
1. Promote the original error up to the retry logic. Keep the `raise ConnectionError(...) from e` change so the root cause (and its errno) is available, and please make cause preservation consistent across the retryable wrap sites in connection.py (the socket read/write wraps currently build ConnectionError from `e.args` without `from e`, so `__cause__` is lost there). Any cause-inspecting checker needs the root error to actually be reachable.

2. Add a new optional property on the Retry object — a custom "is this error retryable?" checker. When set, it is applied on top of the existing supported-errors check and composed as an AND: an error is retried only if it is one of the supported error types AND the custom checker returns True. When it is not set, behavior is unchanged.

3. Make it opt-in. We don't want to change the default retry behavior; the ENOENT/unrecoverable-errno logic should be something a user configures via this checker (we can ship it as a reusable helper), not an always-on default. This also means the classification then applies uniformly to every retry site that uses the Retry object, including the lazy reconnect path, instead of only the explicit connect() call.

Please mirror this in both the sync and async Retry/connection stacks, and add regression tests that fail on base and pass on the branch — including a real (non-mock) missing-socket path. One scoping note: this addresses the ENOENT case, not the original #4026 report (an unavailable TCP server / socket_connect_timeout as a total cap), so let's keep #4026 open and retitle this PR accordingly. Once the checker lives on the Retry object and composes as described, this should be ready for another review.

eeshsaxena (commented):
The classification logic looks safe. `_is_retryable_connect_error` inspects `error.__cause__` and only returns `False` when it is an `OSError` whose errno is in `_UNRETRYABLE_CONNECT_ERRNOS` (currently just `ENOENT`). Since the connect wrapper raises `ConnectionError(...) from e`, `__cause__` is the original `OSError`, so the errno is actually reachable at the point this predicate runs.

Transient failures like `ECONNREFUSED` during server startup keep their existing retry behavior, and only a genuinely missing unix-socket path fails fast, which matches the intent. If `__cause__` is `None` (unchained error) the `isinstance` check is `False` and it defaults to retryable, which is the safe default. Keeping the unretryable set to just `ENOENT` is appropriately conservative. Looks good.

PR #4167, opened 2026-07-04
Add context manager support to ConnectionPool

### LINKED ISSUE #3669: Redis.from_pool(...) is not thread safe
We are using redis as a cache in a FastAPI app. To not make new connections to redis all the time we are using a Connection pool and initiate it startup. For each request we then initiate a Redis client from the pool, like so.

```python
def redis_server(request: fapi.Request) -> redis.Redis:
    # https://www.starlette.io/lifespan/#lifespan-state
    # This is of type starlette.datastructures.state
    pool = typing.cast("redis.ConnectionPool", request.state.redis_pool)
    return redis.Redis.from_pool(connection_pool=pool)  # type: ignore[no-any-return]
```

This causes all connections to close when we end the request and drop the reference to the client object. Now FastAPI runs "sync" requests in a thread pool. So when we have two simultaneous requests, both create a connection, but if just one finishes both connections are closed. 

We noticed this with fun exceptions like this. 
```
 "exc_type": "ValueError",
      "exc_value": "I/O operation on closed file.",
```
Where the buffer of a connection was was randomly closed in the middle of a healthcheck. 

Now I know that `from_pool` documents it *consumes* the connection pool. But that concept is not really possible in python.

sean-kim05:
### Description of change

Adds context manager support to `ConnectionPool` / `BlockingConnectionPool` (and the async equivalents in `redis.asyncio`), so a pool's lifecycle can be managed explicitly and safely:

```python
with ConnectionPool.from_url(url) as pool:
    r = Redis(connection_pool=pool)
```

```python
async with ConnectionPool.from_url(url) as pool:
    r = Redis(connection_pool=pool)
```

**Motivation — issue #3669 (`Redis.from_pool(...)` is not thread safe).**

`Redis.from_pool()` sets `auto_close_connection_pool = True`, so when the client is closed or garbage-collected it calls `connection_pool.disconnect()`, which defaults to `inuse_connections=True` and disconnects **every** connection in the pool — including connections currently checked out by *other* clients that share the same pool object. In a threaded server (e.g. FastAPI running sync routes in a threadpool) where a client is created per request from a shared pool, one request's client being GC'd tears down connections other in-flight requests are still using, producing errors like `ValueError: I/O operation on closed file` mid-healthcheck. The async client hits the same problem via `aclose()`.

There isn't currently a first-class, safe idiom for sharing one pool across multiple clients. This PR adds the context-manager protocol the issue author suggested, and clarifies the `from_pool()` docstrings (sync + async) to document that the returned client takes exclusive ownership of the pool and that the pool must not be shared — pointing users at the context manager instead.

This change is purely additive: no existing behavior changes.

### Pull Request check-list

- [x] Do tests and lints pass with this change? (`ruff check`, `ruff format --check`, and the new tests pass locally)
- [ ] Do the CI tests pass with this change?
- [x] Is the new or changed code fully tested? (new `test_pool_context_manager` in sync and async pool test suites)
- [x] Is a documentation update included? (expanded `from_pool` docstrings)
- [ ] Is there an example added to the examples folder (if applicable)? (n/a — docstring example included)

Closes #3669

---

> [!NOTE]
> **Low Risk**
> Additive API and documentation only; existing pool/client behavior is unchanged aside from optional context-manager cleanup.
> 
> **Overview**
> **ConnectionPool lifecycle:** Sync and async `ConnectionPool` (including blocking variants via inheritance) now support `with` / `async with`, exiting by calling `close()` / `aclose()` and disconnecting pooled connections—including those still checked out in the new tests.
> 
> **Safer shared-pool guidance:** `Redis.from_pool` docstrings (sync and asyncio) now state that the client **owns** the pool and must not be shared across multiple clients, and point to `Redis(connection_pool=pool)` plus explicit pool management—documented with a `ConnectionPool.from_url` context-manager example—for sharing one pool.
> 
> **Tests:** New `test_pool_context_manager` cases in sync and async pool test suites verify disconnect-on-exit behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b96e110837424b7e954f82bbeb7849f3389ba08d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


petyaslavova:
Hi @sean-kim05, thank you for your contribution! I'll provide a review shortly.

petyaslavova (changes requested):
Thanks for the contribution — the context-manager support and the `from_pool` docstring clarification are both welcome, and the sync/async parity is done correctly. Before merging, please address these points:

1. **`BlockingConnectionPool` coverage**: it inherits the new context-manager methods in both stacks, but there's no test for it. Please add a `test_pool_context_manager` to `TestBlockingConnectionPool` in `tests/test_connection_pool.py` and the async equivalent in `tests/test_asyncio/test_connection_pool.py`.
2. **Return type of `__enter__` / `__aenter__`**: they're annotated as `"ConnectionPool"`, so `with BlockingConnectionPool(...) as pool:` types `pool` as the base class. Please use a Self-style annotation via a `TypeVar` bound to `ConnectionPool` (the async client does this with `_RedisT` in `redis/asyncio/client.py`), keeping sync and async annotations identical.
3. **Docstring accuracy**: in the new `from_pool` docstrings, please also state explicitly that the plain constructor, `Redis(connection_pool=pool)`, does *not* take ownership of the pool — that's what makes the recommended example safe, and it's worth spelling out for readers.

With those covered, this should be ready for another review.

sean-kim05:
Thanks for the review! Addressed all three:

1. Added `test_pool_context_manager` to `TestBlockingConnectionPool` in both the sync (`tests/test_connection_pool.py`) and async (`tests/test_asyncio/test_connection_pool.py`) suites.
2. `__enter__`/`__aenter__` now use the existing `_CP` `TypeVar` (bound to `ConnectionPool`), so subclasses type as themselves rather than the base class — identical in sync and async.
3. The `from_pool` docstrings now state explicitly that the plain `Redis(connection_pool=pool)` constructor does not take ownership of the pool, which is what makes the shared-pool context-manager example safe.

`ruff check` / `ruff format --check` are clean and the pool tests pass locally. Ready for another look.

petyaslavova (approved):
LGTM.

PR #4168, opened 2026-07-06
test: add unit tests for cluster key_slot hash-tag logic

eeshsaxena:
## Description

`redis.crc.key_slot` implements the client-side Redis Cluster slot
calculation (CRC16/XMODEM modulo `REDIS_CLUSTER_HASH_SLOTS`) together with
the hash-tag rule: when a key contains a `{...}` with at least one character
inside, only that substring is hashed so related keys can share a slot.

That hash-tag logic is currently only exercised indirectly by the cluster
suites (mostly against a live server). This adds `tests/test_crc.py`, a small
server-free unit test that pins the behavior down directly:

- known CRC16/XMODEM check value (`"123456789"` -> `0x31C3`) and slot range
- plain keys hash the whole key
- keys sharing a hash tag map to the same slot
- only the substring between the first `{` and the first following `}` is used
- `"{{bar}}zap"` hashes `"{bar"` (nested opening brace is part of the tag)
- empty `{}` and unclosed `{` fall back to hashing the whole key
- the custom `bucket` argument is honored

No source changes - documentation-by-test of the existing contract.

## Verification

`pytest tests/test_crc.py` - 9 passed. `ruff check` / `ruff format --check`
clean.

---

> [!NOTE]
> **Low Risk**
> Test-only addition; no runtime or library behavior changes.
> 
> **Overview**
> Adds **`tests/test_crc.py`**, a server-free suite that documents and locks in **`redis.crc.key_slot`** behavior for Redis Cluster slot assignment.
> 
> The tests cover CRC16/XMODEM against a known vector (`"123456789"`), slot bounds, whole-key hashing, hash-tag co-location, first-tag-only parsing (including `{{bar}}zap` → `"{bar"`), fallback when `{}` or an unclosed `{` is present, and the optional **`bucket`** modulo argument. **No production code changes.**
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65bd2ce0ed45874acbd18418ba0e65d71bf65567. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


petyaslavova:
Hey @eeshsaxena, thank you for your contribution! I'll review it shortly.

tests/test_crc.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_unclosed_brace_falls_back_to_whole_key(self):
+        assert key_slot(b"{foo") == crc_hqx(b"{foo", 0) % REDIS_CLUSTER_HASH_SLOTS
+
+    def test_custom_bucket_is_honored(self):
+        assert key_slot(b"foo", 100) == crc_hqx(b"foo", 0) % 100
+        assert 0 <= key_slot(b"123456789", 100) < 100
+
+
+if __name__ == "__main__":
```
petyaslavova: Can you please remove this block.

petyaslavova (changes requested):
Overall, the tests are looking good. Can you please just add the `fixed_client` annotation to the whole test class - this way they will run only with the fixed client action - they don't depend on different protocol or parser configs so we don't need them executed as part of all different test configs.

eeshsaxena:
Thanks for the review @petyaslavova! I've added the `@pytest.mark.fixed_client` marker to the whole `TestKeySlot` class so these run only under the fixed-client config - they're pure client-side CRC/hash-tag checks with no protocol or parser dependency, so there's no need to run them across the other configurations. Pushed in 9d44ed2.

petyaslavova (changes requested):
Hey @eeshsaxena, can you please also remove the `if __name__ == "__main__"` block in the end of the file?

eeshsaxena:
Done - removed the `if __name__ == "__main__"` block at the end of the file. Ruff, black, and the 9 tests are all green locally.

petyaslavova (approved):

PR #4174, opened 2026-07-08
feat: add LMOVEM and BLMOVEM commands for moving multiple list elements

petyaslavova:
## Change summary

This pull request adds client support for the new `LMOVEM` and `BLMOVEM` Redis
commands, which atomically move multiple elements from one end of a source list
to one end of a destination list. They generalize `LMOVE`/`BLMOVE` (which move a
single element) by accepting a `count` and returning the moved elements as an
array.

The commands are implemented once in `BasicKeyCommands` in
`redis/commands/core.py`, which is shared by both the sync `CoreCommands` and
async `AsyncCoreCommands` mixins, so a single addition covers both stacks. Each
command exposes `src`/`dest` end selectors (`LEFT`/`RIGHT`, defaulting to
`LEFT`→`RIGHT`) and three optional modifiers:

- `count` — how many elements to move; when omitted a single element is moved.
- `mode` — `COUNT` (default, move up to `count`) or `EXACTLY` (move exactly
  `count` or nothing, erroring/blocking otherwise).
- `ordering` — `OBO` (push one-by-one, reversing block order / stack semantics)
  or `BULK` (preserve original relative order / queue semantics).

`BLMOVEM` additionally takes a `timeout` (seconds; `0` blocks indefinitely) and
returns `None` on timeout. Argument assembly is conditional — `mode`/`count` are
only appended when `count` is provided, and `ordering` only when set — so the
wire format matches the server's expectations. The change is purely additive:
no existing signatures, defaults, return types, or command semantics change, so
there is no backward-compatibility risk. Type overloads follow the project's
sync/async convention, keeping hints identical across both APIs.

## Test coverage

New tests `test_lmovem` and `test_blmovem` are added to both
`tests/test_commands.py` and `tests/test_asyncio/test_commands.py`, mirrored
between the sync and async suites. They exercise the single-element (no-count)
array reply, `COUNT` with `OBO` reversal and `BULK` order preservation,
up-to-count when fewer elements are available, the empty-source `None` result,
and `EXACTLY` behavior (raising `ResponseError` with too few elements while
leaving the source untouched, then succeeding once enough elements are present).
A few cluster tests are also aded.

---

> [!NOTE]
> **Low Risk**
> Additive command wrappers and tests only; no changes to existing APIs or runtime behavior beyond new methods.
> 
> **Overview**
> Adds **`lmovem`** and **`blmovem`** on the shared list command mixin in `redis/commands/core.py`, wiring **`LMOVEM`** / **`BLMOVEM`** with the same sync/async overload pattern as **`lmove`** / **`blmove`**. Callers can move one or many elements between list ends, with optional **`COUNT`** vs **`EXACTLY`**, **`OBO`** vs **`BULK`** ordering when **`count`** is set, and client-side **`DataError`** checks so **`ordering`** and **`count`** stay paired.
> 
> **`blmovem`** takes a blocking timeout and returns **`None`** on timeout. Integration CI points Redis **8.10** at a newer custom image so jobs can run against a server that exposes these commands.
> 
> New standalone tests (sync and asyncio) cover single-element replies, ordering modes, partial counts, empty source, **`EXACTLY`** semantics, and validation errors. Cluster tests assert same-slot success and cross-slot **`RedisClusterException`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a4fdd2514518f58f094fd65cb04c5c843323063. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/commands/core.py
+++ b/redis/commands/core.py
@@ -3654,17 +3654,25 @@ class BasicKeyCommands(CommandsProtocol):
 
         ``ordering`` controls the order at the destination: ``OBO`` pushes each
         element as it is popped (reversing block order, stack semantics) while
-        ``BULK`` preserves the original relative order (queue semantics).
+        ``BULK`` preserves the original relative order (queue semantics). It is
+        required whenever ``count`` is given.
 
         Returns the array of moved elements, or ``None`` if nothing was moved.
 
         For more information, see https://redis.io/commands/lmovem
         """
+        if count is None:
+            if mode is not None or ordering is not None:
+                raise DataError(
+                    "``count`` is required when ``mode`` or ``ordering`` is set"
+                )
+        elif ordering is None:
+            raise DataError(
+                "``ordering`` (OBO or BULK) is required when ``count`` is set"
+            )
         pieces: list[EncodableT] = [first_list, second_list, src, dest]
         if count is not None:
-            pieces.extend([mode or "COUNT", count])
-            if ordering is not None:
-                pieces.append(ordering)
+            pieces.extend([mode or "COUNT", count, ordering])
         return self.execute_command("LMOVEM", *pieces)
 
     @overload
@@ -3714,13 +3722,23 @@ class BasicKeyCommands(CommandsProtocol):
         holds at least ``count`` elements, then moves exactly ``count``
         atomically. On timeout ``None`` is returned and nothing is moved.
 
+        As with ``lmovem``, ``ordering`` (OBO or BULK) is required whenever
+        ``count`` is given.
+
         For more information, see https://redis.io/commands/blmovem
         """
+        if count is None:
+            if mode is not None or ordering is not None:
+                raise DataError(
+                    "``count`` is required when ``mode`` or ``ordering`` is set"
+                )
+        elif ordering is None:
+            raise DataError(
+                "``ordering`` (OBO or BULK) is required when ``count`` is set"
+            )
         pieces: list[EncodableT] = [first_list, second_list, src, dest, timeout]
         if count is not None:
-            pieces.extend([mode or "COUNT", count])
-            if ordering is not None:
-                pieces.append(ordering)
+            pieces.extend([mode or "COUNT", count, ordering])
         return self.execute_command("BLMOVEM", *pieces)
 
     @overload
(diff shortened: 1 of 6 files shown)


redis/commands/core.py
```diff
@@ -3605,6 +3605,124 @@ def blmove(
...
+        ``BULK`` preserves the original relative order (queue semantics).
+
+        Returns the array of moved elements, or ``None`` if nothing was moved.
+
+        For more information, see https://redis.io/commands/lmovem
+        """
+        pieces: list[EncodableT] = [first_list, second_list, src, dest]
+        if count is not None:
+            pieces.extend([mode or "COUNT", count])
+            if ordering is not None:
```
elena-kolevska: The ordering is required when a count is used ([PR in core](https://github.com/redis/redis/pull/15405/changes#diff-14d4391345129ebfb88a40c4868cd1eec3e0abc6957fc95992487f2b521628aaR1510)).

elena-kolevska (changes requested):
Looks good, except for that one small comment on ordering.

redis/commands/core.py
petyaslavova: Fixed.

redis/commands/core.py
```diff
@@ -3605,6 +3605,142 @@ def blmove(
...
+        src: str = "LEFT",
+        dest: str = "RIGHT",
+        count: int | None = None,
+        mode: Literal["COUNT", "EXACTLY"] | None = None,
+        ordering: Literal["OBO", "BULK"] | None = None,
+    ) -> Awaitable[list[bytes | str] | None]: ...
+
+    def lmovem(
+        self,
+        first_list: str,
```
vladvildanov, line 3634: Maybe it makes sense to rename it to `src_list` and `dest_list`, so it would be more obvious which list is a source and which destination

vladvildanov (changes requested):

redis/commands/core.py
petyaslavova: All other commands are using this pattern with first and second and they have additional config that sets which one is the src and which one is the destination. 
Having just two args that are src and dest for all commands is really more user-friendly, but having a different pattern for just those two commands would be confusing.

vladvildanov (approved):

elena-kolevska (approved):

PR #4179, opened 2026-07-10
feat: add COLLECT reducer support to search aggregations

petyaslavova:
## Change summary

This pull request adds support for the `COLLECT` aggregation reducer to the
search-and-query command surface. `COLLECT` gathers the rows of each `GROUPBY`
group, projects a chosen set of fields from every row, and returns them as an
array of per-entry maps under the reducer alias, with optional deduplication,
sorting, and limiting. It is a preview feature gated behind the server's
`search-enable-unstable-features` flag. The change is purely additive: a new
`collect` reducer class alongside the existing reducers, with no modification to
any existing reducer or public signature.

The implementation lives entirely in `redis/commands/search/reducers.py`. A new
`collect(Reducer)` class builds the `FT.AGGREGATE ... REDUCE COLLECT` argument
vector from four constructor options: `fields` (either the literal `"*"` or a
name / list of names), `distinct`, `sort_by` (a name, an `Asc`/`Desc` instance,
or a list of them), and `limit` as an `(offset, count)` pair. A small module-level
helper, `_ensure_at_prefix`, normalizes every field and sort key to a single
leading `@` on the wire so callers may pass bare names or `@`-prefixed names
interchangeably. Empty or blank field specifications raise `ValueError` early
rather than emitting a malformed command.

A few edge cases are handled explicitly: `sort_by` accepts a single value or a
list and coerces bare names to ascending order via `Asc`/`Desc`; the `FIELDS`
and `SORTBY` clauses emit their argument counts as the wire protocol requires;
and `distinct` is wired through as a forward-compatible `DISTINCT` token even
though the server does not yet implement it (documented in the docstring as
currently producing a server error). Reducers are shared between the sync and
async stacks — both import the same `redis/commands/search/reducers.py` — so
there is no separate async runtime mirror to update; sync/async parity here is
about test coverage, which is provided on both sides. There is no
backward-compatibility or public-API risk, since nothing existing changes.

## Test coverage

New tests are added in both `tests/test_search.py` and
`tests/test_asyncio/test_search.py`, kept in lockstep, with shared helpers
(`collect_entry_to_dict`, `collect_groups`) and a common `_setup_collect_index`
fixture. They exercise the full option matrix against a live index: descending
and ascending `sort_by`, `limit` as top-N selection, multiple sort keys, sparse
projection where some fields are absent from rows, projecting a key field,
`FIELDS *` (project-all), single-string-field projection, and passing field
names with and without the `@` prefix to confirm normalization.

Argument-validation behavior is covered by a parametrized
`test_aggregations_collect_invalid_fields` (sync and async) asserting `ValueError`
for empty/blank field inputs, which needs no server connection. The integration
tests rely on the `COLLECT` preview feature and therefore require a server with

---

> [!NOTE]
> **Low Risk**
> Additive search aggregation API and tests only; no changes to existing reducers or command paths. Preview feature is opt-in via server config and Redis 8.9+ in tests.
> 
> **Overview**
> Adds client support for the preview **COLLECT** `FT.AGGREGATE` reducer so grouped pipelines can return per-row field maps as an array under a reducer alias.
> 
> In `reducers.py`, a new `collect` class builds the `REDUCE COLLECT` wire args from `fields` (`*` or named projections), optional `distinct`, `sort_by` (`Asc`/`Desc`), and `limit` `(offset, count)`. **`_ensure_at_prefix`** normalizes field and sort keys to a single `@` on the wire; empty or blank field specs raise **`ValueError`** before the command is sent. **`distinct`** is emitted for forward compatibility but is documented as not yet supported server-side.
> 
> Sync and async integration tests share fruit fixtures, a `collect_index` fixture that toggles **`search-enable-unstable-features`**, and helpers that normalize COLLECT results across RESP2/RESP3. Coverage includes sort direction, top-N limits, sparse projection, `@__key`, `FIELDS *`, and client-side validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a5e0fc09f59e3f404c2ca4505b441503add87ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/commands/search/reducers.py
+++ b/redis/commands/search/reducers.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Iterable, Union
 
 from .aggregation import Asc, Desc, Reducer, SortDirection
 
@@ -202,26 +202,25 @@ class collect(Reducer):
 
     def __init__(
         self,
-        fields: Union[str, list[str]] = "*",
+        fields: str | Iterable[str] = "*",
         distinct: bool = False,
-        sort_by: Union[str, Asc, Desc, list[Union[str, Asc, Desc]], None] = None,
-        limit: Union[tuple[int, int], None] = None,
+        sort_by: Asc | Desc | Iterable[Asc | Desc] | None = None,
+        limit: tuple[int, int] | None = None,
     ) -> None:
         """
         ### Parameters
 
         - **fields**: The fields to project from each collected row. Either the
             literal ``"*"`` (project every field present in the pipeline at this
-            stage) or a field name / list of field names. Names are normalized
-            to a single leading ``@`` on the wire; output map keys are the bare
-            names.
+            stage) or a field name / iterable of field names. Names are
+            normalized to a single leading ``@`` on the wire; output map keys
+            are the bare names.
         - **distinct**: When ``True``, emit ``DISTINCT`` to deduplicate entries
             with identical projected fields. Forward-compatible: this option is
             not yet implemented by the server and currently produces a server
             error when sent.
-        - **sort_by**: One field name, an ``Asc``/``Desc`` instance, or a list
-            of them, used to order the collected entries within each group.
-            Bare names default to ascending order.
+        - **sort_by**: An ``Asc``/``Desc`` instance or an iterable of them, used
+            to order the collected entries within each group.
         - **limit**: An ``(offset, count)`` pair. Returns at most ``count``
             entries per group after skipping ``offset``. With ``sort_by`` this
             acts as a top-N selection.
@@ -232,18 +231,11 @@ class collect(Reducer):
         if fields == "*":
             args += ["FIELDS", "*"]
         else:
-            if isinstance(fields, str):
-                if not fields.strip():
-                    raise ValueError(
-                        "collect fields must be '*' or a non-empty list of names"
-                    )
-                names = [fields]
-            else:
-                names = list(fields)
-                if not names:
-                    raise ValueError(
-                        "collect fields must be '*' or a non-empty list of names"
-                    )
+            names = [fields] if isinstance(fields, str) else list(fields)
+            if not names or any(not n.strip() for n in names):
+                raise ValueError(
+                    "collect fields must be '*' or a non-empty list of names"
+                )
             names = [_ensure_at_prefix(n) for n in names]
             args += ["FIELDS", str(len(names))] + names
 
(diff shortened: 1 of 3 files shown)


redis/commands/search/reducers.py
```diff
@@ -180,3 +185,90 @@ def __init__(self, field: str, size: int) -> None:
         args = [field, str(size)]
         super().__init__(*args)
         self._field = field
+
+
+class collect(Reducer):
```
vladvildanov, line 190: Class name should start with capital letter, isn't it?

redis/commands/search/reducers.py
```diff
@@ -180,3 +185,90 @@ def __init__(self, field: str, size: int) -> None:
...
+    ``COLLECT`` is a preview feature gated behind
+    ``search-enable-unstable-features``. See
+    `FT.AGGREGATE <https://redis.io/commands/ft.aggregate>`_.
+    """
+
+    NAME = "COLLECT"
+
+    def __init__(
+        self,
+        fields: Union[str, list[str]] = "*",
```
vladvildanov: API wise `Iterable[str]` would be better here, so it allows multiple subtypes to be passed

vladvildanov: AFAIR, we switched from `Union` to `|` operator?

redis/commands/search/reducers.py
```diff
@@ -180,3 +185,90 @@ def __init__(self, field: str, size: int) -> None:
...
+        distinct: bool = False,
+        sort_by: Union[str, Asc, Desc, list[Union[str, Asc, Desc]], None] = None,
```
vladvildanov: Passing `str` as value doesn't seems correct. Maybe it's better to keep just `Iterable[Asc | Desc]`?

tests/test_asyncio/test_search.py
```diff
@@ -1645,7 +1645,71 @@ async def test_config_with_removed_ftconfig(self, decoded_r: redis.Redis):
...
+    return {
+        attrs[group_field]: [collect_entry_to_dict(e) for e in attrs[alias]]
+        for attrs in rows
+    }
+
+
 class TestAggregations(AsyncSearchTestsBase):
+    async def _setup_collect_index(self, client):
+        """Enable the preview feature and index the shared fruit documents."""
+        await client.config_set("search-enable-unstable-features", "yes")
```
vladvildanov: Should cleanup unstable feature flag after test execution

vladvildanov (changes requested):

redis/commands/search/reducers.py
petyaslavova: This one follows the example of the other reducers classes.

redis/commands/search/reducers.py
petyaslavova: fixed

petyaslavova: fixed.

redis/commands/search/reducers.py
petyaslavova: done.

tests/test_asyncio/test_search.py
petyaslavova: done.

redis/commands/search/reducers.py
```diff
@@ -180,3 +185,90 @@ def __init__(self, field: str, size: int) -> None:
...
+                    )
+                names = [fields]
+            else:
+                names = list(fields)
+                if not names:
+                    raise ValueError(
+                        "collect fields must be '*' or a non-empty list of names"
+                    )
+            names = [_ensure_at_prefix(n) for n in names]
+            args += ["FIELDS", str(len(names))] + names
```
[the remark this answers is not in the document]
petyaslavova: fixed

vladvildanov (approved):

PR #4184, opened 2026-07-13
fix: add close()/aclose() to Sentinel to release sentinel client pools

### LINKED ISSUE #3201: asyncio Sentinel  doesn't provide a proper way to clear its  child connection pools
redis-py=-5.0.3, master

python 3.9
Windows 10/ Ubuntu 22.04

```python
from redis.sentinel import Sentinel
# Sentinel.__init__ creates a list of Redis clients - one for each host. Every Redis client contains its own connection pool.
# https://github.com/redis/redis-py/blob/07fc339b4a4088c1ff052527685ebdde43dfc4be/redis/asyncio/sentinel.py#L215
sentinel = Sentinel([('host1', 26379), ('host2', 26379)], socket_timeout=0.1)
# master_for() creates a new SentinelConnectionPool
# https://github.com/redis/redis-py/blob/07fc339b4a4088c1ff052527685ebdde43dfc4be/redis/asyncio/sentinel.py#L350
master = sentinel.master_for('mymaster', socket_timeout=0.1)
await master.set('foo', 'bar')
# Closes a single Redis ConnectionPool instance
# https://github.com/redis/redis-py/blob/07fc339b4a4088c1ff052527685ebdde43dfc4be/redis/asyncio/client.py#L576
await master.aclose(close_connection_pool=True)
```

Since, not all connection pools were closed, testing it with`IsolatedAsyncioTestCase` we get the following warning:
```
/usr/local/lib/python3.9/site-packages/_pytest/unraisableexception.py:78: PytestUnraisableExceptionWarning: Exception ignored in: <function AbstractConnection.__del__ at 0x772f81b31550>
  
    File "/usr/local/lib/python3.9/site-packages/redis/asyncio/connection.py", line 216, in __del__
      self._close()
    File "/usr/local/lib/python3.9/site-packages/redis/asyncio/connection.py", line 223, in _close
      self._writer.close()
```

A workaround is to close all connections in the sentinel.sentinels list too:
```python
for s in sentinel.sentinels:
    await s.aclose()
```

Brumbelow:
### Description of change

`Sentinel` (both sync and async) builds a `Redis` client per sentinel node in `__init__` (`self.sentinels`), each with its own connection pool, but exposes no way to close them. Letting a `Sentinel` go out of scope leaves those pools open — which surfaces as `Exception ignored in <AbstractConnection.__del__>` warnings on GC (e.g. under `IsolatedAsyncioTestCase`) and gives tests / shutdown paths no clean teardown. The only workaround today is to reach into `sentinel.sentinels` and close each one by hand.

This adds `close()` (sync) / `aclose()` (async), plus context-manager support, which close every sentinel client and its pool — mirroring the `Redis` client API:

```python
with Sentinel([("localhost", 26379)]) as sentinel:
    master = sentinel.master_for("mymaster")
    ...
```

Clients returned by `master_for`/`slave_for` are owned by the caller and are left untouched.

### Pull Request check-list

  - [x] Do the CI tests pass with this change?
  - [ ] Is a documentation update included?
  - [ ] Is there an example added to the examples folder?

Fixes redis/redis-py#3201

---

> [!NOTE]
> **Low Risk**
> Additive lifecycle API with bounded behavior; no changes to discovery, failover, or command execution paths.
> 
> **Overview**
> **`Sentinel` now has explicit teardown** for the per-node `Redis` clients it creates in `self.sentinels`, which previously had no public shutdown path and could leak pools (e.g. GC warnings in async tests).
> 
> Sync **`close()`** and async **`aclose()`** shut down every internal sentinel client and its pool. If one close fails, the rest still run and the first error is re-raised afterward. **`master_for` / `slave_for` clients are not closed**—callers still own those.
> 
> **Context manager support** (`with Sentinel(...)` / `async with Sentinel(...)`) calls `close()` / `aclose()` on exit, aligned with the main `Redis` client API.
> 
> Tests cover context-manager teardown, error propagation with full close attempts, and calling close twice.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30616fab405ed2952a6c100438637e734476cbe8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/sentinel.py
+++ b/redis/asyncio/sentinel.py
@@ -269,11 +269,19 @@ class Sentinel(AsyncSentinelCommands):
         Close all sentinel clients created by this Sentinel and their
         connection pools.
 
+        Each client is closed independently: if one raises, the remaining
+        clients are still closed and the first error is re-raised afterwards.
+
         Clients returned by ``master_for``/``slave_for`` are owned by the
         caller and are not closed here.
         """
-        for sentinel in self.sentinels:
-            await sentinel.aclose()
+        resp = await asyncio.gather(
+            *(sentinel.aclose() for sentinel in self.sentinels),
+            return_exceptions=True,
+        )
+        exc = next((r for r in resp if isinstance(r, BaseException)), None)
+        if exc:
+            raise exc
 
     async def __aenter__(self) -> "Sentinel":
         return self
--- a/redis/asyncio/sentinel.py
+++ b/redis/asyncio/sentinel.py
@@ -264,6 +264,23 @@ class Sentinel(AsyncSentinelCommands):
             f"(sentinels=[{','.join(sentinel_addresses)}])>"
         )
 
+    async def aclose(self) -> None:
+        """
+        Close all sentinel clients created by this Sentinel and their
+        connection pools.
+
+        Clients returned by ``master_for``/``slave_for`` are owned by the
+        caller and are not closed here.
+        """
+        for sentinel in self.sentinels:
+            await sentinel.aclose()
+
+    async def __aenter__(self) -> "Sentinel":
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback) -> None:
+        await self.aclose()
+
     def check_master_state(self, state: dict, service_name: str) -> bool:
         if not state["is_master"] or state["is_sdown"] or state["is_odown"]:
             return False
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -294,6 +294,23 @@ class Sentinel(SentinelCommands):
             f"(sentinels=[{','.join(sentinel_addresses)}])>"
         )
 
+    def close(self) -> None:
+        """
+        Close all sentinel clients created by this Sentinel and their
+        connection pools.
+
+        Clients returned by ``master_for``/``slave_for`` are owned by the
+        caller and are not closed here.
+        """
+        for sentinel in self.sentinels:
+            sentinel.close()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+
     def check_master_state(self, state, service_name):
         if not state["is_master"] or state["is_sdown"] or state["is_odown"]:
             return False
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -299,11 +299,21 @@ class Sentinel(SentinelCommands):
         Close all sentinel clients created by this Sentinel and their
         connection pools.
 
+        Each client is closed independently: if one raises, the remaining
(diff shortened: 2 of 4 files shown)


eeshsaxena (commented):
Confirmed the gap: `Sentinel` (sync and async) currently exposes no `close`/`aclose` and no context-manager protocol, so the sentinel clients it builds are never released.

The ownership call is the subtle part here, and I think you got it right. In the constructor:

```python
self.sentinels = [
    Redis(hostname, port, **self.sentinel_kwargs)
    for hostname, port in sentinels
]
```

the `Sentinel` **constructs** those clients, so it owns them and closing them in `close()` is correct. Conversely `master_for`/`slave_for` build a fresh `redis_class` instance and hand it back to the caller without retaining a reference, so they are genuinely not the `Sentinel`'s to close - the docstring saying so is accurate and worth having spelled out. Using `aclose()` (rather than the deprecated `close()`) on the async path is also the right choice.

One robustness suggestion. Both loops close sequentially and are not exception-safe:

```python
for sentinel in self.sentinels:
    sentinel.close()
```

If any one client raises on close (a half-broken socket during a sentinel outage is exactly when you'd be tearing down), the loop aborts and every remaining sentinel client leaks - which is the failure this PR exists to prevent. Wrapping each in `contextlib.suppress(Exception)`, or collecting errors and re-raising after the loop, would make cleanup all-or-nothing. `__exit__`/`__aexit__` are the most important place for that, since they run during exception unwinding.

Nice that both sync and async got tests. Two cases worth adding if they aren't already covered: `close()` being called twice (idempotency, since users will pair it with a `with` block), and one sentinel raising on close while the others still get closed - that pins the behaviour above whichever way you decide it should go.

Deferring to the maintainers.

petyaslavova (changes requested):
Hey @Brumbelow , thank you for your contribution! 
Overall the change looks good and correct and this is the right direction how things should be fixed.
One change before we merge: make the teardown exception-safe. Right now if a single client raises on close — exactly what happens during a sentinel outage, when you're most likely tearing down — the loop aborts and the remaining pools leak, which is the failure this PR is meant to fix. Please close each client independently (e.g. contextlib.suppress(Exception) or collect-and-reraise after the loop), especially in __exit__/__aexit__. Please also add tests for one client raising while the others still get closed, and for calling close()/aclose() twice (idempotency), on both stacks.

Brumbelow:
Thanks @petyaslavova, this is done.  `close()`/`aclose()` now close each client independently and re-raise the first error after the loop, following the same collect-and-reraise pattern the async `ConnectionPool.disconnect` uses. Added tests on both stacks for a client that raises on close (the others still get closed) and for calling close twice.

petyaslavova (approved):
LGTM.

PR #4193, opened 2026-07-15
Fix Sentinel pool capacity loss after failover

### LINKED ISSUE #4187: SentinelConnectionPool permanently loses a pool slot on every master switch -> "Too many connections" forever
### Version

redis-py 8.0.0 and 8.0.1; the code is unchanged on `master` as of today.

### Platform

Python 3.13, Linux. Redis Sentinel (3 sentinels, 3 servers).

### Description

`ConnectionPool` tracks capacity with a monotonic `_created_connections`: `make_connection()` increments it, and **nothing** decrements it. That is sound only because of an invariant — a connection handed out is always handed back to `_available_connections` by `release()`. This is stated explicitly by a maintainer in #1247:

> Created connections are never destroyed. `release()` just returns the connection to the pool so that it can be reused. Therefore `_created_connections` never gets decremented.

`SentinelConnectionPool` is the one pool that breaks that invariant. Its `owns_connection()` additionally requires the connection's `(host, port)` to equal the pool's **current** `master_address`:

```python
# redis/sentinel.py
def owns_connection(self, connection):
    check = not self.is_master or (
        self.is_master and self.master_address == (connection.host, connection.port)
    )
    parent = super()
    return check and parent.owns_connection(connection)
```

So after a failover, every connection that was *checked out at that moment* comes back "unowned", and `ConnectionPool.release()` destroys it — leaving the counter incremented:

```python
# redis/connection.py
else:
    # Pool doesn't own this connection, do not add it back
    # to the pool.
    # The created connections count should not be changed,
    # because the connection was not created by the pool.
    # Still need to decrement USED since it was counted in get_connection()
    connection.disconnect()
    ...
    return
```

The premise in that comment does not hold in the Sentinel case: the pool **did** create this connection. It is discarded merely for pointing at the previous master.

The slot is then gone for the life of the process. After `max_connections` such drops the pool raises `MaxConnectionsError("Too many connections")` on **every** subsequent command, permanently — even though it holds no connections at all and the server is idle. Only restarting the process recovers.

Two things suggest this is an oversight rather than intent:

* `BlockingConnectionPool.release()` handles the identical "not owned" case correctly — it pushes a `None` placeholder back into the queue, "*which will cause the pool to recreate the connection if its needed*". Capacity is preserved.
* The asyncio `ConnectionPool` cannot express the bug at all: it has no counter and gates on `len(self._in_use_connections)`, a live count.

Only the sync `ConnectionPool` + `SentinelConnectionPool` combination leaks.

#### Impact

This took down a dramatiq worker in production. Its broker pool (`max_connections=20`, peak demand ~6) accumulated dead slots across master switches until every command failed; the worker's consumer threads then logged `Consumer encountered a connection error: Too many connections`, restarted 3 seconds later, failed again, and repeated — for four days — while the process stayed up and looked healthy. Nothing recovers on its own, because a burned slot never comes back.

### Reproduction

No Redis server needed — the bug is in the pool's accounting.

```python
import os
from unittest.mock import MagicMock

import redis
from redis.sentinel import SentinelConnectionPool

MASTER_A, MASTER_B = ("master-a", 6379), ("master-b", 6379)


class FakeConnection:
    """Minimal stand-in for SentinelManagedConnection."""

    def __init__(self, **kwargs):
        self.host, self.port = MASTER_A
        self.pid = os.getpid()

    def connect(self): pass
    def disconnect(self, *args): pass
    def can_read(self, timeout=0): return False
    def re_auth(self): pass


pool = SentinelConnectionPool(
    "mymaster", MagicMock(), connection_class=FakeConnection, max_connections=3
)
pool.proxy.master_address = MASTER_A

print(f"redis-py {redis.__version__}, max_connections={pool.max_connections}\n")
for i in range(1, 4):
    conn = pool.get_connection()          # a command is in flight...
    pool.proxy.master_address = MASTER_B  # ...and sentinel promotes a new master
    pool.release(conn)                    # owns_connection() is False -> connection destroyed
    pool.proxy.master_address = MASTER_A  # the switch settles
    print(f"after failover {i}: _created_connections={pool._created_connections} "
          f"available={len(pool._available_connections)} in_use={len(pool._in_use_connections)}")

print("\nThe pool now believes it is full, and holds nothing:")
try:
    pool.get_connection()
except redis.exceptions.MaxConnectionsError as e:
    print(f"  {type(e).__name__}: {e}")
```

Output:

```
redis-py 8.0.0, max_connections=3

after failover 1: _created_connections=1 available=0 in_use=0
after failover 2: _created_connections=2 available=0 in_use=0
after failover 3: _created_connections=3 available=0 in_use=0

The pool now believes it is full, and holds nothing:
  MaxConnectionsError: Too many connections
```

Each iteration is one master switch with a single command in flight. In a real worker the loss per switch is the number of connections checked out at that instant, so a pool sized for a handful of threads can be exhausted by a handful of failovers.

### Expected behaviour

Discarding a connection should give its slot back, preserving `_created_connections == len(_available_connections) + len(_in_use_connections)`. In the discard branch of `ConnectionPool.release()`, decrement when the pool actually created the connection — i.e. when `connection.pid == self.pid`. (The pid check is what keeps the pre-fork case correct: `reset()` has already zeroed the counter for those, so they must *not* decrement.)

Happy to open a PR with the fix and a regression test if that shape is agreeable.

Sanjays2402:
### Description of change

Fixes #4187. During a Sentinel master failover, the sync pool can reject an in-flight connection it originally created, but releasing it did not restore `_created_connections`, permanently consuming capacity. The pool now reclaims that slot only for connections created in its current process, and the regression test fails before the fix and passes after it.

### Pull Request check-list

- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)? (Pending upstream CI.)
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? (Not applicable; no API change.)
- [x] Is there an example added to the examples folder (if applicable)? (Not applicable.)

_NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open._

---

> [!NOTE]
> **Low Risk**
> Small, targeted change to sync `ConnectionPool.release()` for the non-owned-connection path; guarded by PID check to avoid incorrect decrements after fork.
> 
> **Overview**
> Fixes **permanent connection pool shrink** after Sentinel master failover (#4187). When `SentinelConnectionPool` rejects a connection on `release()` because it no longer points at the current master, the pool now decrements `_created_connections` (guarded by `connection.pid == self.pid`) so the slot is not lost forever.
> 
> Adds **`test_master_failover_reclaims_discarded_connection_slot`**, which simulates failover between two master addresses and asserts the pool can still allocate connections after repeated get/release cycles.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72581ec6974b3df32eca473e2d24042a57ee7698. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -3301,10 +3301,11 @@ class ConnectionPool(MaintNotificationsAbstractConnectionPool, ConnectionPoolInt
             else:
                 # Pool doesn't own this connection, do not add it back
                 # to the pool.
-                # The created connections count should not be changed,
-                # because the connection was not created by the pool.
                 # Still need to decrement USED since it was counted in get_connection()
                 connection.disconnect()
+                # Subclasses can reject a connection created by this pool.
+                if connection.pid == self.pid:
+                    self._created_connections -= 1
                 record_connection_count(
                     pool_name="unknown_pool",
                     connection_state=ConnectionState.USED,
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -3303,7 +3303,10 @@ class ConnectionPool(MaintNotificationsAbstractConnectionPool, ConnectionPoolInt
                 # to the pool.
                 # Still need to decrement USED since it was counted in get_connection()
                 connection.disconnect()
-                # Subclasses can reject a connection created by this pool.
+                # Subclasses such as SentinelConnectionPool can override
+                # owns_connection() with a comparison different from local PID
+                # ownership. When such a subclass rejects a connection, also require
+                # connection.pid == self.pid before reclaiming its slot.
                 if connection.pid == self.pid:
                     self._created_connections -= 1
                 record_connection_count(
--- a/tests/test_sentinel.py
+++ b/tests/test_sentinel.py
@@ -1,3 +1,4 @@
+import os
 import socket
 from unittest import mock
 
@@ -257,6 +258,46 @@ def test_slave_round_robin(cluster, sentinel, master_ip):
     assert next(rotator) == (master_ip, 6379)
     with pytest.raises(SlaveNotFoundError):
         next(rotator)
+
+
+@pytest.mark.onlynoncluster
+def test_master_failover_reclaims_discarded_connection_slot():
+    master_a = ("master-a", 6379)
+    master_b = ("master-b", 6379)
+
+    class FakeConnection:
+        def __init__(self, **kwargs):
+            self.host, self.port = master_a
+            self.pid = os.getpid()
+
+        def connect(self):
+            pass
+
+        def disconnect(self):
+            pass
+
+        def can_read(self, timeout=0):
+            return False
+
+        def should_reconnect(self):
+            return False
+
+    pool = SentinelConnectionPool(
+        "mymaster",
+        mock.MagicMock(),
+        connection_class=FakeConnection,
+        max_connections=2,
+    )
+    pool.proxy.master_address = master_a
+
+    for _ in range(pool.max_connections):
+        connection = pool.get_connection()
+        pool.proxy.master_address = master_b
+        pool.release(connection)
+        pool.proxy.master_address = master_a
+
+    assert pool._created_connections == 0
+    pool.get_connection()
 
 
 @pytest.mark.onlynoncluster
--- a/tests/test_sentinel.py
+++ b/tests/test_sentinel.py
@@ -260,6 +260,7 @@ def test_slave_round_robin(cluster, sentinel, master_ip):
         next(rotator)
 
 
+@pytest.mark.fixed_client
 @pytest.mark.onlynoncluster
 def test_master_failover_reclaims_discarded_connection_slot():
     master_a = ("master-a", 6379)


eeshsaxena (commented):
Confirmed, and this one is worth taking - the pool ends up permanently unusable, not just leaky. I reproduced it on current `master` with your scenario:

```
_created_connections after 2 failover releases: 2  (expected 0)
get_connection() -> RAISED: MaxConnectionsError Too many connections
```

So after `max_connections` in-flight connections happen to span a failover, the pool is exhausted forever even though it holds no connections.

The mechanism matches your description. `ConnectionPool.owns_connection()` is just `connection.pid == self.pid` (connection.py:3316), while `SentinelConnectionPool.owns_connection()` is `address_matches and super().owns_connection(...)` (sentinel.py:190). So `release()`'s `else` branch is reachable for two *different* reasons:

1. the connection genuinely came from another process/pool (pid differs) - the count must not be touched, which is what the old comment was protecting;
2. this pool created the connection but the subclass rejected it because the master address moved - here the slot was counted in `make_connection()` and must be given back.

Gating the decrement on `connection.pid == self.pid` is exactly the predicate that separates (2) from (1), so the fix looks correct to me, and the old comment's intent is preserved for case (1).

One small suggestion: `connection.pid == self.pid` is literally the body of the base `ConnectionPool.owns_connection()`. Since `self.owns_connection` is overridden here, you can't call it, but `ConnectionPool.owns_connection(self, connection)` would express the intent directly and keep the two in sync if the base ownership rule ever changes. Purely cosmetic.

For scope: the asyncio pool tracks capacity with `_in_use_connections` / `_available_connections` rather than a `_created_connections` counter, so it doesn't look affected by this particular leak.

The regression test is well targeted - it fails on master with `MaxConnectionsError` and passes with the fix. Deferring to the maintainers.

redis/connection.py
```diff
@@ -3301,10 +3301,11 @@ def release(self, connection: "Connection") -> None:
             else:
                 # Pool doesn't own this connection, do not add it back
                 # to the pool.
-                # The created connections count should not be changed,
-                # because the connection was not created by the pool.
                 # Still need to decrement USED since it was counted in get_connection()
                 connection.disconnect()
+                # Subclasses can reject a connection created by this pool.
```
petyaslavova: Please add a more detailed comment here - the key info is that subclasses (like SentinelConnectionPool, worth mentioning the concrete class as it often is missed as another client we have) can override `owns_connection` method, which will return its output based on a different comparison than the local one, and in this case we should also compare `connection.pid == self.pid`.

tests/test_sentinel.py
```diff
@@ -259,6 +260,46 @@ def test_slave_round_robin(cluster, sentinel, master_ip):
         next(rotator)
 
 
+@pytest.mark.onlynoncluster
```
petyaslavova, line 264: Please also add `fixed_client` marker.

petyaslavova (changes requested):
Hey @Sanjays2402, thank you for your contribution! Two minor changes need to be handled, and it will be ready for merging.

petyaslavova (approved):
LGTM.

PR #4195, opened 2026-07-17
Eliminate GC refcounting cycle on _connect exception

rschlaikjer:
### Description of change

Currently, in `_connect`, if an exception is thrown while trying different address infos, it is cached to be thrown only if _no_ connection succeeds. However, storing a caught exception inside the local stack frame creates a circular reference: the exception's traceback contains a pointer to the current frame, which contains the `err` local, which points to the exception again. This creates garbage that can only be collected by the fallback mark and sweep python GC. 

Break this cycle by clearing the local exception value on re-raise, allowing the error to be reclaimed via refcounting. This is useful for applications that disable or otherwise limit the python garbage collector.

### Pull Request check-list

---

> [!NOTE]
> **Low Risk**
> Localized TCP connect error-handling cleanup with no API or connection-semantics change; tests lock in frame-local clearing behavior.
> 
> **Overview**
> Fixes a **reference cycle** in `Connection._connect` when iterating over `getaddrinfo` results: a failed attempt stores the caught `OSError` in `err` for possible re-raise, but that exception’s traceback keeps the stack frame alive, and the frame still holds `err`—so cleanup depended on the cyclic GC instead of refcounting.
> 
> On **success** after an earlier failure, **`err` is set to `None`** before returning the socket. When **all addresses fail**, re-raise is wrapped in **`try` / `finally`** so **`err` is cleared** after `raise err`, matching the intent described in `read_response` (drop locals that pin exceptions).
> 
> Two unit tests assert that the `_connect` frame’s **`err` local is `None`** after total failure and after fail-then-succeed multi-address connect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33aae85399d05041b2614c8ab16c8dc36153ac96. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -1555,6 +1555,14 @@ class Connection(AbstractConnection):
         # we want to mimic what socket.create_connection does to support
         # ipv4/ipv6, but we want to set options prior to calling
         # socket.connect()
+
+        # Last caught connection error.
+        # Re-thrown if we are unable to connect to any of the options returned
+        # by getaddrinfo.
+        # Note that we must clear this variable before returning - otherwise,
+        # a caught err's traceback points to this frame, which points to err.
+        # Clearing this lets refcounting reclaim the exception immediately
+        # without deferring to the python garbage collector.
         err = None
 
         for res in socket.getaddrinfo(
@@ -1581,6 +1589,10 @@ class Connection(AbstractConnection):
 
                 # set the socket_timeout now that we're connected
                 sock.settimeout(self.socket_timeout)
+
+                # If a previous connection attempt failed, clear the error
+                err = None
+
                 return sock
 
             except OSError as _:
@@ -1593,7 +1605,11 @@ class Connection(AbstractConnection):
                     sock.close()
 
         if err is not None:
-            raise err
+            try:
+                raise err
+            finally:
+                # Ensure we clear local references to caught exceptions
+                err = None
         raise OSError("socket.getaddrinfo returned an empty list")
 
     def _host_error(self):
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -574,6 +574,35 @@ class TestConnection:
         mock_sock.close.assert_called_once()
         assert conn._sock is None
 
+    def test_connect_breaks_exception_reference_cycle(self):
+        """
+        On connection failure, _connect must not leave its frame retaining the
+        raised exception.
+        The exception's traceback references the frame, so a retained local
+        would form a cycle only reclaimable by the GC.
+        The finally clause in _connect breaks it by clearing the local.
+        """
+        conn = Connection(host="localhost", port=6379)
+        addr_info = (socket.AF_INET, socket.SOCK_STREAM, 0, "", ("127.0.0.1", 6379))
+        with (
+            patch.object(socket, "getaddrinfo", return_value=[addr_info]),
+            patch.object(socket, "socket") as socket_factory,
+        ):
+            socket_factory.return_value.connect.side_effect = OSError("refused")
+            with pytest.raises(OSError) as exc_info:
+                conn._connect()
+
+        # Locate the _connect frame in the propagated traceback and confirm its
+        # err local was cleared, proving no exception<->frame cycle survives.
+        connect_frame = None
+        tb = exc_info.value.__traceback__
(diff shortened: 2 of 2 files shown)


sean-kim05 (commented):
This matches the idiom CPython uses in `socket.create_connection` itself (`try: raise err finally: err = None`), so it's well-precedented. 👍

One thing worth surfacing for reviewers: the added `err = None` just before `return sock` isn't redundant with the `finally`. It covers the case where an *earlier* address from `getaddrinfo` fails (populating `err`) and a *later* one succeeds — without it, the returning frame still holds that earlier exception, whose traceback references the frame, keeping the cycle alive until the GC runs. So both clears pull their weight; might be worth a one-line comment on that second one since its purpose is less obvious than the `finally`.

petyaslavova:
Hey @rschlaikjer, thank you for your contribution! I'll review it shortly.

eeshsaxena (approved):
The 	ry: raise err finally: err = None pattern is the right call here. CPython's refcount can reclaim the exception immediately once the local variable is cleared, instead of waiting for the cyclic GC to detect the frame->exception->traceback->frame cycle.

One small thing worth noting: the except OSError as _: block on the original code automatically deletes _ at the end of the except clause (Python 3 always unbinds s-target exception variables at block exit), so that part was never part of the cycle. The cycle was only through the named err variable. The PR targets exactly the right variable.

The err = None on the success path (after sock.settimeout) is also correct - without it, a retry that eventually succeeds would still hold the previous failed connection's exception alive until _connect returns. Minor, but thorough.

LGTM.

Sanjays2402:
same pattern is live in NodesManager.initialize: `exception = e` (cluster.py:2706) stays bound in that frame and then gets attached as `__cause__` on the RedisClusterException raised at 2775, so you get the identical exception -> traceback -> frame -> local loop, on the connect path too. worth clearing it there in the same PR, or do you want to keep this one narrow?

petyaslavova (changes requested):
Hey @rschlaikjer, the change looks good! 
Before merging, can you please also cover the case where the first `sock.connect` has failed, but the next one succeeds? It will be good to have this other flow covered as well...

rschlaikjer:
@petyaslavova, thanks for the review. Added a test for the failure followed by success case.

eeshsaxena (commented):
The cycle break looks right and matches CPython's own `create_connection` idiom. `try: raise err finally: err = None` re-raises the saved `OSError` and then clears the local, so the frame no longer holds a reference back to the exception (whose traceback references the frame), turning the reference cycle into a plain refcount drop instead of leaving it for the GC. Clearing `err = None` on the success path before `return sock` covers the same thing for the connect-succeeds-after-an-earlier-failure case. The empty-getaddrinfo `raise OSError(...)` path is unaffected. Correct.

petyaslavova (approved):
LGTM.

clemenslosbichler-cloud:
One thing about the new tests in #4195, not the fix itself.

Both end in:

    assert connect_frame.f_locals.get("err") is None

dict.get returns None when the fix cleared the local, and also when there's no
local by that name. So renaming err while dropping both clears, which would put
the cycle straight back, keeps both tests green. Deleting either clear on its
own does fail them, so the likely regression is covered; it's the rename that
passes vacuously.

The second test also grabs its frame with sys._getframe(1) without asserting
f_code.co_name == "_connect", so relocating the getaddrinfo call would point
the assertion at an unrelated frame.

If you'd rather assert the property than the name: with gc.disable(), take a
weakref to the OSError from the failing address (needs an OSError subclass,
the builtin isn't weak-referenceable) and assert it's dead right after _connect
raises, and again after the fail-then-succeed path returns. Otherwise adding
assert connect_frame.f_code.co_name == "_connect" and
assert "err" in connect_frame.f_locals would at least make a rename fail loudly.

Found with an automated reviewer I'm building, run against a fork with #4195
recreated on it.

petyaslavova:
Hey @clemenslosbichler-cloud, thanks for the note and for the efforts to make the library code and tests better and more reliable! Are you interested in adding a fix for the discovered test gaps? I'll be happy to review it :)

PR #4199, opened 2026-07-18
fix: use list() to copy dict items before iteration in PubSub.run()

mangodxd:
## WHAT
Guard `self.channels` and `self.patterns` with `list()` before iteration in `PubSub.run()` and `run_in_thread()` validation loops.

## WHY
If `handle_message()` calls `self.channels.pop()` (or `self.patterns.pop()`) during the startup validation loop, the live dict view mutates mid-iteration, causing `RuntimeError: dictionary changed size during iteration`.

This is a real bug, though separate from #3748 (which is an asyncio heap race in `loop.call_at()` — see review discussion).

## HOW
Wrap `.items()` calls with `list()` to snapshot the dict before iterating.

## RISK
Only iteration source changes. Subscription semantics and message ordering unchanged. No allocation in hot path — validation runs once at startup.

## VERIFY
Existing PubSub tests pass.

## NOTE
This is a preventive fix for dict-mutation-during-iteration. It does not fix the asyncio `_scheduled` heap race reported in #3748.

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -1641,10 +1641,10 @@ class PubSub:
             >>> task.cancel()
             >>> await task
         """
-        for channel, handler in self.channels.items():
+        for channel, handler in list(self.channels.items()):
             if handler is None:
                 raise PubSubError(f"Channel: '{channel}' has no handler registered")
-        for pattern, handler in self.patterns.items():
+        for pattern, handler in list(self.patterns.items()):
             if handler is None:
                 raise PubSubError(f"Pattern: '{pattern}' has no handler registered")
 
--- a/redis/client.py
+++ b/redis/client.py
@@ -1608,13 +1608,13 @@ class PubSub:
         pubsub=None,
         sharded_pubsub: bool = False,
     ) -> "PubSubWorkerThread":
-        for channel, handler in self.channels.items():
+        for channel, handler in list(self.channels.items()):
             if handler is None:
                 raise PubSubError(f"Channel: '{channel}' has no handler registered")
-        for pattern, handler in self.patterns.items():
+        for pattern, handler in list(self.patterns.items()):
             if handler is None:
                 raise PubSubError(f"Pattern: '{pattern}' has no handler registered")
-        for s_channel, handler in self.shard_channels.items():
+        for s_channel, handler in list(self.shard_channels.items()):
             if handler is None:
                 raise PubSubError(
                     f"Shard Channel: '{s_channel}' has no handler registered"


redis/asyncio/client.py
```diff
@@ -1641,10 +1641,10 @@ async def run(
             >>> task.cancel()
             >>> await task
         """
-        for channel, handler in self.channels.items():
+        for channel, handler in list(self.channels.items()):
```
Sanjays2402, line 1644: How does copying `self.channels` address #3748? That traceback is from `asyncio.base_events.call_at()` mutating the event loop's `_scheduled` heap, and it occurs in `get_message()`, not either of these startup validation loops. This change doesn't run on the reported path unless the caller uses `PubSub.run()`, so I don't think it can fix the issue as filed.

mangodxd:
You're right — thank you for catching this.

After tracing the full stack, #3748's error originates at `asyncio/base_events.py:781` inside `heapq.heappush(self._scheduled, timer)`, called from `async_timeout.__aenter__()` → `loop.call_at()`. That's the event loop's internal `_scheduled` heap, not any redis-py dictionary. The reported path is `get_message()` → `parse_response()` → `read_response()`, which never touches the validation loops in `run()` or `run_in_thread()`.

So these are two separate bugs:

1. **#3748** — asyncio heap race condition in `loop.call_at()`. This is an asyncio/`async_timeout` internals issue, not something redis-py can fix by copying dicts.

2. **The dict-mutation bug this PR actually fixes** — if `handle_message()` calls `self.channels.pop()` during the validation loop in `run()` (line 1644) or `run_in_thread()` (line 1611), the live dict view mutates mid-iteration → `RuntimeError: dictionary changed size during iteration`. This is a real bug, just not the one filed in #3748.

I'll decouple this PR from #3748 and update the description to reflect the actual scope. The dict-mutation fix is still worth merging as a preventive measure, but it doesn't address the reported traceback.

petyaslavova:
Thanks for investigating this, and for the candid follow-up in the thread separating this change from #3748.

I traced it further and don't think we should merge it. #3748's traceback ends in asyncio's own internals — heapq.heappush(self._scheduled, timer) via async_timeout.__aenter__ → loop.call_at — so the container changing size is the event loop's _scheduled heap, not a redis-py dict; copying self.channels/self.patterns can't affect it. 

Separately, the dict-iteration case this PR guards can't actually occur: handle_message() is only reached from get_message()/listen(), and in async run() the two validation loops contain no await, so the event loop cannot switch tasks mid-iteration. In sync run_in_thread() the loops finish before the worker thread starts, so a concurrent pop() requires driving one PubSub from multiple threads — which is unsupported, since PubSub is not thread-safe.

I'm going to close this. #3748 stays open — if you'd like to keep helping, a minimal reproduction plus your concurrency setup (are multiple tasks/threads sharing one connection or event loop?) is exactly what we need to find the real root cause.

PR #4200, opened 2026-07-18
fix: add recursion depth limit to RESP2 and RESP3 parsers

mangodxd:
## Problem

A malicious or misconfigured Redis server can send deeply nested RESP responses (e.g., `*1\r\n` repeated thousands of times) that cause the pure-Python parsers to recurse without limit, crashing the client with a `RecursionError`.

This is a client-side denial-of-service vulnerability affecting any application using the redis-py library with an untrusted Redis server.

**Issue:** #4116

## Root Cause

The `_read_response()` method in `resp2.py` and `resp3.py` recursively parses aggregate types (arrays `*`, sets `~`, maps `%`, push `>`) without any depth guard. A server sending thousands of nested `*1\r\n` lines exhausts Python's call stack.

## Fix

Add a `_MAX_RESP_DEPTH = 512` constant and a `_depth` parameter to `_read_response()` that increments on each recursive call. When depth exceeds 512, an `InvalidResponse` exception is raised instead of allowing unbounded recursion.

### Files Changed

- `redis/_parsers/resp2.py` - Sync and async `_read_response()` (arrays)
- `redis/_parsers/resp3.py` - Sync and async `_read_response()` (arrays, sets, maps, push)
- `tests/test_resp_recursion_depth.py` - New test file covering all 4 parser variants

### Scope

All 4 pure-Python parsers are protected:

| Parser | Sync | Async |
|--------|------|-------|
| RESP2  | Yes  | Yes   |
| RESP3  | Yes  | Yes   |

The hiredis C parser is not affected (it has its own buffer management).

## Testing

All new tests pass. No regressions on existing unit tests. Full integration tests require a running Redis server (not available in CI environment).

```bash
pytest tests/test_resp_recursion_depth.py -v
# 7 passed
```

## Notes

- Depth limit of 512 is well above any legitimate Redis response nesting depth
- The fix is backward-compatible: no public API changes
- The `_depth` parameter is private (prefixed with underscore)

---

> [!NOTE]
> **Medium Risk**
> Touches core wire-protocol parsing used on every Redis read; wrong depth accounting could reject legitimate RESP3 push traffic or still allow stack exhaustion, though tests target push continuations explicitly.
> 
> **Overview**
> Adds a **`_MAX_RESP_DEPTH` (200)** guard on the pure-Python **RESP2 and RESP3** parsers (sync and async) so deeply nested aggregate replies raise **`InvalidResponse`** instead of blowing the Python stack when talking to a malicious or broken server ([#4116](https://github.com/redis/redis-py/issues/4116)).
> 
> **`_read_response`** now tracks a private **`_depth`** counter and increments it when recursing into arrays (`*`), and in RESP3 also sets (`~`), maps (`%`), and push payloads (`>`). **RESP3 push continuations** after handling a `>` reply re-enter parsing at the **same** depth so long bursts of top-level push notifications do not count as nesting.
> 
> New regression tests in **`tests/test_resp_recursion_depth.py`** cover deep nesting for all four parsers, shallow happy paths, and push-vs-nested behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e1f0ce7fc83fc40b51f11a22336aebd7a34d129. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/_parsers/resp2.py
+++ b/redis/_parsers/resp2.py
@@ -6,6 +6,10 @@ from ..utils import SENTINEL
 from .base import _AsyncRESPBase, _RESPBase
 from .socket import SERVER_CLOSED_CONNECTION_ERROR
 
+# Maximum nesting depth for RESP aggregate replies to prevent RecursionError
+# from malicious or malformed responses. 512 is well above any legitimate use.
+_MAX_RESP_DEPTH = 512
+
 
 class _RESP2Parser(_RESPBase):
     """RESP2 protocol implementation"""
@@ -27,8 +31,13 @@ class _RESP2Parser(_RESPBase):
             return result
 
     def _read_response(
-        self, disable_decoding=False, timeout: Union[float, object] = SENTINEL
+        self, disable_decoding=False, timeout: Union[float, object] = SENTINEL,
+        _depth: int = 0,
     ):
+        if _depth > _MAX_RESP_DEPTH:
+            raise InvalidResponse(
+                f"Protocol Error: RESP nesting depth exceeds {_MAX_RESP_DEPTH}"
+            )
         raw = self._buffer.readline(timeout=timeout)
         if not raw:
             raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
@@ -64,7 +73,10 @@ class _RESP2Parser(_RESPBase):
             return None
         elif byte == b"*":
             response = [
-                self._read_response(disable_decoding=disable_decoding, timeout=timeout)
+                self._read_response(
+                    disable_decoding=disable_decoding, timeout=timeout,
+                    _depth=_depth + 1,
+                )
                 for i in range(int(response))
             ]
         else:
@@ -92,8 +104,12 @@ class _AsyncRESP2Parser(_AsyncRESPBase):
         return response
 
     async def _read_response(
-        self, disable_decoding: bool = False
+        self, disable_decoding: bool = False, _depth: int = 0
     ) -> Union[EncodableT, ResponseError, None]:
+        if _depth > _MAX_RESP_DEPTH:
+            raise InvalidResponse(
+                f"Protocol Error: RESP nesting depth exceeds {_MAX_RESP_DEPTH}"
+            )
         raw = await self._readline()
         response: Any
         byte, response = raw[:1], raw[1:]
@@ -128,7 +144,9 @@ class _AsyncRESP2Parser(_AsyncRESPBase):
             return None
         elif byte == b"*":
             response = [
-                (await self._read_response(disable_decoding))
+                (await self._read_response(
+                    disable_decoding, _depth=_depth + 1
+                ))
                 for _ in range(int(response))  # noqa
             ]
         else:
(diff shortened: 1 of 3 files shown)


redis/_parsers/resp2.py
```diff
@@ -6,6 +6,10 @@
 from .base import _AsyncRESPBase, _RESPBase
 from .socket import SERVER_CLOSED_CONNECTION_ERROR
 
+# Maximum nesting depth for RESP aggregate replies to prevent RecursionError
+# from malicious or malformed responses. 512 is well above any legitimate use.
+_MAX_RESP_DEPTH = 512
```
Sanjays2402: On Python 3.10/3.11 this guard still loses to the interpreter: every nested array adds an `_read_response` frame plus a list-comprehension frame, so the default recursion limit is hit around depth 500 before `_depth` reaches 513. The cap needs to be lower than that; the new depth-600 tests should currently raise a raw `RecursionError` on those versions.

redis/_parsers/resp2.py
```diff
@@ -6,6 +6,10 @@
 from .base import _AsyncRESPBase, _RESPBase
 from .socket import SERVER_CLOSED_CONNECTION_ERROR
 
+# Maximum nesting depth for RESP aggregate replies to prevent RecursionError
```
mangodxd, line 9: Here is the frame breakdown after measuring:

- **Python 3.10/3.11**: each nested array adds 2 frames (list-comprehension frame + recursive _read_response frame), so rames = 2*N + 1. At depth 512 that is 1025 frames, over the default recursion limit of 1000.
- **Python 3.12+**: PEP 709 inlines list comprehensions (no extra frame), so rames = N + 2. Depth 512 is safe here.

Since redis-py supports >=3.10, the cap must account for the 3.10/3.11 frame cost. Lowered _MAX_RESP_DEPTH from 512 to **200**:

- 3.10/3.11: 403 frames at trigger (margin 597)
- 3.12+: 203 frames at trigger (margin 797)
- Still far above any legitimate Redis nesting (real-world max ~10-20)

Updated tests from depth=600 to depth=250 (just above the new cap). All 7 tests pass.

redis/_parsers/resp3.py
```diff
@@ -153,6 +174,7 @@ def _read_response(
             return self._read_response(
                 disable_decoding=disable_decoding,
                 push_request=push_request,
+                _depth=_depth + 1,
```
Sanjays2402: This continuation is not nested inside the push; it is the next top-level reply. A burst of 201 push notifications before the command response will now trip the depth guard, so this call should preserve `_depth` rather than increment it (same in async).

mangodxd:
Fixed. Push notifications are top-level replies, not nested. The continuation call now preserves _depth instead of incrementing it. Added tests for burst push notifications (must not raise) and push followed by deeply nested array (must still raise).

petyaslavova:
Hey @mangodxd, thanks for the fix, and especially for the push-continuation handling and its test — that scenario is worth keeping. The underlying issue (#4116) is valid and a depth guard is the right approach.

We have a second PR for the same issue, #4144, using the same depth-guard design but with regression tests that drive the reporter's exact payload through the real SocketBuffer path (including a 5000-level case that must raise InvalidResponse rather than RecursionError). We're going to consolidate on that PR to avoid two overlapping guards, and carry over your push-notification-burst test idea there.

I'm going to close this one as superseded. Two notes for reference: the description still mentions a 512 limit while the code uses 200, and please run `ruff format` — the multi-line argument lists likely won't pass the formatter. Thanks again for the contribution.

PR #4201, opened 2026-07-19
Decode ACL LOG string values on the default RESP3 legacy callback

chuenchen309:
`parse_acl_log_resp3_to_resp2_legacy` decodes the ACL LOG entry's map **keys** but leaves the string **values** as bytes, so the default client returns bytes where the RESP2 client returns `str`:

```python
Redis(protocol=2).acl_log()   # {'reason': 'auth', 'context': 'toplevel', 'object': 'AUTH', 'username': 'someuser', ...}
Redis().acl_log()             # {'reason': b'auth', 'context': b'toplevel', 'object': b'AUTH', 'username': b'someuser', ...}
```

`Redis()` defaults to the RESP3 wire with `legacy_responses=True` (`get_response_callbacks(None, True)`), and the server sends each ACL LOG entry as a map on RESP3, so `parse_acl_log_resp3_to_resp2_legacy` always runs. Its documented contract (the legacy/unified split added in #4052) is to match the RESP2 legacy shape — but four bulk-string fields (`reason`, `context`, `object`, `username`) come back as `bytes`.

**Root cause** (`redis/_parsers/helpers.py`): the map branch does `{str_if_bytes(k): v for ...}` — keys only. The base RESP2 callback uses `pairs_to_dict(log, True, True)` (decode keys **and** values) and the sibling `parse_acl_log_resp3_unified` also decodes values; this branch is the one that doesn't.

**Fix** (1 line): decode the values too, mirroring the siblings. Non-string values (the integer `count`/`entry-id`/timestamps) pass through `str_if_bytes` unchanged; `age-seconds`/`client-info` are overwritten afterward exactly as before.

```python
-            log_data = {str_if_bytes(k): v for k, v in log.items()}
+            log_data = {str_if_bytes(k): str_if_bytes(v) for k, v in log.items()}
```

**Verification** (re-runnable from the diff): added `test_parse_acl_log_resp3_legacy_decodes_string_values` in `tests/test_parsers/test_helpers.py` (pure unit test, no server) — it builds the RESP3 map entry, asserts the four fields decode to `str`, and asserts the result equals `parse_acl_log` on the equivalent RESP2 flat list. It fails on the current tree (`assert b'auth' == 'auth'`) and passes with the fix. `pytest tests/test_parsers/` → 10 passed; `ruff check`/`ruff format --check` clean. `helpers.py` is shared by the sync and async stacks, so the single change covers both.

---

*Disclosure*: This PR was authored by an AI coding agent (Claude Code) running on this account: the AI found the bug, ran the repro, wrote the test, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.

---

> [!NOTE]
> **Low Risk**
> Single-line parser fix in shared helpers with a focused unit test; no auth or wire-protocol behavior beyond ACL LOG response shaping.
> 
> **Overview**
> Fixes a **RESP3 + `legacy_responses=True`** mismatch for **`ACL LOG`**: map entries were decoded on keys only, so fields like `reason`, `context`, `object`, and `username` stayed as `bytes` instead of matching RESP2’s `str` shape.
> 
> `parse_acl_log_resp3_to_resp2_legacy` now runs **`str_if_bytes` on values** in the RESP3 map branch (same idea as `pairs_to_dict(..., decode_string_values=True)` on RESP2). Post-processing for `client-info` and `age-seconds` is unchanged.
> 
> Adds **`test_parse_acl_log_resp3_legacy_decodes_string_values`** to assert those fields are `str` and that the parsed entry equals `parse_acl_log` on the equivalent RESP2 wire list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed5f5af9f8f21a89056422e4e4bf429988b0315d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/_parsers/helpers.py
+++ b/redis/_parsers/helpers.py
@@ -1341,7 +1341,7 @@ def parse_acl_log_resp3_to_resp2_legacy(response, **options):
     data = []
     for log in response:
         if isinstance(log, dict):
-            log_data = {str_if_bytes(k): v for k, v in log.items()}
+            log_data = {str_if_bytes(k): str_if_bytes(v) for k, v in log.items()}
         else:
             log_data = pairs_to_dict(log, True, True)
         client_info = log_data.get("client-info", "")
--- a/tests/test_parsers/test_helpers.py
+++ b/tests/test_parsers/test_helpers.py
@@ -1,6 +1,8 @@
 import pytest
 
 from redis._parsers.helpers import (
+    parse_acl_log,
+    parse_acl_log_resp3_to_resp2_legacy,
     parse_client_list,
     parse_command,
     parse_info,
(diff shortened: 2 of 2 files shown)


petyaslavova:
Hey @chuenchen309, thank you for your contribution!

I'm curious about the disclaimer 🙂 When you say the AI found the bug, how did it happen? Was it intentionally prompted to look for this kind of issue, or did it discover it while troubleshooting something else related to this command?

I'm asking because if there's a good prompt or approach for identifying these kinds of discrepancies in callbacks, it would be really helpful to use it to check for similar issues elsewhere.

petyaslavova (approved):
LGTM.

chuenchen309:
Happy to explain, and sorry for the slow reply.

It was a deliberate hunt for that class of bug rather than something I ran into while troubleshooting — I'm not a heavy Redis user. `redis/_parsers/` was the target because its callbacks can be driven with raw bytes in a unit test, so a whole class of bug is verifiable without a live server.

The heuristic is: **find places where the same logical response is produced by more than one function, then look for asymmetry in the normalisation steps.** If two siblings are supposed to produce the same shape and don't, at least one is wrong — and unlike a style difference it usually has a decidable answer, so it's cheap to adjudicate and hard to argue with.

`redis-py` has an unusually clean instance of that shape. `_parsers/response_callbacks.py` keeps six command→callback tables, and `_RedisCallbacksRESP3toRESP2Legacy` exists precisely to reproduce what the RESP2 table produces — so every entry in it is a standing claim that two functions agree. `ACL LOG` was where they didn't: the legacy path did `{str_if_bytes(k): v for ...}`, while `parse_acl_log` used `pairs_to_dict(..., decode_string_values=True)` and the sibling `parse_acl_log_resp3_unified` also decoded values. Two of the three decoded, so the odd one out read as a slip rather than a decision — and #4052 had already made the RESP2-compatible shape a documented contract, which is what turns "inconsistent" into "wrong".

If you want to sweep the rest, that table is the surface. It currently holds 30 commands; 18 of them have a genuinely different callback on each side, which is where the risk lives:

`ACL GETUSER`, `ACL LOG`, `CLIENT TRACKINGINFO`, `CLUSTER LINKS`, `CONFIG GET`, `FUNCTION LIST`, `GEOPOS`, `HRANDFIELD`, `LCS`, `MEMORY STATS`, `SENTINEL MASTER`, `SENTINEL MASTERS`, `SENTINEL SENTINELS`, `SENTINEL SLAVES`, `XINFO CONSUMERS`, `XINFO GROUPS`, `ZDIFF`, `ZRANDMEMBER`

(The other 12 share `str_if_bytes` or an identical lambda on both sides, so there's nothing to disagree about.)

I've only actually checked `ACL LOG` — I'm not claiming any of the others are broken. But each one reduces to a single mechanical test: push the RESP3 shape through the legacy callback, push the RESP2 shape through the RESP2 callback, assert the results are equal. No server needed, and it's uniform enough to do in bulk.

*AI-assisted, and the disclaimer is doing real work here: an agent did the hunt above, and the pairing came out of reading the callback tables rather than from experience with the protocol.*

PR #4204, opened 2026-07-22
fix: honor read_from_replicas for sharded pubsub subscriptions in ClusterPubSub

### LINKED ISSUE #3266: Ssubscribe not honoring `read_from_replicas` flag
Thanks for wanting to report an issue you've found in redis-py. Please delete this text and fill in the template below.  
It is of course not always possible to reduce your code to a small test case, but it's highly appreciated to have as much data as possible. Thank you!

**Version**: 5.0.4

**Platform**: Python 3.9 on Macbook Pro running Sonoma 14.5 (23F79) (applies to other platforms as well)

**Description**: Sharded subscribe is not honoring the `read_from_replicas` flag for the [`class RedisCluster`](https://redis-py.readthedocs.io/en/stable/_modules/redis/cluster.html#RedisCluster). In particular, it's calling:

```
node = self.cluster.get_node_from_key(s_channel)
```

which has the following signature:

```
    def get_node_from_key(self, key, replica=False):
        """
        Get the node that holds the key's slot.
        If replica set to True but the slot doesn't have any replicas, None is
        returned.
        """
```

As such, only primary replicas are used for sharded subscribe. It'd be great if `ssubscribe` behaves in the same way as `subscribe` to be consistent with the definition of the `read_from_replicas` flag. Thanks!

Brumbelow:
### Description of change

Fixes #3266

`ClusterPubSub` routes every sharded subscription to the slot's primary: `ssubscribe`, the `sunsubscribe` fallback, and `reinitialize_shard_subscriptions` resolve nodes via `get_node_from_key()`, which ignores `read_from_replicas` and `load_balancing_strategy` — unlike regular pub/sub node selection in `ClusterPubSub.execute_command`, which honors both.

Shard-channel resolution now goes through the same `nodes_manager.get_node_from_slot(slot, read_from_replicas, load_balancing_strategy)` in the sync and async clients. Since a shard subscription may now live on a replica, the keep-vs-migrate decisions (slot-migration reconciliation and the lazy re-route in `ssubscribe`) check whether the tracked node still serves the channel's slot instead of comparing against a fresh pick, so subscriptions don't bounce between shard siblings on topology refreshes.

With the default configuration (`read_from_replicas=False`, no load balancing strategy) behavior is unchanged: subscriptions resolve to and strictly follow the slot primary, and `SlotNotCoveredError` is still raised for uncovered slots.

### Pull Request check-list

- [N/A] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [N/A] Is there an example added to the examples folder (if applicable)?

---

> [!NOTE]
> **Medium Risk**
> Touches cluster pub/sub routing and migration/reconcile paths in both sync and async clients; behavior changes only when replica reads or load balancing is enabled, but mistakes could drop or duplicate shard subscriptions during failover.
> 
> **Overview**
> Fixes **#3266** by aligning **shard pub/sub** node selection in sync and async `ClusterPubSub` with regular cluster routing: `ssubscribe`, `sunsubscribe` fallback, and `reinitialize_shard_subscriptions` now use `nodes_manager.get_node_from_slot(slot, read_from_replicas, load_balancing_strategy)` instead of `get_node_from_key()` (primary-only).
> 
> A new **`_eligible_subscription_nodes`** helper defines which nodes may hold a subscription for a slot (replicas included only when replica reads or load balancing is enabled). **Keep-vs-migrate** logic no longer compares the tracked node to a fresh balancer pick; it skips migration when the tracked node still serves the slot, avoiding churn between primary and replica on re-subscribe or topology refresh. **`sunsubscribe`** without a reverse-index entry scans eligible slot nodes for the pubsub that actually holds the channel.
> 
> Default config (`read_from_replicas=False`) behavior is unchanged. Tests are updated for the new resolution path and extended for replica routing, sibling stability, and unsubscribe fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c2f60eef1f34b233cf60cc28a84997cccbcedb21. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -3742,6 +3742,19 @@ class ClusterPubSub(PubSub):
                 return None
         return message
 
+    def _eligible_subscription_nodes(self, slot: int) -> Dict[str, "ClusterNode"]:
+        """
+        Nodes a shard subscription may live on for ``slot``, keyed by node
+        name. Mirrors regular pubsub node selection: replicas are eligible
+        only when replica reads are enabled.
+        """
+        nodes = self.cluster.nodes_manager.slots_cache.get(slot) or []
+        if not (
+            self.cluster.read_from_replicas or self.cluster.load_balancing_strategy
+        ):
+            nodes = nodes[:1]
+        return {node.name: node for node in nodes}
+
     async def ssubscribe(
         self, *args: ChannelT | Subscription, **kwargs: PubSubHandler
     ) -> None:
@@ -3759,25 +3772,38 @@ class ClusterPubSub(PubSub):
         # re-acquire this lock (asyncio.Lock is non-reentrant).
         async with self._shard_state_lock:
             for s_channel, handler in s_channels.items():
-                node = self.cluster.get_node_from_key(s_channel)
+                slot = self.cluster.keyslot(s_channel)
+                node = self.cluster.nodes_manager.get_node_from_slot(
+                    slot,
+                    self.cluster.read_from_replicas,
+                    self.cluster.load_balancing_strategy,
+                )
                 if not node:
                     continue
                 # Lazy re-route: if this channel is already tracked against a
-                # different node (e.g. after a slot migration), migrate it now
-                # so the caller's intent is applied on the current owner.
+                # different node, migrate it now so the caller's intent is
+                # applied on the current owner — unless the tracked node
+                # still serves the slot, in which case leave it alone.
                 normalized_key = next(iter(self._normalize_keys({s_channel: None})))
                 old_name = self._shard_channel_to_node.get(normalized_key)
                 if old_name and old_name != node.name:
-                    # Match PubSub.ssubscribe() dict.update() semantics: the
-                    # caller's newly supplied handler (including None) always
-                    # overrides any previously registered handler.
-                    await self._migrate_shard_channel(
-                        normalized_key,
-                        handler,
-                        old_name,
-                        node,
-                    )
-                    continue
+                    tracked_node = self._eligible_subscription_nodes(slot).get(old_name)
+                    if tracked_node is not None:
+                        # The tracked node still serves the channel's slot
+                        # (the balancer merely picked a sibling); keep the
+                        # subscription where it is to avoid churn.
+                        node = tracked_node
+                    else:
+                        # Match PubSub.ssubscribe() dict.update() semantics: the
+                        # caller's newly supplied handler (including None) always
+                        # overrides any previously registered handler.
+                        await self._migrate_shard_channel(
+                            normalized_key,
+                            handler,
+                            old_name,
+                            node,
+                        )
+                        continue
                 pubsub = self._get_node_pubsub(node)
                 if handler:
                     await pubsub.ssubscribe(Subscription(s_channel, handler))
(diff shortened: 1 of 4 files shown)


redis/cluster.py
```diff
@@ -3131,10 +3157,21 @@ def sunsubscribe(self, *args):
...
+                    # the subscription for this slot (with replica reads
+                    # enabled that is not necessarily the primary).
+                    slot = self.cluster.keyslot(s_channel)
+                    eligible = self._eligible_subscription_nodes(slot)
+                    if not eligible:
+                        raise SlotNotCoveredError(
+                            f'Slot "{slot}" is not covered by the cluster.'
+                        )
+                    name = next(
+                        (n for n in eligible if n in self.node_pubsub_mapping), None
```
Sanjays2402: If the reverse index is missing and both primary and replica already have pubsub objects, this picks the first mapped node even if that node’s `shard_channels` does not contain `s_channel`. The unsubscribe can go to the primary while the live subscription remains on a replica; should this fallback select by membership in each pubsub’s `shard_channels` instead?

Brumbelow: Good catch — with both pubsubs mapped the first eligible node won regardless of who held the subscription. Changed the fallback to select by membership in each pubsub's `shard_channels` and to skip when no mapped pubsub holds the channel, rather than sending a no-op SUNSUBSCRIBE to the primary. Covered in `test_sunsubscribe_fallback_considers_all_slot_nodes` (sync and async).

redis/asyncio/cluster.py
```diff
@@ -4006,18 +4048,28 @@ async def sunsubscribe(self, *args: Any) -> None:
...
             for channel, handler in list(self.shard_channels.items()):
+                slot = self.cluster.keyslot(channel)
+                old_name = self._shard_channel_to_node.get(channel)
+                if old_name and old_name in self._eligible_subscription_nodes(slot):
+                    # The tracked node still serves the channel's slot under
+                    # the current routing configuration; nothing to do.
+                    continue
                 try:
-                    new_node = self.cluster.get_node_from_key(channel)
+                    new_node = self.cluster.nodes_manager.get_node_from_slot(
```
Sanjays2402, line 4106: the async `get_node_from_slot` only catches `(IndexError, TypeError)`, so an uncovered slot comes out of `self.slots_cache[slot]` as a bare KeyError, not `SlotNotCoveredError` — the `except SlotNotCoveredError` deferral below never fires and reconciliation aborts for the remaining channels. `get_node_from_key` used `slots_cache.get(slot)` and raised properly, so this is a behavior change on the async side only (sync `get_node_from_slot` checks `.get(slot) is None` first). Same applies to the new `_eligible_subscription_nodes` callers in `ssubscribe`.

Brumbelow: Thanks @Sanjays2402. I've noted the comments. Since final approval and direction rest with the maintainers, I’m going to wait for their review before making further revisions.

petyaslavova:
Hey @Brumbelow, thank you for your contribution!

You're right that the server side allows this — Redis forwards shard messages to all nodes in the shard, so a client may subscribe to a shard channel on the slot's primary or on any of its replicas, and we already send `READONLY` on node connections whenever `read_from_replicas`/`load_balancing_strategy` is set.

Before we take this toward merge there is a design question to settle on our side: those two options are documented as read-command routing knobs (and `read_from_replicas` is deprecated), so folding shard-subscription placement into them silently changes the delivery path for existing users — a replica that is reachable by the client but partitioned from its primary stops receiving shard messages with nothing to notice it. We'd also want placement sticky per shard rather than round-robin per channel, so channels in one slot don't spread across siblings and enlarge the polled pubsub set.

Two points are blockers regardless of that decision:

- Async `NodesManager.get_node_from_slot` only converts `IndexError`/`TypeError`, so an uncovered slot now surfaces as a bare `KeyError`. The `except SlotNotCoveredError` deferral in `reinitialize_shard_subscriptions` no longer fires and a whole reconcile pass aborts, and `ssubscribe` no longer matches the sync error type. The updated unit test hides this because the mock raises `SlotNotCoveredError` itself. Bugbot flagged this one as well.
- The new coverage is mock-only, so nothing proves the actual behavior this PR relies on. Our CI cluster runs with one replica per shard and we already have `onlycluster` shard-pubsub tests — please add an end-to-end test that subscribes with replica routing enabled and receives an `SPUBLISH` published to the shard.

Please keep this as a draft in the meantime. I'll follow up on #3266 for the operational scenario behind the request, which is what decides whether we do this at all and whether it should be an explicit opt-in.

Brumbelow:
Thanks @petyaslavova , both blockers are pushed and addressed. 

- Async NodesManager.get_node_from_slot now raises SlotNotCoveredError when the slot is missing from slots_cache, matching the sync nodes manager, so the deferral in reinitialize_shard_subscriptions fires again and ssubscribe raises the same error type as sync.
- Added end-to-end coverage (TestClusterPubSubReplicaRouting, sync and async): with read_from_replicas=True, two channels in the same slot are subscribed so one lands on the shard's replica, then both receive a PUBLISH

Will keep as draft until you all tell me otherwise. Let me know if I can help!

PR #4207, opened 2026-07-25
fix(asyncio): fully disconnect removed cluster nodes in NodesManager.set_nodes

### LINKED ISSUE #4053: Async cluster: removed-node cleanup in NodesManager.set_nodes leaks in-use connections (extends #3572)
## Summary

`NodesManager.set_nodes(remove_old=True)` (introduced in #3917, released in v7.2.0) cleans up nodes that are removed from the cluster topology by calling `disconnect_free_connections()` on them and marking active connections via `update_active_connections_for_reconnect()`. The marking step is a no-op for fully removed nodes: `disconnect_if_needed()` only fires on `acquire_connection()`, but since the node has just been removed from `nodes_cache`, no future acquisition will ever happen on it. Active connections are released back into `_free` (just appended — `release()` does not check the flag) with `is_connected=True` until GC reclaims the `ClusterNode`, at which point `ClusterNode.__del__` logs `Unclosed ClusterNode object`.

#3572 is still open and tracks the broader symptom; #3917 partially addressed it for the *node-preserved* case but did not address the *node-removed* case. Filing this as a focused follow-up so it can be closed precisely.

Tested against `redis==7.4.0`.

## Affected code

```python
# redis/asyncio/cluster.py v7.4.0, NodesManager.set_nodes (lines ~1647-1668)
if remove_old:
    for name in list(old.keys()):
        if name not in new:
            removed_node = old.pop(name)
            removed_node.update_active_connections_for_reconnect()  # only flags in-use conns
            task = asyncio.create_task(
                removed_node.disconnect_free_connections()           # only disconnects idle conns
            )
            self._background_tasks.add(task)
            task.add_done_callback(self._background_tasks.discard)
```

`update_active_connections_for_reconnect` calls `connection.mark_for_reconnect()`. The mark is honored by `disconnect_if_needed`, which is only called from `acquire_connection`. A removed node will never see another `acquire_connection`, so flagged connections are released back into `_free` after their command finishes (`release()` just appends — no flag check) and stay connected indefinitely.

## Reproducer

```python
import asyncio
from redis.asyncio.cluster import ClusterNode, NodesManager

class FakeConn:
    def __init__(self):
        self.is_connected = True
        self.disconnect_called = False
        self._reconnect = False

    def mark_for_reconnect(self):
        self._reconnect = True

    def should_reconnect(self):
        return self._reconnect

    async def disconnect(self):
        self.disconnect_called = True
        self.is_connected = False

async def main():
    nm = NodesManager(
        startup_nodes=[], require_full_coverage=False, connection_kwargs={}
    )
    node = ClusterNode("10.0.0.1", 6379, "primary")

    free_conn = FakeConn()
    in_use_conn = FakeConn()
    node._connections = [free_conn, in_use_conn]
    node._free.append(free_conn)  # only free_conn is idle; in_use_conn is "checked out"

    nm.nodes_cache = {node.name: node}

    nm.set_nodes(nm.nodes_cache, {}, remove_old=True)

    # Wait for the disconnect_free_connections background task
    await asyncio.gather(*nm._background_tasks)

    print(f"free_conn.disconnect_called   = {free_conn.disconnect_called}")
    print(f"in_use_conn.is_connected      = {in_use_conn.is_connected}")
    print(f"in_use_conn._reconnect        = {in_use_conn._reconnect}")
    print(f"in_use_conn.disconnect_called = {in_use_conn.disconnect_called}")

asyncio.run(main())
```

Output on `redis==7.4.0`:

```
Unclosed ClusterNode object
client: [host=10.0.0.1, port=6379, name=10.0.0.1:6379, server_type=primary]
free_conn.disconnect_called   = True
in_use_conn.is_connected      = True
in_use_conn._reconnect        = True
in_use_conn.disconnect_called = False
```

The in-use connection is flagged but never disconnected. The `Unclosed ClusterNode object` warning fires from `ClusterNode.__del__` when the orphaned node is GC'd.

## Expected behavior

When a node is fully removed from the cluster topology, all of its connections (free *and* in-use) should be disconnected before the `ClusterNode` becomes eligible for GC. In-flight commands on those connections will surface as `ConnectionError`, which the cluster retry layer at `_execute_command` already handles by reinitializing topology and retrying.

## Suggested fix

In the `if name not in new` branch of `set_nodes`, call full `disconnect()` instead of `disconnect_free_connections()`:

```python
if remove_old:
    for name in list(old.keys()):
        if name not in new:
            removed_node = old.pop(name)
            task = asyncio.create_task(removed_node.disconnect())
            self._background_tasks.add(task)
            task.add_done_callback(self._background_tasks.discard)
```

The node-preserved branch (slot rebalance, where the same `ClusterNode` stays in cache) should keep its current `mark_for_reconnect` behavior — that's correct because future acquisitions will run `disconnect_if_needed`.

## Related

- #3572 — original report, still open
- #3917 — partial fix in v7.2.0, addressed the slot-rebalance case
- valkey-io/valkey-py#283 — the fork's fix (different approach, awaits disconnects fully)

eeshsaxena:
When removed from the cluster topology, old nodes only had their free (idle) connections cleaned up. In-use connections were left open indefinitely - the node is gone from nodes_cache, so acquire_connection() is never called again on it, meaning those sockets never get closed. This causes an "Unclosed ClusterNode object" warning and a real connection leak in long-running clients.

The fix calls disconnect() on each removed node instead of only disconnect_free_connections(). This closes all connections on removal, including ones still marked as in-use.

Note: in-flight commands on a removed node will fail with a connection error on this path, but that was already the case on reconnect attempts. The trade-off is correct - a node that left the topology should not keep sockets alive.

Test added in test_cluster.py verifies that both free and in-use connections are disconnected when a node is removed from the topology.

Fixes #4053

---

> [!NOTE]
> **Medium Risk**
> Changes connection teardown during topology refresh; in-flight work on removed nodes fails immediately instead of finishing, but only affects nodes already dropped from the cluster.
> 
> **Overview**
> Fixes a **connection leak** when asyncio cluster topology refresh drops a node: removed nodes were only closing idle connections while in-use sockets stayed open indefinitely (triggering "Unclosed ClusterNode object" warnings in long-lived clients).
> 
> **`NodesManager.set_nodes`** now schedules **`ClusterNode.disconnect()`** for each removed node instead of marking active connections for lazy reconnect and only disconnecting free ones. All pooled connections close immediately when a node leaves the cache.
> 
> The cluster test was renamed and updated to assert both free and in-use connections are disconnected on removal; in-flight commands on a removed node may fail with a connection error rather than completing after a deferred disconnect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aecbaf7fae1688734a3e2f1f00be746c9f30105a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1930,16 +1930,11 @@ class NodesManager:
             for name in list(old.keys()):
                 if name not in new:
                     # Node is removed from cache before disconnect starts,
-                    # so it won't be found in lookups during disconnect
-                    # Mark active connections so in-flight commands can
-                    # finish, then disconnect them when their current
-                    # operation completes. Free connections can be
-                    # disconnected immediately.
+                    # so it won't be found in lookups during disconnect.
+                    # We fully disconnect the node so that all its connections
+                    # (in-use and free) are closed immediately, preventing leaks.
                     removed_node = old.pop(name)
-                    removed_node.update_active_connections_for_reconnect()
-                    task = asyncio.create_task(
-                        removed_node.disconnect_free_connections()
-                    )
+                    task = asyncio.create_task(removed_node.disconnect())
                     self._background_tasks.add(task)
                     task.add_done_callback(self._background_tasks.discard)
 
--- a/tests/test_asyncio/test_cluster.py
+++ b/tests/test_asyncio/test_cluster.py
@@ -3140,11 +3140,12 @@ class TestNodesManager:
         assert nodes_cache_names == [node2.name, node1.name, node3.name]
 
     @pytest.mark.fixed_client
-    async def test_set_nodes_disconnects_removed_node_connections_after_use(
+    async def test_set_nodes_disconnects_removed_node_connections_immediately(
         self,
+        default_host: str,
     ) -> None:
         """
-        Test removed nodes let in-flight commands finish, then disconnect.
+        Test removed nodes are fully disconnected immediately, including in-use connections.
         """
 
         class FakeConnection:
@@ -3201,18 +3202,15 @@ class TestNodesManager:
         assert nodes_manager.nodes_cache == {}
         assert free_conn.disconnect_called is True
         assert free_conn.is_connected is False
-        assert active_conn.reconnect is True
-        assert active_conn.disconnect_called is False
-        assert active_conn.is_connected is True
-
-        active_conn.response_ready.set()
-        assert await command == b"OK"
-
         assert active_conn.disconnect_called is True
         assert active_conn.is_connected is False
-        assert active_conn in node._free
-        assert free_conn.reconnect is False
-        assert active_conn.reconnect is False
+
+        # Cancel the hanging command task since FakeConnection.disconnect doesn't interrupt read_response
+        command.cancel()
+        try:
+            await command
+        except asyncio.CancelledError:
+            pass
 
     @pytest.mark.fixed_client
     async def test_move_node_to_end_of_cached_nodes_nonexistent(self) -> None:
--- a/tests/test_asyncio/test_cluster.py
+++ b/tests/test_asyncio/test_cluster.py
@@ -3142,7 +3142,6 @@ class TestNodesManager:
     @pytest.mark.fixed_client
     async def test_set_nodes_disconnects_removed_node_connections_immediately(
         self,
-        default_host: str,
     ) -> None:
         """
         Test removed nodes are fully disconnected immediately, including in-use connections.


redis/asyncio/cluster.py
```diff
@@ -1930,16 +1930,11 @@ def set_nodes(
...
-                    # disconnected immediately.
+                    # so it won't be found in lookups during disconnect.
+                    # We fully disconnect the node so that all its connections
+                    # (in-use and free) are closed immediately, preventing leaks.
                     removed_node = old.pop(name)
-                    removed_node.update_active_connections_for_reconnect()
-                    task = asyncio.create_task(
-                        removed_node.disconnect_free_connections()
-                    )
+                    task = asyncio.create_task(removed_node.disconnect())
```
Sanjays2402, line 1937: ClusterNode.disconnect() gathers with return_exceptions=True and then re-raises the first exception, unlike disconnect_free_connections() which swallowed them. Since this is a fire-and-forget task whose only done_callback is discard, a node that goes away mid-topology-change (socket close raising) now turns into an unretrieved "Task exception was never retrieved" on the loop instead of a quiet cleanup. Probably worth wrapping the create_task target in a small helper that logs instead of raising.

eeshsaxena:
Pushed a fix for the red asyncio jobs. The new test took `default_host` as a parameter, but in `tests/test_asyncio/test_cluster.py` that is a module-level constant rather than a fixture (unlike the sync suite, where `tests/conftest.py` does define a fixture by that name), so pytest errored at setup with `fixture 'default_host' not found`. Dropped the parameter - the constant resolves fine. Test passes locally.

redis/asyncio/cluster.py
Sanjays2402, line 1937: ClusterNode.disconnect() re-raises the first exception out of its gather, disconnect_free_connections() swallowed them (return_exceptions=True, no re-raise). This is a bare create_task whose only done_callback is the set discard, so nothing ever retrieves the result — a socket that errors on close now surfaces as "Task exception was never retrieved" on the loop instead of being ignored like before. Probably want to wrap it in a coroutine that logs and drops.

redis/asyncio/cluster.py
```diff
@@ -1930,16 +1930,11 @@ def set_nodes(
...
                 if name not in new:
                     # Node is removed from cache before disconnect starts,
-                    # so it won't be found in lookups during disconnect
-                    # Mark active connections so in-flight commands can
-                    # finish, then disconnect them when their current
-                    # operation completes. Free connections can be
-                    # disconnected immediately.
+                    # so it won't be found in lookups during disconnect.
+                    # We fully disconnect the node so that all its connections
+                    # (in-use and free) are closed immediately, preventing leaks.
```
petyaslavova, line 1935: Hey @eeshsaxena, the trade-off is not correct :) This was intentional, and it is related to maintenance notifications handling- in-use connections should have a chance to complete the in-flight commands. With enterprise server maintenance events those commands still have a few seconds within which they can complete successfully and we don't want to fail them. 
After a command fails or succeeds, it should try to release the connection back - this is where we should handle those connections states.

petyaslavova (changes requested):

eeshsaxena:
Thanks @petyaslavova, that makes sense — I hadn't accounted for the maintenance-notification path where in-use connections need those few seconds to finish in-flight commands, and releasing on command completion is the right place to handle their state. My change to eagerly `disconnect()` removed nodes would break that intentional behavior (and, as the bot noted, re-raise out of the fire-and-forget task). The original `update_active_connections_for_reconnect()` + `disconnect_free_connections()` approach is correct, so I'll close this. Sorry for the noise, and thanks for the clear explanation!

petyaslavova:
Hey @eeshsaxena, if you are still willing to dig into this problem, a new PR with a fix focused on proper connection release will be very welcome, and I'll be happy to review it :) 
One more thing that will be really helpful for the fix evaluation is a good code example of how this can be reproduced...

PR #4210, opened 2026-07-26
fix(asyncio): keep pipeline packing off the event loop

### LINKED ISSUE #3416: Pipeline `execute` in async blocks as it packs the commands
**Version**:
redis-py: 5.11

**Platform**:
Python 3.12/Ubuntu 24.04

**Description**:
Executing a pipeline of many commands in asyncio blocks whilst it packs the commands
https://github.com/redis/redis-py/blob/d6ddb0d66bc4bd28c99d9cf148c9085117cc8ce9/redis/asyncio/client.py#L1365

I'm not really sure how to build a repl to demonstrate the issue, but we have a case where we are sending many (200k) commands to a pipeline as a transaction, it blocks for several seconds.

lh0156:
## Summary

Large async pipelines currently call `Connection.pack_commands()` synchronously from the event loop. Packing a large transaction or pipeline can therefore block unrelated coroutines for seconds.

This change:

- offloads command packing for transactional and non-transactional async pipelines with `asyncio.to_thread`
- preserves the existing wire packing and response handling
- adds a deterministic regression test covering both pipeline paths

Fixes #3416.

## Validation

- TDD regression test was red before the production change and green after it
- `pytest -q tests/test_asyncio/test_pipeline.py -k 'pipeline_packing_does_not_block_event_loop or TestAsyncPipelineOperationDurationMetricsRecording'` — 7 passed
- RESP3 async pipeline suite — 71 passed, 2 skipped
- RESP2 async pipeline suite — 71 passed, 2 skipped
- `invoke linters` — passed

The full Docker/fixed-client matrix was not completed because the local Docker credential helper and required cluster/maintenance services were unavailable.

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -1928,8 +1928,9 @@ class Pipeline(Redis):  # lgtm [py/init-calls-subclass]
         pre: CommandT = (("MULTI",), {})
         post: CommandT = (("EXEC",), {})
         cmds = (pre, *commands, post)
-        all_cmds = connection.pack_commands(
-            args for args, options in cmds if EMPTY_RESPONSE not in options
+        all_cmds = await self._pack_commands(
+            connection,
+            (args for args, options in cmds if EMPTY_RESPONSE not in options),
         )
         await connection.send_packed_command(all_cmds)
         errors = []
@@ -2004,7 +2005,7 @@ class Pipeline(Redis):  # lgtm [py/init-calls-subclass]
         self, connection: Connection, commands: CommandStackT, raise_on_error: bool
     ):
         # build up all commands into a single request to increase network perf
-        all_cmds = connection.pack_commands([args for args, _ in commands])
+        all_cmds = await self._pack_commands(connection, [args for args, _ in commands])
         await connection.send_packed_command(all_cmds)
 
         response = []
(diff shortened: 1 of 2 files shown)


petyaslavova (changes requested):
Thanks for tackling this — the underlying issue in #3416 is real: for large async pipelines/transactions, pack_commands runs synchronously on the event loop and can block it for seconds, and moving packing off the loop is a reasonable direction.

There is, though, one design point to settle: the offload is applied unconditionally, so every async pipeline — including small ones — now pays a `to_thread` dispatch and uses the loop's shared default executor. 
It will be better to gate the offload behind a batch-size threshold (so the common small-pipeline path keeps its current cost) or share a quick benchmark showing the small-pipeline overhead is negligible.

It would also help to note whether async cluster pipelines (redis/asyncio/cluster.py, which pack the same way) are in scope here or a follow-up.

Once that's decided and the PR is out of draft, it should be ready for another review.

eeshsaxena (commented):
Moving `pack_commands` onto a thread via `asyncio.to_thread` looks correct. The generator built in `_execute_transaction` is only ever consumed inside the worker thread, and nothing on the event loop touches it concurrently, so there is no shared-iterator hazard. `pack_commands` runs against the per-connection encoder, whose `encode` produces fresh bytes and holds no mutable shared state, and the connection is held exclusively for the duration of the pipeline execute, so no two threads pack against the same encoder at once. The offloading is transparent to the wire format since the same `pack_commands` output is then sent as-is. Reads fine.

PR #4211, opened 2026-07-26
feat(ssl): support custom SSLContext

### LINKED ISSUE #3599: redis.exceptions.ConnectionError: Error while reading from redis.hyperreal.dev:3003 : (1, '[SSL: TLSV13_ALERT_CERTIFICATE_REQUIRED] tlsv13 alert certificate required (_ssl.c:2570)')
I'm trying to connect to an SSL protected server with a certificate issued by a private CA and its certs available in the system.

I am getting: ```redis.exceptions.ConnectionError: Error while reading from redis.hyperreal.dev:3003 : (1, '[SSL: TLSV13_ALERT_CERTIFICATE_REQUIRED] tlsv13 alert certificate required (_ssl.c:2570)')```

In other python modules I am usually able to solve this specifying ssl_context like this:

```ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)```

and set ```ssl_context=ssl_context``` as a parameter to a client.

But redis does not allow setting ssl_context parameter, unfortunately. Setting ssl_ca_certs or ssl_certfile does not solve the issue. In result, I am not able to connect to my server using the python module at all. 

In my opinion ssl_context parameter should get added allowing larger flexibility with SSL operations.

lh0156:
## Summary

redis-py currently exposes SSL configuration as individual keyword arguments, which prevents applications from supplying a pre-configured `ssl.SSLContext`. This is required for deployments that need TLS settings not represented by the current keyword arguments, such as custom certificate stores or private CA handling.

This change:

- accepts a pre-configured `SSLContext` for synchronous and asynchronous connections
- propagates it through standalone, URL-based, and cluster clients
- uses the supplied context without mutating it
- preserves the existing context-building path when no custom context is provided
- adds focused regression tests for direct clients, URL connection pools, and cluster configuration

Fixes #3599.

## Validation

- TDD regression tests failed before the implementation and passed afterward
- custom-context coverage: 8 passed
- existing SSL URL parsing regression coverage: 9 passed
- `invoke linters`: passed
- `python -m compileall -q redis tests`: passed
- `git diff --check`: passed

The full TLS/integration matrix was not completed because this checkout has no TLS fixture certificates and the local Redis instance does not enable the `DEBUG` command used by several unrelated connection tests.

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -97,8 +97,9 @@ from redis.utils import (
 )
 
 if TYPE_CHECKING and SSL_AVAILABLE:
-    from ssl import TLSVersion, VerifyFlags, VerifyMode
+    from ssl import SSLContext, TLSVersion, VerifyFlags, VerifyMode
 else:
+    SSLContext = None
     TLSVersion = None
     VerifyMode = None
     VerifyFlags = None
@@ -293,6 +294,7 @@ class Redis(
         ssl_min_version: "TLSVersion | None" = None,
         ssl_ciphers: str | None = None,
         ssl_password: str | None = None,
+        ssl_context: "SSLContext | None" = None,
         max_connections: int | None = None,
         single_connection_client: bool = False,
         health_check_interval: int = 0,
@@ -449,6 +451,7 @@ class Redis(
                             "ssl_min_version": ssl_min_version,
                             "ssl_ciphers": ssl_ciphers,
                             "ssl_password": ssl_password,
+                            "ssl_context": ssl_context,
                         }
                     )
             maint_notifications_enabled = (
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -133,8 +133,9 @@ from redis.utils import (
 )
 
 if SSL_AVAILABLE:
-    from ssl import TLSVersion, VerifyFlags, VerifyMode
+    from ssl import SSLContext, TLSVersion, VerifyFlags, VerifyMode
 else:
+    SSLContext = None
     TLSVersion = None
     VerifyMode = None
     VerifyFlags = None
@@ -443,6 +444,7 @@ class RedisCluster(
         ssl_keyfile: str | None = None,
         ssl_min_version: "TLSVersion | None" = None,
         ssl_ciphers: str | None = None,
+        ssl_context: "SSLContext | None" = None,
         protocol: int | None = None,
         legacy_responses: bool = True,
         address_remap: Callable[[Tuple[str, int]], Tuple[str, int]] | None = None,
@@ -510,6 +512,7 @@ class RedisCluster(
                     "ssl_keyfile": ssl_keyfile,
                     "ssl_min_version": ssl_min_version,
                     "ssl_ciphers": ssl_ciphers,
+                    "ssl_context": ssl_context,
                 }
             )
 
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1507,6 +1507,7 @@ class SSLConnection(Connection):
         ssl_min_version: Optional[TLSVersion] = None,
         ssl_ciphers: Optional[str] = None,
         ssl_password: Optional[str] = None,
+        ssl_context: Optional[SSLContext] = None,
         **kwargs,
     ):
         if not SSL_AVAILABLE:
@@ -1525,6 +1526,7 @@ class SSLConnection(Connection):
             min_version=ssl_min_version,
             ciphers=ssl_ciphers,
             password=ssl_password,
+            context=ssl_context,
         )
         super().__init__(**kwargs)
 
@@ -1601,6 +1603,7 @@ class RedisSSLContext:
         min_version: Optional[TLSVersion] = None,
         ciphers: Optional[str] = None,
         password: Optional[str] = None,
+        context: Optional[SSLContext] = None,
     ):
         if not SSL_AVAILABLE:
             raise RedisError("Python wasn't built with SSL support")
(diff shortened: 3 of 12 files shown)


petyaslavova (changes requested):
Thanks for taking this on — the underlying gap is real. The `ssl_*` keywords can only express file-based PEM trust and client certificates, so anything needing a pre-built context (OS trust store via `truststore`, in-memory or PKCS#11 keys, ALPN, SNI callbacks) currently has no supported route, and accepting an `SSLContext` will cover that gap.

I'd like to take it in a slightly different shape though, which also shrinks the diff considerably.
Rather than branching inside `_wrap_socket_with_ssl`, let the existing `RedisSSLContext` own the
decision, and share one holder between both stacks:

1. Move `RedisSSLContext` to `redis/connection.py` and import it into `redis/asyncio/connection.py`.
   Async-imports-from-sync is the direction we already use (see `redis/asyncio/cluster.py`), and the import binding keeps `from redis.asyncio.connection import RedisSSLContext` working. This gives one class object, so `isinstance` and the annotation mean the same thing on both sides.
2. Give `RedisSSLContext` a `context` argument. Please keep `context` readable and writable — it's pre-existing public state — so store it in a `_context` slot behind a `context` property (a slot and a property can't share a name, and `__slots__` has no `__dict__`, so `_context` needs adding there). The setter should validate `ssl.SSLContext | None`, and the getter should keep today's behaviour of returning `None` until the context is built or supplied.
3. When `context` is supplied, ignore the other fields; otherwise build from them as today. With that in the holder, `_wrap_socket_with_ssl` collapses to `context = self.ssl_context.get()` and the duplicated builder disappears. This also means the context is built once and reused on the sync side rather than rebuilt on every reconnect, which is what we want. The sync-only OCSP block stays where it is — it runs after `wrap_socket` and needs the live socket.
4. Public `ssl_context` should accept `ssl.SSLContext | RedisSSLContext | None`, default `None`.
   A holder is used as-is, a raw context is wrapped, anything else raises `RedisError` at
   construction rather than at first connect. The holder's own `context` argument stays
   `ssl.SSLContext | None` — it shouldn't accept another holder. Please also annotate the sync
   `SSLConnection` parameter; it's currently bare `ssl_context=None`.
5. Honour it only when `ssl=True`, consistent with the other SSL options, and document that explicitly.
6. Add the three missing async properties — `ca_path`, `ciphers`, `password` — so both stacks expose the same set of concepts. Please don't rename anything: sync keeps
   `ssl_include_verify_flags`, `certificate_password`, `ssl_min_version`, and async keeps its existing property names. `tests/test_ssl.py` and  `tests/test_asyncio/test_ssl.py` read several of these directly.
7. Docstrings should state that a supplied context is used as-is, that the other `ssl_*` options are then ignored, and that the caller owns full configuration — mirroring how we word "Argument is ignored when connection_pool is provided".
8. Please keep the parameter in the same position across `Redis`, `redis.asyncio.Redis`,
   `RedisCluster` and `redis.asyncio.RedisCluster`. It's currently appended in `redis/client.py` but inserted mid-signature in `redis/asyncio/cluster.py`, whose parameters are all positional-or-keyword, so insertion shifts six later ones(this can break existing apps providing the config args as positional).

One detail to resolve while merging the two builders: they guard the same options with inverted conditions — sync uses `is not None` for the CA fields and truthiness for ciphers, async the opposite. Since an empty value is never meaningful for any of them (`ca_certs=""` → `FileNotFoundError`, `ca_path=""` → `INVALID_DIRECTORY`, `ca_data=""` → `Empty certificate data`, `ciphers=""` → `No cipher can be selected`), please reject empty strings — and empty `bytes`, since `ca_data` accepts DER — up front with a `RedisError` naming the option. The builder conditions can
then simply use truthiness, since only `None` or non-empty values can reach them. Note `ssl_password=""` should stay accepted, and `min_version` keeps its `is not None` check.

On tests, the current ones assert object identity or a mocked `wrap_socket`, which doesn't show the context reaching the handshake. `tests/test_connect.py::test_tcp_ssl_connect` plus
`tests/ssl_utils.get_tls_certificates` already give you a local TLS server, so please add real handshake coverage on both stacks for both input forms, plus cases for: the holder ignoring other fields when `context` is set, `ssl=False` leaving the context unused, an invalid `ssl_context` type, empty-value rejection per option, cluster and `from_url` propagation, and the new async properties.
An example in `docs/examples/ssl_connection_examples.ipynb` would be a nice addition for discoverability, though we don't consider it blocking.

Finally, could you retarget the description to `Refs #3599` rather than `Fixes`? The reporter's `TLSV13_ALERT_CERTIFICATE_REQUIRED` is a mutual-TLS failure that `ssl_certfile`/`ssl_keyfile` should already cover, and we already load the system trust store via `ssl.create_default_context()`, so we want to confirm their case separately before closing that issue.

Happy to review a WIP push if you'd like feedback before it's complete.

eeshsaxena (commented):
Traced the context handling and it does the right thing. `RedisSSLContext.get()` returns the provided `context` as-is when one is passed and only builds from the individual `ssl_*` params when `context is None`, so a caller-supplied `SSLContext` is used verbatim for `open_connection(ssl=...)` rather than being rebuilt or merged. Threading `ssl_context` through the client/cluster/connection kwargs is consistent across the sync and async stacks, and the import fallback (`SSLContext = None` when SSL isn't available) matches the existing pattern for the other ssl symbols.

Two things that are probably worth a doc note rather than code changes:

- When `ssl_context` is given, the individual params (`ssl_cert_reqs`, `ssl_ca_certs`, `ssl_check_hostname`, etc.) are silently ignored. That is sensible precedence, but a caller who sets both might expect them to merge, so it is worth spelling out.
- The introspection properties (`check_hostname`, `cert_reqs`, `ca_certs`, ...) read from the `RedisSSLContext` fields, so with a custom context they report the RedisSSLContext defaults rather than the context's actual settings. Only matters for repr/introspection since the real TLS comes from `get()`.

PR #4212, opened 2026-07-26
Add opt-in latency-aware cluster read balancing

### LINKED ISSUE #4185: ClusterClient: latency-aware read load balancing (power-of-two-choices + peak-EWMA)
All four `LoadBalancingStrategy` options in `redis/cluster.py` are latency-blind: a degraded replica (GC pause, saturated host, cross-AZ hop) keeps its full read share until it fails outright, so read tail latency spikes long before errors appear.

Proposal: a `LATENCY_BASED` strategy using power-of-two-choices + peak-sensitive EWMA — pick two candidate nodes at random, route to the lower score = peak-EWMA of recent command latency × (in-flight + 1). O(1) per read, no probing; latency comes from command round-trips the client already performs. Zero required config, and slow cross-AZ replicas naturally deprioritize themselves.

I've implemented this algorithm in Apache brpc (apache/brpc#3367) and ClickHouse (ClickHouse/ClickHouse#109234); in brpc's benchmark with 1 degraded backend of 4, p99 dropped from 6.5ms (round-robin) to 337µs. Happy to open a PR with the strategy, tests, and docs if maintainers are open to it.

lh0156:
Fixes #4185

## Summary

- Add opt-in `LoadBalancingStrategy.LATENCY_BASED` for cluster reads.
- Select two eligible nodes with power-of-two choices and prefer the lower peak-sensitive EWMA score:
  `peak_ewma * (in_flight + 1)`.
- Keep existing round-robin and random strategies unchanged.
- Record per-node latency and in-flight state for sync and asyncio commands.
- Record MOVED/ASK attempts against the node that handled each attempt.
- Record one node-level sample for grouped pipeline execution when the batch contains reads.
- Document the strategy and add a small cluster benchmark.

## Implementation notes

The latency state is implemented once in the shared `LoadBalancer`, so sync and asyncio clients use the same selection and update logic. The selection path is protected by the existing bounded lock; no await is introduced into asyncio selection.

Latency instrumentation is enabled only for the new strategy. Existing strategies do not pay for request timing or state updates.

## Verification

- TDD RED: node-manager tests initially failed because the candidate node list was not supplied to `LoadBalancer`.
- TDD GREEN: focused sync tests: 8 passed.
- Focused asyncio tests: 3 passed.
- `invoke linters`: passed.
- `python -m compileall`: passed.
- `git diff --check`: passed.

The full cluster test files require local Redis cluster services. They were attempted, but this environment has no Redis cluster listening on the test ports, so connection-dependent tests could not complete.

The benchmark is available at `benchmarks/cluster_latency_load_balancing.py`; it requires a Redis cluster with at least one replica for the selected key slot and reports delayed-replica selection share and p99 latency against round-robin.

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1205,6 +1205,8 @@ class RedisCluster(
 
         while ttl > 0:
             ttl -= 1
+            tracked_node_name = None
+            attempt_start_time = None
             try:
                 if asking:
                     target_node = self.get_node(node_name=redirect_addr)
@@ -1223,6 +1225,15 @@ class RedisCluster(
                     )
                     moved = False
 
+                if (
+                    self.load_balancing_strategy == LoadBalancingStrategy.LATENCY_BASED
+                    and command in READ_COMMANDS
+                ):
+                    tracked_node_name = target_node.name
+                    attempt_start_time = time.monotonic()
+                    self.nodes_manager.read_load_balancer.start_request(
+                        tracked_node_name
+                    )
                 response = await target_node.execute_command(*args, **kwargs)
                 await self._record_command_metric(
                     command_name=command,
@@ -1370,6 +1381,11 @@ class RedisCluster(
                     error=e,
                 )
                 raise
+            finally:
+                if tracked_node_name is not None and attempt_start_time is not None:
+                    self.nodes_manager.read_load_balancer.record_latency(
+                        tracked_node_name, time.monotonic() - attempt_start_time
+                    )
 
         e = ClusterError("TTL exhausted.")
         e.connection = target_node
@@ -2066,7 +2082,10 @@ class NodesManager:
                 # get the server index using the strategy defined in load_balancing_strategy
                 primary_name = self.slots_cache[slot][0].name
                 node_idx = self.read_load_balancer.get_server_index(
-                    primary_name, len(self.slots_cache[slot]), load_balancing_strategy
+                    primary_name,
+                    len(self.slots_cache[slot]),
+                    load_balancing_strategy,
+                    nodes=self.slots_cache[slot],
                 )
                 return self.slots_cache[slot][node_idx]
             return self.slots_cache[slot][0]
@@ -2793,10 +2812,27 @@ class PipelineStrategy(AbstractStrategy):
         # Start timing for observability
         start_time = time.monotonic()
 
+        async def execute_node_pipeline(node, commands):
+            track_latency = (
+                client.load_balancing_strategy == LoadBalancingStrategy.LATENCY_BASED
+                and any(command.args[0] in READ_COMMANDS for command in commands)
+            )
+            if not track_latency:
+                return await node.execute_pipeline(commands)
+
+            client.nodes_manager.read_load_balancer.start_request(node.name)
+            start_time = time.monotonic()
+            try:
+                return await node.execute_pipeline(commands)
+            finally:
+                client.nodes_manager.read_load_balancer.record_latency(
+                    node.name, time.monotonic() - start_time
+                )
+
         errors = await asyncio.gather(
             *(
-                asyncio.create_task(node[0].execute_pipeline(node[1]))
-                for node in nodes.values()
+                asyncio.create_task(execute_node_pipeline(node, commands))
+                for node, commands in nodes.values()
             )
         )
 
--- a/redis/cluster.py
+++ b/redis/cluster.py
@@ -9,6 +9,7 @@ from abc import ABC, abstractmethod
 from collections import OrderedDict, defaultdict
 from concurrent.futures import Future, ThreadPoolExecutor
 from copy import copy
+from dataclasses import dataclass
 from enum import Enum
 from itertools import chain
 from types import MethodType
@@ -20,6 +21,7 @@ from typing import (
     List,
     Literal,
     Optional,
+    Sequence,
     Set,
     Tuple,
     Type,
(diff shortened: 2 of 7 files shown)


redis/cluster.py
```diff
@@ -1883,6 +1896,10 @@ def _execute_command(self, target_node, *args, **kwargs):
                 )
                 raise e
             finally:
+                if tracked_node_name is not None and attempt_start_time is not None:
```
petyaslavova, line 1899: This block will record inaccurate latencies. Executed in `finally` clause it will record the fast failing error situation as best execution times.

_Rewarded (fast failure → lowest score → node becomes preferred):_
- MovedError / AskError — sub-millisecond ResponseErrors. A node with a stale slot view scores ≈0 and keeps winning selection, so the redirect loop is self-reinforcing until the slot cache rebuilds. The PR body presents per-attempt MOVED attribution as a feature; the effect is the opposite.
- TryAgainError, ClusterDownError, SlotNotCoveredError, AuthenticationError, and ConnectionError raised from send_command/parse_response on an already-broken socket — all immediate, all rewarded.

redis/cluster.py
```diff
@@ -2051,25 +2068,44 @@ class LoadBalancingStrategy(Enum):
...
         self.start_index: int = start_index
         self._lock: threading.Lock = threading.Lock()
+        self._latency_state: dict[str, _NodeLatencyState] = {}
 
     def get_server_index(
         self,
         primary: str,
         list_size: int,
         load_balancing_strategy: LoadBalancingStrategy = LoadBalancingStrategy.ROUND_ROBIN,
+        nodes: Optional[Sequence["ClusterNode"]] = None,
```
petyaslavova, line 2101: Newly created or existing touched methods should use PEP604 standard for type annotations --> `Optional[Sequence["ClusterNode"]]` should become `Sequence["ClusterNode"] | None`

redis/asyncio/cluster.py
```diff
@@ -1370,6 +1381,11 @@ async def _execute_command(
                     error=e,
                 )
                 raise
+            finally:
+                if tracked_node_name is not None and attempt_start_time is not None:
```
petyaslavova, line 1385: Situations that will produce fake good and prefered latency records:

- MaxConnectionsError. Raised synchronously by acquire_connection, which is inside the tracked window because async starts the timer before target_node.execute_command(). So the most saturated node scores ≈0, attracts more reads, and produces more MaxConnectionsError. That's a strict inversion of the feature's purpose.

- connect failure. A refused TCP connect fails in microseconds, so a dead node becomes top-ranked until the slot map refreshes.

petyaslavova (changes requested):
This needs changes in three areas before it can be reviewed as a candidate for merge.

Sampling. record_latency runs in a finally block, so failure durations become "this node is fast"
evidence. MOVED/ASK, TRYAGAIN, CLUSTERDOWN and auth errors all return in well under a millisecond, so
a node with a stale slot view scores near zero and keeps winning selection — the redirect loop
reinforces itself. On the async side the timer also covers connection acquisition, so
MaxConnectionsError (raised synchronously by acquire_connection) and a refused TCP connect both record
near-zero latency: the most saturated node, and a dead one, become the preferred targets. In the other
direction, a TimeoutError records the full socket_timeout — client configuration rather than node
service time — and task cancellation records the caller's wait. Keep the in-flight release in finally,
but record a sample only on a successful response, and skip redirects, cluster-state and auth errors,
pool exhaustion, connect failures and cancellation. If timeouts should penalize a node, make that an
explicit capped penalty.

Sync/async parity. The two stacks measure different intervals: sync starts timing after
get_connection() — so pool wait, connect and the health-check reconnect are excluded, which is exactly
where a degraded node burns time — while async starts before it. Settle on one definition of a node
latency sample and apply it identically in both stacks.

Algorithm. The peak score only decays when the node completes another request, but a penalized node is
never selected again once a healthier peer exists. A single transient outlier therefore excludes a node
from reads permanently, until the next NodesManager.initialize(), which also clears all state. That is
worse than round-robin in the recovery case. Decay by elapsed time since the last observation, give
unmeasured nodes a genuinely neutral score rather than 0.0 (currently the minimum, so cold nodes always
win), and decide explicitly how the state should survive initialize(). Related: the primary is an
eligible candidate but only READ_COMMANDS are instrumented, so a write-saturated primary looks idle and
attracts reads.

Tests. None of the current cases reach the power-of-two-choices path — all four use two nodes and take
the len(nodes) <= 2 shortcut, so random.sample is never called and the patch is inert. Add a case with
three or more nodes, one asserting in_flight returns to zero after success, failure and a MOVED retry,
and one covering decay with no traffic (the current decay test drives record_latency directly, which is
precisely what cannot happen for a starved node). Note also that patch("redis.cluster.random.sample")
patches the stdlib random module globally rather than a local alias. Separately, the sync pipeline
finally block rebinds start_time, shadowing the value later passed to _raise_first_error.

Benchmark. #4185 asked for measured numbers from a redis-py cluster; the PR states the script was never
run. `invoke devenv` provides a cluster. The script also injects a client-side asyncio.sleep into a
globally monkeypatched ClusterNode.execute_command and then times its own injected delay, so it does not
measure server-side degradation — use DEBUG SLEEP on the replica or netem instead.

Rebase on master as well; the hunks in _execute_command and _send_cluster_commands conflict.

petyaslavova:
Due to no activity by the contributor and the amount of changes needed to complete this PR, I'm gonna close it for now. Please reopen it or create a new PR if you are still willing to work on this change with the requested changes.

PR #4213, opened 2026-07-26
Make cluster scans best-effort with partial coverage

### LINKED ISSUE #3411: cluster - scan do not respect require_full_coverage at false
**Version**: 5.1.1

**Platform**: python 3.11

**Description**: 

when I run a scan ( with scan_iter helper ) on a redis cluster with require_full_coverage at false

and a node is missing then the scan command fail 

```log
  File "/home/raphael/REPO/xx/src/redis_clean.py", line 27, in main
    keys_to_delete = [x for x in r_client.scan_iter(match_regex, 100)]
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/raphael/REPO/xx/src/redis_clean.py", line 27, in <listcomp>
    keys_to_delete = [x for x in r_client.scan_iter(match_regex, 100)]
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/raphael/REPO/xx/venv/lib/python3.11/site-packages/redis/commands/cluster.py", line 793, in scan_iter
    cursors, data = self.scan(match=match, count=count, _type=_type, **kwargs)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/raphael/REPO/xx/venv/lib/python3.11/site-packages/redis/commands/core.py", line 3017, in scan
    return self.execute_command("SCAN", *pieces, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/raphael/REPO/xx/venv/lib/python3.11/site-packages/redis/cluster.py", line 1115, in execute_command
    raise e
  File "/home/raphael/REPO/xx/venv/lib/python3.11/site-packages/redis/cluster.py", line 1101, in execute_command
    res[node.name] = self._execute_command(node, *args, **kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/raphael/REPO/xx/venv/lib/python3.11/site-packages/redis/cluster.py", line 1174, in _execute_command
    raise e
  File "/home/raphael/REPO/xx/venv/lib/python3.11/site-packages/redis/cluster.py", line 1144, in _execute_command
    connection = get_connection(redis_node, *args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/raphael/REPO/xx/venv/lib/python3.11/site-packages/redis/cluster.py", line 51, in get_connection
    return redis_node.connection or redis_node.connection_pool.get_connection(
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/raphael/REPO/xx/venv/lib/python3.11/site-packages/redis/connection.py", line 1074, in get_connection
    connection.connect()
  File "/home/raphael/REPO/xx/venv/lib/python3.11/site-packages/redis/connection.py", line 283, in connect
    raise ConnectionError(self._error_message(e))
redis.exceptions.ConnectionError: Error 113 connecting to 172.20.0.4:6379. No route to host.
```

I would expect the command to run in best effort on available nodes.

lh0156:
Fixes #3411

## What changed

- Make cluster `scan_iter()` best-effort when `require_full_coverage=False`.
- Skip only per-node `ConnectionError` and `TimeoutError` failures while continuing to scan reachable nodes.
- Preserve fail-fast behavior when `require_full_coverage=True`.
- Continue to propagate Redis command errors such as `ResponseError`.
- Keep synchronous and asynchronous cluster clients consistent.
- Add regression tests for both clients and the failure boundaries.

## Verification

- `pytest -q tests/test_cluster.py -k 'cluster_scan_iter'` — 4 passed
- `pytest -q tests/test_asyncio/test_cluster.py -k 'cluster_scan_iter'` — 5 passed
- `invoke linters`
- `ruff format --check`
- `git diff --check`
- `python -m compileall`

The full Redis Cluster integration suite requires local Redis cluster services, which are not available in this workspace.

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1069,6 +1069,10 @@ class RedisCluster(
         command = args[0]
         target_nodes = []
         target_nodes_specified = False
+        best_effort = (
+            kwargs.pop("_best_effort", False)
+            and not self.nodes_manager.require_full_coverage
+        )
         retry_attempts = self.retry.get_retries()
 
         passed_targets = kwargs.pop("target_nodes", None)
@@ -1137,7 +1141,18 @@ class RedisCluster(
 
                 if len(target_nodes) == 1:
                     # Return the processed result
-                    ret = await self._execute_command(target_nodes[0], *args, **kwargs)
+                    try:
+                        ret = await self._execute_command(
+                            target_nodes[0], *args, **kwargs
+                        )
+                    except (ConnectionError, TimeoutError):
+                        if not best_effort:
+                            raise
+                        if command in self.result_callbacks:
+                            return self.result_callbacks[command](command, {}, **kwargs)
+                        return self._policies_callback_mapping[
+                            command_policies.response_policy
+                        ]({})
                     if command in self.result_callbacks:
                         ret = self.result_callbacks[command](
                             command, {target_nodes[0].name: ret}, **kwargs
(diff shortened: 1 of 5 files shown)


petyaslavova:
Thanks for picking this up, and for scoping it explicitly in the issue first.

I traced the change through both cluster clients, and I don't think we can take this design.
`except (ConnectionError, TimeoutError)` also captures `AuthenticationError`,
`AuthorizationError`, `MaxConnectionsError` and `BusyLoadingError`, which both stacks
deliberately re-raise as *not* node failures — so a wrong password, an exhausted pool, or a
node still loading its dataset would make `scan_iter()` return a partial or empty key set with
no error at all. The skip also happens inside the `ERRORS_ALLOW_RETRY` loop, so a transient
blip or an in-progress failover that the client recovers from today would instead silently
return incomplete results. And `require_full_coverage` is documented as construction-time slot
coverage, not runtime failure tolerance, so keying behavior off it changes existing users
silently and only for `scan_iter`.

The underlying use case is already supported per shard: iterate `get_primaries()` and call
`scan_iter(match=..., target_nodes=node)` inside your own try/except. That is covered by
existing tests and, unlike an implicit skip, it tells the caller which shards were missed.

I'm going to close this for now. If there is a concrete case the per-shard approach can't
handle, we can revisit it as an explicit opt-in argument on `scan_iter` that applies only after
retries are exhausted, matches the exact connection/timeout types, and either reports the
skipped nodes or raises when no node succeeded.

PR #4214, opened 2026-07-26
Handle ASK redirects to slotless nodes during resharding

lh0156:
Fixes #4015

## What changed

- Preserve an ASK redirect target for the current retry even when it is absent from the refreshed slots cache.
- Create a temporary primary `ClusterNode` only when the redirect target is unknown.
- Keep the existing ASK sequence unchanged: send `ASKING`, then retry the original command.
- Do not patch slot ownership or retain the temporary node in `nodes_cache`.
- Apply the same behavior to synchronous and asynchronous cluster clients.
- Add deterministic regression tests for an ASK target with no slots.

## Verification

- `pytest -q tests/test_cluster.py -m onlycluster -k 'ask_redirection_to_node_without_slots'` — 1 passed
- `pytest -q tests/test_asyncio/test_cluster.py -k 'ask_redirection_to_node_without_slots'` — 1 passed
- `invoke linters`
- `ruff format --check`
- `git diff --check`
- `python -m compileall`

The full Redis Cluster integration suite requires local Redis cluster services, which are not available in this workspace.

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1198,7 +1198,7 @@ class RedisCluster(
         self, target_node: "ClusterNode", *args: Union[KeyT, EncodableT], **kwargs: Any
     ) -> Any:
         asking = moved = False
-        redirect_addr = None
+        redirected_node = None
         ttl = self.RedisClusterRequestTTL
         command = args[0]
         start_time = time.monotonic()
@@ -1207,7 +1207,7 @@ class RedisCluster(
             ttl -= 1
             try:
                 if asking:
-                    target_node = self.get_node(node_name=redirect_addr)
+                    target_node = redirected_node
                     await target_node.execute_command("ASKING")
                     asking = False
                 elif moved:
@@ -1327,7 +1327,11 @@ class RedisCluster(
                     connection=target_node,
                 )
             except AskError as e:
-                redirect_addr = get_node_name(host=e.host, port=e.port)
+                redirected_node = self.get_node(host=e.host, port=e.port)
+                if redirected_node is None:
+                    redirected_node = ClusterNode(
+                        e.host, e.port, PRIMARY, **self.nodes_manager.connection_kwargs
+                    )
                 asking = True
                 await self._record_command_metric(
                     command_name=command,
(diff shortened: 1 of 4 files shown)


petyaslavova (changes requested):
Thanks for picking this up, and for the precise scoping — the diagnosis is right. `initialize()` builds `nodes_cache` purely from `CLUSTER SLOTS`, so a slotless importing node is never in it and the ASK retry dereferences `None` in both stacks. That part of the fix is the direction we want.

Before we can merge it, the redirect target needs an owner. As written, the temporary `ClusterNode` is intentionally kept out of `nodes_cache` and `startup_nodes`, which are the only collections `aclose()` / `set_nodes(remove_old=True)` ever tear down, so every ASK-redirected command leaves a connection behind. On the async side `ClusterNode.__del__` can only warn (`"Unclosed ClusterNode object"`) — it cannot close the transport. On the sync side each redirect builds a fresh `Redis` client, connection pool, and handshake via `create_redis_connections()`. During a real reshard that is one orphaned connection per redirected command, in exactly the sustained-traffic scenario the issue describes.

Could you please:

- make the ad-hoc node manager-owned and short-lived — reused for repeated redirects to the same address, and disconnected once it is no longer needed;
- route the redirect address through `remap_host_port()` so `address_remap` users are covered;
- extend the tests so they exercise real connection acquisition and release and assert that nothing is left behind after the command. The current tests patch out `get_redis_connection` / `ClusterNode.execute_command`, so they prove the `None` dereference is gone but not the resource behavior.

A rebase is also needed — the branch currently conflicts with `master` — and the PR is still a draft. Happy to review again once those are in.

eeshsaxena (commented):
The slotless-node handling looks right. On the sync side `ClusterNode(e.host, e.port, PRIMARY)` is created with `redis_connection=None`, and the first time it is used `get_redis_connection` -> `nodes_manager.create_redis_connections([node])` -> `create_redis_node(..., **self.connection_kwargs)`, so TLS/auth/timeout config is still applied even though the node was built ad hoc for the ASK target. The async side passes `**self.nodes_manager.connection_kwargs` at construction since the async `ClusterNode` owns its own pool. Different mechanism, same outcome: both inherit the cluster's connection settings.

This also fixes the previous behavior where `get_node(node_name=...)` returned `None` for a node not in the slot map, and the next loop iteration would call `execute_command("ASKING")` on `None`. Looks good.

PR #4215, opened 2026-07-26
Avoid retrying client-initiated SHUTDOWN

### LINKED ISSUE #3704: Shutdown from redis client takes a long time
If we run a redis-server from the shell with no persistence (in this case redis-7.2.3):
```sh
redis-server --save "" --appendonly no --loglevel verbose
```
Then, on the same machine hop into a python shell and run (redis-py 6.2.0):
```python3
from redis import Redis
redis.shutdown()
```
The redis server itself instantly exits, but the `redis.shutdown()` command does not return for about 5 seconds.  It appears that the redis client is waiting for some form of network timeout before returning.
This is unaffected by any combination of the parameters (nosave, now, force).

This may be a regression from earlier versions, but we are not entirely sure.

lh0156:
Fixes #3704

## Motivation

When Redis shuts down, the connection used to send `SHUTDOWN` is expected to close. The default connection retry policy currently retries the resulting `ConnectionError`, which delays `Redis.shutdown()` by the configured backoff/retry period.

## Solution

- Add an internal per-command no-retry option to the synchronous and asynchronous clients.
- Use it only for `shutdown()`, leaving retry behavior for all other commands unchanged.
- Preserve the existing `ConnectionError` handling that treats the expected disconnect as a successful shutdown.
- Close the failed connection and keep existing operation/error metrics behavior.

## Tests

- Verify sync and async `shutdown()` pass the internal option for NOSAVE, SAVE/NOW/FORCE, and ABORT.
- Verify sync and async command execution performs exactly one attempt even when the connection retry policy allows retries.
- Existing client retry and metrics tests remain green.

Validation:
- `pytest -q tests/test_commands.py -k 'test_shutdown'`
- `pytest -q tests/test_asyncio/test_commands.py -k 'shutdown_disables_command_retries'`
- `pytest -q tests/test_client.py`
- `pytest -q tests/test_asyncio/test_client.py`
- `invoke linters`
- `python -m compileall -q redis ...`

### DIFF
--- a/redis/asyncio/client.py
+++ b/redis/asyncio/client.py
@@ -872,6 +872,7 @@ class Redis(
         await self.initialize()
         pool = self.connection_pool
         command_name = args[0]
+        no_retry = options.pop("_no_retry", False)
         conn = self.connection or await pool.get_connection()
 
         # Start timing for observability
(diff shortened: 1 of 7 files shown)


petyaslavova:
Thank you for picking this up and for the clear write-up — the diagnosis is correct, and the reworked `test_shutdown` assertions are a real improvement over the previous ones.

We are not going to take this approach, though. Skipping retries per command is the right idea, but we plan to expose it through a designed, public per-command options argument (a `commandOptions`-style parameter available to every command, with "no retry" as one of its options) rather than a private `_no_retry` key smuggled through `**options`. Because `execute_command` is public, that key would immediately become a de-facto public API we would have to support, and it is not honored or stripped by the cluster and pipeline executors, so `RedisCluster.shutdown()` would keep the current behavior. Separately, `SHUTDOWN ABORT` should not skip retries: it does not close the connection, so a transient error would be reported as a successful abort.

I am closing this PR in favor of that work. #3704 stays open and will be resolved by it.

PR #4216, opened 2026-07-26
Refresh pooled connections before credential expiry

### LINKED ISSUE #3895: Restrict total connection lifetime to prevent using invalidated credentials
At his point in time, this library does not offer any mechanism to control the total connection duration.
This is problematic as, when using a rotating token, such as in the case of a Workload Identity in Azure, the token is invalidated when the token rotates. **As a result, after 24h, the session is invalidated and the connection fails (with an invalid username/password error)**

There should be :
- A configurable option to control the total connection lifetime. All connections exceeding this timeout should be closed
- Connections should be re-opened when being close to reaching this timeout. A pool should not distribute connections that are too close to this timeout, and this delta should be configurable.

Note that :
- **This issue is separate from the use of the StreamingCredentialProvider - connections should still be opened with a valid token**
- The timeout options that exist are only valid for setting a delta limit on the response after a query. **It does not control the total lifetime of the connection**

As a workaround, we implemented a custom Connection Pool with a thread closing connections reaching a defined total lifetime (set to the remaining lifetime of the token last obtained from Azure). The use of both a StreamingCredentialProvider and a custom cache token ensure all connections are always opened with a valid token, and old ones are closed.

lh0156:
Fixes #3895

## Motivation

Connection pools can keep a socket alive beyond the lifetime of the credentials used to authenticate it. This is especially problematic for rotating credentials such as Azure Workload Identity tokens: the pool continues to hand out an established connection, which eventually fails authentication after the token expires.

## Solution

- Add an opt-in `max_connection_lifetime` connection setting; `None` preserves current behavior.
- Record the lifetime after a connection's handshake completes.
- On the next checkout after the lifetime is reached, close the idle socket and reconnect it before returning it to the caller.
- Refresh credentials through the existing credential provider during reconnect.
- Apply the same lifecycle behavior to sync, asyncio, and Sentinel-managed connections.
- Accept the setting in Redis URLs and document the checkout semantics.
- Never interrupt an in-use command; refresh occurs when the connection is checked out again.

## Tests

- TDD regression coverage for expired and non-expired sync and async connections.
- Validation for positive lifetime values and URL parsing in both implementations.
- Full sync connection suite: 95 passed, 2 skipped.
- Full async connection suite: 66 passed, 2 skipped.
- `invoke linters`, `compileall`, and `git diff --check` passed.

The local pool/Sentinel integration suites require services not available in this checkout: the local Redis server has DEBUG disabled and no Sentinel is listening on the configured ports.

### DIFF
--- a/docs/connections.rst
+++ b/docs/connections.rst
@@ -95,6 +95,11 @@ Connection
 .. autoclass:: redis.connection.Connection
     :members:
 
+For deployments that use rotating credentials, ``max_connection_lifetime``
+can be set in seconds. A connection that reaches this age is refreshed the
+next time it is checked out from the pool, allowing the configured credential
+provider to obtain fresh credentials without interrupting an active command.
+
 Connection (Async)
 ==================
 .. autoclass:: redis.asyncio.connection.Connection
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -561,6 +561,8 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
         "retry_on_timeout",
         "retry_on_error",
         "health_check_interval",
+        "max_connection_lifetime",
+        "_connection_created_at",
         "next_health_check",
         "last_active_at",
         "encoder",
@@ -597,6 +599,7 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
         parser_class: Type[BaseParser] = DefaultParser,
         socket_read_size: int = DEFAULT_SOCKET_READ_SIZE,
         health_check_interval: float = 0,
+        max_connection_lifetime: float | None = None,
         client_name: str | None = None,
         lib_name: str | object | None = SENTINEL,
         lib_version: str | object | None = SENTINEL,
@@ -627,6 +630,11 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
 
         Parameters
         ----------
+        max_connection_lifetime : float, optional
+            Maximum time in seconds to reuse a connected socket. When the limit
+            is reached, the next connection attempt reconnects the socket so
+            credential providers can refresh their credentials. ``None``
+            disables the limit.
         driver_info : DriverInfo, optional
             Driver metadata for CLIENT SETINFO. If provided, lib_name and lib_version
             are ignored. If not provided, a DriverInfo will be created from lib_name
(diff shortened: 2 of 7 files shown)


petyaslavova:
Thanks for picking this up, and for posting the design note on #3895 before starting — that's appreciated.

Before going further on the implementation, I'd like to settle the underlying need. 
`redis-py` already refreshes rotating credentials on live pooled connections: with a `StreamingCredentialProvider` (e.g. redis-entraid's EntraIdCredentialsProvider), `TokenManager` renews ahead of expiry and  `ConnectionPool.re_auth_callback` re-AUTHs idle connections immediately, while in-use ones re-authenticate right after their current command. That covers standalone pools, pipelines, cluster nodes and RESP3 pub/sub, and is exercised by the `Entra ID` tests in `tests/test_credentials.py`.
Could you share a reproduction where a `StreamingCredentialProvider` is configured and a pooled connection still fails authentication? That's the evidence we'd need to justify a new connection setting.

Two things about the current shape are worth flagging regardless. `max_connection_lifetime` isn't in `REDIS_ALLOWED_KEYS`, so `RedisCluster` silently drops it (and async RedisCluster.from_url raises TypeError), and it isn't accepted by `Redis.__init__` at all — only via `from_url`. Because expiry is only checked in `connect_check_health`, the setting also never applies to `single_connection_client=True` clients or pub/sub connections, which are the sockets most likely to outlive a token.

Separately, we're leaning towards solving this at the pool layer instead: an age/idle sweep over pooled connections that reuses the existing mark_for_reconnect() / should_reconnect() mechanism, so idle and in-use connections are retired consistently across standalone, cluster and sentinel, and the pool can also shrink back after a burst. If you'd like to re-target the PR onto that design I'd be glad to review it. I'll leave this as a draft in the meantime.

eeshsaxena (commented):
Traced the lifetime handling on both stacks and it lines up:

- `_connection_created_at` is set after a successful connect, checked on checkout via `connect_check_health` (`is_connection_expired` -> `disconnect` -> reconnect), and reset to `None` on `disconnect`. So an aged connection is refreshed before it is handed out for the next command rather than being torn down mid-command, which matches the stated intent for rotating credentials.
- Sync and async stay consistent. The async `AbstractConnection` correctly adds `max_connection_lifetime` and `_connection_created_at` to `__slots__`, and the sync `Connection` does not use `__slots__`, so neither path raises on the new attribute assignments.
- `max_connection_lifetime <= 0` is rejected at construction, and `None` disables the check via the short-circuit in `is_connection_expired`.

Reads correctly to me.

petyaslavova:
Closing this PR due to no activity by contributor.

PR #4217, opened 2026-07-26
Normalize asyncio writer close races

### LINKED ISSUE #3546: self._write_ready() 'NoneType' object is not callable
```
  File "/backend/src/services/redis_service.py", line 94, in get_cached_profiles
    fetched_profiles = await self.client.json().mget(keys, Path.root_path())
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/redis/asyncio/client.py", line 616, in execute_command
  File "/opt/venv/lib/python3.12/site-packages/redis/asyncio/retry.py", line 59, in call_with_retry
    return await do()
           ^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/redis/asyncio/client.py", line 589, in _send_command_parse_response
  File "/opt/venv/lib/python3.12/site-packages/redis/asyncio/connection.py", line 505, in send_command
    await self.send_packed_command(
  File "/opt/venv/lib/python3.12/site-packages/redis/asyncio/connection.py", line 480, in send_packed_command
    self._writer.writelines(command)
  File "/usr/local/lib/python3.12/asyncio/streams.py", line 349, in writelines
    self._transport.writelines(data)
  File "/usr/local/lib/python3.12/asyncio/selector_events.py", line 1182, in writelines
    self._write_ready()
TypeError: 'NoneType' object is not callable 
```

_**I am experiencing this issue especially in peak traffic.**_

My redis service is initialized this way:

```
import redis.asyncio as redis
@lru_cache()
def get_redis_pool():
    settings = get_settings()
    return redis.ConnectionPool.from_url(settings.REDIS_URL_WITH_PASSWORD, decode_responses=True)


@lru_cache()
def get_redis_client():
    redis_pool = get_redis_pool()
    client = redis.Redis(connection_pool=redis_pool)
    return client


# Dependency to create RedisService
async def get_redis_service(
    redis_client: redis.Redis = Depends(get_redis_client),
    firestore_db: firestore.AsyncClient = Depends(get_firestore_db),
):
    from src.services.redis_service import RedisService

    return RedisService(redis_client, firestore_db)
```

Am I missing something in my redis client?

I saw an issue similar to this [here](https://github.com/python/cpython/issues/115514) but upgrading python to 3.12.9 didn't help.

python **3.12.9**
redis **5.2.1**

lh0156:
Fixes #3546

## Problem

When an asyncio transport is closed concurrently with a write, Python can raise a low-level `TypeError` from `StreamWriter.writelines()` after its internal `_write_ready` callback has been cleared. redis-py currently lets that raw `TypeError` escape, so the existing retry policy cannot recognize the broken connection as retryable.

## Solution

- Keep a local writer reference for the complete write/drain operation.
- Convert writer-close `AttributeError`/`TypeError` failures into redis-py's `ConnectionError`.
- Reuse the existing disconnect path, so the connection is discarded and configured retries can handle the failure.
- Route both timeout and no-timeout writes through the same helper to keep behavior consistent.

## Tests

- Add a deterministic async regression test for the closed-writer race.
- Cover both `socket_timeout=None` and finite `socket_timeout` paths.
- Full async connection suite: 65 passed, 2 skipped.
- Async client suite: 7 passed.
- `invoke linters`, `compileall`, and `git diff --check` passed.

The local integration skips are unchanged and are unrelated to this focused writer error path.

---

> [!NOTE]
> **Low Risk**
> Focused error normalization in asyncio connection I/O with targeted regression tests; behavior change is limited to exception type mapping on close races.
> 
> **Overview**
> Fixes concurrent-close races where `StreamWriter.writelines()` could raise raw `AttributeError`/`TypeError` instead of a retryable connection failure.
> 
> **Write path:** Adds `_send_packed_command()` with a local writer reference, maps known close-race exceptions to `ConnectionError("Connection closed while writing")`, and routes both timed and untimed `send_packed_command()` through that helper.
> 
> **Read path:** `_socket_is_empty()` now guards a missing reader and maps buffer access failures to `ConnectionError("Connection closed while reading")`, including during invalidation processing.
> 
> **Tests:** New async tests cover closed-writer races (with/without `socket_timeout`), closed reader during invalidation, and that unrelated `TypeError`/`AttributeError` from bad commands still propagate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c3a84066f319d101a0892e9f4971b992fe915ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1143,8 +1143,14 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
             )
 
     async def _send_packed_command(self, command: Iterable[bytes]) -> None:
-        self._writer.writelines(command)
-        await self._writer.drain()
+        writer = self._writer
+        if writer is None:
+            raise ConnectionError("Connection closed while writing")
+        try:
+            writer.writelines(command)
+            await writer.drain()
+        except (AttributeError, TypeError) as error:
+            raise ConnectionError("Connection closed while writing") from error
 
     async def send_packed_command(
         self, command: Union[bytes, str, Iterable[bytes]], check_health: bool = True
@@ -1164,8 +1170,7 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
                     self._send_packed_command(command), self.socket_timeout
                 )
             else:
-                self._writer.writelines(command)
-                await self._writer.drain()
+                await self._send_packed_command(command)
         except asyncio.TimeoutError:
             await self.disconnect(nowait=True)
             raise TimeoutError("Timeout writing to socket") from None
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1398,7 +1398,13 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
 
     def _socket_is_empty(self):
         """Check if the socket is empty"""
-        return len(self._reader._buffer) == 0
+        reader = self._reader
+        if reader is None:
+            raise ConnectionError("Connection closed while reading")
+        try:
+            return len(reader._buffer) == 0
+        except AttributeError as error:
+            raise ConnectionError("Connection closed while reading") from error
 
     async def process_invalidation_messages(self):
         while not self._socket_is_empty():
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1151,6 +1151,10 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
             await writer.drain()
         except AttributeError as error:
             raise ConnectionError("Connection closed while writing") from error
+        except TypeError as error:
+            if str(error) != "'NoneType' object is not callable":
+                raise
+            raise ConnectionError("Connection closed while writing") from error
 
     async def send_packed_command(
         self, command: Union[bytes, str, Iterable[bytes]], check_health: bool = True
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1150,6 +1150,8 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
             writer.writelines(command)
             await writer.drain()
         except AttributeError as error:
+            if str(error) != "'NoneType' object has no attribute 'writelines'":
+                raise
             raise ConnectionError("Connection closed while writing") from error
         except TypeError as error:
             if str(error) != "'NoneType' object is not callable":
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1149,7 +1149,7 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
         try:
             writer.writelines(command)
             await writer.drain()
-        except (AttributeError, TypeError) as error:
+        except AttributeError as error:
             raise ConnectionError("Connection closed while writing") from error
 
     async def send_packed_command(
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -116,6 +116,21 @@ async def test_invalid_packed_command_type_error_is_not_connection_error():
     conn._writer = None
 
 
+async def test_closed_writer_type_error_is_connection_error():
+    conn = Connection()
+    conn._reader = mock.Mock()
+    conn._writer = mock.Mock()
+    conn._writer.writelines.side_effect = TypeError("'NoneType' object is not callable")
+    conn.disconnect = mock.AsyncMock()
+
+    with pytest.raises(ConnectionError, match="Connection closed while writing"):
+        await conn.send_packed_command([b"PING\r\n"], check_health=False)
+
+    conn.disconnect.assert_awaited_once_with(nowait=True)
+    conn._reader = None
+    conn._writer = None
+
+
 async def test_closed_reader_during_invalidation_processing_is_connection_error():
     conn = Connection()
     conn._reader = None
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -88,10 +88,27 @@ async def test_closed_writer_during_write_is_connection_error(socket_timeout):
     conn = Connection(socket_timeout=socket_timeout)
     conn._reader = mock.Mock()
     conn._writer = mock.Mock()
-    conn._writer.writelines.side_effect = TypeError("'NoneType' object is not callable")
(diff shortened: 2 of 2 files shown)


vladvildanov:
Please make sure resolve Codex comments first

redis/asyncio/connection.py
```diff
@@ -1143,8 +1143,14 @@ async def check_health(self):
...
     async def _send_packed_command(self, command: Iterable[bytes]) -> None:
-        self._writer.writelines(command)
-        await self._writer.drain()
+        writer = self._writer
+        if writer is None:
+            raise ConnectionError("Connection closed while writing")
+        try:
+            writer.writelines(command)
+            await writer.drain()
+        except (AttributeError, TypeError) as error:
```
Sanjays2402: the AttributeError/TypeError catch is broader than the None-writer case: pack_command aside, send_packed_command takes a caller-supplied iterable, so a bad element (str in the list) raises TypeError from writelines and now surfaces as ConnectionError, which retry.call_with_retry treats as retryable. that turns a programming error into a silent reconnect loop. since you already null-check writer above, is the except clause still buying anything beyond the writer-set-to-None-mid-await race?

lh0156: Addressed in d7dd61c5. The broad TypeError catch was removed, so a caller-supplied invalid packed iterable is no longer reported as ConnectionError. The existing closed-writer coverage now models the AttributeError path explicitly.

redis/asyncio/connection.py
```diff
@@ -1143,8 +1143,14 @@ async def check_health(self):
...
+            raise ConnectionError("Connection closed while writing") from error
```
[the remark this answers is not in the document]
lh0156: Addressed in d7dd61c5. The inner write wrapper now translates only AttributeError from a closed writer; caller/data TypeError values are allowed to propagate unchanged. Added a malformed packed-command regression test. The asyncio connection suite passes (67 passed, 1 skipped).

redis/asyncio/connection.py
```diff
@@ -1143,8 +1143,14 @@ async def check_health(self):
...
     async def _send_packed_command(self, command: Iterable[bytes]) -> None:
-        self._writer.writelines(command)
-        await self._writer.drain()
+        writer = self._writer
+        if writer is None:
+            raise ConnectionError("Connection closed while writing")
+        try:
+            writer.writelines(command)
+            await writer.drain()
+        except AttributeError as error:
```
Sanjays2402, line 1152: the AttributeError catch here covers the writer half of the race, but disconnect() nulls _reader in the same finally block and the read side is still bare: _socket_is_empty() does len(self._reader._buffer), so a close landing during process_invalidation_messages() raises AttributeError straight out to the caller instead of ConnectionError. worth normalizing that one too while you're in here, or is it out of scope for this PR?

lh0156: Addressed in fa2279f7. _socket_is_empty() now translates a missing or invalid reader buffer into ConnectionError("Connection closed while reading"), covering the reader side of the same disconnect race. Added a process_invalidation_messages() regression test; the asyncio connection suite passes (68 passed, 1 skipped), and ruff checks pass.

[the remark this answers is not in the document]
lh0156: Addressed in 4588fa12. _send_packed_command now translates the exact closed-transport TypeError (NoneType object is not callable) into ConnectionError while re-raising unrelated TypeErrors unchanged. Added test_closed_writer_type_error_is_connection_error; the malformed-command regression remains covered. The focused connection suite passes: 69 passed, 1 skipped.

redis/asyncio/connection.py
```diff
@@ -1143,8 +1143,18 @@ async def check_health(self):
...
+            raise ConnectionError("Connection closed while writing") from error
```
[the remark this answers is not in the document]
lh0156: Addressed in 7c3a8406. The write path now translates only the specific closed-writer AttributeError (NoneType object has no attribute writelines) and re-raises other AttributeErrors from the packed command or writer unchanged. Added test_invalid_packed_command_attribute_error_is_not_connection_error; the full focused asyncio connection suite passes: 70 passed, 1 skipped.

petyaslavova (changes requested):
Thanks for digging into this — the underlying problem is real. CPython clears
_SelectorSocketTransport._write_ready in _call_connection_lost on 3.12+, and writelines()
calls it without the _conn_lost guard that write() has, so a write on a peer-reset transport
raises a raw TypeError that our retry policy cannot classify. Mapping it to ConnectionError
in the async write path is the right layer.

Three things before we can merge it:

1. Classify by transport state rather than by exception message. str(error) comparisons
   against CPython wording are not a stable contract, and they still rewrite a malformed
   caller-supplied iterable into a retried ConnectionError (the open Codex comment).
   Checking `writer.transport is None or writer.transport.is_closing()` before writelines()
   is race-free — _force_close()/close() set _closing before scheduling _call_connection_lost,
   and there is no suspension point between the check and the write. Re-checking is_closing()
   inside a narrow `except (AttributeError, TypeError)` covers the drain() window.

2. Please drop the _socket_is_empty()/process_invalidation_messages() change. Those methods
   have no callers in the library or the test suite, and the added blanket `except
   AttributeError` conflicts with the deliberate "fail loudly if the private buffer API
   changes" invariant in redis/_parsers/base.py.

3. Add one regression test against real asyncio objects instead of mocks that supply the
   CPython message text — e.g. connect to a local asyncio.start_server, call
   `conn._writer.transport.abort()`, `await asyncio.sleep(0)`, then send a command. The
   current tests would keep passing even if the real path stopped being detected.

With those in place this should be ready for another review.

eeshsaxena (commented):
The normalization looks right. `_send_packed_command` snapshots `self._writer` into a local and null-checks it before use, so a concurrent `disconnect()` that sets `self._writer = None` becomes a clean `ConnectionError` instead of an `AttributeError` mid-write.

The `except AttributeError` / `except TypeError` fallbacks are gated on exact message matching and re-raise anything else, so a real `TypeError` from an invalid packed-command payload still propagates as itself rather than being masked as a connection error (the two negative tests cover that distinction well). `_socket_is_empty` gets the same None-reader treatment. The one caveat is that the message-string matching is fragile if CPython ever rewords those messages, but the tests pin the current strings, so that would fail loudly rather than silently. Reads correctly.

petyaslavova:
Closing this PR since there was no activity by the contributor after requested changes.

PR #4218, opened 2026-07-26
Add a separate timeout for Redis Cluster topology discovery

### LINKED ISSUE #3800: Timeout errors and new connections spiking
Hello. We are seeing a strange behavior with our Redis cache cluster where we occasionally get a burst of timeout errors on one node, and at the same time we see a huge spike in new connections on that node. We're not sure yet in what order those things are happening. The cluster is hosted in AWS, and we usually see a "network bandwidth exceeded" metric firing on the node in question at the same time.

I'm reaching out because the errors have been pretty confounding to debug, so if this behavior makes sense to any contributors it might help steer me in the right direction.

We use a `socket_connect_timeout` of 1s and a `socket_timeout` of 100ms. We have `retry` set to to an exponential backoff with a base of 10ms, a cap of 200ms, but currently only a single retry attempt.

Some questions/thoughts I have:
* Does the retry policy apply to establishing a connection, or just to executing commands?
* Does the client try to re-establish connections when a certain number of errors have been encountered? I'm trying to track down if the new connections we're seeing are a result of the timeouts or vice-versa. If the errors are **not** causing re-connection, are there any other situations that the client tries to re-connect?

lh0156:
## Summary

Fixes #3800.

Redis Cluster topology discovery currently uses the regular command/socket read timeout for `CLUSTER SLOTS`. A short application timeout can therefore abort a slow topology response and cause repeated connection attempts.

This change adds `cluster_slots_timeout` to both synchronous and asyncio `RedisCluster` clients. When configured, it applies only to `CLUSTER SLOTS` during initial discovery and topology refreshes. Application commands continue to use `socket_timeout`, and the default behavior is unchanged.

## Testing

- Added sync and asyncio regression tests for the configured and default paths.
- Added read-timeout propagation tests for both parser paths.
- `pytest -m fixed_client tests/test_cluster.py tests/test_asyncio/test_cluster.py`: 71 passed.
- `pytest tests/test_connection.py tests/test_asyncio/test_connection.py`: 153 passed, 4 skipped.
- `pytest tests/test_client.py`: 10 passed.
- `ruff check`, `compileall`, and `git diff --check` passed.

The local environment does not provide the Redis Cluster fixture, so the tests marked `onlycluster` were not used for validation.

### DIFF
--- a/docs/clustering.rst
+++ b/docs/clustering.rst
@@ -61,6 +61,11 @@ commands cache contains all the server supported commands that were
 retrieved using the Redis ‘COMMAND’ output. See *RedisCluster specific
 options* below for more.
 
+The ``cluster_slots_timeout`` option can be used when topology discovery is
+expected to take longer than regular application commands. It overrides the
+read timeout only for ``CLUSTER SLOTS`` during initialization and topology
+refreshes; when omitted, ``socket_timeout`` continues to apply.
+
 RedisCluster instance can be directly used to execute Redis commands.
 When a command is being executed through the cluster instance, the
 target node(s) will be internally determined. When using a key-based
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -449,6 +449,8 @@ class RedisCluster(
         event_dispatcher: EventDispatcher | None = None,
         policy_resolver: AsyncPolicyResolver = AsyncStaticPolicyResolver(),
         maint_notifications_config: MaintNotificationsConfig | None = None,
+        # Read timeout used only for CLUSTER SLOTS topology discovery.
+        cluster_slots_timeout: float | None = None,
     ) -> None:
         if db:
             raise RedisClusterException(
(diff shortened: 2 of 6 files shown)


petyaslavova:
Thanks for the investigation and the patch — your diagnosis on #3800 is correct: topology discovery shares socket_timeout with application commands, and there is no supported way to give it a separate deadline today.

That said, I am going to close this PR. We are planning a general per-command options mechanism that will include a per-command timeout, and that is the layer this belongs in: a dedicated cluster_slots_timeout constructor argument would be a permanent public contract for a case the generic option will cover, and we would then be maintaining both. If topology discovery still needs a longer deadline once that lands, it becomes a small internal default applied to CLUSTER SLOTS rather than new API surface.

For completeness, a few points in the current implementation also pointed the same way. The timeout is threaded as a private _read_timeout option through the base sync Redis.parse_response, which makes it an undocumented per-command timeout for every sync client — essentially a narrow version of the generic mechanism above. The async side instead swaps connection.socket_timeout around the read, which can revert a maintenance-relaxed timeout installed while the read is in flight. The URL form (?cluster_slots_timeout=5) arrives as a string and raises TypeError from socket.settimeout. And only CLUSTER SLOTS is covered, while commands_parser.initialize() in the same discovery sequence still uses socket_timeout.

Thank you for the time you put into this, including the sync and async coverage — please do not read the closure as a rejection of the problem. Once per-command options are in place I would be glad to have you revisit the discovery path on top of them.

PR #1899, opened 2022-01-25
Add Async Support

Andrew-Chen-Wang:
Adds initial async support from aioredis.

Things to do:
- [x] Debate on whether to add the parallel test_commands.py from aioredis (I don't think it's necessary; lots of duplicate code) when logic is the same. The only async test cases needed for commands is when something is not implemented
- [ ] Add Redis modules testing
- [ ] also make sure this stuff actually works. Ensure uvloop env is running
- [ ] Add cluster support

TODO in other PRs:
- De-duplicate code
- Implement 100% coverage of commands on async side

PR #4221, opened 2026-07-26
Add configurable encoder classes to standalone clients

### LINKED ISSUE #3181: Configurable Encoder
https://github.com/redis/redis-py/blob/ddff7b54b5db4ace94203dc141151b3f08060d2a/redis/_parsers/encoders.py#L4

Feature request: Make Encoder configurable to allow users customize the transformation of keys and values.
- namespacing (key prefix)
- different encode/decode algorithm
- encode/decode user-defined types, like dict or numpy.ndarray

```py
class MyEncoder(Encoder):
  ...

client = Redis(host=..., encoder=MyEncoder())
```

lh0156:
## What

Adds support for configuring a custom `Encoder` subclass on standalone sync and asyncio clients and their connection pools.

Closes #3181

## Why

The default encoder rejects application-specific Python value types. Users can now implement custom serialization/deserialization while retaining redis-py's existing encoding configuration.

## Design

- Adds `encoder_class` to sync and asyncio `Redis` constructors.
- Propagates the class to the connection pool and connection instances.
- Instantiates an encoder per connection/helper instead of sharing mutable encoder instances.
- Exposes `Encoder` from `redis` and `redis.asyncio`.
- Preserves the default `Encoder` behavior when no custom class is supplied.
- Documents the connection-isolation contract.
- Limits this change to standalone clients and pools; Cluster/Sentinel behavior is unchanged.

## Tests

- Added sync and asyncio regression coverage for:
  - custom encoding and decoding
  - pool-created connection encoders
  - pool helper encoders being distinct instances
  - client-to-pool configuration propagation
- `157 passed, 4 skipped` for `tests/test_connection.py tests/test_asyncio/test_connection.py`
- `17 passed` for `tests/test_client.py tests/test_asyncio/test_client.py`
- Ruff check/format, compileall, and `git diff --check` pass.

This is a draft while CI and maintainer feedback run.

### DIFF
--- a/docs/advanced_features.rst
+++ b/docs/advanced_features.rst
@@ -23,6 +23,32 @@ a separate connection pool) for each database.
 
 It is not safe to pass PubSub or Pipeline objects between threads.
 
+Custom encoders
+---------------
+
+Redis clients use an ``Encoder`` to serialize command arguments
+and deserialize responses. Applications that need to support additional
+Python value types can provide an ``encoder_class``. The class is
+instantiated separately for each connection, so it must accept the same
+``encoding``, ``encoding_errors``, and ``decode_responses`` arguments as the
+default encoder:
+
+.. code:: python
+
+   from redis import Encoder, Redis
+
+   class ApplicationEncoder(Encoder):
+       def encode(self, value):
+           if isinstance(value, ApplicationValue):
+               return value.to_bytes()
+           return super().encode(value)
+
+   client = Redis(encoder_class=ApplicationEncoder)
+
+The same option is available on ``redis.asyncio.Redis`` and both standalone
+connection-pool classes. Passing an encoder class, rather than a shared
+encoder instance, keeps connection-specific encoder state isolated.
+
 Pipelines
 ---------
 
--- a/redis/__init__.py
+++ b/redis/__init__.py
@@ -2,6 +2,7 @@ from redis import asyncio  # noqa
 from redis.backoff import default_backoff
 from redis.client import Redis, StrictRedis
 from redis.driver_info import DriverInfo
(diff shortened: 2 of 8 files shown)


petyaslavova:
Thanks for the detailed work here, and for scoping it to the standalone clients.

I traced the encoder through both serializer and parser backends before reviewing the diff, and unfortunately the Encoder is not the only serialization boundary in the client. When the hiredis extra is installed — a supported configuration we test in CI — Connection._construct_command_packer returns HiredisRespSerializer, which calls hiredis.pack_command() directly and never invokes Encoder.encode, and _HiredisParser builds hiredis.Reader from encoding/encoding_errors/decode_responses only, so Encoder.decode is never called either. The ApplicationEncoder example added in docs/advanced_features.rst would raise DataError for those users rather than encoding the value. The added tests call encoder.encode()/encoder.decode() on the encoder object directly, so they pass without exercising that path.

Separately, the async stack has accepted encoder_class on AbstractConnection and ConnectionPool since #1899, so redis.asyncio.Redis.from_url(url, encoder_class=MyEncoder) already works today; the genuinely missing piece is the sync side. And encoder_class is not in REDIS_ALLOWED_KEYS, so redis.RedisCluster(encoder_class=...) would be silently dropped.

Given that, I'm going to close this PR. If we revisit the sync/async asymmetry, we'd want a much narrower change — encoder_class on the sync AbstractConnection and ConnectionPool.get_encoder only, mirroring async — together with a decision on whether the hiredis pack/decode paths route through the encoder or the limitation is documented. If you'd like to pursue that narrower scope, please open a fresh PR with a regression test that packs a real command and round-trips a response under both the pure-Python and hiredis backends.

Issue #4208, opened 2026-07-26
URL query values are percent-decoded twice


winklemad:
**Version**: `master` at `88d16d0db10c852d1d292628b2c832f1ed117056`

**Redis version**: Not applicable; this occurs before connecting to Redis.

**Platform**: Python 3.10.9 on macOS

**Description**:

`ConnectionPool.from_url` decodes URL query values twice:

```python
import redis

pool = redis.ConnectionPool.from_url(
    "redis://localhost?client_name=worker%2520name"
)
print(pool.connection_kwargs["client_name"])
```

Expected:

```text
worker%20name
```

Actual:

```text
worker name
```

This mutates literal percent-encoded sequences in query options, so callers
cannot preserve values such as the intended client name.

vladvildanov:
@winklemad Hey! Thanks for addressing this, would take care of it

vladvildanov:
@winklemad #4222 should address this issue, feel free to leave your comments on that

PR #4222, opened 2026-07-27
fix: Fixed double decoding issue with unquote()

vladvildanov:
### Description of change

`ConnectionPool.from_url` (and everything that routes through parse_url — Redis.from_url, the sync/async cluster clients, and Sentinel) percent-decodes URL query values twice, corrupting any value that contains a literal percent-encoded sequence.

`urllib.parse.parse_qs()` already percent-decodes the values it returns, but parse_url then called `urllib.parse.unquote()` on the result a second time:
                                                           
```                                                                                                                                                                                                                 
for name, value in parse_qs(url.query).items():                                                                                                                                                                
    if value and len(value) > 0:                                                                                                                                                                               
        value = unquote(value[0])   # ← second, erroneous decode
```

### Fix

Drop the redundant second decode in both stacks and use the value parse_qs already returned:                                                                                                                   
                                                                                                                                                                                                               
- redis/connection.py — parse_url()                                                                                                                                                                            
- redis/asyncio/connection.py — parse_url()                                                                                                                                                                    
                                                                                                                                                                                                                 
The unquote() calls on username, password, path, and hostname are intentionally left unchanged: urlparse() does not decode those components, so those are the correct single decode. + handling is also unaffected — parse_qs already applies form-encoding semantics (+ → space), and unquote never touched +.

### Pull Request check-list

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Low Risk**
> Narrow URL parsing fix with targeted tests; behavior change only for query values that were incorrectly double-decoded.
> 
> **Overview**
> **Fixes double percent-decoding of query-string values** in `parse_url()` for sync and async connection URLs (`redis://`, `rediss://`, `unix://`), including `ConnectionPool.from_url` / `Redis.from_url` and other callers of that helper (#4208).
> 
> Query parameters were passed through `urllib.parse.unquote()` after `parse_qs()` had already decoded them, so values like `client_name=worker%2520name` became a space instead of the intended literal `%20`. The redundant `unquote` is removed; **username, password, path, and hostname** still use `unquote` because `urlparse` does not decode those parts.
> 
> Sync and asyncio connection-pool tests cover single decode (`worker%20name` → space) and no second decode (`worker%2520name` → `worker%20name`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2af8435108e9d3b6638fc57444124236704f5cba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1749,7 +1749,10 @@ def parse_url(url: str) -> ConnectKwargs:
 
     for name, value_list in parse_qs(parsed.query).items():
         if value_list and len(value_list) > 0:
-            value = unquote(value_list[0])
+            # parse_qs() already percent-decodes query values, so use the value
+            # as-is; unquoting again here would double-decode (e.g. "%2520" ->
+            # "%20" -> " "). See issue #4208.
+            value = value_list[0]
             parser = URL_QUERY_ARGUMENT_PARSERS.get(name)
             if parser:
                 try:
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -2291,7 +2291,10 @@ def parse_url(url):
 
     for name, value in parse_qs(url.query).items():
         if value and len(value) > 0:
-            value = unquote(value[0])
+            # parse_qs() already percent-decodes query values, so use the value
+            # as-is; unquoting again here would double-decode (e.g. "%2520" ->
+            # "%20" -> " "). See issue #4208.
+            value = value[0]
             parser = URL_QUERY_ARGUMENT_PARSERS.get(name)
             if parser:
                 try:
--- a/tests/test_asyncio/test_connection_pool.py
+++ b/tests/test_asyncio/test_connection_pool.py
@@ -617,6 +617,21 @@ class TestConnectionPoolURLParsing:
         pool = redis.ConnectionPool.from_url("redis://location?client_name=test-client")
         assert pool.connection_kwargs["client_name"] == "test-client"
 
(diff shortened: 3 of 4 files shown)


winklemad:
Thanks for the quick turnaround, @vladvildanov. The fix looks correct to me. `parse_qs` already percent-decodes the values it returns (and folds `+` to a space), so the follow-up `unquote()` was a genuine second decode. Walking the exact change through with the issue's input: `parse_qs("client_name=worker%2520name")` gives back `worker%20name`, the old `unquote()` turned that into `worker name`, and dropping it keeps `worker%20name` — which is the result the issue expected. The two regression tests pin down both the decode-once and don't-decode-twice cases.

I also went through the surrounding paths to be sure the scope is right: the `username`/`password`/`host` and the unix `path` branches still call `unquote()` once each, and that's correct — `urlparse` hands those back still-encoded, so they each need exactly one pass. The query loop was the only spot getting a decode from `parse_qs` *and* a second one from `unquote`, so this is the right (and only) place to touch.

One pre-existing thing I noticed while reading, not for this PR: since `parse_qs` folds `+` into a space, a literal `+` in a query value comes out as a space, whereas a `+` in the userinfo stays a `+`. That's standard form-decoding and out of scope here — just flagging it in case it's ever worth a line in the docs. LGTM.

PR #4232, opened 2026-07-31
fix: Using lazy-logging in pub/sub push handler with more efficient string conversion

### LINKED ISSUE #4230: issue: Pub/sub push handler builds a debug string for EVERY message regardless of log level (25% of process CPU in production)
## Summary

`handle_pubsub_push_response` concatenates the full stringified push message into a debug log line as an *argument expression*, so `str(response)` is evaluated on every pub/sub message even when DEBUG logging is disabled. The result is then discarded.

```python
async def handle_pubsub_push_response(self, response):
    logger = getLogger("push_response")
    logger.debug("Push response: " + str(response))   # eager: runs at every log level
    return response
```

Because the concatenation happens before `logger.debug` is called, **no logging configuration avoids this cost**. Setting the `push_response` logger level, disabling the logger, or calling `logging.disable()` all still pay for the `str()` and the concatenation.

## Affected code

Four identical copies, sync and async, in both parsers (line numbers from 8.1.0):

- `redis/_parsers/hiredis.py:129` — `_HiredisParser.handle_pubsub_push_response`
- `redis/_parsers/hiredis.py:284` — `_AsyncHiredisParser.handle_pubsub_push_response`
- `redis/_parsers/resp3.py:28` — `_RESP3Parser.handle_pubsub_push_response`
- `redis/_parsers/resp3.py:173` — `_AsyncRESP3Parser.handle_pubsub_push_response`

Since it is in both parsers, switching between hiredis and the pure-Python parser does not avoid it. `DEFAULT_RESP_VERSION = 3` (`redis/utils.py`) means the RESP3 push path is the default, so any pub/sub consumer on redis-py 8.x is affected.

## Impact

Found while profiling a production deployment (16 replicas of a chat application using `python-socketio`'s `AsyncRedisManager` for pub/sub fan-out) with py-spy. A 15-minute sampled profile attributed **25.8% of all process CPU** to this single line. The largest consumer in the process by a wide margin:

```
run (runners.py:118)
  _thread (async_pubsub_manager.py:232)
    _listen (async_redis_manager.py:171)
      listen (client.py:1451)
        parse_response (client.py:1295)
          read_response (connection.py:792)
            read_response (hiredis.py:306)
              handle_push_response (base.py:431)
                handle_pubsub_push_response (hiredis.py:219)   <-- 25.8% of total CPU
```

Because every replica subscribes to the same channel, each one stringifies every message published by any replica, so the cost scales with `message_rate x replica_count x payload_size`.

## Reproduction

```python
import json, timeit, logging
from logging import getLogger
logging.basicConfig(level=logging.WARNING)   # DEBUG disabled

def payload(n):
    body = json.dumps({"type": "message", "data": {"content": "x" * n}})
    return [b"message", b"channel", body.encode()]

for n in (200, 2000, 20000):
    r = payload(n)
    eager = timeit.timeit(
        lambda: getLogger("push_response").debug("Push response: " + str(r)),
        number=2000) / 2000
    lazy = timeit.timeit(
        lambda: getLogger("push_response").debug("Push response: %s", r),
        number=2000) / 2000
    print(f"payload {len(r[2]):>6} B   eager {eager*1e6:7.1f} us   "
          f"lazy {lazy*1e6:5.1f} us   ratio {eager/lazy:6.1f}x")
```

Result (CPython 3.11, DEBUG disabled):

| payload | eager (current) | lazy (`%s`) | ratio |
|--------:|----------------:|------------:|------:|
|   298 B |          2.0 us |      0.6 us |  3.3x |
|  2098 B |          7.8 us |      0.6 us | 12.5x |
| 20098 B |         64.2 us |      0.6 us | 115.8x |

Lazy formatting is constant because the message is never formatted; the eager version scales linearly with payload size.

## Proof that the work is wasted, and that the fix is behaviour-preserving

```python
import logging, io
from logging import getLogger

calls = []
class Sentinel:
    def __init__(self, tag): self.tag = tag
    def __str__(self):
        calls.append(self.tag)
        return f"<payload {self.tag}>"

def run(level, label):
    calls.clear()
    root = logging.getLogger()
    for h in root.handlers[:]: root.removeHandler(h)
    buf = io.StringIO()
    root.addHandler(logging.StreamHandler(buf)); root.setLevel(level)

    logger = getLogger("push_response")
    logger.debug("Push response: " + str(Sentinel("EAGER")))   # current
    logger.debug("Push response: %s", Sentinel("LAZY"))        # proposed

    print(f"{label}: __str__ called for {calls or 'nothing'}; "
          f"emitted {buf.getvalue().strip()!r}")

run(logging.WARNING, "WARNING (default)")
run(logging.DEBUG,   "DEBUG")
```

Output:

```
WARNING (default): __str__ called for ['EAGER']; emitted ''
DEBUG:             __str__ called for ['EAGER', 'LAZY']; emitted 'Push response: <payload EAGER>\nPush response: <payload LAZY>'
```

At the default log level the current form stringifies the payload and the result is
discarded; the lazy form does no work at all. With DEBUG enabled both produce identical
output. `getLogger("push_response").isEnabledFor(logging.DEBUG)` is `False` under default
logging configuration, so the wasteful path is the one that runs in production.

## Proposed fix

Use lazy `%s` formatting, which is the standard idiom and produces byte-identical output when DEBUG is enabled (`%s` calls `str()`):

```python
logger.debug("Push response: %s", response)
```

Optionally, hoist the logger to module level as well. `getLogger()` is a dict lookup under a module-level lock and accounts for most of the remaining cost once the formatting is lazy: measured 0.53 us -> 0.15 us per call with a module-level logger.

Happy to open a PR for this if the approach looks right.

## Environment

- redis-py 8.0.1 (verified the same code is present in 8.1.0)
- hiredis parser in use; the pure-Python RESP3 parser has the same issue
- RESP3 (default)
- Consumer: `python-socketio` 5.16.2 `AsyncRedisManager`

vladvildanov:
### Description of change

Closes #4230 

### Pull Request check-list

---

> [!NOTE]
> **Low Risk**
> Logging-only performance tweak on parser internals with no API or protocol behavior changes.
> 
> **Overview**
> Optimizes the **pub/sub push response** debug path in `hiredis.py` and `resp3.py` (sync and async parsers) so high-volume pub/sub traffic does less work when debug logging is off.
> 
> A module-level `_pubsub_push_logger` replaces per-message `getLogger("push_response")`, and debug lines use **`%s` lazy formatting** instead of concatenating with `str(response)`, so stringification only happens when DEBUG is enabled. **Behavior is unchanged** for callers; only logging cost on the hot path is reduced.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e9d2b4385ab110daaf5fb9cf70201057eb55fbc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/_parsers/hiredis.py
+++ b/redis/_parsers/hiredis.py
@@ -19,6 +19,11 @@ from .socket import (
     SERVER_CLOSED_CONNECTION_ERROR,
 )
 
+# Resolved once at import time rather than per message: the pub/sub push handler
+# is on the hot path (called for every message) and ``getLogger`` is a locked
+# dict lookup.
+_pubsub_push_logger = getLogger("push_response")
+
 # Used to signal that hiredis-py does not have enough data to parse.
 # Using `False` or `None` is not reliable, given that the parser can
 # return `False` or `None` for legitimate reasons from RESP payloads.
@@ -126,8 +131,9 @@ class _HiredisParser(BaseParser, PushNotificationsParser):
             pass
 
     def handle_pubsub_push_response(self, response):
-        logger = getLogger("push_response")
-        logger.debug("Push response: " + str(response))
+        # Lazy ``%s`` formatting: ``str(response)`` is only paid when DEBUG is
+        # actually enabled, not on every message.
+        _pubsub_push_logger.debug("Push response: %s", response)
         return response
 
     def on_connect(self, connection, **kwargs):
@@ -281,8 +287,9 @@ class _AsyncHiredisParser(AsyncBaseParser, AsyncPushNotificationsParser):
         self._hiredis_PushNotificationType = None
 
     async def handle_pubsub_push_response(self, response):
-        logger = getLogger("push_response")
-        logger.debug("Push response: " + str(response))
+        # Lazy ``%s`` formatting: ``str(response)`` is only paid when DEBUG is
+        # actually enabled, not on every message.
+        _pubsub_push_logger.debug("Push response: %s", response)
         return response
 
     def on_connect(self, connection):
--- a/redis/_parsers/resp3.py
+++ b/redis/_parsers/resp3.py
@@ -12,6 +12,11 @@ from .base import (
 )
 from .socket import SERVER_CLOSED_CONNECTION_ERROR
 
+# Resolved once at import time rather than per message: the pub/sub push handler
+# is on the hot path (called for every message) and ``getLogger`` is a locked
+# dict lookup.
+_pubsub_push_logger = getLogger("push_response")
+
 
 class _RESP3Parser(_RESPBase, PushNotificationsParser):
     """RESP3 protocol implementation"""
@@ -25,8 +30,9 @@ class _RESP3Parser(_RESPBase, PushNotificationsParser):
         self.invalidation_push_handler_func = None
 
     def handle_pubsub_push_response(self, response):
-        logger = getLogger("push_response")
-        logger.debug("Push response: " + str(response))
+        # Lazy ``%s`` formatting: ``str(response)`` is only paid when DEBUG is
+        # actually enabled, not on every message.
+        _pubsub_push_logger.debug("Push response: %s", response)
         return response
 
     def read_response(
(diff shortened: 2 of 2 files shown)


Classic298:
@vladvildanov Thanks for the quick turnaround! One thing on the getLogger hoist though.

It moves logger creation from "first message" to "redis import". `dictConfig` has `disable_existing_loggers=True` by default, so if the app configures logging after importing redis (the normal order), the logger gets disabled and push_response DEBUG
output never shows up even when you explicitly turn it on.

uvicorn's built-in config sets `disable_existing_loggers: False` so stock deployments are fine, but a custom `--log-config` that omits the key gets `True`.

Measured the hoist at 0.53us -> 0.15us per message. The `%s` change alone saves 2-64us depending on payload. Your call, just flagging the hoist isn't free the way `%s` is.

vladvildanov:
@Classic298 I'll closing this one and proceed with your fix

PR #4234, opened 2026-08-01
test: add unit tests for pairs_to_dict and pairs_to_dict_typed

eeshsaxena:
### Description of change

`pairs_to_dict` and `pairs_to_dict_typed` in `redis/_parsers/helpers.py`
turn a flat `[k1, v1, k2, v2]` reply into a dict and are used throughout the
response callbacks, but they were only exercised indirectly through
integration tests that need a live server.

This adds direct unit tests (no server required) covering:

- `pairs_to_dict`: `None`/empty input, the `decode_keys` / `decode_string_values`
  flags (individually and together), odd-length input (the unpaired trailing
  item is dropped by `zip`, not an error), non-bytes values passing through
  `str_if_bytes` unchanged, and duplicate keys.
- `pairs_to_dict_typed`: type coercion for known keys, the fallback that keeps
  the raw value when coercion raises, untyped keys left untouched, mixed
  typed/untyped keys, and odd-length input.

No production code is changed.

### Pull Request check-list

- [x] Is the new/changed code covered by tests?
- [ ] Does this change require documentation? (no — tests only)

---

> [!NOTE]
> **Low Risk**
> Test-only change with no modifications to library behavior or production paths.
> 
> **Overview**
> Adds **direct unit tests** in `tests/test_parsers/test_helpers.py` for `pairs_to_dict` and `pairs_to_dict_typed` from `redis/_parsers/helpers.py`, so flat `[k1, v1, k2, v2]` reply parsing is covered **without a live Redis server** (previously only hit via integration tests).
> 
> For **`pairs_to_dict`**, tests cover `None`/empty input, passthrough without decode flags, `decode_keys` / `decode_string_values` alone and together, and non-bytes values staying unchanged when decoding values.
> 
> For **`pairs_to_dict_typed`**, tests cover empty input, coercion for keys in `type_info`, keeping the raw value when coercion fails, unknown keys unchanged, and mixed typed/untyped keys.
> 
> **No production code changes.**
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6916db8c119948f3102cb547f4a61769a000d415. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_parser_helpers.py
```diff
@@ -0,0 +1,83 @@
...
+
+``pairs_to_dict`` and ``pairs_to_dict_typed`` turn a flat ``[k1, v1, k2, v2]``
+reply into a dict and are used all over the response callbacks, but they were
+only exercised indirectly through integration tests. These cover their
+behavior (decoding flags, odd-length input, type coercion and its fallback)
+without needing a live server.
+"""
+
+from redis._parsers.helpers import pairs_to_dict, pairs_to_dict_typed
+
```
petyaslavova: All those tests need to be marked as fixed_client tests.

tests/test_parser_helpers.py
petyaslavova, line 1: Please move both test classes into the existing tests/test_parsers/test_helpers.py, which is
   already the dedicated module for redis._parsers.helpers, and drop the new
   tests/test_parser_helpers.py. Note that file uses module-level test functions rather than classes.

tests/test_parser_helpers.py
```diff
@@ -0,0 +1,83 @@
+"""Unit tests for the pure response-shaping helpers in
```
petyaslavova: This comment is more appropriate to be added to the concrete class - since the tests need to be moved to the appropriate test module and they won't be the only tests there, it won't be suitable for a module-level docstring.

tests/test_parser_helpers.py
```diff
@@ -0,0 +1,83 @@
...
+        # Values are decoded to str; keys are left untouched.
+        assert pairs_to_dict([b"a", b"1"], decode_string_values=True) == {b"a": "1"}
+
+    def test_decode_keys_and_values(self):
+        result = pairs_to_dict(
+            [b"a", b"1", b"b", b"2"], decode_keys=True, decode_string_values=True
+        )
+        assert result == {"a": "1", "b": "2"}
+
+    def test_odd_length_drops_unpaired_trailing_item(self):
```
petyaslavova: `test_odd_length_drops_unpaired_trailing_item` and `test_duplicate_keys_last_value_wins`
pin zip()/dict() artifacts rather than a documented contract, so they would fail if we ever decided a
malformed odd-length reply should raise.

petyaslavova (changes requested):
Hey @eeshsaxena, thank you for your contribution! I have a few minor comments that need to be applied before merging this.

eeshsaxena:
Thanks for the detailed feedback, @petyaslavova. Applied all of it:

- Moved both test groups into `tests/test_parsers/test_helpers.py` as module-level functions matching that file's style, and deleted the standalone `tests/test_parser_helpers.py`.
- Marked each test with `@pytest.mark.fixed_client`.
- Dropped the module-level docstring.
- Removed the odd-length and duplicate-key tests, since as you noted they pinned zip()/dict() artifacts rather than a documented contract.

petyaslavova (approved):
LGTM.

PR #4235, opened 2026-08-01
fix(zadd): correct the mutually-exclusive-option error message

eeshsaxena:
### Description of change

When `ZADD` is called with `nx` together with `gt` or `lt`, the raised
`DataError` names a non-existent option `'gr'`:

```python
>>> r.zadd("k", {"m": 1}, nx=True, gt=True)
redis.exceptions.DataError: Only one of 'nx', 'lt', or 'gr' may be defined.
```

`'gr'` is a typo for `'gt'`, so the message points users at an option that
does not exist. This corrects it to `'gt'` (and orders it `'nx', 'gt', 'lt'`
to match the check `if nx and (gt or lt)`).

The error type and the condition that triggers it are unchanged - only the
message text is fixed. The message lives in the single shared `zadd`
definition in `redis/commands/core.py`, so sync and async are covered
together.

### Pull Request check-list

- [x] Is the new/changed code covered by tests?
- [ ] Does this change require documentation? (no)

---

> [!NOTE]
> **Low Risk**
> Single-string fix to a client-side validation error message with no logic or runtime behavior changes.
> 
> **Overview**
> Fixes a typo in the **`ZADD`** client validation error when **`nx`** is combined with **`gt`** or **`lt`**: the message incorrectly referred to **`'gr'`** instead of **`'gt'`**, and now lists **`'nx', 'gt', 'lt'`** in an order that matches the validation check.
> 
> Validation behavior is unchanged—only the **`DataError`** text in **`redis/commands/core.py`** is updated, which applies to both sync and async clients.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7da131217719723164307020dd14090cdf32c61a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_zadd_validation.py
petyaslavova, line 1: Please remove this test file and the tests. The options are already validated in `tests/test_commands.py`

petyaslavova (changes requested):
Hey @eeshsaxena, thanks for the msg fix! Please take a look at the requested changes, once applied and we have a green CICD, the PR will be ready for merging.

eeshsaxena:
Thanks for the review, @petyaslavova. Removed `tests/test_zadd_validation.py` as suggested, since those option combinations are already covered in `tests/test_commands.py`. The PR is now just the error-message fix in `core.py`.

petyaslavova (approved):
LGTM.

PR #4241, opened 2026-08-02
Distinguish connection health check timeouts

### LINKED ISSUE #4240: on_connect_check_health raises indistinguishable TimeoutError from pipeline response read timeout
## Bug

`pipeline.execute()` can raise `redis.exceptions.TimeoutError: Timeout reading from <host>`
from two completely different code paths that have opposite semantics:

1. **Response read timeout** (`_execute_pipeline` → `read_response`): The write commands already landed in Redis. The timeout occurred while reading the response back. The write succeeded.

2. **Connection health check timeout** (`get_connection` → `ensure_connection` → `connect` → `connect_check_health` → `on_connect_check_health` → `read_response`): The connection could not be established. The write never reached Redis.

Both raise identical exceptions:
```
redis.exceptions.TimeoutError: Timeout reading from <host>:6379
```

There is no attribute, subclass, or message difference to distinguish them.

## Impact

Any caller that needs to treat these differently (e.g. to decide whether to retry a write or treat it as already applied) must resort to walking `exc.__traceback__` and inspecting frame names — a fragile approach that breaks if redis-py internal function names change.

## Repro

The traceback below comes from `get_connection` during pipeline execution — the write never reached Redis, but the exception is indistinguishable from a response read timeout:

```
File "redis/asyncio/client.py", line 1662, in execute
    conn = await self.connection_pool.get_connection()
  File "redis/asyncio/connection.py", line 1259, in get_connection
    await self.ensure_connection(connection)
  ...
  File "redis/asyncio/connection.py", line 355, in connect_check_health
    await self.on_connect_check_health(check_health=check_health)
  File "redis/asyncio/connection.py", line 512, in on_connect_check_health
    await self.read_response()
  File "redis/asyncio/connection.py", line 660, in read_response
    raise TimeoutError(f"Timeout reading from {host_error}")
redis.exceptions.TimeoutError: Timeout reading from <host>:6379
```

## Proposed Fix

`on_connect_check_health` (and `connect_check_health`) should catch `TimeoutError` from `read_response` and re-raise as `ConnectionError`, which already semantically represents "failed to establish a usable connection":

```python
async def on_connect_check_health(self, check_health: bool = True) -> None:
    ...
    try:
        await self.read_response()
    except TimeoutError as e:
        raise ConnectionError(f"Timeout during connection health check to {self.host}:{self.port}") from e
```

This makes `TimeoutError` from `pipeline.execute()` unambiguously mean "response read timed out after write was sent."

## Version

redis-py 7.2.0

lh0156:
## What

Convert a timeout raised while establishing and health-checking a connection into `ConnectionError`, while preserving `TimeoutError` for response reads after a command has been sent. This lets callers distinguish a write that may have reached Redis from a connection that was never usable.

The behavior is mirrored in sync and asyncio connection implementations, and the existing disconnect cleanup path is preserved.

## Tests

- Added deterministic sync and asyncio regression tests for handshake read timeouts.
- `tests/test_connection.py`: 93 passed, 1 skipped
- `tests/test_asyncio/test_connection.py`: 65 passed, 1 skipped
- Ruff check, format, and `git diff --check` pass.

Fixes #4240

---

> [!NOTE]
> **Medium Risk**
> Changes error types on the connect path, which can affect retry and exception-handling logic for callers that only caught TimeoutError during connect.
> 
> **Overview**
> **Connect-time read timeouts** during the post-socket handshake (`on_connect_check_health` / custom `redis_connect_func`) are now turned into **`ConnectionError`** with a clear “connection health check” message, after **`disconnect()`**—matching the existing cleanup for other handshake failures.
> 
> **`TimeoutError` is unchanged** for timeouts while reading a command response, so callers can tell “connection never became usable” from “a write may have reached Redis.”
> 
> The same behavior is applied in **sync** (`redis/connection.py`) and **asyncio** (`redis/asyncio/connection.py`). **Regression tests** cover both stacks when `read_response` times out during `connect_check_health`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36e0eac05c1fee437193c3d609d86ca8d0d0bae1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -888,6 +888,11 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
                     if asyncio.iscoroutinefunction(self.redis_connect_func)
                     else self.redis_connect_func(self)
                 )
+        except TimeoutError as error:
+            await self.disconnect()
+            raise ConnectionError(
+                f"Timeout during connection health check to {self._host_error()}"
+            ) from error
         except RedisError:
             # clean up after any error in on_connect
             await self.disconnect()
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -1084,6 +1084,11 @@ class AbstractConnection(MaintNotificationsAbstractConnection, ConnectionInterfa
             else:
                 # Use the passed function redis_connect_func
                 self.redis_connect_func(self)
+        except TimeoutError as error:
+            self.disconnect()
+            raise ConnectionError(
+                f"Timeout during connection health check to {self._host_error()}"
+            ) from error
         except RedisError:
             # clean up after any error in on_connect
             self.disconnect()
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -256,6 +256,22 @@ async def test_client_setinfo_skipped_with_explicit_none(connection_kwargs):
     conn.read_response.assert_not_awaited()
 
 
+async def test_connect_check_health_timeout_is_connection_error():
+    conn = Connection(protocol=3, driver_info=None)
+    conn._connect = mock.AsyncMock()
+    conn._parser.on_connect = mock.Mock()
+    conn.send_command = mock.AsyncMock()
+    conn.read_response = mock.AsyncMock(
+        side_effect=TimeoutError("Timeout reading from localhost:6379")
+    )
+    conn.disconnect = mock.AsyncMock()
+
+    with pytest.raises(ConnectionError, match="connection health check"):
+        await conn.connect_check_health(retry_socket_connect=False)
+
+    conn.disconnect.assert_awaited_once_with()
+
+
 @pytest.mark.onlynoncluster
 async def test_invalid_response(create_redis):
     r = await create_redis(single_connection_client=True)
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -631,6 +631,21 @@ class TestConnection:
             assert _connect.call_count == 1
             self.clear(conn)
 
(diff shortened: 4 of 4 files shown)


petyaslavova:
Thanks for the report and the patch. The ambiguity you found is real — the handshake read in
on_connect_check_health and a command response read both raise
TimeoutError("Timeout reading from <host>:<port>"), byte for byte.

I don't think this change gets us to the invariant it's aiming for, though. Three other timeouts
still surface as plain TimeoutError before your commands are written: the TCP connect timeout
("Timeout connecting to server"), the periodic health-check PING in check_health() -> _send_ping()
(identical "Timeout reading from ..." message whenever health_check_interval is set), and
"Timeout writing to socket" from send_packed_command, which fires inside the sendall loop and so
means the batch was only partially written. Converting just the handshake case to ConnectionError
also splits connect() across two error types and changes a released exception type, so
`except redis.TimeoutError` around connect stops matching(which is unacceptable at this point, a breaking change).

There's also a design problem with any fix keyed on the connection phase: it would have to cover
connect, handshake, health-check PING and arbitrary caller-supplied redis_connect_func work under one
label. The distinction that actually matters is whether the pending command's bytes were written, not
which phase failed — and note that redis-py's default Retry already retries commands after a
response-read timeout, so the client itself doesn't currently honour this distinction.

I'm going to close this PR and leave #4240 open. If you can share a concrete case where treating
every timeout as "outcome unknown" (and adjusting the retry policy accordingly) is not workable,
we can revisit it as a command-lifecycle signal covering all four sites in both stacks.

PR #4243, opened 2026-08-03
Fix return_field ignoring decode_field for aliased fields

ritsth:
### Description of change

`Query.return_field()` records the per-field encoding under the field
identifier, but the server returns an aliased field under its alias, so the
lookup in `Result` never matches and `decode_field=False` is ignored:
```python
Query("*").return_field("$.vector_emb", as_field="vector_emb", decode_field=False)
# _return_fields_decode_as == {'$.vector_emb': None}, response key is 'vector_emb'
```
Because `to_string` decodes with `errors="ignore"`, a binary field is not just
decoded but silently truncated — `b"\x00\x01\x82\xff"` comes back as
`'\x00\x01'`, with no error raised. The same call without `as_field` works.

`HybridQuery._set_load_field_encodings` already keys this same map by the alias
when `AS` is present; this makes `return_field` consistent with it.

Keying by the alias fixes it. The generated command arguments are unchanged.
Added a test covering both the RESP2 and RESP3 result paths.

### Pull Request check-list

---

> [!NOTE]
> **Low Risk**
> Small, localized fix to search query result decoding with added tests; no API or command-shape changes.
> 
> **Overview**
> Fixes **`Query.return_field()`** so `_return_fields_decode_as` is keyed by the response field name: the **`AS` alias** when present, otherwise the original field id. Redis returns aliased fields under the alias, so the old key caused **`decode_field=False`** (and custom encodings) to be ignored and binary payloads to be decoded/truncated via `errors="ignore"`.
> 
> Command arguments are unchanged; behavior now matches **`HybridQuery._set_load_field_encodings`**. New unit tests cover RESP2/RESP3 **`Result`** decoding, plus a **`redismod`** integration test for aliased binary vector fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef13aa7cc7344aa36a39e5b8b6b8aa82d262ff11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -1500,6 +1500,55 @@ class TestBaseSearchFunctionality(SearchTestsBase):
         assert res.docs[0].vector_emb == raw
 
     @pytest.mark.redismod
+    def test_binary_field_survives_alias_with_decode_field_false(self, client):
+        # Regression test: return_field() used to key the decode-encoding map
+        # by the field identifier instead of the AS alias, so an aliased
+        # binary field was decoded (and silently truncated) instead of being
+        # kept raw. Covered against a real server across the protocol /
+        # legacy_responses matrix that CI runs this suite under.
+        fake_vec = np.array([0.1, 0.2, 0.3, 0.4], dtype=np.float32)
+
+        index_name = "aliased_binary_index"
+        client.hset(f"{index_name}:1", mapping={"vector_emb": fake_vec.tobytes()})
+
+        client.ft(index_name).create_index(
+            fields=(
+                VectorField(
+                    "vector_emb",
+                    algorithm="HNSW",
+                    attributes={
+                        "TYPE": "FLOAT32",
+                        "DIM": 4,
+                        "DISTANCE_METRIC": "COSINE",
+                    },
+                ),
+            ),
+            definition=IndexDefinition(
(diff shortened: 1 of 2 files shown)


tests/test_search.py
```diff
@@ -1473,6 +1473,32 @@ def test_binary_and_text_fields(self, client):
             "The text field is not decoded correctly"
         )
 
+    def test_return_field_encoding_is_keyed_by_alias(self):
```
petyaslavova, line 1476: It will be better to add a real integration test to cover the functionality  - this way you will validate it with all the possible combinations of `legacy_responses` and `protocol` settings.

petyaslavova (changes requested):
Hey @ritsth, thank you for your contribution! Please check my comment about the tests. Once the requested changes are applied, I think this PR will be ready for another review round!

tests/test_search.py
ritsth: @petyaslavova got it!

ritsth:
@petyaslavova Added a real integration test as requested that indexes a binary vector field and queries it through the aliased return_field(..., decode_field=False) path against a live server, so it's covered across the protocol/legacy_responses matrix.

petyaslavova (approved):
LGTM.

PR #4244, opened 2026-08-03
Pass zrevrange key as a list in options["keys"] to match sibling range commands

uttam12331:
## Summary

`SortedSetCommands.zrevrange` stores its key in `options["keys"]` as a **bare string**, while every sibling range command passes it as a **list**:

```python
# zrange / _zrange
options["keys"] = [name]
# zrangebyscore
options["keys"] = [name]
# zrevrangebyscore
options["keys"] = [name]
# zrevrange  <-- inconsistent
options["keys"] = name
```

## Impact

`options["keys"]` is consumed as a sequence of keys. When it is a bare string, consumers iterate it **character by character**. The clearest case is client-side caching in `redis/connection.py`:

```python
self._current_command_cache_key = CacheKey(
    command=args[0], redis_keys=tuple(kwargs.get("keys")), redis_args=args
)
```

For `zrevrange("myzset", 0, -1)` this builds `redis_keys=('m', 'y', 'z', 's', 'e', 't')` instead of `('myzset',)`, so the cache key / invalidation mapping for `zrevrange` is wrong.

## Fix

Wrap the key in a list, matching the other range commands:

```diff
-        options["keys"] = name
+        options["keys"] = [name]
```

## Tests

Added `test_zrevrange_passes_key_as_list`, which mocks `execute_command` and asserts `zrevrange` passes `keys=["myzset"]`. It fails on the previous behavior (`"myzset" != ["myzset"]`) and passes with the fix. Verified locally without a server.

---

> [!NOTE]
> **Low Risk**
> One-line behavioral fix in command options plus cache regression tests; no auth, security, or broad API changes.
> 
> **Overview**
> **`zrevrange`** now sets `options["keys"]` to `[name]` instead of a bare string, aligning with **`zrange`**, **`zrangebyscore`**, and other range commands.
> 
> That fixes client-side caching: `redis_keys` is built from `tuple(kwargs["keys"])`, so a string key was iterated **per character** (`('m','y','z',...)`) rather than as a single Redis key (`('myzset',)`), breaking cache lookup and invalidation for **`ZREVRANGE`**.
> 
> A regression test exercises cached **`zrevrange`** results and invalidation when the sorted set is updated from another client (single-connection and pool modes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f0d891ab733310178e36734f4498a72914709fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -5222,16 +5222,6 @@ class TestRedisCommands:
             [[b"a3", "3.0"], [b"a2", "2.0"]],
         )
 
-    def test_zrevrange_passes_key_as_list(self, r):
-        # zrevrange must pass its key in a list, like the sibling range
-        # commands (zrange, zrangebyscore, zrevrangebyscore). Passing a bare
-        # string means downstream consumers that iterate options["keys"]
-        # (e.g. client-side cache key construction) walk it character by
-        # character instead of treating it as a single key.
-        with mock.patch.object(r, "execute_command", return_value=[]) as execute_command:
-            r.zrevrange("myzset", 0, -1)
-        assert execute_command.call_args.kwargs["keys"] == ["myzset"]
-
     def test_zrevrangebyscore(self, r):
         r.zadd("a", {"a1": 1, "a2": 2, "a3": 3, "a4": 4, "a5": 5})
         assert r.zrevrangebyscore("a", 4, 2) == [b"a4", b"a3", b"a2"]
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -5222,6 +5222,16 @@ class TestRedisCommands:
             [[b"a3", "3.0"], [b"a2", "2.0"]],
         )
 
+    def test_zrevrange_passes_key_as_list(self, r):
+        # zrevrange must pass its key in a list, like the sibling range
+        # commands (zrange, zrangebyscore, zrevrangebyscore). Passing a bare
+        # string means downstream consumers that iterate options["keys"]
+        # (e.g. client-side cache key construction) walk it character by
+        # character instead of treating it as a single key.
+        with mock.patch.object(r, "execute_command", return_value=[]) as execute_command:
+            r.zrevrange("myzset", 0, -1)
+        assert execute_command.call_args.kwargs["keys"] == ["myzset"]
+
     def test_zrevrangebyscore(self, r):
         r.zadd("a", {"a1": 1, "a2": 2, "a3": 3, "a4": 4, "a5": 5})
         assert r.zrevrangebyscore("a", 4, 2) == [b"a4", b"a3", b"a2"]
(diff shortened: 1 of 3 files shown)


tests/test_commands.py
```diff
@@ -5222,6 +5222,16 @@ def test_zrevrange(self, r):
...
 
+    def test_zrevrange_passes_key_as_list(self, r):
+        # zrevrange must pass its key in a list, like the sibling range
+        # commands (zrange, zrangebyscore, zrevrangebyscore). Passing a bare
+        # string means downstream consumers that iterate options["keys"]
+        # (e.g. client-side cache key construction) walk it character by
+        # character instead of treating it as a single key.
+        with mock.patch.object(r, "execute_command", return_value=[]) as execute_command:
+            r.zrevrange("myzset", 0, -1)
+        assert execute_command.call_args.kwargs["keys"] == ["myzset"]
```
petyaslavova: It will be great if you also add a test that goes through the client-side caching case.. - this one can go into tests/test_cache.py

petyaslavova (changes requested):
Hey @uttam12331, thank you for your contribution!
That was a nice catch - the issue is a real defect, and it's really nice that you have found and fixed it!

Two things before we can merge:

1. Please check linters errors.
2. Please add the regression test at the level where the defect is observable, rather than asserting the internal `keys` kwarg. `tests/test_cache.py::test_get_from_given_cache` is a good model: cache a `zrevrange` result, mutate the sorted set from a second client, and assert both the CacheKey shape (`redis_keys=("myzset",)`) and that the entry is invalidated. That test fails on master and passes with your fix, and it protects the behavior rather than the plumbing.

Separately, and please do not expand this PR for it, in case you are interested in another contribution :) -->ZRANK and ZREVRANK are in the cache allow list but pass no `keys` at all, so they currently will raise ValueError("Cannot create cache key.") when client-side caching is enabled.

uttam12331:
Thanks for the review @petyaslavova, and glad it was a useful catch!

Addressed both points:
- **Linters:** fixed (removed the over-long line; `ruff check` + `ruff format --check` now clean).
- **Regression test:** moved it to the behavior level in `tests/test_cache.py` (`test_zrevrange_cache_key_uses_whole_key`, modeled on `test_get_from_given_cache`). It caches a `zrevrange` result, asserts the entry is stored under `redis_keys=("myzset",)` (rather than per-character), and that mutating the sorted set from a second client invalidates it. Removed the old test that asserted the internal `keys` kwarg.

And thanks for the ZRANK/ZREVRANK pointer — I'd be happy to take that on in a separate PR so this one stays focused.

petyaslavova (approved):
LGTM.

PR #4245, opened 2026-08-05
Clarify connection health check timeouts

daleselaji-dev:
## Summary\n\nFixes #4240\n\nA timeout while reading a response during connection establishment or a connection health PING is currently reported as the same TimeoutError used for command response reads after a write. This change converts only health-check/connection-establishment response timeouts to ConnectionError with an explicit message, preserving the original timeout as the exception cause.\n\nThe sync and asyncio connection implementations now share the same behavior.\n\n## Tests\n\n- python -m pytest tests/test_connection.py tests/test_asyncio/test_connection.py -k 'connection_health_check_timeout or connect_timeout_error_without_retry' -q\n- python -m ruff check redis/connection.py redis/asyncio/connection.py tests/test_connection.py tests/test_asyncio/test_connection.py\n- python -m compileall -q redis/connection.py redis/asyncio/connection.py\n\nThe full connection test files require a local Redis server; the relevant unit tests pass without one.

---

> [!NOTE]
> **Low Risk**
> Narrow exception-mapping change on connect/PING paths only, with sync/async parity and targeted unit tests; no change to post-connect command timeout behavior.
> 
> **Overview**
> **Connection establishment and health PING reads** now go through `_read_response_with_health_check()` instead of calling `read_response()` directly. If a read times out in those paths, redis-py raises **`ConnectionError`** with a message like *"Timeout during connection health check to …"*, with the original **`TimeoutError`** chained as `__cause__`.
> 
> The same helper is wired through the connect handshake (HELLO/AUTH, CLIENT SETNAME/SETINFO, SELECT) and **`_send_ping`** in both **`redis/connection.py`** and **`redis/asyncio/connection.py`**. Ordinary command response reads are unchanged and still surface **`TimeoutError`** when the socket read times out.
> 
> Unit tests assert the new exception type, message, and cause for sync and asyncio connections.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c2e5905c93c1042c30dd3b6e143795555fdea675. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -969,7 +969,7 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
             await self.send_command(
                 "HELLO", self.protocol, "AUTH", *auth_args, check_health=False
             )
-            response = await self.read_response()
+            response = await self._read_response_with_health_check()
             if response.get(b"proto") != int(self.protocol) and response.get(
                 "proto"
             ) != int(self.protocol):
@@ -980,14 +980,14 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
             await self.send_command("AUTH", *auth_args, check_health=False)
 
             try:
-                auth_response = await self.read_response()
+                auth_response = await self._read_response_with_health_check()
             except AuthenticationWrongNumberOfArgsError:
                 # a username and password were specified but the Redis
                 # server seems to be < 6.0.0 which expects a single password
                 # arg. retry auth with just the password.
                 # https://github.com/andymccurdy/redis-py/issues/1274
                 await self.send_command("AUTH", auth_args[-1], check_health=False)
-                auth_response = await self.read_response()
+                auth_response = await self._read_response_with_health_check()
 
             if str_if_bytes(auth_response) != "OK":
                 raise AuthenticationError("Invalid Username or Password")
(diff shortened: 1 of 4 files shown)


petyaslavova:
Hey @daleselaji-dev , thanks for picking this up, and for mirroring the change in both stacks.

This is the same design as #4241, which we closed recently, and the objections from that review still apply.
Wrapping `_send_ping` covers the periodic health-check PING, which is a real improvement over #4241, but three problems remain: a TCP connect timeout still raises `TimeoutError("Timeout connecting to server")` (`connect_check_health`), a timeout while *writing* HELLO/AUTH/SELECT still raises `TimeoutError("Timeout writing to socket")` (`send_packed_command`), and the `CLIENT MAINT_NOTIFICATIONS` read inside `on_connect_check_health` is not routed through the new helper. So `TimeoutError` out of `pipeline.execute()` still does not mean "the write reached Redis". Changing the exception type for a released connect path is also a breaking change we can't take outside a major release.

Also note that the added tests call `_read_response_with_health_check()` with a mocked `read_response`, so they assert the helper's own five lines rather than the reported scenario reaching `ConnectionError` through `get_connection()`.

I'm closing this PR and leaving #4240 open. It's still waiting on the question I asked there: since a timed-out response read gives no guarantee the command was processed server-side, we'd like a concrete case where treating every timeout as "outcome unknown" is not workable. With that, we can look at a command-lifecycle signal that covers all four timeout sites in both stacks instead of an exception-type change.

PR #4246, opened 2026-08-05
fix(asyncio): reject connection URLs missing the scheme separator

Arunendra21:
## Summary

The async `parse_url` in `redis/asyncio/connection.py` validated the URL scheme with `urlparse(url).scheme`, which accepts a URL that has a valid scheme name but is missing the `://` separator, for example `redis:foo.bar.com:12345`. In that case `urlparse` reports the scheme as `redis` with an empty netloc, so the function returned partial kwargs (no host/port) and the async client silently connected to the default host instead of raising.

The sync `parse_url` in `redis/connection.py` already guards against this up front and has a dedicated test (`test_invalid_scheme_raises_error_when_double_slash_missing`), but the async copy did not, so `redis.from_url("redis:foo.bar.com:12345")` and `redis.asyncio.from_url("redis:foo.bar.com:12345")` behaved differently.

## Changes

- Add the same up-front scheme check that the sync version uses, so a URL missing the `redis://`, `rediss://`, or `unix://` prefix raises `ValueError` with the same message. The now-unreachable trailing `else` branch is removed, matching the structure of the sync implementation.
- Add `test_invalid_scheme_raises_error_when_double_slash_missing` to the async connection-pool URL-parsing tests, mirroring the existing sync test.

## Testing

Verified that async `parse_url` / `ConnectionPool.from_url` now raises for `redis:foo.bar.com:12345`, `localhost`, and `http://localhost`, while valid `redis://`, `rediss://`, and `unix://` URLs (including host, port, db, and querystring options) still parse exactly as before. The sync and async implementations now return matching results for these inputs.

---

> [!NOTE]
> **Low Risk**
> Localized URL parsing only; behavior change is stricter rejection of bad URLs and broader acceptance of uppercase schemes on sync, with no auth or wire-protocol impact.
> 
> **Overview**
> **Sync and async `parse_url` now validate connection URLs the same way** before `urlparse`: the string must start with `redis://`, `rediss://`, or `unix://` (case-insensitive). Malformed values such as `redis:foo.bar.com:12345` raise `ValueError` instead of being accepted—async previously treated them as a valid `redis` scheme with no host and could connect to defaults.
> 
> The async parser gains the sync-style upfront check and drops the redundant trailing invalid-scheme branch. **Sync** `parse_url` switches from case-sensitive `startswith` to `url.lower().startswith(...)`, so **`REDIS://`, `REDISS://`, and `UNIX://` are now accepted** on the sync client as well.
> 
> Tests cover the missing-`://` case for async and uppercase schemes for both sync and async `ConnectionPool.from_url`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7137e006820a659cde2ac128f13a85aab689978d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Behavior change (for release notes)

While aligning the sync and async parsers and addressing review feedback, the checks are now case-insensitive per RFC 3986. As a result the **sync** `redis.from_url()` now also accepts uppercase schemes such as `REDIS://`, `REDISS://`, and `UNIX://` (the `urlparse`-based async parser already accepted these). This is a released-behavior change for the sync parser worth noting in the release notes; the missing-separator case (e.g. `redis:foo.bar.com:12345`) is still rejected.

Requiring the `://` separator also tightens two undocumented input forms on the **async** side, which the sync parser has rejected since #2343: the single-slash `unix:/tmp/redis.sock` form (only `unix:///tmp/redis.sock` is documented) and URLs with leading whitespace or control characters, e.g. a value read from a `.env` file as `"\nredis://host"`. Both previously worked on async because `urlparse` tolerated them, and both now raise `ValueError` with the same "Redis URL must specify one of the following schemes" message. Upgraders hitting that error should add the missing slash or strip the URL; trailing whitespace is unaffected.

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1766,11 +1766,10 @@ class ConnectKwargs(TypedDict, total=False):
 
 
 def parse_url(url: str) -> ConnectKwargs:
-    if not (
-        url.startswith("redis://")
-        or url.startswith("rediss://")
-        or url.startswith("unix://")
-    ):
+    # Scheme names are case-insensitive (RFC 3986), so normalize before the
+    # prefix check; the "://" is required so a URL like "redis:foo" (which
+    # urlparse would still report as the "redis" scheme) is rejected.
+    if not url.lower().startswith(("redis://", "rediss://", "unix://")):
         raise ValueError(
             "Redis URL must specify one of the following schemes "
             "(redis://, rediss://, unix://)"
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1766,6 +1766,16 @@ class ConnectKwargs(TypedDict, total=False):
 
 
 def parse_url(url: str) -> ConnectKwargs:
+    if not (
+        url.startswith("redis://")
+        or url.startswith("rediss://")
+        or url.startswith("unix://")
+    ):
+        raise ValueError(
+            "Redis URL must specify one of the following schemes "
+            "(redis://, rediss://, unix://)"
+        )
+
     parsed: ParseResult = urlparse(url)
     kwargs: ConnectKwargs = {}
 
@@ -1795,7 +1805,7 @@ def parse_url(url: str) -> ConnectKwargs:
             kwargs["path"] = unquote(parsed.path)
         kwargs["connection_class"] = UnixDomainSocketConnection
 
-    elif parsed.scheme in ("redis", "rediss"):
+    else:  # implied:  parsed.scheme in ("redis", "rediss")
         if parsed.hostname:
             kwargs["host"] = unquote(parsed.hostname)
         if parsed.port:
@@ -1811,12 +1821,6 @@ def parse_url(url: str) -> ConnectKwargs:
 
         if parsed.scheme == "rediss":
             kwargs["connection_class"] = SSLConnection
-
-    else:
-        valid_schemes = "redis://, rediss://, unix://"
-        raise ValueError(
-            f"Redis URL must specify one of the following schemes ({valid_schemes})"
-        )
 
     return kwargs
 
--- a/redis/connection.py
+++ b/redis/connection.py
@@ -2328,11 +2328,10 @@ URL_QUERY_ARGUMENT_PARSERS = {
 
 
 def parse_url(url):
-    if not (
-        url.startswith("redis://")
-        or url.startswith("rediss://")
-        or url.startswith("unix://")
-    ):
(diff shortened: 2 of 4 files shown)


Arunendra21:
Good catch, thanks. Updated the check to normalize the scheme case with `url.lower().startswith((...))`, so `REDIS://`, `Rediss://`, and `UNIX://` are accepted again while the missing-separator case (e.g. `redis:foo.bar.com:12345`) still raises. Applied the same change to the sync `parse_url` so the two stay consistent, and added an uppercase-scheme test to both.

petyaslavova (changes requested):
Hey @Arunendra21, thank you for your contribution! 

Two items before merge:

1. Please extend the uppercase coverage to `REDISS://` (asserting `SSLConnection`) and `UNIX://` (asserting `UnixDomainSocketConnection`) in both `tests/test_connection_pool.py` and `tests/test_asyncio/test_connection_pool.py`. Case-insensitive acceptance is new for the sync parser, and the TLS variant is the one worth pinning with a test.
2. Please mention in the PR description that sync `from_url` now accepts uppercase schemes — it is a released-behavior change we want captured in the release notes.

Arunendra21:
Thanks @petyaslavova! Both addressed:

1. Extended `test_uppercase_scheme_is_accepted` in both `tests/test_connection_pool.py` and `tests/test_asyncio/test_connection_pool.py` to also assert `REDISS://` → `SSLConnection` and `UNIX://` → `UnixDomainSocketConnection`.
2. Added a "Behavior change (for release notes)" section to the PR description noting that sync `from_url` now accepts uppercase schemes, while the missing-separator case is still rejected.

eeshsaxena (commented):
Reads correctly. The early `url.lower().startswith(("redis://", "rediss://", "unix://"))` guard rejects scheme-without-separator inputs like `redis:foo.bar.com:12345` that `urlparse` would otherwise report as scheme `redis`, which is the actual bug. Switching to `.lower()` also makes the check case-insensitive, so an uppercase `REDIS://...` is now accepted and lines up with `urlparse` lowercasing the scheme anyway.

The async refactor from `elif parsed.scheme in ("redis", "rediss")` to `else` is safe: the early guard constrains the scheme set to those three, and the `if parsed.scheme == "unix"` branch peels off unix, so `else` can only be redis/rediss. Sync and async stay in step.

petyaslavova (approved):
LGTM.

PR #4248, opened 2026-08-05
docs: add sliding window rate limiter example notebook

YashwinReddy29:
## What this adds

A new Jupyter notebook `docs/examples/rate_limiter_sliding_window.ipynb` 
showing two sliding window rate limiter implementations using Redis:

- Sorted set approach using ZADD and ZREMRANGEBYSCORE
- Atomic Lua script version that eliminates race conditions in 
  multi-instance deployments
- Async version using redis.asyncio

## Why it's useful

The existing examples cover data structures, search, pipelines, and 
connections. There is no example showing rate limiting — a very common 
Redis use case. The sliding window pattern is more accurate than fixed 
window and is the pattern most teams reach for in production.

## Context

Built this pattern in production for a distributed rate limiting 
microservice handling high-throughput concurrent traffic.

---

> [!NOTE]
> **Low Risk**
> Documentation-only addition; no runtime, library, or production code paths change.
> 
> **Overview**
> Adds **`docs/examples/rate_limiter_sliding_window.ipynb`**, a new documentation example for sliding-window rate limiting with Redis.
> 
> The notebook walks through a **sorted-set** limiter (`ZREMRANGEBYSCORE` / `ZCARD` / `ZADD` with key TTL), an **atomic Lua** variant via `register_script` for multi-instance safety, and an **`redis.asyncio`** pipeline version. Each section includes a small demo (e.g. 5 requests per 10 seconds) and documents expected allow/reject output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af4f39ecf144e17c845581e0e9b794678a373ba2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/examples/rate_limiter_sliding_window.ipynb
+++ b/docs/examples/rate_limiter_sliding_window.ipynb
@@ -1,0 +1,270 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Sliding Window Rate Limiter\n",
+    "\n",
+    "A sliding window rate limiter tracks requests within a rolling time window.\n",
+    "Unlike a fixed window (which resets at a set interval), a sliding window\n",
+    "prevents burst traffic at window boundaries.\n",
+    "\n",
+    "This example shows two implementations:\n",
+    "- **Sorted set approach** — precise sliding window using `ZADD` / `ZREMRANGEBYSCORE`\n",
+    "- **Lua script approach** — atomic version safe for concurrent multi-instance deployments"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Setup"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import time\n",
+    "import uuid\n",
+    "\n",
+    "import redis\n",
+    "\n",
+    "r = redis.Redis(host='localhost', port=6379, decode_responses=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sorted Set Implementation\n",
+    "\n",
+    "Each request is stored as a member of a sorted set with the timestamp as score.\n",
+    "On every request:\n",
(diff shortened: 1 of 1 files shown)


petyaslavova:
Hey @YashwinReddy29, thanks for putting this together, and for the detailed write-up!

Looking at the redis-py side of the notebook, the pieces it demonstrates are already documented: pipelines and transactions in `docs/examples/pipeline_examples.ipynb`, `register_script`/`Script` in `docs/lua_scripting.rst`, and the async client and async pipelines in `docs/examples/asyncio_examples.ipynb`. What is left is the sliding-window pattern itself, which is application-level design rather than client API usage, and that is better placed in the Redis patterns documentation on redis.io than in this repository's client examples.

Two things also stand out about the file as committed. All code cells are unexecuted, and nbsphinx's default `auto` mode executes any notebook with no outputs, so `invoke build-docs` would try to run it against `localhost:6379` and our Docs CI job has no Redis available. The notebook is also not added to the `docs/examples.rst` toctree, so it would not appear in the generated docs at all. Separately, `is_allowed_atomic` ignores its `client` argument and uses the module-level script bound to `r`, and the async section returns to the non-atomic read-then-write sequence that the Lua section explains is unsafe — `register_script` works the same way on `redis.asyncio`.

I am going to close this for now. If there is demand for rate limiting to be covered here, a short section in `docs/lua_scripting.rst` or a script under `doctests/` (which CI lints and can carry assertions) would be the lighter-maintenance home, and I would be happy to look at that.

YashwinReddy29:
Hi @petyaslavova, thank you for the detailed review and for taking the time
to explain exactly what works and what does not. The points about
unexecuted cells breaking CI, the toctree omission, and the
is_allowed_atomic bug ignoring the client argument are all valid — I should
have caught those before opening.

The suggestion about docs/lua_scripting.rst or a doctest script makes
sense. The sliding window pattern does sit at the application level rather
than the client API level, and a focused section in lua_scripting.rst would
be cleaner and more maintainable than a standalone notebook.

I will resubmit as a new PR targeting lua_scripting.rst with a
self-contained doctest that CI can actually run. Happy to have you take a
look at that when it is ready.

Thanks again.

PR #4256, opened 2026-08-09
Fix async cluster release race for closed connections

### LINKED ISSUE #4247: Async cluster: `ClusterNode.release()` defers returning an already-closed connection to `_free`, creating a transient false `MaxConnectionsError`
### Affected

- redis-py 8.1.0 and earlier (present on master @ `1627c98a`)
- **Async cluster only** (`redis/asyncio/cluster.py`). The sync cluster and both standalone pools do not have this window — see "Why the other pools don't have this" below.
- Requires an explicitly configured `max_connections` small enough for a node pool to reach capacity.

### Summary

`ClusterNode.release()` returns a connection marked for reconnect to the pool through a
background task instead of returning it inline. Between the `release()` call and the task's
first scheduling step, the connection is still in `self._connections` — so it counts against
`max_connections` — but is not in `self._free`, so it cannot be acquired. A concurrent command
arriving in that window gets `MaxConnectionsError` even though capacity is about to be
returned. The connection has already been closed at that point, so the task performs no I/O:
its only effect is deferring a `self._free.append(connection)` by one event-loop iteration.

### Code path

`redis/asyncio/cluster.py`:

```python
# ClusterNode.release
def release(self, connection: Connection) -> None:
    if connection.should_reconnect():
        task = asyncio.create_task(self._disconnect_and_release(connection))   # 1759
        self._background_tasks.add(task)
        task.add_done_callback(self._background_tasks.discard)
        return
    self._free.append(connection)

# ClusterNode._disconnect_and_release
async def _disconnect_and_release(self, connection: Connection) -> None:
    try:
        await connection.disconnect()      # no-op: connection is already disconnected
    except Exception as exc:
        ...
        return
    self._free.append(connection)          # the only real work, deferred by one loop turn
```

`acquire_connection()` raises `MaxConnectionsError` whenever `_free` is empty and
`len(self._connections) >= self.max_connections`, which is exactly the state during that window.

### Why the connection is always already closed when this branch is taken

Every caller of `release()` first awaits a disconnect — `execute_command`,
`execute_pipeline`, `TransactionStrategy` error path and `reset()`,
`_ClusterNodePoolAdapter.get_connection`/`release` - and
`AbstractConnection.disconnect()` calls `reset_should_reconnect()`
(`redis/asyncio/connection.py`) before its first suspension point, with its inner `finally`
nulling `_reader`/`_writer` even on the timeout path.

So `release()` can only observe `should_reconnect() == True` if the connection was
**re-marked during that awaited `disconnect()`**, at which point `is_connected` is already
`False` and `_disconnect_and_release()` early-returns at `if not self.is_connected`.

### Reachable triggers for the re-mark

1. `update_active_connections_for_reconnect()`, called from the
   `except (ConnectionError, TimeoutError)` handler in `_execute_command`. Requires two
   error events on the same node closely spaced, the second landing inside the first
   connection's `wait_closed()` suspension.
2. The SMIGRATED maintenance-notification handler
   (`redis/asyncio/maint_notifications.py`), which marks all in-use connections on affected
   nodes. **Note:** this requires `CLIENT MAINT_NOTIFICATIONS` support
   (`redis/asyncio/connection.py`), so it does not apply to deployments that do not
   negotiate it.

Both need the node pool to be simultaneously at `max_connections`.

### Impact

One transient `MaxConnectionsError` per occurrence, recoverable on the caller's next attempt.
It is **not** self-retried: `ERRORS_ALLOW_RETRY` (`redis/cluster.py`) is matched with
`type(e) in ...`, and `MaxConnectionsError` is a `ConnectionError` *subclass*, so the exact-type
check excludes it — and `_execute_command` re-raises it deliberately without reinitializing
(`redis/asyncio/cluster.py`). So it surfaces to application code.

Under high concurrency one event-loop iteration can contain many acquisition attempts, so the
error is observable in practice; it is a spurious error, not a stuck client. This is distinct
from #4110 (a monotonic slot leak that never self-healed, fixed in 8.1.0 via #4111).

### Why the other pools don't have this

Both `ConnectionPool.release()` implementations make the connection available only *after*
disconnecting, so no window exists:

- sync `redis/connection.py` — disconnects inline while holding `self._lock`
- async `redis/asyncio/connection.py` — `await connection.disconnect()` inside
  `async with self._lock`

The async cluster needs the background task only because `ClusterNode.release()` is `def`,
not `async def`.

### Proposed fix

Since the connection is guaranteed closed at that point, stop deferring: have `release()`
append it to `_free` directly (guarded on `is_connected` so a still-open connection is not
silently returned), or make `release()` a coroutine that awaits the disconnect inline, matching
the two pools above. Either removes the window, and the first also removes the
`_background_tasks` machinery from this path — a net reduction in complexity rather than a new
public contract.

A regression test should drive the real path (`execute_command` with a concurrent re-mark),
not construct the state by calling `node.release()` directly.

### Related

- PR #4219 proposes solving this by adding a second, waiting acquisition API
  (`acquire_connection_async`). That direction adds public surface, an untimed wait that works
  against the deliberate fast-fail above, and a new `await` in the transaction path, which runs
  inside `asyncio.run()` on a worker thread while the owning loop is blocked. Filing this issue
  to capture the underlying defect independently of that approach.
- #4110 / #4111 — the monotonic slot leak with the same symptom, fixed in 8.1.0.
- #3929 — original vague report; its persistent variant was #4110.

### Secondary: hardening, no demonstrated reproduction

Stated explicitly as an invariant gap rather than a confirmed bug: `ClusterNode._free` is
`collections.deque(maxlen=self.max_connections)` (1668) and `release()` carries no ownership or
duplicate-release guard, unlike sync `ConnectionPool.release()` which drops unowned connections
via `_in_use_connections.remove()` / `owns_connection()`. If any path ever released the same
connection twice, `append` on a full bounded deque silently evicts from the opposite end —
losing a pool slot permanently — or the same connection could be handed to two concurrent
tasks. I did not find a reachable double-release path; recording it so the invariant is
asserted rather than assumed.

LubuSeb:
### Description of change

Fixes #4247.

An async cluster connection can be marked for reconnect while its first disconnect is still waiting for the socket to close. That leaves the connection closed but marked again. `release()` previously treated every marked connection as needing another background disconnect, so at `max_connections=1` the only connection stayed outside `_free` for another event-loop turn. A concurrent acquire could then raise a false `MaxConnectionsError`.

This change keeps the existing background disconnect for marked connections that are still open. If the connection is already closed, it clears the stale reconnect flag and returns the connection to `_free` immediately.

The regression test drives the real `execute_command()` path with a deterministic interleaving, then verifies that the next acquire reuses the same connection without a second disconnect.

### Validation

- `TestClusterNodeConnectionHandling`: 13 passed on Windows and Ubuntu WSL
- `invoke linters`: passed
- Ruff check, format, and vulture: passed on Ubuntu WSL
- `git diff --check`: passed
- The complete integration matrix requires the repository's Redis Cluster services and will be covered by CI

### Pull Request check-list

- [x] Relevant tests and lints pass with this change
- [ ] CI tests pass with this change
- [x] The changed behavior has regression coverage
- [x] No public API or documentation changes are needed
- [x] No example is needed for this internal connection-pool fix

---

> [!NOTE]
> **Medium Risk**
> Touches async cluster connection-pool release logic on a hot path; behavior change is narrow (closed + reconnect-flagged) but incorrect pooling can cause false MaxConnectionsError or connection leaks under concurrency.
> 
> **Overview**
> Fixes a race in async `ClusterNode.release()` where a connection could be **re-marked for reconnect** while its first background disconnect was still in flight. The connection was already closed but still flagged, so `release()` always scheduled another async disconnect and withheld it from `_free` for an extra event-loop turn. At **`max_connections=1`**, a concurrent acquire could hit a false **`MaxConnectionsError`**.
> 
> **`release()`** now only spawns the background disconnect when the connection is still **`is_connected`**. If it is already closed, it clears the stale reconnect flag via **`reset_should_reconnect()`** and appends the connection to **`_free`** synchronously.
> 
> Adds a regression test that drives **`execute_command()`** with a deterministic interleaving (remark during in-flight disconnect) and asserts the next acquire reuses the same connection without a second disconnect or stray background tasks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39addac7a61ce8fa7c8e964f834999bb068e43ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1756,10 +1756,12 @@ class ClusterNode:
         returning it to the free queue.
         """
         if connection.should_reconnect():
-            task = asyncio.create_task(self._disconnect_and_release(connection))
-            self._background_tasks.add(task)
-            task.add_done_callback(self._background_tasks.discard)
-            return
+            if connection.is_connected:
+                task = asyncio.create_task(self._disconnect_and_release(connection))
+                self._background_tasks.add(task)
+                task.add_done_callback(self._background_tasks.discard)
+                return
+            connection.reset_should_reconnect()
         self._free.append(connection)
 
     async def _disconnect_and_release(self, connection: Connection) -> None:
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -1753,8 +1753,9 @@ class ClusterNode:
     def release(self, connection: Connection) -> None:
         """
         Release connection back to free queue.
-        If the connection is marked for reconnect, disconnect it before
-        returning it to the free queue.
+        If a connected connection is marked for reconnect, disconnect it before
+        returning it to the free queue. An already-closed connection can be
+        returned immediately after clearing its reconnect flag.
         """
         if connection.should_reconnect():
             if connection.is_connected:
@@ -1762,6 +1763,7 @@ class ClusterNode:
                 self._background_tasks.add(task)
                 task.add_done_callback(self._background_tasks.discard)
                 return
+            # It may have been re-marked while its own disconnect was in progress.
             connection.reset_should_reconnect()
         self._free.append(connection)
 
--- a/tests/test_asyncio/test_cluster.py
+++ b/tests/test_asyncio/test_cluster.py
@@ -3642,13 +3642,11 @@ class TestClusterNodeConnectionHandling:
             max_connections=1,
             connection_class=RaceConnection,
         )
+        connection = RaceConnection(host=default_host, port=7000)
+        node._connections = [connection]
+        node._free.append(connection)
 
         async def remark_during_disconnect() -> None:
-            while not node._connections:
-                await asyncio.sleep(0)
-
-            connection = node._connections[0]
-
             await connection.send_started.wait()
             node.update_active_connections_for_reconnect()
             connection.response_ready.set()
@@ -3661,8 +3659,6 @@ class TestClusterNodeConnectionHandling:
 
         assert await node.execute_command("GET", "key") == b"value"
         await remark_task
-
-        connection = node._connections[0]
 
         assert node.acquire_connection() is connection
         assert node._background_tasks == set()
--- a/tests/test_asyncio/test_cluster.py
+++ b/tests/test_asyncio/test_cluster.py
@@ -3589,6 +3589,85 @@ class TestClusterNodeConnectionHandling:
         # Should not raise
         await node.disconnect_free_connections()
 
+    async def test_execute_command_reuses_closed_remarked_connection_inline(
+        self,
+    ) -> None:
+        class RaceConnection:
+            def __init__(self, **kwargs: Any) -> None:
+                self.host = kwargs["host"]
+                self.port = kwargs["port"]
+                self.db = kwargs.get("db", 0)
+                self.is_connected = True
+                self._reconnect = False
+                self.send_started = asyncio.Event()
+                self.response_ready = asyncio.Event()
+                self.disconnect_started = asyncio.Event()
+                self.allow_disconnect = asyncio.Event()
+                self.disconnect_calls = 0
+
+            def pack_command(self, *args: Any) -> bytes:
+                return b"packed-command"
+
+            async def send_packed_command(self, command: Any) -> None:
+                self.send_started.set()
+
+            async def read_response(self, *args: Any, **kwargs: Any) -> bytes:
+                await self.response_ready.wait()
+                return b"value"
+
+            def mark_for_reconnect(self) -> None:
+                self._reconnect = True
+
+            def should_reconnect(self) -> bool:
+                return self._reconnect
+
+            def reset_should_reconnect(self) -> None:
+                self._reconnect = False
+
+            async def disconnect(self) -> None:
+                self.disconnect_calls += 1
+                self.reset_should_reconnect()
+
+                if not self.is_connected:
+                    return
+
+                self.disconnect_started.set()
+                await self.allow_disconnect.wait()
+                self.is_connected = False
+
+        node = ClusterNode(
+            default_host,
+            7000,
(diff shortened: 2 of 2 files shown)


petyaslavova (changes requested):
Hey @LubuSeb, thanks for the contribution!

Two things need to be handled before merging:

1. Please replace the `while not node._connections: await asyncio.sleep(0)` spin in the new test. It never runs today — `acquire_connection()` is synchronous, so `_connections` is already populated before `execute_command()` reaches its first suspension point — but if a future change moves connection creation behind an `await`, an unbounded `sleep(0)` loop busy-spins rather than failing cleanly, and it is sensitive to loop-turn counts under `--uvloop`. Constructing the connection up front and seeding `node._connections` / `node._free` before calling `execute_command()` removes the loop and keeps the rest of the interleaving event-driven and deterministic; other tests in this class already seed node state that way.

2. Please update the `release()` docstring, which still describes only the "disconnect before returning to the free queue" branch, and add a short comment explaining why an already-closed connection can be holding a set reconnect flag (it was re-marked during its own `disconnect()`). The `is_connected` guard is a non-obvious invariant and this file documents that kind of reasoning inline elsewhere.

After those and a green CI run this looks ready.

LubuSeb:
Addressed both review points in 2e543c5: the regression test now seeds the connection up front and remains entirely event-driven, and `release()` now documents the connected and already-closed branches plus the re-marking race. The 13 focused connection-handling tests pass on Windows and under uvloop, and Ruff/format checks pass. Fresh hosted workflows are awaiting maintainer approval.

petyaslavova (approved):
LGTM.

PR #4260, opened 2026-08-10
fix(asyncio): raise on EOF in can_read() so pools replace server-closed connections

### LINKED ISSUE #4252: Unexpected Connection closed by server behavior on redis==8.1.0
```
[BUG] asyncio: ConnectionError: Connection closed by server. surfaces to the caller when a pooled connection is closed while idle (regression in 8.1.0)
Version
redis-py: 8.1.0 (works on 8.0.1)
Python: 3.13
Server: Redis (docker redis:latest), standalone, RESP3
Platform: Linux container / macOS
hiredis: reproduced both with and without hiredis installed
Description

With redis.asyncio and a ConnectionPool configured with health_check_interval and a Retry, a connection that the server closed while it sat idle in the pool causes the next command to raise ConnectionError: Connection closed by server. to the application, instead of being transparently retried on a fresh connection.

On 8.0.1 the identical script recovers and returns the value. On 8.1.0 it raises. No application-side change; only the redis-py version differs.

The failure surfaces from the PING inside check_health(): the health check detects the dead socket (correctly), but the resulting ConnectionError propagates out of Redis.execute_command rather than being retried.

Client config

import redis.asyncio as redis
from redis.backoff import ExponentialBackoff
from redis.retry import Retry

pool = redis.ConnectionPool.from_url(
    url=f"redis://{host}:{port}/0",
    encoding="utf-8",
    decode_responses=True,
    health_check_interval=10,
    retry=Retry(ExponentialBackoff(), 3),
    retry_on_error=[redis.ConnectionError, redis.TimeoutError, RuntimeError],
    socket_connect_timeout=5,
    retry_on_timeout=True,
    socket_keepalive=True,
)
client = redis.Redis.from_pool(connection_pool=pool)
I verified the retry object actually reaches the connection — on both versions conn.retry.get_retries() == 3 and the supported errors are (TimeoutError, ConnectionError, RuntimeError, asyncio.TimeoutError). So this is not a case of the retry policy being dropped at construction.

Reproduction
Standalone script. It kills the pooled connection server-side with CLIENT KILL, which is what an idle-timeout / server-side reap looks like to the client. next_health_check = 0 just forces the health check to be due immediately, so you don't have to wait out health_check_interval.


import asyncio
import redis
import redis.asyncio as aredis
from redis.backoff import ExponentialBackoff
from redis.retry import Retry


async def client_id(conn):
    await conn.send_command("CLIENT", "ID", check_health=False)
    return await conn.read_response()


async def main():
    pool = aredis.ConnectionPool.from_url(
        url="redis://localhost:6379/0",
        decode_responses=True,
        health_check_interval=10,
        retry=Retry(ExponentialBackoff(), 3),
        retry_on_error=[aredis.ConnectionError, aredis.TimeoutError, RuntimeError],
        socket_connect_timeout=5,
        retry_on_timeout=True,
        socket_keepalive=True,
    )
    c = aredis.Redis.from_pool(connection_pool=pool)

    await c.hset("h", "k", "v")
    print("before:", await c.hget(name="h", key="k"))

    # The connection is idle in the pool. Make the health check due, then have
    # the server close it — same as an idle timeout / server-side reap.
    conn = pool._available_connections[0]
    conn.next_health_check = 0

    admin = aredis.Redis.from_url("redis://localhost:6379/0", decode_responses=True)
    await admin.execute_command("CLIENT", "KILL", "ID", str(await client_id(conn)))
    await admin.aclose()

    try:
        print("after: ", await c.hget(name="h", key="k"))
        print("RESULT: RECOVERED")
    except Exception as e:
        print(f"RESULT: RAISED {type(e).__name__}: {e}")

    await c.aclose()


print("redis-py", redis.__version__)
asyncio.run(main())
Actual (8.1.0)

redis-py 8.1.0
before: v
RESULT: RAISED ConnectionError: Connection closed by server.
Expected — and what 8.0.1 actually does

redis-py 8.0.1
before: v
after:  v
RESULT: RECOVERED
Traceback (from the real application)

    customer = await self.redis_client.hget(name=hset_name, key=hashed_api_key)
  File "/usr/local/lib/python3.13/site-packages/redis/asyncio/client.py", line 946, in execute_command
    result = await conn.retry.call_with_retry(
  File "/usr/local/lib/python3.13/site-packages/redis/asyncio/client.py", line 870, in _send_command_parse_response
  File "/usr/local/lib/python3.13/site-packages/redis/asyncio/connection.py", line 1214, in send_command
    await self.send_packed_command(
  File "/usr/local/lib/python3.13/site-packages/redis/asyncio/connection.py", line 1177, in send_packed_command
    await self.check_health()
  File "/usr/local/lib/python3.13/site-packages/redis/asyncio/connection.py", line 1163, in check_health
    await self.retry.call_with_retry(
  File "/usr/local/lib/python3.13/site-packages/redis/asyncio/connection.py", line 1148, in _send_ping
    if str_if_bytes(await self.read_response()) != "PONG":
  File "/usr/local/lib/python3.13/site-packages/redis/asyncio/connection.py", line 1286, in read_response
    response = await self._read_response_from_parser(
  File "/usr/local/lib/python3.13/site-packages/redis/asyncio/connection.py", line 1335, in _read_response_from_parser
    return await self._parser.read_response(
  File "/usr/local/lib/python3.13/site-packages/redis/_parsers/resp3.py", line 185, in read_response
    response = await self._read_response(
  File "/usr/local/lib/python3.13/site-packages/redis/_parsers/resp3.py", line 197, in _read_response
    raw = await self._readline()
  File "/usr/local/lib/python3.13/site-packages/redis/_parsers/base.py", line 589, in _readline
Notes from narrowing it down
A few things I checked so they can be ruled out:

Not the retry configuration. conn.retry carries retries=3 and includes ConnectionError in its supported errors on both versions.
Not retry.py. redis/retry.py and redis/asyncio/retry.py are byte-identical between 8.0.1 and 8.1.0.
Not the parser. My first comparison was confounded by hiredis being absent in the 8.1.0 venv (so it picked _AsyncRESP3Parser vs _AsyncHiredisParser). After pinning hiredis==3.3.1 in both, the parser matches and 8.1.0 still fails — so the parser is not the cause.
Not should_reconnect(). Forcing it to False on 8.1.0 does not change the outcome.
Every function in the traceback (check_health, _send_ping, _ping_failed, send_packed_command) is byte-identical between the two versions, and calling conn.check_health() directly raises the same ConnectionError on both. So the difference is not in raising the error — it is in whether the caller retries it.
Instrumenting Retry.call_with_retry shows the difference. On 8.0.1 the failing command drives repeated health checks and recovers; on 8.1.0 _ping_failed is never invoked and the error escapes on the first failure:


=== 8.0.1 ===                        === 8.1.0 ===
disconnect(...)                      check_health
check_health                         disconnect(...)
check_health                         -> ConnectionError raised
check_health
check_health
-> recovered
So on 8.1.0 the ConnectionError raised by the health-check PING is escaping the retry loop in execute_command instead of triggering a reconnect-and-retry.


```

Brumbelow:
### Description of change

Fixes #4252.

In `redis.asyncio`, a pooled connection that the server closed while it sat idle (idle timeout, `CLIENT KILL`, server-side reap) is handed out to the next command instead of being replaced, and the command fails with `ConnectionError: Connection closed by server.`. Regression in 8.1.0 under RESP3 (the default); 8.0.1 recovers transparently.

Root cause: the `and not self.maint_notifications_enabled()` guard that #4177 added to `ensure_connection()`'s checkout probe. The async parsers' non-destructive `can_read()` (#4063, shipped in 8.0.1, which recovers correctly) reports a server-closed stream by returning `True` — EOF folded into "has readable data" — and in 8.0.1 that `True` unconditionally hits the `Connection has data` branch, which disconnects and reconnects. With RESP3 over TCP the pool auto-creates `MaintNotificationsConfig(enabled="auto")`, so since #4177 the exemption is on by default and the reconnect branch is never entered: the dead connection passes checkout. RESP2 connections get no maintenance-notifications handler, which is why RESP2 still recovers.

The sync side doesn't have this problem: `SocketBuffer.can_read()` raises `ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)` on EOF, which the pool's `except (ConnectionError, TimeoutError, OSError)` branch turns into disconnect + reconnect regardless of any exemption.

This change makes the async parsers (`_AsyncRESPBase`, `_AsyncHiredisParser`) behave the same way: raise `ConnectionError` on `at_eof()` instead of returning `True`, with already-buffered data still reported first (matching the sync parsers' buffer-first ordering, so a pending reply or push notification is delivered rather than discarded). `ensure_connection()`'s existing reconnect branch then replaces the dead connection whether or not the pending-push-data exemption applies — the closed-connection signal can no longer be swallowed by that guard, rather than only reordering the guard itself. The async `parser.can_read()` is reached only through `Connection.can_read()`, whose only caller is the pool checkout path, so the blast radius is that seam.

Verified with the reproduction from #4252 (server closes the idle pooled connection, health check due): before this change RESP3 raises and RESP2 recovers; after it both recover, and 8.0.1 behavior is restored.

One additional observation from the issue's repro, out of scope here: it passes the sync `redis.retry.Retry` into the async pool. Awaiting the sync `call_with_retry` silently degrades it to a single attempt with no failure callback, which is why the health-check `PING` failure surfaced unretried. With `redis.asyncio.retry.Retry` the client recovers even without this fix (at the cost of a failed attempt and reconnect churn). Filed separately as #4262.

### Pull Request check-list

- [N/A] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? 
- [N/A] Is there an example added to the examples folder (if applicable)?

### DIFF
--- a/redis/_parsers/base.py
+++ b/redis/_parsers/base.py
@@ -547,13 +547,16 @@ class _AsyncRESPBase(AsyncBaseParser):
         # connection state, not only whether application data can be read.
         if not self._connected:
             raise OSError("Buffer is closed.")
+        # buffered data wins over EOF, like the sync SocketBuffer: a pending
+        # response or push notification must stay readable even if the server
+        # has since closed the connection.
+        if self._buffer:
+            return True
         if self._stream.at_eof():
             # Raise like the sync SocketBuffer does on a server-closed
             # connection, so callers that tolerate pending data (push
             # notifications) can't mistake EOF for a readable connection.
             raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
-        if self._buffer:
-            return True
         # asyncio.StreamReader has no public non-destructive API for checking
         # buffered bytes. Preserve dirty-connection detection for the Python
         # parser and fail loudly if the private buffer API changes.
--- a/redis/_parsers/base.py
+++ b/redis/_parsers/base.py
@@ -547,12 +547,17 @@ class _AsyncRESPBase(AsyncBaseParser):
         # connection state, not only whether application data can be read.
         if not self._connected:
             raise OSError("Buffer is closed.")
+        if self._stream.at_eof():
+            # Raise like the sync SocketBuffer does on a server-closed
+            # connection, so callers that tolerate pending data (push
+            # notifications) can't mistake EOF for a readable connection.
+            raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
         if self._buffer:
             return True
         # asyncio.StreamReader has no public non-destructive API for checking
         # buffered bytes. Preserve dirty-connection detection for the Python
         # parser and fail loudly if the private buffer API changes.
-        return bool(self._stream._buffer) or self._stream.at_eof()
+        return bool(self._stream._buffer)
 
     async def _read(self, length: int) -> bytes:
         """
--- a/redis/_parsers/hiredis.py
+++ b/redis/_parsers/hiredis.py
@@ -323,8 +323,12 @@ class _AsyncHiredisParser(AsyncBaseParser, AsyncPushNotificationsParser):
         # connection state, not only whether application data can be read.
         if not self._connected:
             raise OSError("Buffer is closed.")
-        # EOF means the connection is closed and not safe to reuse.
-        if self._reader.has_data() or self._stream.at_eof():
+        if self._stream.at_eof():
+            # Raise like the sync SocketBuffer does on a server-closed
+            # connection, so callers that tolerate pending data (push
+            # notifications) can't mistake EOF for a readable connection.
+            raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
+        if self._reader.has_data():
             return True
         # asyncio.StreamReader has no public non-destructive API for checking
         # buffered bytes. Preserve dirty-connection detection for hiredis; tests
--- a/redis/_parsers/hiredis.py
+++ b/redis/_parsers/hiredis.py
@@ -323,13 +323,16 @@ class _AsyncHiredisParser(AsyncBaseParser, AsyncPushNotificationsParser):
         # connection state, not only whether application data can be read.
         if not self._connected:
             raise OSError("Buffer is closed.")
-        if self._stream.at_eof():
-            # Raise like the sync SocketBuffer does on a server-closed
-            # connection, so callers that tolerate pending data (push
-            # notifications) can't mistake EOF for a readable connection.
-            raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
+        # buffered data wins over EOF, like the sync parser: a pending
+        # response or push notification must stay readable even if the
+        # server has since closed the connection.
         if self._reader.has_data():
             return True
+        if self._stream.at_eof():
+            # Raise like the sync parser does on a server-closed connection,
+            # so callers that tolerate pending data (push notifications)
+            # can't mistake EOF for a readable connection.
+            raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
         # asyncio.StreamReader has no public non-destructive API for checking
         # buffered bytes. Preserve dirty-connection detection for hiredis; tests
         # with a real StreamReader guard this private buffer API in CI.
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -109,6 +109,18 @@ async def test_async_hiredis_can_read_raises_on_eof():
     assert stream.read_called is False
 
 
+async def test_async_hiredis_can_read_prefers_buffered_data_over_eof():
+    # a reply or push notification the reader already buffered must stay
+    # readable even if the server has since closed the connection,
+    # matching the sync parser's has_data-first ordering
+    stream = DummyAsyncStream(eof=True)
+    parser = make_async_hiredis_parser(stream, response=b"OK", has_data=True)
+
+    assert await parser.can_read() is True
+    assert await parser.read_response() == b"OK"
+    assert stream.read_called is False
+
+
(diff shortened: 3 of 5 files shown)


petyaslavova (changes requested):
Hey @Brumbelow, thank you for your contribution!


A few things need to be addressed before we can merge this PR:

1. Please correct the root-cause attribution in the description. #4063 shipped in 8.0.1, which recovers correctly with the same "EOF ⇒ True" parser code — the async `can_read()` implementations are byte-identical between `v8.0.1` and `master`, so it isn't a co-cause. The regression is only the `and not self.maint_notifications_enabled()` guard added in #4177: with RESP3 over TCP the pool auto-creates `MaintNotificationsConfig(enabled="auto")`, so the exemption is on by default and the reconnect branch is never entered. RESP2 gets no handler, which is why it still recovers.
2. The comment in `tests/test_asyncio/test_scenario/test_maint_notifications.py` that explains its `except RedisConnectionError` still describes the old behaviour ("can_read() reports True on EOF ... so the drain enters read_response()"). Please update it.
3. Parametrize the new pool test over `_AsyncHiredisParser` as well, and add a real-server regression test for #4252 using `client_kill_filter(_id=...)` on an idle pooled connection — the current test asserts the branch, not end-to-end recovery.


And yes, please do file the sync-`Retry`-in-an-async-pool degradation separately — good catch, and it deserves its own guard.

petyaslavova (approved):
LGTM.

NicolasFerec:
We are still hitting this regression in production with redis-py 8.1.0 (`redis.asyncio`), with `ConnectionError: Error UNKNOWN while writing to socket. Connection lost.` when an idle pooled connection is reused. We have added an async retry and health check as a temporary workaround, but we are still waiting for a release that includes this merged fix. Is there an ETA for the next patch release? Thanks!

PR #4265, opened 2026-08-13
Feat/register custom read commands

SantoshDhaladhuli:
### Description of change

This PR refactors how `redis-py` determines if a command is safe to route to a cluster replica, shifting from a hardcoded set to a dynamic, extensible metadata resolver. This allows third-party modules (e.g., FalkorDB, RediSearch, RedisJSON) to natively route their custom read-only commands to replicas without mutating internal data structures.

**Key Enhancements:**
1. **Dynamic Replica Routing via MetadataResolver**: Replaced the static `command in READ_COMMANDS` check in `_determine_nodes`, `_execute_pipeline_by_slot`, and keyless routing with a new `is_replica_safe` API on `MetadataResolver` and `AsyncMetadataResolver`.
2. **Metadata is the only source**: `is_replica_safe` answers from `metadata.is_readonly` alone; `READ_COMMANDS` is not consulted. It stays exported and is now deprecated, so editing it changes nothing. The single exception is `TOUCH`, which the server reports readonly even though it refreshes key idle time: it is seeded replica-unsafe into every resolver's memo at construction, which also stops a resolver chained behind the static one from re-admitting it. Caller-supplied metadata remains authoritative for every ordinary core and module read.
3. **Async Cluster Support**: Wired the `metadata_resolver` into the `redis.asyncio.RedisCluster` constructor, bringing parity with the sync client. Async routing internally awaits the resolver while maintaining the synchronous API contract for public callers of `get_random_primary_or_all_nodes`.
4. **Centralized policy resolution**: Both stacks gained `_resolve_command_policies`, which decides a command's request and response policies once per `execute_command` and returns the name they were decided by, so the flag tables, the result callbacks and the routing decision all agree on one spelling. Keyless routing moved to `get_keyless_target_node`, with `get_random_primary_or_all_nodes` kept as a deprecated alias of the name public since 7.1.0.
5. **Static metadata coverage**: The table now carries every command on the 7.1.0 client-side-cache allow list plus their module counterparts, and deliberately withholds routing policies for `COMMAND`, `DBSIZE`, `KEYS`, `RANDOMKEY` and `SCAN` so their existing `COMMAND_FLAGS` routing survives now that resolved metadata outranks the flag table.

**Use Case:**
Library authors can now pass a custom `DynamicMetadataResolver` (containing their module's command metadata) into the `RedisCluster` constructor. `redis-py` will automatically respect the `is_readonly` flags of those custom commands and route them to replicas when `read_from_replicas=True` or a `load_balancing_strategy` is configured.

### Backward compatibility

Checked in both directions. Every member of `READ_COMMANDS` still resolves replica-safe, so nothing that reached a replica stops doing so; 18 commands become newly eligible, all of them genuinely readonly search, timeseries and `_ro` reads. Changes worth a release-note line:

- Keyless and split-multi-key reads now reach replicas when only `load_balancing_strategy` is set, matching what the keyed path already did, and a replicas-only strategy (`ROUND_ROBIN_REPLICAS`, `RANDOM_REPLICA`) is honored rather than falling through to a uniform pick that could land on a primary. A cluster with no replicas falls back to the whole node set.
- An explicit `policy_resolver` still decides which nodes a command targets, but no longer supplies a response policy when the caller names its targets. Replica safety and cache eligibility always resolve through `metadata_resolver`, since a `CommandPolicies` record carries no `is_readonly` flag.
- Client-side-cache eligibility differs from the deprecated `CacheConfig.DEFAULT_ALLOW_LIST` by a documented delta: `DIGEST`, the `*EXPIRETIME` family, `SDIFFCARD`, `SUNIONCARD` and the two suggestion reads are added; `XPENDING`, `TS.INFO` and `XREAD` are dropped, all three confirmed by the server as not cacheable.
- `redis/asyncio/retry.py` widens `Retry` to `AbstractRetry[Exception]` to match the sync class, and `digest_local` now hashes with the client's configured encoder instead of assuming UTF-8.

Sync and async are changed together throughout, with mirrored signatures, docstrings and tests.

### Test coverage

- Metadata resolvers: `is_replica_safe` across both stacks for core reads, module reads, writes, unknown names and non-string input; the `TOUCH` memo seed pinned in both directions; the newly eligible commands parametrized; and the cache-eligibility delta against the legacy allow list pinned from one shared constant.
- Cluster routing: `SCAN` to all primaries, `DBSIZE`/`KEYS`/`RANDOMKEY` to the default node, container commands routed and dispatched by the two-word name, `SLOT_ID` commands routed by slot for any spelling, explicit targets not applying a command's aggregation, and `_determine_nodes` raising `RedisClusterException` rather than `KeyError` when no policy resolved.
- Keyless selection: a replicas-only strategy never returns a primary, a replica-less cluster still returns a node instead of raising, the remaining strategies and bare `read_from_replicas` keep returning any node, and every `LoadBalancingStrategy` member is classified.
- Deprecation surface: `get_random_primary_or_all_nodes` warns, delegates, answers from the metadata resolver rather than `READ_COMMANDS`, and stays synchronous on the async client both inside a running loop and off it. The `READ_COMMANDS` fallback on the command mixin is covered for a third-party host class that supplies no resolver.
- `digest_local` pins that the client encoding decides the hashed bytes, and the three client-side-cache invalidation tests share one bounded-wait helper in place of a hand-rolled poll loop and two fixed sleeps.

Verified against a live cluster and standalone server: `tests/test_cluster.py` 339 passed, `tests/test_asyncio/test_cluster.py` 325 passed, `tests/test_command_metadata.py` 97 passed, `tests/test_asyncio/test_command_metadata.py` 63 passed, `tests/test_cache.py` 105 passed, `tests/test_commands.py` 636 passed. `ruff check`, `ruff format --check` and `vulture` clean.

### Pull Request check-list

- [x] Is there an example added to the examples folder (N/A - cluster routing API update)?

---

> [!NOTE]
> **High Risk**
> Changes default cluster replica routing, policy resolution, and static metadata/CSC eligibility across sync and async paths—behavior shifts for custom modules, keyless reads, and commands like TOUCH even when `policy_resolver` is overridden.
> 
> **Overview**
> **Cluster replica routing** no longer uses the static `READ_COMMANDS` set. Sync and async `RedisCluster` call `metadata_resolver.is_replica_safe()` everywhere slot/keyless routing decides replica eligibility, including MOVED retries and multi-key pipelines. `READ_COMMANDS` is documented as deprecated and unused by the default path; host mixins only fall back to it with a deprecation warning.
> 
> **`metadata_resolver` on async cluster**: `redis.asyncio.RedisCluster` accepts `metadata_resolver` (default `AsyncStaticMetadataResolver`). An explicit `policy_resolver` still wins for *which nodes* to target; replica safety and CSC eligibility always come from the metadata resolver. Keyless routing moves to `get_keyless_target_node` (async), with `get_random_primary_or_all_nodes` deprecated (sync alias; async uses a thread bridge).
> 
> **Routing policy resolution** is centralized in `_resolve_command_policies` (normalized command names, explicit `target_nodes` skips cluster aggregation, shared by execute and pipelines). Fixes include case-insensitive `SLOT_ID` handling, `int` slots from string args, and consistent result-callback command names.
> 
> **Static command metadata** gains `is_replica_safe` on resolvers (memoized; `TOUCH` forced replica-unsafe), expanded `_STATIC_COMMAND_METADATA` (withheld routing for `SCAN`/`DBSIZE`/etc., more cacheable reads), and fully explicit `CommandMetadata` fields to avoid accidental CSC side effects.
> 
> **Smaller API fixes**: `DIGEST` / `*EXPIRETIME` pass `keys=` for client-side cache; `digest_local` hashes via the client encoder; async `Retry` types broadened to `Exception`; cache invalidation tests poll instead of fixed sleeps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f17054770dbf64a9ed7b53262cc8ed94be5fc1a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Mukller (commented):
Reviewed with local verification of the parts that can be checked offline (Python 3.13, branch checkout):

**1. The `.upper()` fix in `get_random_primary_or_all_nodes` is a genuine robustness fix — worth calling out in the changelog.** On master, `READ_COMMANDS` is an all-uppercase frozenset (96 entries: `{'BITCOUNT', 'BITFIELD_RO', ..., 'EVALSHA_RO', ...}`), and membership is case-sensitive. Any caller that reaches node resolution with a non-uppercase name — `execute_command("dbsize")`, or a custom command registered lowercase, which is exactly this PR's feature — silently misroutes reads to the primary even with `read_from_replicas=True`. The new `_get_random_primary_or_all_nodes_for_command` wrapper normalizing the case fixes that class of misses. Verified `READ_COMMANDS` contents and the call chain (`_policies_callback_mapping[DEFAULT_KEYLESS] → get_random_primary_or_all_nodes(command)`) on master before this change.

**2. Wiring is consistent**: on the branch, `AsyncStaticPolicyResolver.__init__` accepts `metadata_resolver`, `AsyncStaticMetadataResolver` resolves, and `RedisCluster(metadata_resolver=...)` defaults correctly when omitted. The docstring + `docs/resp3_features.rst` correction (async cluster *does* accept `metadata_resolver`, async standalone does not) matches the code.

**3. What I could not verify offline** — flagging honestly rather than guessing: end-to-end custom-command registration and replica routing needs a live cluster; the server-dependent tests in my environment fail at connection setup identically to master.

**4. One question about the test changes**: `test_cache.py` loosens invalidation waits from `sleep(0.1)` to `sleep(0.5)` plus a 3-second polling loop, and the comment changed from "SSL connection" to "Cluster connection". Is the underlying issue just CI variance, or is cluster-mode invalidation propagation actually slower/lossier than 100ms in practice? If the latter, it might deserve its own tracking issue — loosened sleeps can hide a real race for future readers.

Given (3), leaving the final approval to someone who can run the cluster suite.

Mukller (commented):
Reviewed statically and with offline probes (no live cluster in this environment — noted honestly below).

**Design coherence** — this mirrors the existing `policy_resolver` architecture nicely, and the doc update to resp3_features.rst matches the implementation (async `RedisCluster` now accepts `metadata_resolver`; plain async `Redis` still doesn't).

**Verified by direct probe on the branch**:

```python
r = AsyncStaticMetadataResolver()
await r.is_replica_safe("GET")   # True  (replica-safe)
await r.is_replica_safe("SET")   # False
await r.is_replica_safe("get")   # True  - case normalization handled
```

**The `DEFAULT_KEYLESS` behavioral change checks out**: master's mapping was

```python
RequestPolicy.DEFAULT_KEYLESS: lambda command_name: [self.get_random_primary_or_all_nodes(command_name)]
```

i.e. already returned a single-element list. The new branch code `[await self._get_random_primary_or_all_nodes_for_command(command)]` preserves that shape exactly while making the replica-safety decision awaitable (resolver-based instead of the static `READ_COMMANDS` membership). Both the old sync method and the new async variant now normalize case consistently (`command_name.upper()` vs resolver internals).

**What I could not verify here**: the live-routing behavior (`test_cache.py` needs a running Redis; local run just hangs on connection retries), and CI was still mid-run at review time (freshly rebased). The remaining diff hunks (`retry.py` +3, sync `cluster.py` +9) are small and read consistently with the async side.

One question for maintainers: with `is_replica_safe` being dynamic per-resolver, should `get_nodes_from_slot`'s load-balancing-strategy selection also consult the resolver for *write* commands routed during failover, or is primary-only routing there intentionally unchanged? Not a blocker — just want to confirm it's a conscious boundary.

Happy to re-check once CI completes if anything looks off.

redis/commands/metadata.py
```diff
@@ -39,6 +39,8 @@
 from enum import Enum
 from types import MappingProxyType
 
+from redis.commands.cluster import READ_COMMANDS
```
petyaslavova: We should completely stop using this set - all the commands metadata should be added to the static metadata in `_STATIC_COMMAND_METADATA`. This way we can decide for each command is it replica safe only using the metadata. The `READ_COMMANDS` constant is public and we can't just delete it, but we can deprecate it - check how it was done for the cachable commands constant

redis/commands/metadata.py
```diff
@@ -325,6 +333,18 @@ def is_cacheable(self, command_name: str) -> bool:
         """
         pass
 
+    def is_replica_safe(self, command_name: str) -> bool:
```
petyaslavova, line 335: Since we don't want to consult with the old READ_COMMANDS and we want to use the metadata instead - this method should just define the method in the interface with empty body

redis/commands/metadata.py
```diff
@@ -389,6 +409,18 @@ async def is_cacheable(self, command_name: str) -> bool:
         """
         pass
 
+    async def is_replica_safe(self, command_name: str) -> bool:
```
petyaslavova, line 412: Same note as for the sync interface.

redis/commands/metadata.py
```diff
@@ -501,6 +534,35 @@ def is_cacheable(self, command_name: str) -> bool:
...
         return cacheable
 
+    def is_replica_safe(self, command_name: str) -> bool:
+        if not isinstance(command_name, str):
+            return False
+
+        memo_key = command_name.lower()
+
+        try:
+            return self._replica_safe[memo_key]
```
petyaslavova, line 546: You can feed the local `_replica_safe` cache with the forbidden commands listed in `_REPLICA_UNSAFE_COMMANDS` on `init` - this way, you would not need to jump between `lower` and `upper` letters all the time.

redis/commands/metadata.py
```diff
@@ -501,6 +534,35 @@ def is_cacheable(self, command_name: str) -> bool:
...
+        except ValueError:
+            metadata = None
+
+        normalized_name = command_name.upper()
+        if normalized_name in _REPLICA_UNSAFE_COMMANDS:
+            replica_safe = False
+        elif normalized_name in READ_COMMANDS:
+            replica_safe = True
+        else:
+            replica_safe = metadata.is_readonly if metadata is not None else False
```
petyaslavova: Please expose local private helper method similar to `_is_client_side_cacheable(metadata)` that will decide if a command is replica safe based on the readonly flag and metadata availability. If the `_REPLICA_UNSAFE_COMMANDS` are loaded into the `self._replica_safe` field we should not worry about those ones - they should have been checked already above.

redis/commands/metadata.py
```diff
@@ -602,6 +665,35 @@ async def is_cacheable(self, command_name: str) -> bool:
 
         return cacheable
 
+    async def is_replica_safe(self, command_name: str) -> bool:
```
petyaslavova, line 666: The same notes posted for sync implementation apply here as well.

redis/commands/metadata.py
petyaslavova, line 1116: This data should be enriched with the missing commands that are in the READ_COMMANDS list.
### Filling missing `_STATIC_COMMAND_METADATA` entries from the latest Redis release

#### 1. Find the gaps
Every name in `READ_COMMANDS` (`redis/commands/cluster.py:63`) with no entry in `_STATIC_COMMAND_METADATA` (`redis/commands/metadata.py`).

#### 2. Get the data (pick one)
- Live: start the image CI pins (`CURRENT_REDIS_VERSION` in `.github/workflows/integration.yaml`), then `redis-cli COMMAND INFO ttl scan dbsize ...`
- Offline: `src/commands/<cmd>.json` in the redis/redis repo at the release tag (`command_flags`, `command_tips`, `key_specs`)

#### 3. Map to fields
Same mapping `_to_command_metadata` uses (`redis/_parsers/commands.py:748`):

| Reply | Field |
|---|---|
| flag `readonly` | `is_readonly` |
| flag `blocking` | `is_blocking` |
| flag `script_runner` | `is_script_runner` |
| tip `nondeterministic_output` (exact match) | `has_nondeterministic_output` |
| tip `dont_cache` | `is_dont_cache` |
| key spec without `not_key` (fallback: `first_key > 0 and step > 0`) | `has_key_argument` |
| tips `request_policy:*` / `response_policy:*`; flag `movablekeys` = withhold policies | routing policies |

#### 4. Don't build records by hand
Match each command to an existing shape constant in `metadata.py` (`_CACHEABLE_KEYED`, `_READONLY_KEYLESS`, `_DONT_CACHE_KEYED`, `_CACHEABLE_MOVABLE_KEYS`, ...). Most of the ~22 fit one.

#### 5. Finish
- Lowercase keys, subcommands space-joined (`"memory usage"`)
- Update the provenance note (`redis/commands/metadata.py:955`) with the Redis version used
- Run `pytest tests/test_command_metadata.py tests/test_asyncio/test_command_metadata.py`

redis/asyncio/cluster.py
```diff
@@ -873,11 +890,33 @@ def get_random_primary_or_all_nodes(self, command_name):
...
+
+        if self.read_from_replicas and await self._is_replica_safe(command_name):
+            return self.get_random_node()
+
+        return self.get_random_primary_node()
+
+    async def _is_replica_safe(self, command_name):
+        if self._routing_uses_metadata:
+            return await self._metadata_resolver.is_replica_safe(command_name)
+        return isinstance(command_name, str) and command_name.upper() in READ_COMMANDS
```
petyaslavova: We should always consult with the metadata resolver to identify if the command is replica safe except in the case where it is not a RedisCluster instance

redis/asyncio/cluster.py
```diff
@@ -889,11 +928,12 @@ async def get_nodes_from_slot(self, command: str, *args):
         Returns a list of nodes that hold the specified keys' slots.
         """
         # get the node that holds the key's slot
+        replica_safe = await self._is_replica_safe(command)
```
petyaslavova: I don't think we want to consult with the metadata resolver here. The method explicitly states that it looks only for primary nodes, so in this case we don't care if the command is replica safe or not.

redis/commands/cluster.py
```diff
@@ -426,7 +427,8 @@ async def _execute_pipeline_by_slot(
     ) -> List[Any]:
         if self._initialize:
             await self.initialize()
-        read_from_replicas = self.read_from_replicas and command in READ_COMMANDS
+        replica_safe = await self._is_replica_safe(command)
+        read_from_replicas = self.read_from_replicas and replica_safe
```
petyaslavova: It will be better performance wise to keep those on one line - evaluate the result of `self._is_replica_safe(command)` only when self.read_from_replicas is True

redis/commands/cluster.py
```diff
@@ -204,7 +204,8 @@ def _partition_pairs_by_slot(
     def _execute_pipeline_by_slot(
         self, command: str, slots_to_args: Mapping[int, Iterable[EncodableT]]
     ) -> List[Any]:
-        read_from_replicas = self.read_from_replicas and command in READ_COMMANDS
+        replica_safe = self._is_replica_safe(command)
```
petyaslavova: The mixin itself should provide a default `_is_replica_safe` implementation with the legacy behavior (`command.upper() in READ_COMMANDS`) - otherwise third-party clients composing `RedisClusterCommands` directly will fail with `AttributeError`, since `ClusterCommandsProtocol` neither requires nor provides this method. Since we are deprecating `READ_COMMANDS`, this default should also emit the deprecation warning when used. The same applies to the async mixin - the default there should be `async def` so the `await` call sites keep working.

petyaslavova (changes requested):
Hey @SantoshDhaladhuli, thank you for applying all these changes :) Sorry for the delayed review!

redis/asyncio/cluster.py
```diff
@@ -876,11 +891,31 @@ def get_random_primary_or_all_nodes(self, command_name):
         """
         Returns random primary or all nodes depends on READONLY mode.
         """
-        if self.read_from_replicas and command_name in READ_COMMANDS:
+        if self.read_from_replicas and command_name.upper() in READ_COMMANDS:
```
petyaslavova: This check should use the metadata_resolver.

redis/asyncio/cluster.py
```diff
@@ -1289,12 +1326,11 @@ async def _execute_command(
...
                     slot = await self._determine_slot(*args)
+                    replica_safe = await self._is_replica_safe(args[0])
                     target_node = self.nodes_manager.get_node_from_slot(
                         slot,
-                        self.read_from_replicas and args[0] in READ_COMMANDS,
-                        self.load_balancing_strategy
-                        if args[0] in READ_COMMANDS
-                        else None,
+                        self.read_from_replicas and replica_safe,
+                        self.load_balancing_strategy if replica_safe else None,
```
petyaslavova: This one can be transformed into the following(of course, formatted properly :) ), with this we don't always evaluate the call to `self._is_replica_safe(args[0])` and save some time on the hot path when reading from replicas is not enabled: 
```
replica_safe = (
                        self.read_from_replicas
                        or self.load_balancing_strategy is not None
                    ) and await self._is_replica_safe(args[0])
                    target_node = self.nodes_manager.get_node_from_slot(
                        slot,
                        replica_safe,
                        self.load_balancing_strategy if replica_safe else None,
                    )
```

redis/commands/cluster.py
```diff
@@ -201,17 +209,34 @@ def _partition_pairs_by_slot(
...
+        return isinstance(command_name, str) and command_name.upper() in READ_COMMANDS
+
     def _execute_pipeline_by_slot(
         self, command: str, slots_to_args: Mapping[int, Iterable[EncodableT]]
     ) -> List[Any]:
-        read_from_replicas = self.read_from_replicas and command in READ_COMMANDS
+        replica_safe = self._is_replica_safe(command)
+        read_from_replicas = (
+            getattr(self, "read_from_replicas", False) and replica_safe
+        )
```
petyaslavova: Please combine those two, so that the call to `self._is_replica_safe(command)` will be evaluated only when getattr(self, "read_from_replicas", False)  is True

redis/commands/cluster.py
```diff
@@ -421,19 +446,36 @@ async def _split_command_across_slots(self, command: str, *keys: KeyT) -> int:
...
     async def _execute_pipeline_by_slot(
         self, command: str, slots_to_args: Mapping[int, Iterable[EncodableT]]
     ) -> List[Any]:
         if self._initialize:
             await self.initialize()
-        read_from_replicas = self.read_from_replicas and command in READ_COMMANDS
+        replica_safe = await self._is_replica_safe(command)
+        read_from_replicas = (
+            getattr(self, "read_from_replicas", False) and replica_safe
+        )
```
petyaslavova: Those can also be combined

redis/asyncio/cluster.py
```diff
@@ -892,11 +927,12 @@ async def get_nodes_from_slot(self, command: str, *args):
         Returns a list of nodes that hold the specified keys' slots.
         """
         # get the node that holds the key's slot
+        replica_safe = await self._is_replica_safe(command)
```
petyaslavova: Optimise this as already described in one of the other comments.

petyaslavova (changes requested):
Hey @SantoshDhaladhuli, thanks for the quick turnaround. The core of it landed well: the interface methods are abstract-only, `_REPLICA_UNSAFE_COMMANDS` is seeded into the memo, `_is_replica_safe(metadata)` mirrors `_is_client_side_cacheable`, and every `READ_COMMANDS` entry is now in `_STATIC_COMMAND_METADATA` with a readonly record. A few things still need to happen before this can merge, in addition to the inline notes.

**1. The same short-circuit applies in five places:** `get_nodes_from_slot` in both stacks, the MOVED refresh in `redis/cluster.py`, and `_execute_pipeline_by_slot` in `redis/commands/cluster.py` (sync and async). For example, the async `get_nodes_from_slot` becomes:

    replica_safe = (
        self.read_from_replicas or self.load_balancing_strategy is not None
    ) and await self._is_replica_safe(command)
    return [
        self.nodes_manager.get_node_from_slot(
            await self._determine_slot(command, *args),
            replica_safe,
            self.load_balancing_strategy if replica_safe else None,
        )
    ]

Note that `replica_safe` is passed straight through - `self.read_from_replicas and replica_safe` is redundant once the guard is in the assignment, since `get_node_from_slot` only acts on the resolved `load_balancing_strategy`. In `_execute_pipeline_by_slot` the guard has a second benefit: the mixin's default `_is_replica_safe` emits a `DeprecationWarning`, so today every `mset_nonatomic`/`exists`/`delete` warns even for clients that never read from replicas. Please also replace the `getattr(self, "read_from_replicas", False)` lookups there with plain attribute access - those attributes are always present, and the default hides a real misconfiguration.

**2. Two comments lost their endings during editing.** The provenance note in `redis/commands/metadata.py` now stops mid-sentence and then claims the table was generated against Redis 8.6-rc1, while the paragraph above it says 8.10.0 and `STATIC_TABLE_SERVER_VERSION` is still 8.10.0. The comment above `StaticPolicyResolver` in `redis/cluster.py` also lost its closing clause. That file is our provenance record, so please restore both and settle on one version.

**3. Docs that need updates.** The eligibility-delta comment on `CacheConfig.DEFAULT_ALLOW_LIST` still says the difference is `+FT.SUGGET`/`+FT.SUGLEN`, but this PR makes seven more commands cacheable (`DIGEST`, `EXPIRETIME`, `PEXPIRETIME`, `HEXPIRETIME`, `HPEXPIRETIME`, `SDIFFCARD`, `SUNIONCARD`). The `metadata_resolver` docstring in `redis/commands/policies.py` and the async note in `AGENTS.md` both still say the async clients take no `metadata_resolver`.

**4. One test we need.** `_determine_nodes` now only consults `COMMAND_FLAGS` when the resolved `request_policy` is `None`, so correctness depends on every static-table entry that also appears in the cluster `COMMAND_FLAGS` withholding its routing policies. That holds today, and the last two routing regressions here were exactly that case, so please add a test derived from the table rather than a hand-written list. Relatedly, `request_policy` is now typed `Optional[RequestPolicy]` but `self._policies_callback_mapping[request_policy]` raises `KeyError: None` for any command outside `COMMAND_FLAGS` - either handle `None` or keep the argument required.

One last thing for the description: because replica safety is now derived from `is_readonly`, 19 commands that were never in `READ_COMMANDS` become replica-eligible - `FT.AGGREGATE`, `FT.SPELLCHECK`, `FT.TAGVALS`, `FT.SYNDUMP`, `FT.DICTDUMP`, `FT.EXPLAINCLI`, `TS.GET`, `TS.RANGE`, `TS.REVRANGE`, `SORT_RO`, `GEORADIUS_RO`, `GEORADIUSBYMEMBER_RO`, `SUBSTR` and others. That follows from letting metadata decide and I think it is the right outcome, but it is a user-visible routing change that should be called out explicitly and covered by a test for at least the `FT.*` group.

redis/cluster.py
```diff
@@ -1020,7 +1024,7 @@ class in the private `redis._parsers` package. Note that a server-derived resolv
             Union[RequestPolicy, ResponsePolicy], Callable
         ] = {
             RequestPolicy.DEFAULT_KEYLESS: lambda command_name: [
-                self.get_random_primary_or_all_nodes(command_name)
+                self.get_keyless_target_node(command_name)
             ],
```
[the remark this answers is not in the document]
petyaslavova: Rejecting this one - the sync change is intentional, and it is the point of the change rather than a side effect.

**Why the sync mapping was changed too.** Before this PR the two stacks disagreed: sync keyless routing went through `get_random_primary_or_all_nodes()` and async did not. Restoring the sync call site would preserve that asymmetry, which is exactly what this PR set out to remove.

**Why "route through the compatibility method" is not implementable.** `get_keyless_target_node()` is `async def` on the async stack, because replica safety now resolves through an awaitable metadata resolver. The deprecated name there is a synchronous shim that runs that coroutine on a private event loop in a worker thread (`_run_coroutine_in_thread`). Calling it from the routing path would block the client's loop, and for a caller-supplied `AsyncMetadataResolver` whose `is_replica_safe` awaits I/O it would await a loop from a thread that has blocked it, and hang. That hazard is documented on the method itself. So the suggestion can only be applied to the sync stack, and a sync-only fix is the divergence again.

**Why it would also be wrong on the sync stack alone.** `deprecated_function` emits the warning on every call, so routing internally through the deprecated name would raise a `DeprecationWarning` for every keyless command - `PING`, `SCRIPT LOAD`, `FT.SEARCH` - at users who never touched the old name. Moving internal call sites to the new name at the same time the alias is introduced is what keeps the warning meaningful.

**On the override scenario.** Overriding a `get_*` selector to pin a command to a node relied on an internal call site, not a documented contract, and it was never expressible on the async stack. The supported ways to do this are `target_nodes=` per call and `policy_resolver=` / `metadata_resolver=` at client construction, which is the extension point this PR centralizes.

**Acknowledged residual.** A subclass that overrides `get_random_primary_or_all_nodes()` stops influencing routing, and the `DeprecationWarning` fires at the call site rather than at the override. That is accepted for the deprecation window and will be called out in the release notes.

redis/cluster.py
```diff
@@ -5329,7 +5414,7 @@ def _parse_target_nodes(self, target_nodes):
         return nodes
 
     def _determine_nodes(
-        self, *args, request_policy: RequestPolicy, **kwargs
```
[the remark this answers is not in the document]
petyaslavova: Accurate, and it is a known gap rather than an oversight - it is tracked as a follow-up from the earlier review round on this PR.

Some detail on why it was not consolidated here. The sync pipeline copy is not a verbatim duplicate of `_resolve_command_policies`, which is why it could not simply be swapped out:

1. It special-cases `EVAL`/`EVALSHA` to the keyed record directly, where the shared helper reaches the same outcome indirectly through `determine_slot`, which has its own `EVAL`/`EVALSHA` branch.
2. It resolves keys with `_get_command_keys(*args)` where the shared helper calls `determine_slot(*args)`.

Those two routes agree today for the commands that reach them, but they are not interchangeable in general. `determine_slot` raises `RedisClusterException` for an unflagged command with no keys, whereas the pipeline copy falls back to keyless routing. Pointing the sync pipeline at the shared helper would therefore make pipelines start raising where they currently route keyless. That asymmetry predates this PR, so it is a behavior change to make deliberately and with its own tests, not a drive-by inside a refactor of this size.

So the plan is: prove the two key-resolution routes agree for every unflagged command, decide the raise-versus-keyless question on its own merits, then delete the copy. Until then the duplication is documented so any edit to the fallback chain is made in both places on the sync side.

redis/commands/metadata.py
```diff
@@ -325,6 +338,19 @@ def is_cacheable(self, command_name: str) -> bool:
         """
         pass
 
+    @abstractmethod
+    def is_replica_safe(self, command_name: str) -> bool:
```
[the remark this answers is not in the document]
petyaslavova: Not valid - this is a completely new functionality that is not yet released.

petyaslavova (approved):

PR #4267, opened 2026-08-17
Assert the reclamation property in the _connect cycle tests

clemenslosbichler-cloud:
### Description of change

Follows up on the note in #4195, per @petyaslavova's ask there.

Both tests ended in `assert connect_frame.f_locals.get("err") is None`. `dict.get` returns `None` when the local was cleared and also when there is no local by that name, so renaming `err` while dropping both clears would have put the cycle straight back and left the tests green.

The fail-then-succeed case can assert the property instead of the name: with the GC disabled, refcounting alone has to reclaim the caught exception once `_connect` returns. It raises from a factory so the test holds no strong reference of its own, and it no longer mentions `err` anywhere.

The all-addresses-fail case cannot do that. The exception escapes to the caller, so it is legitimately still alive and there is nothing to assert dead. That one keeps frame introspection, with `assert "err" in connect_frame.f_locals` added before the value check so a rename fails loudly rather than passing vacuously.

Checked both ways: with the two `err = None` clears reverted, both tests fail; with them in place, both pass. `tests/test_connection.py` is otherwise unchanged from master on my machine (3 failed / 86 passed / 5 errors before and after, all needing a live server).

Both gaps were reported by [Context Goblin](https://github.com/apps/contextgoblin), an automated reviewer I'm building, run against a fork with #4195 recreated on it.

### Pull Request check-list

Tests only, so no docs or examples. CI box left unchecked since I have not enabled actions on the fork.

---

> [!NOTE]
> **Low Risk**
> Test-only changes to connection reference-cycle coverage; no library behavior is modified.
> 
> **Overview**
> **Tightens regression tests** for `Connection._connect` clearing the `err` local so failed connection attempts do not leave exception↔frame reference cycles.
> 
> For the all-addresses-fail path, the test now requires `"err"` to exist in the `_connect` frame before asserting it is `None`, so a renamed or removed local cannot pass via `f_locals.get("err")`.
> 
> For fail-then-succeed, frame introspection is replaced with **GC disabled** and a **weak reference** to a custom `OSError` subclass: after `_connect` returns the working socket, the caught exception must be gone without relying on the local’s name. The test is **skipped on PyPy** (refcount semantics). Docstrings clarify why the two cases use different strategies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit abd66b2dbcafe07f61c69437653caf0ac9405377. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1,4 +1,5 @@
 import copy
+import gc
 import os
 import platform
 import select
@@ -9,6 +10,7 @@ import sys
 import threading
 import time
 import types
+import weakref
 from errno import ECONNREFUSED, EWOULDBLOCK
 from typing import Any
 from unittest import mock
@@ -582,6 +584,10 @@ class TestConnection:
         The exception's traceback references the frame, so a retained local
         would form a cycle only reclaimable by the GC.
         The finally clause in _connect breaks it by clearing the local.
+
+        The exception escapes to the caller here, so it cannot be asserted
+        dead; the frame is what has to be checked. See the sibling test for
+        the reclamation property, which does not depend on the local's name.
         """
         conn = Connection(host="localhost", port=6379)
         addr_info = (socket.AF_INET, socket.SOCK_STREAM, 0, "", ("127.0.0.1", 6379))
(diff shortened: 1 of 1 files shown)


petyaslavova (changes requested):
Hey @clemenslosbichler-cloud, thank you for your contribution!

Thanks for following up on the note in #4195!

Could you please address the points raised in the Codex review comments, and fix the linter error reported by the build?

Once CI is green, I'll take another look.

Mukller (approved):
Test-only PR, verified locally (Python 3.13):

- The two targeted tests pass: `pytest tests/test_connection.py -k "reference_cycle or reclaim"` — 2 passed.
- The redesign is a genuine improvement over the previous frame-peeking approach: instead of asserting an implementation detail ("a local named `err` was cleared"), the new sibling test asserts the actual *property* — the raised exception becomes unreachable after `_connect` returns — via `weakref` with `gc.disable()`/`finally gc.enable()`. That keeps passing even if someone renames the local or restructures the `finally`, which is what a regression test should want.
- Details are handled correctly: the `WeakReferenceableOSError` subclass (OSError isn't weak-referenceable), and clearing the test frame's own local in the factory so the assertion measures `_connect` and not the harness.
- The `"err" in connect_frame.f_locals` guard added to the first test closes a real hole — `.get("err") is None` also holds when the local doesn't exist at all.

For completeness I ran the full `tests/test_connection.py`: besides these, remaining reds are `ConnectionRefusedError` setup errors from server-dependent tests — identical on upstream main without a local Redis, i.e. environment, not this diff (+41/-18 touches only that one test file).

petyaslavova (commented):
Thanks for the quick turnaround! The updates look good: the F401 on the dropped `sys` import is gone, the reclamation assertion is now skipped on PyPy in line with the other interpreter-specific guards in this file, and the collector state is recorded and restored rather than unconditionally re-enabled, so a harness that already disabled cyclic GC is not affected.

The previous run failed on exactly those two points (the linter and both PyPy `fixed-clients` jobs), so I am approving the workflow runs on the latest commit. Once CI is green I will merge this.

petyaslavova (approved):

PR #4272, opened 2026-08-18
feat: decide client-side cache eligibility from command metadata

petyaslavova:
## Change summary

Client-side-cache eligibility came from the hardcoded `CacheConfig.DEFAULT_ALLOW_LIST` while
cluster routing came from `STATIC_POLICIES` - two hand-maintained views of the same `COMMAND`
reply. They had drifted: the allow list admits `XPENDING`, `TS.INFO` and `XREAD`, all three of
which the server reports as not cacheable. The new `redis/commands/metadata.py` makes command
metadata the single source for both. `CommandMetadata` carries the routing policies plus the
cacheability inputs (`readonly`, `blocking`, key argument, `nondeterministic_output`,
`script_runner`, `dont_cache`, metadata completeness); `_is_client_side_cacheable` is the only
implementation of the rules and every field fails closed, so unknown commands are never cached.
`MetadataResolver` serves records by name, chains through `with_fallback`, and answers routing
(`resolve_policies`) and caching (`is_cacheable`) off the same record, memoized per name.
`_STATIC_COMMAND_METADATA` is the shipped table, validated against a live Redis 8.10 stack reply
(and slightly updated to reflect some library restrictions in the current code).

`redis/_parsers/commands.py` gains `get_command_metadata()` and builds both views from one
traversal, so they cannot disagree; the policy resolvers become the routing projection of a
metadata resolver. `metadata_resolver=` is plumbed through `Redis`, `ConnectionPool`,
`RedisCluster`, `NodesManager` and `ClusterPipeline`, built once per client, shared with every
node client, and injected into `CacheConfig` - into a copy for `cache_config=`, in place for
`cache=`/`cache_factory=`.
The resolver classes are mirrored in async, but the async clients take no `metadata_resolver` yet, since the async stack has no client-side cache available. The metadata logic is used the same way for policy resolvers in async as in sync implementation.

Compatibility, against 7.1.0:

- Cacheable set changes by `+FT.SUGGET`, `+FT.SUGLEN` and `-XPENDING`, `-TS.INFO`, `-XREAD`.
- `CacheProxyConnection.send_command` no longer raises `ValueError("Cannot create cache key.")`
  when an eligible command carries no `keys`; it sends normally and caches nothing.
- `STATIC_POLICIES`, `DEFAULT_ALLOW_LIST` and `get_command_policies()` are deprecated and unread.
- `RequestPolicy`, `ResponsePolicy`, `CommandPolicies`, `PolicyRecords` moved to
  `redis.commands.metadata`, re-exported from `redis._parsers.commands`.
- `policy_resolver` defaults to `None` and is built per client; an explicit one still wins over
  `metadata_resolver` for routing.
- `RedisCluster._get_command_keys` raises `RedisClusterException` instead of `AttributeError`
  when there is no default node.

## Test coverage

New `tests/test_command_metadata.py` and its async mirror cover each eligibility rule as an
independent veto, the unknown/incomplete-metadata refusals, withheld routing policies for the
`movablekeys` reads, case-insensitive lookup, memoization and its cap, fallback chains, and the
deprecation warnings; the resolver tests from `tests/test_command_policies.py` moved here and
were extended. Live-server tests assert the shipped table still matches a real `COMMAND` reply
and that each recorded divergence is still needed. `tests/test_cache.py` and
`tests/test_connection.py` cover the eligibility path and the resolver-injection matrix,
`tests/test_cluster.py` and its async mirror cover per-client resolvers, pipeline routing,
`policy_resolver` precedence and the new missing-default-node error. Docs updated in
`docs/resp3_features.rst`.

---

> [!NOTE]
> **High Risk**
> Touches cluster command routing and client-side cache eligibility, including which commands are cached and how movablekeys reads are dispatched. Compatibility shims exist, but wrong metadata would misroute or cache incorrectly.
> 
> **Overview**
> Makes **command metadata** the single source of truth for both **client-side cache eligibility** and **cluster routing**, replacing the drifted `CacheConfig.DEFAULT_ALLOW_LIST` and unread `STATIC_POLICIES` table.
> 
> Adds `redis.commands.metadata` with frozen `CommandMetadata` records, fail-closed `_is_client_side_cacheable` rules, and chainable `MetadataResolver`s (static/dynamic, sync/async). Policy resolvers become a routing projection of that metadata; `movablekeys` reads withhold policies so the client still resolves keys itself.
> 
> `metadata_resolver=` is plumbed through sync `Redis`, `ConnectionPool`, `RedisCluster`, `NodesManager`, and pipelines (async clients skip the arg until CSC exists). Cache eligibility now asks the resolver; missing `keys=` on a cacheable command **skips caching** instead of raising. Cacheable set shifts by **+FT.SUGGET/FT.SUGLEN** and **-XPENDING/TS.INFO/XREAD**.
> 
> `get_command_policies()`, `STATIC_POLICIES`, and `DEFAULT_ALLOW_LIST` are deprecated. Policy types move to `metadata.py` with re-exports. `RedisCluster._get_command_keys` raises `RedisClusterException` when there is no default node.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5310f9f474bacf9b9e8a30ddde304cb50a2610f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redis/cache.py
```diff
@@ -458,7 +502,11 @@ def is_exceeds_max_size(self, count: int) -> bool:
         return count > self._max_size
 
     def is_allowed_to_cache(self, command: str) -> bool:
-        return command in self.DEFAULT_ALLOW_LIST
+        # Fails closed on everything the resolver cannot decide: an unknown command, a name
+        # the record tables cannot be keyed by, and a record built from incomplete metadata
+        # all resolve to False. The verdict is memoized per command name, so this is a dict
+        # hit on the command execution path.
+        return self._metadata_resolver.is_cacheable(command)
```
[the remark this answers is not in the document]
petyaslavova: We can't integrate this list with the metadata resolver. Even if a consumer would edit this list - those additional commands would not work because the commands should provide additional keys kwarg, and lately we identified that even commands listed in this list are not all providing it. Consider this as invalid feedback.

redis/connection.py
```diff
@@ -2986,6 +3005,7 @@ def __init__(
         max_connections: Optional[int] = None,
         cache_factory: Optional[CacheFactoryInterface] = None,
         maint_notifications_config: Optional[MaintNotificationsConfig] = None,
+        metadata_resolver: Optional[MetadataResolver] = None,
```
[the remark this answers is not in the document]
petyaslavova: Invalid. This is a newly introduced arg

redis/commands/metadata.py
```diff
@@ -0,0 +1,1186 @@
...
+            self.request_policy == other.request_policy
+            and self.response_policy == other.response_policy
+        )
+
+    def __hash__(self) -> int:
+        # Defined alongside ``__eq__`` because Python would otherwise set it to None, which
+        # would make a type that shipped hashable in 7.1.0 unhashable.
+        return hash((self.request_policy, self.response_policy))
+
+    def __repr__(self) -> str:
```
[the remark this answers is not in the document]
petyaslavova: invalid

redis/commands/metadata.py
```diff
@@ -0,0 +1,1186 @@
...
+    rebuilding it. Caller-supplied records carry whatever spelling the caller chose, and a
+    resolver that silently answers nothing for a table keyed ``"ZCOUNT"`` would be worse than
+    one that normalizes it.
+    """
+    if all(
+        module_name == module_name.lower()
+        and all(command == command.lower() for command in commands)
+        for module_name, commands in records.items()
+    ):
+        return records
```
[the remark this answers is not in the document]
petyaslavova: The snapshot is the documented contract, not an accident. Invalid

redis/_parsers/commands.py
```diff
@@ -210,189 +197,57 @@ def _get_moveable_keys(self, redis_conn, *args):
...
-            )
-        else:
-            command_details = self.commands.get(command_name, None)
-            if command_details is not None:
-                return command_details["first_key_pos"] <= 0
-
-            raise ValueError(f"Command {command_name} not found in commands")
-
+    @deprecated_function(
+        version="8.1.0",
```
vladvildanov: 8.1.0 has already been released, should be 8.2.0

redis/commands/metadata.py
```diff
@@ -0,0 +1,1186 @@
...
+# =============================================================================
+# This table is seeded from old ``CacheConfig.DEFAULT_ALLOW_LIST`` and ``redis.commands.policies.STATIC_POLICIES``.
+#
+# Every entry below was validated against a live ``COMMAND`` reply from Redis 8.10.0 with
+# search, timeseries, ReJSON, bf and vectorset loaded (the ``redislabs/client-libs-test``
+# stack image).
+#
+# The table is not exhaustive: the server reports 127 cacheable commands and this covers
+# 73 of them. The uncovered ones are whole module surfaces the CSC allow-list never
+# carried - ``bf.*``, ``cf.*``, ``cms.*``, ``topk.*``, ``tdigest.*``, the vectorset reads
```
vladvildanov, line 967: According to design we shouldn't skip modules command and should rely only on metadata

redis/commands/metadata.py
```diff
@@ -0,0 +1,1186 @@
...
+        try:
+            metadata = self.resolve(command_name)
+        except ValueError:
+            # More than one module prefix, which ``_split_command_name`` refuses.
+            metadata = None
+
+        cacheable = _is_client_side_cacheable(metadata)
+
+        if len(self._cacheable) < _MEMO_MAX_ENTRIES:
+            self._cacheable[command_name] = cacheable
```
vladvildanov: Command names requires normalisation to avoid duplicates like `get` and `GET`

vladvildanov (changes requested):

redis/_parsers/commands.py
petyaslavova: Fixed.

redis/commands/metadata.py
petyaslavova: I would prefer to add the rest and have full coverage in separate PR - after the multi-shard and movable keys are properly handled in the request/response policies. With this change we cover all commands that have been covered before.

redis/commands/metadata.py
petyaslavova: Fixed.

vladvildanov (approved):

redis/commands/metadata.py
```diff
@@ -0,0 +1,1219 @@
...
+                "explain": _DONT_CACHE_READONLY_KEYLESS,
+                "explaincli": _DONT_CACHE_READONLY_KEYLESS,
+                "info": _DONT_CACHE_READONLY_KEYLESS,
+                "profile": _DONT_CACHE_READONLY_KEYLESS,
+                "search": _DONT_CACHE_READONLY_KEYLESS,
+                "spellcheck": _DONT_CACHE_READONLY_KEYLESS,
+                "sugadd": _DONT_CACHE_WRITE_KEYED,
+                "sugdel": _DONT_CACHE_WRITE_KEYED,
+                "sugget": _CACHEABLE_KEYED,
+                "suglen": _CACHEABLE_KEYED,
```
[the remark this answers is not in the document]
petyaslavova: Out of the scope of this PR

PR #3797, opened 2025-10-14
Add handling of empty spaces during CLIENT LIST response parsing

peperon:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [ ] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

This change is addressing an issue where we may have white space in the value of some the attributes returned by `CLIENT LIST`. For example, when runing redis with unix socket like:
```
redis-server --port 0 \
  --unixsocket "/tmp/redis sock/redis.sock" \
  --unixsocketperm 777 \
  --save "" --appendonly no
```
Where the socket file have white space in the path - `/tmp/redis sock/redis.sock`. With the current implementation, in this case we getting exception like - `ValueError: dictionary update sequence element #n has length 1; 2 is required.`.

PR #4275, opened 2026-08-22
Don't crash parsing CLIENT INFO with '=' or spaces in a field value

eeshsaxena:
parse_client_info splits each field token on '=' with no maxsplit and tokenizes on arbitrary whitespace, so a value containing '=' (a client name like "foo=bar", which Redis allows) or a space (a unix-socket addr like "/tmp/redis sock/redis.sock") raises ValueError. This trips up both CLIENT INFO responses and ACL LOG parsing, which embeds the same client-info.

The sibling parse_client_list already handles both cases, so this reuses that same tokenization (split on '=' once, reattach space-separated continuations) and adds a regression test.

---

> [!NOTE]
> **Low Risk**
> Localized response-parser change with regression tests; no auth, persistence, or wire-protocol behavior changes beyond fixing parsed dict shapes.
> 
> **Overview**
> Fixes **CLIENT INFO** / **ACL LOG** parsing when Redis returns `key=value` blobs where a value includes **`=`** (e.g. client name `test=name`) or **spaces** (e.g. Unix-socket `addr` paths). The old `parse_client_info` split every token on `=` with no limit and used whitespace tokenization, which could raise **ValueError** or mis-parse fields.
> 
> The PR adds shared **`_parse_client_info_fields`** (split only on the first `=`, reattach space-separated continuation tokens, safe empty/leading-junk handling) and wires both **`parse_client_info`** and **`parse_client_list`** through it so behavior stays consistent. **Unit tests** cover complex blobs, empty client-info, and malformed **CLIENT LIST** lines; **sync/async integration tests** assert a client name with `=` round-trips via **`client_setname`** / **`client_info`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab231d82ce3ba5753430a69feffd1fa2f24c8383. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Mukller (commented):
Verified locally against the PR branch (installed redis 8.1.0 venv, ran from repo checkout so the patched module shadows site-packages):

**Tests**: `pytest tests/test_parsers/test_helpers.py` — 22/22 green, including the two new cases.

**Adversarial matrix through the new `parse_client_info`**:

| input | result |
|---|---|
| `id=7 addr=/tmp/redis sock/redis.sock:0` | `addr` = `/tmp/redis sock/redis.sock:0` ✓ (the unix-socket case) |
| `name=foo=bar=baz age=42` | `name` = `foo=bar=baz`, `age` coerced to int ✓ |
| `lib-ver=` (empty value) | stored as `""` ✓ |
| `totally-broken-token id=3` | junk token skipped instead of `ValueError` (old code crashed) |

The root fix is right: `split("=", 1)` + reattachment via `last_key`, with the `last_key is None` sentinel correctly keeping empty ACL-LOG client-info as `{}`.

One minor nit, non-blocking: because `.split(" ")` no longer collapses whitespace like the old `.split()`, a **double space between two pairs** attaches an empty token to the previous value:

```python
parse_client_info("x=1  y=2")
# -> {'x': '1 ', 'y': '2'}     # note trailing space on 'x'
```

Real `CLIENT INFO` output separates pairs with single spaces, so this shouldn't occur in practice — but if you want strict parity with the old parser's whitespace collapsing, you could skip empty tokens in the reattachment branch (`if not token: continue`). Either way fine by me.

CI question: the failing job is the single combo "Redis 8.4.3; Python 3.14; plain parser; asyncio; **2-unified_responses**". Under RESP3/unified responses, does `CLIENT INFO` still reach `parse_client_info`, or does it arrive pre-parsed as a map (making this failure unrelated)? If unrelated/flaky, a re-run would clean it up.

petyaslavova (changes requested):
Hey @eeshsaxena, thank you for your contribution!

The defect is real and worth fixing: `parse_client_info` unpacks `token.split("=")` into exactly two values, so it raises `ValueError` both for a value containing `=` (Redis allows `=` in a client name) and for a space in a unix-socket `addr`/`laddr`. It reaches `CLIENT INFO` and all three `ACL LOG` parsers, on both stacks and every protocol/topology combination. #3797 fixed exactly this in `parse_client_list`, and this function was simply left behind - so aligning the two is the right call, and the empty-input guard you added in the second commit is a good catch.

One code change before merge: `parse_client_list` and `parse_client_info` now hold identical tokenization that has already diverged - only the new copy guards `last_key is not None`, while the older one would raise `KeyError` on that path. Please fold the loop into a single module-private helper in `redis/_parsers/helpers.py` and call it from both, so the guard applies to `CLIENT LIST` too and the two parsers of the same server-side blob cannot drift again.

Additionally: an integration test that does `client_setname("test=name")` and asserts the name round-trips through `client_info()` would cover the real server path alongside the unit tests.

Please also check the bot's comments.

petyaslavova (changes requested):
Thanks for the quick turnaround - the shared `_parse_client_info_fields` helper and the sync/async round-trip tests are exactly what was needed, and CI is fully green now, including the RESP3 unified-responses job that was failing earlier.

One change left. The `if not token: continue` branch collapses runs of spaces, which silently regresses `CLIENT LIST` for the case #3797 fixed: on master, `addr=/tmp/redis␣␣sock/r.sock:0` round-trips verbatim, but on this branch it comes back as `/tmp/redis␣sock/r.sock:0`. A file name can contain consecutive spaces, so that path is reachable, whereas the double-space-between-pairs nit it was added for cannot occur in real server output. Please drop the `continue` and rely only on the `last_key is not None` guard - the empty-input fix still holds, since the single empty token from `""` or `"   "` hits that guard and yields `{}`, and `CLIENT LIST` stays byte-identical to today. The second assertion in `test_parse_client_list_shares_guard` will need updating; an assertion that a double-space socket path survives both parsers unchanged would be a better fit.

On the remaining bot comment about a value holding both a space and an `=` after it: no action needed. That ambiguity is unresolvable without pinning a closed set of field names, which would break whenever the server adds one. Keeping both parsers consistent is the right call.

petyaslavova (approved):
Thanks for the follow-up - this is what I was after. Dropping the empty-token skip restores the `CLIENT LIST` behavior from #3797 verbatim: a value with consecutive spaces now round-trips unchanged, while `""` and `"   "` still hit the `last_key is None` guard and yield `{}`. The shared `_parse_client_info_fields` means both parsers of the same server blob can no longer drift, and the sync/async `client_setname("test=name")` round-trips cover the real server path.

Agreed on the remaining bot note: a value containing both a space and a later `=` cannot be recovered from the flat blob without pinning a closed set of field names, and keeping both parsers consistent is the right call.

LGTM. I will re-run the one cancelled pypy cluster job (it timed out at 50 minutes, unrelated to this change) and merge from there.

PR #4276, opened 2026-08-22
fix(multidb): stop star-unpacking args in execute_pubsub_method

### LINKED ISSUE #4274: MultiDBClient.pubsub().subscribe() crashes on more than one channel (sync and async)
**Version**: redis-py 8.0.0 (also reproduces on current `master`, 081923b8)

**Platform**: Python 3.13, macOS — not platform-specific, this is a pure Python bug that never reaches the server

**Description**:

`MultiDBClient.pubsub().subscribe()` throws when subscribing to more than one channel, in both the sync and async client:

```python
client = MultiDBClient(config)
pubsub = client.pubsub()
pubsub.subscribe("channel-1", "channel-2")
```

```
TypeError: _execute_with_failure_detection() takes from 2 to 3 positional arguments but 4 were given
```

A single-channel subscribe doesn't crash, but it has a quieter bug: if that subscribe fails, the channel name gets split into individual characters before it reaches the failure detector, so `"channel-1"` is reported as `('c', 'h', 'a', 'n', 'n', 'e', 'l', '-', '1')` instead of `("channel-1",)`.

Root cause is in `DefaultCommandExecutor.execute_pubsub_method`, in `redis/multidb/command_executor.py` (and the identical code in `redis/asyncio/multidb/command_executor.py`):

```python
return self._execute_with_failure_detection(callback, *args)
```

`_execute_with_failure_detection` only takes one optional extra argument (`cmds: tuple = ()`), so star-unpacking `args` into it breaks for more than one positional arg, and silently mangles a single string arg into its characters. `execute_command` and `execute_pipeline`, right next to it in the same class, pass their tuple correctly, unstarred — this looks like a plain typo rather than anything intentional.

I have a one-line fix for both files plus regression tests (covering the crash and the corrupted-tuple case, sync and async) — will open a PR shortly.

pouyashahrdami:
## Summary

Fixes #4274.

`DefaultCommandExecutor.execute_pubsub_method` forwarded its `args` tuple star-unpacked into `_execute_with_failure_detection(callback, cmds: tuple = ())`, which only accepts one extra positional argument:

```python
return self._execute_with_failure_detection(callback, *args)
```

Subscribing to more than one channel through `MultiDBClient.pubsub()` raised `TypeError`. Subscribing to exactly one channel didn't crash, but the channel string was star-unpacked into individual characters before reaching the failure detector on the failure path, so `"channel-1"` got reported as `('c', 'h', 'a', 'n', ...)` instead of `("channel-1",)`.

The fix is passing the tuple unstarred, matching how `execute_command` and `execute_pipeline` already do it right next to this method:

```python
return self._execute_with_failure_detection(callback, args)
```

Same bug, same fix, in both `redis/multidb/command_executor.py` (sync) and `redis/asyncio/multidb/command_executor.py` (async).

## Test coverage

Added to both `tests/test_multidb/test_command_executor.py` and `tests/test_asyncio/test_multidb/test_command_executor.py`:

- `test_execute_pubsub_method_forwards_multiple_args` — subscribing to two channels no longer raises, and the failure detector is registered with the correct args tuple.
- `test_execute_pubsub_method_failure_reports_intact_command_tuple` — on a subscribe failure, the dispatched `OnCommandsFailEvent.commands` is `("channel-1",)`, not exploded into characters.

Confirmed both new tests fail against the pre-fix code (one with the `TypeError`, one with the corrupted tuple) and pass with the fix. Full `tests/test_multidb/` and `tests/test_asyncio/test_multidb/` suites (158 tests, excluding the infra-dependent integration tests) pass, and `ruff check` / `ruff format --check` are clean on all changed files.

---

> [!NOTE]
> **Low Risk**
> One-line argument-passing fix in command execution with tests; no auth, data, or failover-strategy changes.
> 
> **Overview**
> Stops `execute_pubsub_method` from star-unpacking `args` into `_execute_with_failure_detection` in both sync and async MultiDB executors.
> 
> That unpacking caused a `TypeError` when subscribing to more than one channel, and on the failure path a single channel name was exploded into characters in `OnCommandsFailEvent`. The tuple is now passed intact, matching `execute_command` / `execute_pipeline`. Tests cover multi-channel subscribe and intact failure-event commands.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd30a822cae8067299664ee9eb2ed40e88880959. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/multidb/command_executor.py
+++ b/redis/asyncio/multidb/command_executor.py
@@ -280,7 +280,7 @@ class DefaultCommandExecutor(BaseCommandExecutor, AsyncCommandExecutor):
             await self._register_command_execution(args)
             return response
 
-        return await self._execute_with_failure_detection(callback, *args)
+        return await self._execute_with_failure_detection(callback, args)
 
     async def execute_pubsub_run(
         self, sleep_time: float, exception_handler=None, pubsub=None
--- a/redis/multidb/command_executor.py
+++ b/redis/multidb/command_executor.py
@@ -300,7 +300,7 @@ class DefaultCommandExecutor(SyncCommandExecutor, BaseCommandExecutor):
             self._register_command_execution(args)
             return response
 
-        return self._execute_with_failure_detection(callback, *args)
+        return self._execute_with_failure_detection(callback, args)
 
     def execute_pubsub_run(self, sleep_time, **kwargs) -> "PubSubWorkerThread":
         def callback():
--- a/tests/test_asyncio/test_multidb/test_command_executor.py
+++ b/tests/test_asyncio/test_multidb/test_command_executor.py
@@ -4,7 +4,7 @@ from unittest.mock import AsyncMock
 import pytest
 
 from redis.asyncio.multidb.failure_detector import FailureDetectorAsyncWrapper
-from redis.event import EventDispatcher
+from redis.event import EventDispatcher, OnCommandsFailEvent
 from redis.exceptions import ConnectionError
 from redis.asyncio.multidb.command_executor import DefaultCommandExecutor
 from redis.asyncio.retry import Retry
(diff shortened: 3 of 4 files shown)


Mukller (approved):
Verified locally against both `upstream/main` and this branch (no server needed).

**The bug is real and this fixes it.** `_execute_with_failure_detection` declares a tuple contract (`callback, cmds: tuple = ()`), and almost all call sites already follow it (`execute_command` passes `args`, `execute_pipeline` passes `command_stack`). But the two `execute_pubsub_method` call sites still star-unpack, so any multi-piece command through them fails at binding time — reproduced directly on upstream main:

```python
await DefaultCommandExecutor._execute_with_failure_detection(cb, *("SET", "key", "value"))
# TypeError: _execute_with_failure_detection() takes from 2 to 3 positional
# arguments but 5 were given
```

Even a single-argument variant would silently degrade: `cmds` would become a string, and `_on_command_fail(error, *cmds)` would iterate it character by character. So every multidb pubsub path is currently either crashing or corrupting failure reporting — worth noting in the changelog as a bug fix, since it means multidb + PubSub was effectively unusable.

On the branch:
- both sync and async `test_command_executor.py` suites pass (6/6 + 6/6), including the new weighted-fallback parametrized case;
- the change is exactly the two call sites, no signature changes needed — consistent with the sibling methods.

Approving; nice catch.

petyaslavova (commented):
Hey @pouyashahrdami, thank you for your contribution!

The change looks good to merge as-is, and thanks to you for mirroring both the fix and the regression tests across the sync and async stacks. I will approve the CI run for the workflows and merge once the matrix is green.

petyaslavova (approved):

PR #4280, opened 2026-08-24
Fix asyncio Connection mutating caller retry_on_error list

### LINKED ISSUE #4278: redis.asyncio Connection mutates the caller-supplied retry_on_error list (sync version copies it)
`redis.asyncio.Connection` appends its default timeout exceptions to the **caller-supplied** `retry_on_error` list instead of copying it, so constructing connections mutates user data — and each pooled connection adds more duplicates.

**Version:** redis-py 8.1.0 (also on current `master`)
**Python:** 3.13.14

The sync `redis.Connection` copies the argument (`list(retry_on_error)`), so the asyncio behavior is also a parity bug.

## Repro

```python
from redis.asyncio import Connection as AsyncConnection

my_errors = [ValueError]
snapshot = list(my_errors)

AsyncConnection(host="h", retry_on_error=my_errors, retry_on_timeout=True)
print("before:", snapshot)
print("after :", my_errors)

# second connection -> grows again
AsyncConnection(host="h", retry_on_error=my_errors, retry_on_timeout=True)
print("after 2nd connection:", my_errors)
```

Output:

```
before: [<class 'ValueError'>]
after : [<class 'ValueError'>, <class 'redis.exceptions.TimeoutError'>, <class 'TimeoutError'>, <class 'TimeoutError'>]
after 2nd connection: [<class 'ValueError'>, ..., +3 more duplicates]
```

Each `Connection(...)` construction with `retry_on_timeout=True` appends `redis.TimeoutError`, `socket.timeout`, and `asyncio.TimeoutError` (three `.append()` calls in `AbstractAsyncConnection.__init__`) directly into the passed-in list.

## Impact

1. Surprising global side effect: a shared module-level list of "extra errors I care about" silently accumulates unrelated entries.
2. With a connection pool creating many connections the list grows unboundedly (duplicates), inflating every `except (...)` tuple evaluation.
3. Sync/async asymmetry makes code that shares config between both clients behave differently.

## Expected behavior

Copy before mutating, like the sync client does:

```python
retry_on_error = list(retry_on_error) if retry_on_error else []
```

Happy to send a one-line-plus-test PR.

leilei3167:
### Description of change

`AbstractAsyncConnection.__init__` appended timeout exceptions onto the caller-supplied `retry_on_error` list. A shared list (or one reused by a connection pool) therefore grew on every construction, and the sync client already avoided this by copying first.

Copy the list before appending, matching `AbstractConnection` in `redis/connection.py`. Construction no longer mutates the caller's list, including after two connections are created with `retry_on_timeout=True`.

Fixes #4278

### Pull Request check-list

---

> [!NOTE]
> **Low Risk**
> Localized constructor fix matching existing sync behavior, with a focused regression test; no change to retry runtime logic beyond avoiding shared-list mutation.
> 
> **Overview**
> Fixes a bug where **async** `AbstractConnection` appended timeout exception types directly onto the caller’s `retry_on_error` list when `retry_on_timeout=True`, so a shared list (e.g. from pool kwargs) could grow on every connection build. The sync client already copied first; async now does the same with `list(retry_on_error)` before appending `TimeoutError`, `socket.timeout`, and `asyncio.TimeoutError`.
> 
> The `retry_on_error` parameter type is tightened from bare `list` to `Iterable[Type[Exception]] | object` for clearer typing. A regression test ensures two constructions with `retry_on_timeout=True` leave the original list unchanged and give each connection its own copy with the expected timeout types added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 29a3a4342a78d78b393e6b86ded8ba65c7a2d50c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -665,6 +665,9 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
         self.retry_on_timeout = retry_on_timeout
         if retry_on_error is SENTINEL:
             retry_on_error = []
+        else:
+            # Copy so we never mutate the caller-supplied list (parity with sync).
+            retry_on_error = list(retry_on_error)
         if retry_on_timeout:
             retry_on_error.append(TimeoutError)
             retry_on_error.append(socket.timeout)
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -591,7 +591,7 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
         socket_timeout: float | None = DEFAULT_SOCKET_TIMEOUT,
         socket_connect_timeout: float | None = DEFAULT_SOCKET_CONNECT_TIMEOUT,
         retry_on_timeout: bool = False,
-        retry_on_error: list | object = SENTINEL,
+        retry_on_error: Iterable[Type[Exception]] | object = SENTINEL,
         encoding: str = "utf-8",
         encoding_errors: str = "strict",
         decode_responses: bool = False,
--- a/tests/test_asyncio/test_retry.py
+++ b/tests/test_asyncio/test_retry.py
@@ -59,6 +59,21 @@ class TestConnectionConstructorWithRetry:
         assert isinstance(c.retry, Retry)
         assert c.retry._retries == retries
 
+    @pytest.mark.parametrize("Class", [Connection, UnixDomainSocketConnection])
+    def test_retry_on_error_does_not_mutate_caller_list(self, Class):
+        retry_on_error = [ValueError]
+        snapshot = list(retry_on_error)
+
+        # Two constructions must not grow or otherwise mutate the caller's list.
+        c1 = Class(retry_on_error=retry_on_error, retry_on_timeout=True)
+        c2 = Class(retry_on_error=retry_on_error, retry_on_timeout=True)
+
+        assert retry_on_error == snapshot
+        assert retry_on_error is not c1.retry_on_error
+        assert ValueError in c1.retry_on_error
+        assert TimeoutError in c1.retry_on_error
+        assert c1.retry_on_error == c2.retry_on_error
+
     @pytest.mark.parametrize("retries", range(10))
     @pytest.mark.parametrize("Class", [Connection, UnixDomainSocketConnection])
     def test_retry_with_retry_on_error(self, Class, retries: int):


Mukller (approved):
I filed the underlying #4278 and verified this fix locally on the branch (Python 3.13, editable install) — the exact mutation scenario from my report is resolved:

```python
my_errors = [ValueError]
AsyncConnection(host="h", retry_on_error=my_errors, retry_on_timeout=True)  # x2
# master: my_errors grows by 3 entries per construction (TimeoutError/socket.timeout/asyncio.TimeoutError)
# branch: my_errors == [ValueError] after both constructions — caller list untouched
```

The one-line defensive copy is precisely what the sync client already did (`list(retry_on_error)`), restoring sync/async parity as intended. The added regression test covers the important properties: no mutation across repeated constructions, identity separation from the caller's list, and correct content of the internal copy.

`pytest tests/test_asyncio/test_retry.py -k "mutate"` passes locally; CI green.

Approving.

petyaslavova (commented):
Hey @leilei3167, thank you for your contribution!

This is the right fix at the right layer - `AbstractConnection` in `redis/connection.py` already copies, and mirroring that in the asyncio constructor restores parity and stops us from mutating a caller-owned list. Good call keeping the `is SENTINEL` branch intact; the snippet suggested in #4278 would have called `list()` on the sentinel. The regression test parametrized over both connection classes, asserting the caller's list is untouched and that the copy is a distinct object, covers it well.

One clarification for the record: `Retry.update_supported_errors` deduplicates through `set()`, so `_supported_errors` was never actually inflating - the impact is the caller-visible mutation and the unbounded growth of the shared pool kwargs list, which is exactly what this fixes. I will approve this change after we have green CI run.

If you would like to add one more line while you are here, aligning the async hint `retry_on_error: list | object` with the sync `Iterable[Type[Exception]] | object` would finish the parity story now that `list()` accepts any iterable. Entirely optional and not merge-blocking.

petyaslavova (commented):
Thanks for adding the type-hint alignment - the async signature now matches the sync `Iterable[Type[Exception]] | object`, so the parity story is complete.

I have finished reviewing this and it is good to merge as-is: the copy sits at the layer that owned the mutation, no other site in the library mutates `retry_on_error`, and the regression test covers the caller-list invariant across both connection classes. The only thing left is a green CI run, I just started it.

petyaslavova (approved):

PR #4288, opened 2026-08-25
fix(asyncio): raise ConnectionError when writing to a peer-closed connection

### LINKED ISSUE #4287: asyncio: TypeError/AttributeError instead of ConnectionError when writing to a peer-closed connection (CPython 3.12, 3.13.0-3.13.9, 3.14.0)
**Version:** redis-py 8.1.0 (code is unchanged on current `master`), Python 3.12 (all), 3.13.0–3.13.9, 3.14.0

**Platform:** Any (observed on Linux in production, reproduced on macOS), asyncio client, plain and cluster

**Description:**

When the server (or an intermediary) closes a pooled connection while it is idle, the next write on that connection leaks a transport-internal error instead of `ConnectionError`: `TypeError: 'NoneType' object is not callable` on Python 3.12 and 3.14.0, `AttributeError: 'NoneType' object has no attribute '_add_writer'` on 3.13.0–3.13.9 (all measured against a real transport after a genuine peer RST). Neither error is in the retry allowlists (`Retry` supported errors, `RedisCluster.ERRORS_ALLOW_RETRY`) and is not a `RedisError`, so it bypasses every recovery layer and surfaces to application code. We hit this in production on the auth path behind `RedisCluster` pipelines (~1000 events over two weeks); the trigger is simply an idle pooled connection being reset by the peer.

**Mechanism:**

1. The peer resets an idle pooled connection → the event loop runs `connection_lost`, which tears down the transport's internals (on 3.12/3.14.0 `_call_connection_lost` clears the instance attribute `self._write_ready`; on 3.13.x it clears `self._loop`).
2. `AbstractConnection.is_connected` only checks `self._reader is not None and self._writer is not None`, so the dead connection is picked up from the pool as healthy.
3. `send_packed_command` → `_send_packed_command` → `StreamWriter.writelines`. On affected CPython versions, `transport.writelines` is missing the `if self._conn_lost: return` guard that `transport.write` has ([CPython gh-136234](https://github.com/python/cpython/issues/136234)): on 3.12 and 3.14.0 it calls the cleared `self._write_ready` → `TypeError: 'NoneType' object is not callable`; on 3.13.0–3.13.9 `_write_ready` survives but the buffered write path hits the cleared `self._loop` → `AttributeError: 'NoneType' object has no attribute '_add_writer'`.
4. In `send_packed_command`, both errors only match the final `except BaseException` clause, which disconnects and re-raises them as-is — no translation to `ConnectionError`, so no retry engages.

CPython fixed `writelines` via gh-136234 in 3.13.10+ and 3.14.1+ (it now drops the data silently, and the subsequent `drain()` raises `ConnectionResetError`, which redis-py already translates and retries). 3.12 is in security-only mode, so the fix will not be backported — every asyncio user on 3.12 is exposed, plus anyone pinned to 3.13.0–3.13.9 or 3.14.0.

*(Edited: the original report framed this as 3.12-only; measurements on 3.13.0/3.14.0 and the gh-136234 timeline show the ranges above — thanks to @petyaslavova for catching it.)*

Traceback (redis-py 8.1.0, CPython 3.12.14, via `RedisCluster` pipeline):

```
File "redis/asyncio/cluster.py", line 1938, in execute_pipeline
    await connection.send_packed_command(
File "redis/asyncio/connection.py", line 1185, in send_packed_command
    await asyncio.wait_for(
File "redis/asyncio/connection.py", line 1168, in _send_packed_command
    self._writer.writelines(command)
File "asyncio/streams.py", line 349, in writelines
    self._transport.writelines(data)
File "asyncio/selector_events.py", line 1182, in writelines
    self._write_ready()
TypeError: 'NoneType' object is not callable
```

**Reproduction (Python 3.12):**

```python
import asyncio
import socket
import struct

from redis.asyncio.connection import Connection


async def main():
    holder = {}

    async def handle(reader, writer):
        holder["w"] = writer

    server = await asyncio.start_server(handle, "127.0.0.1", 0)
    port = server.sockets[0].getsockname()[1]
    reader, writer = await asyncio.open_connection("127.0.0.1", port)
    await asyncio.sleep(0.1)

    # Peer resets the idle connection. SO_LINGER(1, 0) forces a genuine RST;
    # a plain FIN would take asyncio's half-close path instead, which is fine.
    sock = holder["w"].transport.get_extra_info("socket")
    sock.setsockopt(socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0))
    holder["w"].transport.abort()
    await asyncio.sleep(0.2)  # let the client loop run connection_lost

    # The pool would now hand this connection out: is_connected is still True.
    conn = Connection.__new__(Connection)
    conn._writer = writer
    await conn._send_packed_command([b"PING\r\n"])
    # -> TypeError: 'NoneType' object is not callable   (expected: ConnectionError)


asyncio.run(main())
```

**Expected behavior:** writing on a connection whose transport is closed/closing should raise `redis.exceptions.ConnectionError` so the existing disconnect + retry machinery handles it (as it does for `OSError`).

**Proposed fix:**

Guard the write in `send_packed_command` (covering both the `socket_timeout` path through `_send_packed_command` and the inline `writelines` branch): if `self._writer` is `None` or `self._writer.transport.is_closing()`, raise `ConnectionError` instead of writing into the dead transport. This fails fast into the existing recovery path on every Python version and fixes the `TypeError` leak on 3.12. Note the raised error should be `redis.exceptions.ConnectionError` itself, since `RedisCluster.ERRORS_ALLOW_RETRY` matching uses exact types.

I will open a PR with this fix and a regression test shortly.

Related: #4283 (a different stale-stream issue after disconnect, same subsystem).

abersheeran:
### Pull Request check-list

### Description of change

Fixes #4287, #3546.

When the server (or an intermediary) closes a pooled connection while it is idle, `AbstractConnection.is_connected` still reports the connection as healthy (it only checks that `_reader`/`_writer` are non-`None`), so the dead connection is reused. On CPython 3.12 the subsequent write then raises `TypeError: 'NoneType' object is not callable` instead of `ConnectionError`: 3.12's `_SelectorSocketTransport.writelines` is missing the `_conn_lost` guard that `write` has, and `_call_connection_lost` has already cleared the transport's `_write_ready` callback. The `TypeError` is not translated to `ConnectionError`, so it bypasses `Retry` and `RedisCluster.ERRORS_ALLOW_RETRY` and surfaces to application code. CPython fixed `writelines` in 3.13+, but 3.12 is security-only, so the fix will not be backported.

This change guards the write in `send_packed_command`: if the writer is missing or its transport is closing, raise `ConnectionError` (the exact class, since `ERRORS_ALLOW_RETRY` matching uses exact types) before touching the transport. The guard sits inside the existing `try` block, so the `except BaseException` clause disconnects the dead connection before re-raising, and the existing retry machinery takes over — on every Python version, covering both the `socket_timeout` path and the inline `writelines` path.

Regression tests (mock-based, no live server needed):

- a closed transport makes `send_packed_command` raise exactly `ConnectionError`, without writing, and the connection is disconnected;
- a healthy transport still writes through `writelines`/`drain` on both write paths (`socket_timeout` set and unset).

Verified against a real asyncio transport as well (peer sends a genuine RST via `SO_LINGER(1, 0)`): before this change the write raises the `TypeError` above on CPython 3.12; with it, `ConnectionError`.

---

> [!NOTE]
> **Medium Risk**
> Touches the hot path for every async Redis command write and error classification; behavior change is intentional (TypeError → ConnectionError on dead sockets) but could affect callers that relied on the old exception type.
> 
> **Overview**
> Fixes **#4287** by hardening async command writes when a pooled connection was closed by the server but still looks connected.
> 
> **`_send_packed_command`** now refuses writes if the stream writer is missing or `transport.is_closing()` is true, raising **`ConnectionError`** (exact type, so cluster/retry allowlists still match). It centralizes `writelines`/`drain` for both the `socket_timeout` and non-timeout paths. On older CPython builds where a dead transport can surface **`TypeError`** or **`AttributeError`** from `writelines`, those are mapped to **`ConnectionError`** when the transport is closing; unrelated errors are re-raised. Existing **`send_packed_command`** error handling still disconnects on failure.
> 
> Adds mock-based tests for closed transport, successful writes, translated transport errors, and preserved **`TypeError`** on an open transport.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0cd03beddd9f291a3db4cd0ff61f9d8a9124f86e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1177,6 +1177,10 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
             await self.check_health()
 
         try:
+            # Fail fast because writing to a closed transport can raise TypeError
+            # on CPython 3.12 (#4287).
+            if self._writer is None or self._writer.transport.is_closing():
+                raise ConnectionError("Connection closed by the server before write")
             if isinstance(command, str):
                 command = command.encode()
             if isinstance(command, bytes):
--- a/redis/asyncio/connection.py
+++ b/redis/asyncio/connection.py
@@ -1165,8 +1165,21 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
             )
 
     async def _send_packed_command(self, command: Iterable[bytes]) -> None:
-        self._writer.writelines(command)
-        await self._writer.drain()
+        writer = self._writer
+        if writer is None or writer.transport.is_closing():
+            raise ConnectionError("Connection closed by the server before write")
+        try:
+            writer.writelines(command)
+            await writer.drain()
+        except (TypeError, AttributeError) as e:
+            # CPython gh-136234 adds the missing connection-lost check in 3.13.10+
+            # and 3.14.1+. Python 3.12, 3.13.0-3.13.9, and 3.14.0 can instead
+            # leak TypeError or AttributeError from the transport (#4287).
+            if writer.transport.is_closing():
+                raise ConnectionError(
+                    "Connection closed by the server while writing"
+                ) from e
+            raise
 
     async def send_packed_command(
         self, command: Union[bytes, str, Iterable[bytes]], check_health: bool = True
@@ -1177,10 +1190,6 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
             await self.check_health()
 
         try:
-            # Fail fast because writing to a closed transport can raise TypeError
-            # on CPython 3.12 (#4287).
-            if self._writer is None or self._writer.transport.is_closing():
-                raise ConnectionError("Connection closed by the server before write")
             if isinstance(command, str):
                 command = command.encode()
             if isinstance(command, bytes):
@@ -1190,8 +1199,7 @@ class AbstractConnection(AsyncMaintNotificationsAbstractConnection):
                     self._send_packed_command(command), self.socket_timeout
                 )
             else:
-                self._writer.writelines(command)
-                await self._writer.drain()
+                await self._send_packed_command(command)
         except asyncio.TimeoutError:
             await self.disconnect(nowait=True)
             raise TimeoutError("Timeout writing to socket") from None
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -251,6 +251,41 @@ def test_get_resolved_ip_uses_async_writer_peer_before_dns():
     getaddrinfo.assert_not_called()
 
 
+async def test_send_packed_command_rejects_closed_transport():
+    conn = Connection(health_check_interval=0)
+    writer = mock.Mock()
+    writer.transport.is_closing.return_value = True
+    writer.drain = mock.AsyncMock()
+    conn._reader = mock.Mock()
+    conn._writer = writer
+
+    with pytest.raises(ConnectionError, match="closed") as exc_info:
+        await conn.send_packed_command(b"PING", check_health=False)
+
+    assert type(exc_info.value) is ConnectionError
+    writer.writelines.assert_not_called()
+    writer.drain.assert_not_awaited()
+    writer.close.assert_called_once()
+    assert not conn.is_connected
+
+
+@pytest.mark.parametrize("socket_timeout", [None, 1])
+async def test_send_packed_command_writes_to_open_transport(socket_timeout):
+    conn = Connection(socket_timeout=socket_timeout, health_check_interval=0)
+    writer = mock.Mock()
+    writer.transport.is_closing.return_value = False
+    writer.drain = mock.AsyncMock()
+    conn._reader = mock.Mock()
+    conn._writer = writer
+
+    await conn.send_packed_command(b"PING", check_health=False)
+
+    writer.transport.is_closing.assert_called_once_with()
+    writer.writelines.assert_called_once_with([b"PING"])
+    writer.drain.assert_awaited_once_with()
+    await conn.disconnect(nowait=True)
+
+
 @pytest.mark.fixed_client
 @pytest.mark.parametrize(
     "client_kwargs",
--- a/tests/test_asyncio/test_connection.py
+++ b/tests/test_asyncio/test_connection.py
@@ -286,6 +286,57 @@ async def test_send_packed_command_writes_to_open_transport(socket_timeout):
     await conn.disconnect(nowait=True)
 
 
+@pytest.mark.parametrize(
+    ("error_type", "message"),
+    [
+        (TypeError, "'NoneType' object is not callable"),
+        (AttributeError, "'NoneType' object has no attribute '_add_writer'"),
+    ],
+)
+async def test_send_packed_command_translates_closed_transport_error(
+    error_type, message
+):
+    conn = Connection(health_check_interval=0)
+    writer = mock.Mock()
+    writer.transport.is_closing.side_effect = [False, True]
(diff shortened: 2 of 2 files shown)


petyaslavova (changes requested):
Thanks for the thorough investigation - the diagnosis is right, and raising `ConnectionError` itself rather than a subclass is the correct call given that `ERRORS_ALLOW_RETRY` matches on exact types.

Two things before we merge. First, the guard re-reads `self._writer` afterwards, and `asyncio.wait_for` suspends between the check and `writelines()` on the `socket_timeout` path, so a concurrent `disconnect()` still surfaces `AttributeError: 'NoneType' object has no attribute 'writelines'` and still bypasses retry. Please bind the writer to a local once, guard the local, and write through it - that also covers the variant reported in #3546 and in open PR #4217.

Second, the 3.12-only framing is not accurate: the `_SelectorSocketTransport.writelines` override was added in 3.12 and still calls `_write_ready()` unguarded in 3.13 and 3.14. I measured `TypeError` on 3.12 and 3.14 and `AttributeError: 'NoneType' object has no attribute '_add_writer'` on 3.13, with 3.11 correctly raising `ConnectionError`. Your guard is version-agnostic, so only the code comment needs updating - but it makes the case for handling `AttributeError` too.

fallenmi (approved):
Reviewed exact head `1ea25fc2f50f95d58c9846e05cc917e3f7faea85` against exact base `afe6523ec5e07b49fe15e1748c049cc2ab2644a6` and current `main` `c78ccfcbd05d129993bb7bc61b1ab3c3993e0b23`.

The pre-write guard and single local-writer path address both failure classes without widening exception translation:

- An exact-base harness with a closing transport returned normally and called `writelines`; the exact head raises the concrete `redis.exceptions.ConnectionError` before writing.
- A 10-case exact-head matrix covered closed-before-write, transport-side `TypeError`/`AttributeError` while closing with the cause preserved, a genuine `TypeError` on an open transport with identity preserved, and `_writer` cleared after the health check, with and without `socket_timeout`.
- The six focused upstream cases pass on the exact head and after a clean merge into current `main`. Ruff check/format, compileall, and diff check passed; live CI is 368 successful with 0 failing or pending.

I also ran the full test file: 61 passed and 1 skipped; the remaining 11 failures and 8 setup errors all require the unavailable local Redis endpoint, so I did not count that as a green full-suite result.

Approved.

Review drafted and validated with OpenAI Codex.

petyaslavova (commented):
Thanks for the quick turnaround - both points are addressed. `_send_packed_command` is now the single write path and the writer is bound to a local inside the `wait_for` task with no suspension before `writelines()`, so the concurrent-`disconnect()` variant from #3546 is covered too. Gating the `TypeError`/`AttributeError` translation on `transport.is_closing()` keeps genuine encoding errors intact, and raising exact `redis.exceptions.ConnectionError` is what `ERRORS_ALLOW_RETRY` needs, since the cluster paths match on `type(e)`.

Merging this after a green CI. Two things I am tracking separately, no action needed here: `is_connected` still reports a closed-transport connection as healthy, so the pool keeps handing one out and the first use costs a failed command plus a retry, and the read path has no equivalent guard.

petyaslavova (approved):

Issue #2754, opened 2023-05-08
redis.client.PubSub().parse_response() calling 


stacywsmith:
**Version**: 4.5.5
**Platform**: CPython 3.8.16 on Ubuntu 

**Description**: 
[`redis.client.PubSub().parse_response()`](https://github.com/redis/redis-py/blob/v4.5.5/redis/client.py#L1540) is calling [`redis.connection.PythonParser().read_response()`](https://github.com/redis/redis-py/blob/v4.5.5/redis/connection.py#L344) with a non-existent `disconnect_on_error` keyword argument.

```
08:24:16 .tox/_docker/py/lib/python3.8/site-packages/redis/client.py:1690: in get_message
08:24:16     response = self.parse_response(block=(timeout is None), timeout=timeout)
08:24:16 .tox/_docker/py/lib/python3.8/site-packages/redis/client.py:1542: in parse_response
08:24:16     response = self._execute(conn, try_read)
08:24:16 .tox/_docker/py/lib/python3.8/site-packages/redis/client.py:1518: in _execute
08:24:16     return conn.retry.call_with_retry(
08:24:16 .tox/_docker/py/lib/python3.8/site-packages/redis/retry.py:46: in call_with_retry
08:24:16     return do()
08:24:16 .tox/_docker/py/lib/python3.8/site-packages/redis/client.py:1519: in <lambda>
08:24:16     lambda: command(*args, **kwargs),
08:24:16 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
08:24:16 
08:24:16     def try_read():
08:24:16         if not block:
08:24:16             if not conn.can_read(timeout=timeout):
08:24:16                 return None
08:24:16         else:
08:24:16             conn.connect()
08:24:16 >       return conn.read_response(disconnect_on_error=False)
08:24:16 E       TypeError: read_response() got an unexpected keyword argument 'disconnect_on_error'
08:24:16 
08:24:16 .tox/_docker/py/lib/python3.8/site-packages/redis/client.py:1540: TypeError
```

dvora-h:
@kristjanvalur Can you take a look here?

kristjanvalur:
self.connection is supposed to be an redis.Connection object.

kristjanvalur:
[...]

PR #4292, opened 2026-08-27
fix(sentinel): disconnect on read errors by default

dylanpulver:
### Description of change

`SentinelManagedConnection.read_response()` defaults `disconnect_on_error` to `False` (`redis/sentinel.py:73`), so a connection whose reply was never read goes back to the pool and the next command gets the previous one's response. Against a real sentinel, after an interrupt at the socket read:

```
master.ping()       -> False
master.get("probe") -> b'PONG'
```

Every other `read_response` defaults to `True`, including the async twin at `redis/asyncio/sentinel.py:76` and base `Connection.read_response`, whose `except BaseException` branch exists for this case (#1128, #2499). Both were written in one commit, `35b7e09a` for #2754, which took the value from the `PubSub.parse_response` call site, where `False` is passed explicitly.

`PubSub` passes the value explicitly, so it is unaffected. `ReadOnlyError` too, since the parser returns error replies rather than raising them and they surface after the guarded block.

Why no existing test caught it: they patch out `Connection.read_response`, the only code that reads the flag, so they check the forwarded value and never its effect. Four assertions pinned `False` and now pin `True`; the new tests assert socket state and the next reply.

Ran `tests/` standalone plus asyncio against a local sentinel setup, with a failure set identical to master's (the rest need redis-stack or TLS). `invoke linters` clean on the pinned ruff 0.9.6. Reverting the one-line change fails three of the four new tests; the fourth guards the explicit `False` opt-out and passes either way.

Prepared with Claude (Opus 5).

### Pull Request check-list

---

> [!NOTE]
> **Medium Risk**
> Changes default connection lifecycle for all Sentinel-managed reads except explicit `disconnect_on_error=False`; low blast radius but affects pooled connection reuse after read failures.
> 
> **Overview**
> **Fixes stale replies on Sentinel master connections** when a read is interrupted before the full response is consumed (e.g. `KeyboardInterrupt` at the socket read).
> 
> `SentinelManagedConnection.read_response()` now defaults **`disconnect_on_error` to `True`**, matching base `Connection`, the async Sentinel connection, and the behavior intended in #1128. With the old default of `False`, the socket stayed open and pooled connections could return the previous command’s reply on the next call (e.g. `ping()` → wrong result, `get()` → `PONG`). Callers that need to keep the connection open (e.g. PubSub with an explicit `disconnect_on_error=False`) are unchanged.
> 
> Tests are updated to expect the new default and add sync/async coverage that checks **socket teardown and the next reply**, including an interrupted transactional pipeline recycled through `SentinelConnectionPool`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed39ad8bb548aff89fa2a9fb05b3a272d7247c8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/sentinel.py
+++ b/redis/sentinel.py
@@ -70,7 +70,7 @@ class SentinelManagedConnection(Connection):
         disable_decoding=False,
         *,
         timeout: Union[float, object] = SENTINEL,
-        disconnect_on_error: Optional[bool] = False,
+        disconnect_on_error: Optional[bool] = True,
         push_request: Optional[bool] = False,
     ):
         try:
--- a/tests/test_asyncio/test_sentinel_managed_connection.py
+++ b/tests/test_asyncio/test_sentinel_managed_connection.py
@@ -36,3 +36,29 @@ async def test_connect_retry_on_timeout_error(connect_args):
     assert conn._connect.call_count == 3
     assert connection_pool.get_master_address.call_count == 3
     await conn.disconnect()
+
+
+@pytest.mark.fixed_client
+async def test_read_response_disconnects_on_base_exception(connect_args):
+    """
+    Mirror of the sync test: a BaseException at the socket read leaves the reply
+    unread, so the connection must be closed rather than reused. See #1128.
+    """
+    connection_pool = mock.AsyncMock()
+    connection_pool.get_master_address = mock.AsyncMock(
+        return_value=(connect_args["host"], connect_args["port"])
+    )
+    connection_pool.is_master = True
+    connection_pool.check_connection = False
+    conn = SentinelManagedConnection(connection_pool=connection_pool)
+    await conn.connect()
+    try:
+        await conn.send_command("PING")
+        with mock.patch.object(
+            conn, "_read_response_from_parser", side_effect=KeyboardInterrupt
+        ):
+            with pytest.raises(KeyboardInterrupt):
+                await conn.read_response()
+        assert conn._reader is None
+    finally:
+        await conn.disconnect()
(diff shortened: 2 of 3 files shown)


petyaslavova (changes requested):
Hey @dylanpulver, thank you for your contribution!

Confirmed, and the impact is wider than the description argues. Every `read_response` caller that omits `disconnect_on_error` is a strict command/response pair that needs `True` - `Redis.parse_response`, the `on_connect` handshake, `_send_ping`, `re_auth`. The three callers that genuinely want `False` (PubSub, the pending-push drain, `_process_pending_invalidations`) all pass it explicitly, so the current default has no beneficiary. It was never a design choice: it came from the #2754 hotfix, where the sync half was amended in last (`4598d805d`) and the author noted they were unfamiliar with Sentinel - which is why the async twin got `True` in the same commit.

Two things before merge:

1. Please drop `test_default_disconnect_on_error_matches_base_connection`. Asserting the default through `inspect.signature` restates the source and would still pass if the flag were never forwarded; the four assertions you updated already pin it.
2. Please add a pipeline regression test. `_execute_transaction` reads N+2 replies in a loop, and a `BaseException` at reply k strands the rest: `Retry` does not treat it as retryable, `execute()` catches only `Exception`, `reset()` disconnects only when watching, and `release()` only when `should_reconnect()`. The connection goes back to `_available_connections` desynced, which is a worse failure than the single-command case you reproduced.

PR #4301, opened 2026-09-01
Stabilize MultiDB (Active-Active) client and tests

petyaslavova:
## Change summary

This pull request stabilizes the MultiDB (Active-Active) client and its scenario tests. A temporarily unreachable cluster used to surface as a plain `RedisClusterException` from topology refresh, which the MultiDB command retry did not recognize, so commands failed instead of triggering a failover. `NodesManager.initialize` (sync and async) now raises a new `RedisClusterUnreachableError` - a subclass of `RedisClusterException`, so existing handlers keep working - and `MultiDBClient` registers it as retryable, letting the failover path take over.

`LagAwareHealthCheck` also probed the wrong node in some Redis Enterprise setups: it matched the bdb by the first discovered node, whose advertised address is often neither the endpoint DNS name nor an address the bdb lists. It now matches against all configured database hosts, preferring a DNS-name match over an address match, and health check probes across nodes run in parallel to keep a probe round short. On the test side, the `r_multi_db` fixture now closes the client on teardown (previously its health check loop thread kept running for the rest of the session), and the AA scenario tests bound their failover waits with an explicit deadline and use a tight health check interval so a probe round reliably lands inside the injected transient outage.

## Test coverage

Unit tests were extended on both stacks: `tests/test_multidb/` and `tests/test_asyncio/test_multidb/` cover the new retryable-error registration in `MultiDBClient` and command executor failover on `RedisClusterUnreachableError`, and the async healthcheck tests cover bdb matching by DNS name vs address and the parallel probe rounds. The AA scenario tests (`tests/test_scenario/test_active_active.py`, sync and async) were updated as described above; they require Redis Enterprise infrastructure and are excluded from default runs.

---

> [!NOTE]
> **Medium Risk**
> Changes cluster bootstrap exceptions, MultiDB failover/retry behavior, and health-check probing paths that can mark databases unhealthy or trigger failover; coverage is broad but these are core availability paths.
> 
> **Overview**
> Introduces **`RedisClusterUnreachableError`** when sync/async cluster startup cannot reach any node (connectivity-only; **auth** and **config** failures still raise plain **`RedisClusterException`**). **MultiDB** sync/async clients register that type for **command retry** so cluster outages can **fail over** without treating every **`RedisClusterException`** as retryable.
> 
> **Async MultiDB health checks** are tightened: probe clients keep **TLS**, cluster probes use **all startup nodes**, **`initialize()`** runs before cluster PINGs, nodes are pinged **concurrently**, and connectivity failures **reset** the probe client. **`LagAwareHealthCheck`** matches Redis Enterprise bdbs by **configured hosts** (DNS over address), and REST calls use **per-request absolute URLs** so concurrent database checks do not clobber a shared HTTP **base_url**.
> 
> **Active-active scenario tests** are re-enabled with faster health-check intervals, explicit failover deadlines, fixture teardown that **closes** the MultiDB client, and clearer LagAware credential handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4eae24c034256ac53fd1c306e96fbe907cf746fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -2386,7 +2386,16 @@ class NodesManager:
                     break
 
             if not startup_nodes_reachable:
-                raise RedisClusterUnreachableError(
+                # The unreachable subtype is reserved for connectivity failures:
+                # MultiDB registers it as retryable, so a deterministic
+                # server/configuration error (e.g. cluster mode disabled) must
+                # keep surfacing as a plain RedisClusterException.
+                if isinstance(exception, (ConnectionError, TimeoutError, OSError)):
+                    raise RedisClusterUnreachableError(
+                        f"Redis Cluster cannot be connected. Please provide at least "
+                        f"one reachable node: {str(exception)}"
+                    ) from exception
+                raise RedisClusterException(
                     f"Redis Cluster cannot be connected. Please provide at least "
                     f"one reachable node: {str(exception)}"
                 ) from exception
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -116,6 +116,7 @@ from redis.exceptions import (
     MaxConnectionsError,
     MovedError,
     RedisClusterException,
+    RedisClusterUnreachableError,
     RedisError,
     ResponseError,
     SlotNotCoveredError,
@@ -2385,7 +2386,7 @@ class NodesManager:
                     break
 
             if not startup_nodes_reachable:
-                raise RedisClusterException(
+                raise RedisClusterUnreachableError(
                     f"Redis Cluster cannot be connected. Please provide at least "
                     f"one reachable node: {str(exception)}"
                 ) from exception
--- a/redis/asyncio/cluster.py
+++ b/redis/asyncio/cluster.py
@@ -105,6 +105,8 @@ from redis.event import (
 )
 from redis.exceptions import (
     AskError,
+    AuthenticationError,
+    AuthorizationError,
     BusyLoadingError,
     ClusterDownError,
     ClusterError,
@@ -2388,9 +2390,16 @@ class NodesManager:
             if not startup_nodes_reachable:
                 # The unreachable subtype is reserved for connectivity failures:
                 # MultiDB registers it as retryable, so a deterministic
-                # server/configuration error (e.g. cluster mode disabled) must
-                # keep surfacing as a plain RedisClusterException.
-                if isinstance(exception, (ConnectionError, TimeoutError, OSError)):
+                # server/configuration error (e.g. cluster mode disabled or
+                # invalid credentials - AuthenticationError and
+                # AuthorizationError subclass ConnectionError but cannot be
+                # repaired by a failover) must keep surfacing as a plain
+                # RedisClusterException.
+                if isinstance(
+                    exception, (ConnectionError, TimeoutError, OSError)
+                ) and not isinstance(
+                    exception, (AuthenticationError, AuthorizationError)
+                ):
                     raise RedisClusterUnreachableError(
                         f"Redis Cluster cannot be connected. Please provide at least "
                         f"one reachable node: {str(exception)}"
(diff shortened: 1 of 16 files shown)


tests/test_scenario/test_active_active.py
```diff
@@ -29,22 +88,14 @@ def trigger_network_failure_action(
...
-        )
+    status_result = fault_injector_client.get_operation_result(result["action_id"])
 
     if event:
         event.set()
 
     logger.info(f"Action completed. Status: {status_result['status']}")
 
 
-@pytest.mark.skip(reason="Temporarily disabled")
```
vladvildanov, line 47: How about async tests? They're still excluded

redis/asyncio/multidb/healthcheck.py
```diff
@@ -471,38 +538,37 @@ async def check_health(self, database, hc_client: AsyncRedisClientT) -> bool:
...
             raise ValueError("Could not find a matching bdb")
 
         url = (
-            f"/v1/bdbs/{matching_bdb['uid']}/availability"
+            f"{base_url}/v1/bdbs/{matching_bdb['uid']}/availability"
             f"?extend_check=lag&availability_lag_tolerance_ms={self._lag_aware_tolerance}"
         )
-        await self._http_client.get(url, expect_json=False)
+        await self._http_client.get(
+            url, expect_json=False, timeout=self.health_check_timeout
```
vladvildanov: The timeout to be used here is `http_timeout`. `health_check_timeout` is a deadline for overall health check operations across all nodes and it's used as deadline from higher level

tests/test_asyncio/test_scenario/test_active_active.py
```diff
@@ -20,6 +21,65 @@
 
 logger = logging.getLogger(__name__)
 
+# The injected network failure is transient - the fault injector restores the link a
+# few seconds after the action is triggered. A database is only taken out of service by
+# a health check probe that runs while the link is down, and on the default interval a
+# probe round is short next to the pause that follows it, so most rounds land after the
+# link is already back and no failover is ever initiated. Probing back to back keeps a
+# round inside the outage.
+FAILOVER_HEALTH_CHECK_INTERVAL = 0.1
```
vladvildanov, line 30: Looks a bit aggressive, is it intentional?

vladvildanov (approved):

tests/test_scenario/test_active_active.py
petyaslavova: fixed

redis/asyncio/multidb/healthcheck.py
petyaslavova: fixed.

tests/test_asyncio/test_scenario/test_active_active.py
petyaslavova: Yes, it's intentional. The injected network failure is quite fast - the link is restored a few seconds after the action, and a failover only happens if a probe runs while the link is down. With the default interval most probe rounds miss the outage and the test flakes, so the probes run back-to-back.

redis/cluster.py
```diff
@@ -2886,6 +2888,22 @@ def initialize(
...
+                # server/configuration error (e.g. cluster mode disabled or
+                # invalid credentials - AuthenticationError and
+                # AuthorizationError subclass ConnectionError but cannot be
+                # repaired by a failover) must keep surfacing as a plain
+                # RedisClusterException.
+                if isinstance(
+                    exception, (ConnectionError, TimeoutError, OSError)
+                ) and not isinstance(
+                    exception, (AuthenticationError, AuthorizationError)
+                ):
```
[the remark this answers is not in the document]
petyaslavova: In sync with current behaviour of client instances

redis/asyncio/cluster.py
```diff
@@ -2385,6 +2388,22 @@ async def initialize(
...
+                # RedisClusterException.
+                if isinstance(
+                    exception, (ConnectionError, TimeoutError, OSError)
+                ) and not isinstance(
+                    exception, (AuthenticationError, AuthorizationError)
+                ):
+                    raise RedisClusterUnreachableError(
+                        f"Redis Cluster cannot be connected. Please provide at least "
+                        f"one reachable node: {str(exception)}"
+                    ) from exception
```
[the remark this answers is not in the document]
petyaslavova: Will not change

PR #4310, opened 2026-09-08
fix: raise when only one FILTER_BY_VALUE bound is given

youdie006:
### Description of change

`_append_filer_by_value` (`redis/commands/timeseries/commands.py:2094`) emitted the `FILTER_BY_VALUE` clause only when **both** bounds were set, and silently dropped it otherwise. A caller that supplies one bound gets an unfiltered range back rather than an error.

Measured against `master` on `redis-stack-server:latest`:

```python
r.ts().add("probe", 100, 1.0); r.ts().add("probe", 200, 5.0); r.ts().add("probe", 300, 9.0)

r.ts().range("probe", 0, 500, filter_by_min_value=5)
# -> [(100, 1.0), (200, 5.0), (300, 9.0)]      the filter was never sent

r.ts().range("probe", 0, 500, filter_by_min_value=5, filter_by_max_value=5)
# -> [(200, 5.0)]
```

Three private builders call it — `__range_params` (`:764`), `__n_range_params` (`:1112`) and `__mrange_params` (`:1428`) — so this reaches `range`, `revrange`, `nrange`, `nrevrange`, `mrange` and `mrevrange`.

The docstrings already state the requirement (`commands.py:967-972`): *"Filter result by minimum value (must mention also `filter_by_max_value`)"*. `FILTER_BY_VALUE` takes both bounds or neither, so a one-sided filter cannot be expressed at all — the choice is between silently ignoring it and rejecting it.

**Two siblings in this same class already reject it:**

| helper | on a partial group |
|---|---|
| `_append_block` (`:1985`) | `raise DataError("block_min_count requires block_milliseconds to be set; the BLOCK group is all-or-nothing.")` |
| `_append_insertion_filters` (`:2136`) | `if (a is None) != (b is None): raise ...` |
| `_append_filer_by_value` (`:2094`) | dropped the clause |

`_append_block`'s guard is recent — added in #4170 (`990c11c`, 2026-07-09) — so this is following a call the project already made in this file rather than proposing a new convention. The patch is the same two-statement shape as `_append_insertion_filters`.

I used `DataError` rather than `ValueError`: this module raises `DataError` in 8 of its 9 `raise` statements, and across `redis/commands/` it is 144 to 17. The one `ValueError` is `_append_insertion_filters` from #3228 (2024).

**This is a behaviour change** for anyone passing a single bound today. They are currently getting unfiltered results back, so the change surfaces a bug in their call rather than removing a working feature — but it is worth calling out explicitly, and I am happy to make it a warning instead if you would rather not raise in a patch release.

### Pull Request check-list

- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? — the docstrings already document the requirement; nothing new to describe
- [ ] Is there an example added to the examples folder (if applicable)? — not applicable

### Testing

`test_range_filter_by_value_requires_both_bounds` added to `tests/test_timeseries.py` and `tests/test_asyncio/test_timeseries.py`, placed next to `test_read_block_min_count_requires_milliseconds`, its direct twin. Each covers both one-sided cases for `range` and `revrange` and asserts that supplying both bounds still filters server-side.

- All three fail on `master` with `Failed: DID NOT RAISE DataError`, and pass here.
- Mutation-checked both directions. Deleting the guard reproduces the three failures. Over-guarding it (`(a is None) != (b is None)` changed to `a is None or b is None`, so it raises whenever a bound is absent) fails 22 tests in `test_timeseries.py`, including `test_mrange` and `test_revrange_empty` — so the "neither bound given" path is pinned as still valid.
- `pytest tests/test_timeseries.py tests/test_asyncio/test_timeseries.py` → **109 passed, 102 skipped** against `redis-stack-server:latest`.
- `ruff check tests redis` → All checks passed. `ruff format --check --diff tests redis` → 272 files already formatted. `vulture redis whitelist.py --min-confidence 80` → clean. (The three commands `tasks.py:37-39` runs.)

Existing coverage stops one short here: all 22 tests that touch these parameters supply both halves.

### AI

Per `CONTRIBUTING.md` "AI-driven contributions": this patch was found and written with Claude Code. I have reviewed every line and can explain it without going back to the tool; every number and line of output above is verbatim from running it against `master` and against this branch.

---

> [!NOTE]
> **Medium Risk**
> This is a breaking behavior change for callers that passed a single filter bound and relied on silent no-op filtering; correct usage (both bounds or neither) is unchanged.
> 
> **Overview**
> **TimeSeries `FILTER_BY_VALUE` now fails fast when only one bound is supplied.** Previously `_append_filer_by_value` omitted the clause if `filter_by_min_value` or `filter_by_max_value` was set alone, so callers could get unfiltered results. The helper now raises `DataError` when exactly one bound is present, matching the all-or-nothing behavior of `_append_block` and `_append_insertion_filters`.
> 
> That validation applies everywhere this helper is used (`range`/`revrange`, `nrange`/`nrevrange`, `mrange`/`mrevrange`). Sync and async tests assert both one-sided cases error on `range`/`revrange` and that both bounds still filter correctly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0d38e6ba3b7172fb2dc63a83c6fc5410c4ac29b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


petyaslavova (commented):

Your trace is right and the report is exactly the kind we want: `_append_filer_by_value` dropped a one-sided `FILTER_BY_VALUE` silently, so `range`, `revrange`, `nrange`, `nrevrange`, `mrange` and `mrevrange` returned unfiltered samples for a call that asked to be filtered. `FILTER_BY_VALUE` takes both bounds or neither, so a lone bound was never a valid request in the first place. Current change is a good input validation hardening: rejecting an invalid argument combination client-side, in line with `_append_block` and `_append_insertion_filters` in the same class.

I think the raise would be better choise here, than just logging a warning, and `DataError` is the right type for this module. Placement in the shared helper is also right, since it covers all six entry points at once. The sync and async tests look good.

I will kick off CI on the branch. Assuming it comes back green, this is ready to merge as it stands.

petyaslavova (approved):

PR #3023, opened 2023-10-24
Adding lock_name to LockError

TheBlusky:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [X] Do tests and lints pass with this change?
- [ ] Do the CI tests pass with this change (enable it first in your forked repo and wait for the github action build to finish)?
- [X] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Is there an example added to the examples folder (if applicable)?
- [X] Was the change added to CHANGES file?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._

### Description of change

This PR adds an optional `lock_name` attribute to `LockError`.
The main goal behind this PR is to be able to retrieve which Lock failed when excepting a `LockError` exception.

PR #4317, opened 2026-09-10
fix(asyncio): set lock_name on the errors the async lock raises

dylanpulver:
### Description of change

`LockError.lock_name` is `None` for 8 of the 9 errors the asyncio lock raises.

`redis/lock.py` passes `lock_name=self.name` at all 9 raise sites. `redis/asyncio/lock.py` passes it at 1, so `except LockError as e: e.lock_name` gives the name on the sync lock and `None` on the async one. This adds the argument at the 8 remaining sites, matching the sync twin.

It also moves the assertion in `test_lock_error_gives_correct_lock_name` out of the `pytest.raises` block. It sat after the statement that raises, so it never ran: removing `lock_name` from `redis/lock.py` left the test green.

Verified against Redis 8.6.2. 106 pass across `tests/test_lock.py`, `tests/test_asyncio/test_lock.py` and `tests/test_exceptions.py`. Reverting `redis/asyncio/lock.py` fails the two new tests with `assert None == 'foo'`, and removing `lock_name` from the sync lock now fails the moved assertion.

### Pull Request check-list

https://claude.ai/code/session_011M5uTyCU4WcNTsPvGrErDo

---

> [!NOTE]
> **Low Risk**
> Behavior-only improvement to exception metadata with no change to lock acquisition or release logic; low regression risk beyond error object fields.
> 
> **Overview**
> **Async lock errors now include the Redis key name**, matching the synchronous `Lock` implementation.
> 
> `LockError` and `LockNotOwnedError` raised from `redis/asyncio/lock.py` (context manager acquire failure, release/extend/reacquire validation, and Lua script failures) now pass **`lock_name=self.name`**, so handlers can identify which lock failed instead of getting `None`.
> 
> Tests add asyncio coverage for `lock_name` on both exception types. The sync **`test_lock_error_gives_correct_lock_name`** assertion is moved **outside** the `pytest.raises` block so it actually runs when `__enter__` raises.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4721f0c3e682e2c4d8ab85fe78c56e79ceb05ba3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


petyaslavova (commented):
Hey @dylanpulver, thank you for your contribution!

Confirmed. #3023 added `lock_name` to `LockError` and wired it into all nine raise sites in `redis/lock.py`, but the async twin was never mirrored, so `LockError.lock_name` came back `None` for eight of the nine async cases. Sync/async parity is a repository requirement, and this restores it site-for-site without touching acquire/release logic. Good catch on `test_lock_error_gives_correct_lock_name` as well: the assertion sat inside the `pytest.raises` block after the statement that raises on `__enter__`, so it never ran.

This looks good to merge once the integration workflow has run. One optional follow-up, not blocking: consider mirroring the new `test_lock_not_owned_error_gives_correct_lock_name` into `tests/test_lock.py` so the sync side is guarded too.

petyaslavova (approved):

Issue #4326, opened 2026-09-13
CLIENT LIST on a connection with many thousands of clients causes a large, avoidable memory spike


fcostaoliveira:
### Description

Calling `CLIENT LIST` (or `client_list()`) against a Redis/Redis-compatible server with a large number of connected clients causes resident memory to spike proportionally to the number of connected clients, rather than the number of lines actually needed by the caller. This happens because the entire reply is read and parsed into memory before any of it is returned to application code — there's no way to consume the result incrementally.

### Observed behavior

In testing, resident memory during a single `client_list()` call grew roughly linearly with the number of connected clients — on the order of a few KB of process memory per connected client, but reaching multi-GB territory once the connection count climbs into the tens of thousands (which is realistic for a busy, long-lived server). The growth was reproducible and consistent across repeated runs at the same connection count.

### Steps to reproduce

1. Start a Redis server and open a large number of client connections against it (e.g. 5,000–20,000 plain `CLIENT` connections held open, no traffic needed).
2. From a separate connection, call `client_list()`.
3. Observe process RSS before/after the call — it grows well beyond what's needed to hold the actual returned data structure once you account for the raw reply size.

### Why this happens (as far as I could tell)

- With the default `hiredis`-backed parser, the C reader only begins parsing a bulk-string reply once the *entire* reply is already present in its internal buffer — there's no incremental/streaming path.
- The pure-Python parser doesn't appear to expose a streaming/line-by-line read path for this either, so the same buffer-then-parse pattern applies regardless of parser backend.

### Suggested fix

An opt-in streaming variant (e.g. `client_list_iter()`/similar) that reads and yields one client record at a time as the reply arrives off the socket, bounding peak memory to roughly the socket read chunk size instead of the full client count. Understand this may only be feasible on the pure-Python parser path without a larger change; happy to help test/validate if there's interest.

### Environment

- `redis-py` (current release)
- Reproduced with both the default (hiredis) and pure-Python parser configurations

PR #4328, opened 2026-09-15
Add client_list_iter(): streaming CLIENT LIST prototype

fcostaoliveira:
## Summary

Production Prometheus data across 3,369 Redis Cloud clusters showed a service's RSS climbing linearly with connection count (~104.8 MB + 2.59 KiB/connection). A memray trace pinned the allocation site to `_execute_command → read_response → _read_from_socket`: redis-py's `CLIENT LIST` reply is read as one giant buffered blob and parsed into a full list of dicts before the caller sees anything, so memory use is proportional to the number of connected clients on the server, not to what the caller actually needs.

This PR adds `client_list_iter()`, a prototype that streams and parses the `CLIENT LIST` reply incrementally as bytes arrive off the socket, yielding one dict per client instead of buffering the whole reply:

```python
for client in r.client_list_iter():
    ...
```

Peak memory is bounded by the socket read chunk size, independent of how many clients are connected.

- Only supported on connections using the pure-Python RESP2/RESP3 parser (raises `NotImplementedError` under `hiredis`, where `client_list()` should be used instead).
- Not supported from within a `Pipeline` or on a cluster client (`RedisCluster`) — both raise `NotImplementedError` immediately.
- Prefer closing the returned iterator deterministically with a `with` block: `with r.client_list_iter() as it: ...`.
- This is a **prototype**: it intentionally trades some of `client_list()`'s conveniences (automatic retry, cluster/pipeline support) for the streaming behavior, to keep the initial implementation reviewable. `client_list()` remains fully supported and unaffected for callers that don't need this.

While building this, an unrelated pre-existing bug was also found and fixed: `CLIENT LIST ID <id1> <id2>` sent multiple ids joined into a single space-separated argument, which redis-server rejects outright ("Invalid client ID") for anything but one id — `_client_list_args()` now sends each id as its own argument, fixing multi-id filtering for both `client_list()` and the new `client_list_iter()`.

## Test coverage

- [x] New unit test suite (`tests/test_streaming_client_list.py`) covering the RESP2/RESP3 streaming parser, chunk-boundary correctness (parametrized across chunk sizes and client counts), memory bounds (via `tracemalloc`), the `MAX_UNTERMINATED_LINE_SIZE` safety valve and its edge cases, generator abandonment/cleanup semantics, concurrency/locking behavior in `Redis.close()` and `client_list_iter()` (including structural tests that verify specific code runs under the intended lock, not just timing-based races), connection-pool safety (single_connection_client and pooled modes, double-release, `BlockingConnectionPool` compatibility), and API-surface consistency with `client_list()`.
- [x] Integration tests added to `tests/test_commands.py`, run against a live redis-server (RESP2 and RESP3), including a corrected multi-id `client_list_iter()` test.
- [x] `ruff check` and `ruff format --check` pass on all touched files.
- [x] Verified live against redis-server 8.6.0 under both RESP2 and RESP3 (RESP3 replies to `CLIENT LIST` with a verbatim string, handled explicitly).
- [x] A prior version of this PR had a real algorithmic-complexity issue (the internal accumulator was effectively quadratic-time for a pathological, slowly-delivered unterminated record) — this has been fixed and is covered by a dedicated timing-based regression test.

---

> [!NOTE]
> **Medium Risk**
> Touches connection lifecycle, global locking in `close()`, and low-level RESP parsing; mistakes could corrupt pooled connections or stall unrelated clients, though behavior is heavily tested.
> 
> **Overview**
> Adds **`client_list_iter()`**, a prototype that streams `CLIENT LIST` replies line-by-line so peak memory stays near the socket read size instead of holding the full reply and parsed list. The pure-Python parser path gains **`SocketBuffer.read_bulk_lines()`** and **`Connection.read_response_lines_streaming()`** (RESP2 bulk strings, RESP3 verbatim strings, push/invalidation frames); hiredis, pipelines, cluster clients, and async raise **`NotImplementedError`**.
> 
> **`client_list()`** parsing is aligned with streaming: shared **`parse_client_list_line()`**, newline-only splitting (not `splitlines()`), and refactored **`_client_list_args()`** that fixes multi-ID wire format (each ID as its own argument), rejects combining `TYPE` and `ID`, and encodes `bytes`/`float` IDs correctly.
> 
> **`_ClientListIter`** owns checkout/release, supports `with`/explicit `close()`, and guards concurrent close vs `next()`. **`Redis.close()`** is extended to track iterators via a process-wide **`RLock`**, avoid releasing or pool-disconnecting busy connections, and disconnect abandoned streams safely—without holding the lock across blocking pool I/O.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dab34199967b6b73c3ce3578a168bc8248598be9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vladvildanov:
@fcostaoliveira Duplicates #4327

fcostaoliveira:
Thanks for the quick look, @vladvildanov! I'm the reporter of #4326, so let me clarify the distinction, since the two PRs share a method name but solve different parts of the problem described there.

#4327's `client_list_iter()` still reads the entire `CLIENT LIST` reply off the socket in one piece first (its own docstring says so directly: *"The full CLIENT LIST reply is still read off the socket in one piece before this can start yielding... this only reduces the memory used to hold the parsed records, not the raw reply."*). That's a real, useful improvement to the parsing/materialization step, but it doesn't touch the allocation site #4326 traced with memray (`_execute_command → read_response → _read_from_socket`), which is where the actual multi-GB RSS growth at high connection counts comes from - the raw reply buffering, not the parsed dict list.

This PR's `client_list_iter()` instead streams and parses the reply incrementally as bytes arrive off the socket (`SocketBuffer.read_bulk_lines()` / `Connection.read_response_lines_streaming()`), so peak memory stays bounded to roughly the socket read chunk size regardless of how many clients are connected - verified directly with `tracemalloc` in the test suite, and against a live server. That's the specific behavior #4326 asks for.

Given that, I don't think this is a duplicate in the sense that matters for closing #4326, though I completely understand if maintainers prefer one approach, want to reconcile the two, or would rather I adjust naming/scope to avoid the collision - happy to do whatever's most useful here. Flagging the technical difference mainly so it's clear which approach actually resolves the reported issue.

Issue #1179, opened 2019-06-21
Sentinel connection strategies


ryan-shaw:
I thought it would be good to have an issue to track this as I see 2 PRs that are relatively old, #654 and #770 

We ran into this issue today, we have 3 sentinel instances that we were changing hosts on, these are across 3 AZs, 1a, 1b, and 1c. All our configs list the sentinel starting with 1a.

We started with the upgrade to 1c which involved shutting down the host and moving IP address, all fine and 0 impact. When we got to 1a the sentinel was shutdown and everything worked fine with no latency change, but when the host was shutdown we had a large latency spike to around 1s and only now found out it's because the sentinels list always uses the first server. When the host was shutdown the connect timeouts started to be hit (we set them at 1.2s). 

It would be nice to see different strategy options so behaviour can be more predictable across all sentinels.

BingerCoder:
I push new commit, user can set weight of each instance

RoeyPrat:
sorry for the long post 😄 
### existing implementation:
looking at [discover_master](https://github.com/andymccurdy/redis-py/blob/e06af2a0f648d4241930479884533bc53cc9fdfb/redis/sentinel.py#L200), the current expected behavior is that `Sentinel` instance has a list of sentinel hosts, and will cycle through them until one of them returns a successful response.
Importantly, once a successful response is returned, `Sentinel` instance will "promote" the good sentinel host to be the first in the list - meaning next time the first sentinel will succeed.
This means in case of X out of Y sentinels are down, we will see at most X timeouts per instance, and then no more timeouts until the next sentinel goes down.
The symptoms you speak of seem to be expected, as we see one timeout per sentinel instance, and then back to normal.
Lowering the connect timeout might help with the symptom, but I bet you thought of that.

### alternate strategy 1 - shuffle at init:
With this strategy, we shuffle once during `Sentinel` instance initialization. 
This is the strategy in PR https://github.com/andymccurdy/redis-py/pull/654 commit dc7dd8a, though it is referred to in the PR as "Round-robin".
After the initial shuffle, the `Sentinel` instance will still "stick" to the first sentinel in the list, until it fails.
This might be advantageous to some users - if you have many `Sentinel` instances in one of more clients, and you do a rolling update in the sentinels. each time a sentinel goes down, statistically, only some of the `Sentinel` instances will fail, since they have different orders of the list. 
Another benefit is that the first master in the list is not "hammered" by all the instances - the traffic is somewhat .
[...]

PR #654, opened 2015-09-16
Round-robin sentinel nodes balancer

BingerCoder:
when sentinel init with arg sentinels,the function `discover_master` always make the first node as used, so connections of the  first node is more many than others, so shuffle the nodes of sentinels, can balance the connections. 

In each processing, the sentinels node list is different.

Maybe 20000+ connections is hard to rearch, but in our project and users, it's not hard~     

Thank you~

PR #4331, opened 2026-09-17
Round-robin sentinel discovery traffic across healthy sentinel nodes

### LINKED ISSUE #4330: Sentinel discovery traffic always lands on the same sentinel node, causing load imbalance
**Version**: redis-py 8.1.0

**Platform**: Python 3.13 on Debian Trixie

**Description**:

When using `Sentinel`, all master/replica discovery traffic seems to consistently land on the same sentinel node rather than being distributed across the configured sentinels. In practice this means one node ends up carrying noticeably more load than the others, which sit mostly idle.

This isn't a new observation — it was reported before in #1179, and there were two earlier attempts to address it (#654, #770), both of which stalled and were closed as stale without real maintainer discussion. Since the underlying behavior hasn't changed, I wanted to resurface it in case there's renewed interest in revisiting it, especially for setups running many sentinel-connected clients where the imbalance becomes more noticeable over time.

kokhlo:
## Summary

Every master/replica discovery call started from the head of the `sentinels` list, and while all sentinels were healthy the head never changed: the "put this sentinel at the top" swap in `discover_master()` only fired when a non-first sentinel had to answer. So a long-lived client sent *all* of its discovery traffic to one sentinel node — the imbalance reported in #4330 (and earlier in #1179, where taking the pinned sentinel's host down caused a latency spike).

## Approach

- `discover_master()` / `discover_slaves()` (sync + asyncio) now iterate in `_discovery_order()`:
  - a sentinel that failed during a discovery call is quarantined at the end of the try order for `FAILED_SENTINEL_QUARANTINE` (8) subsequent calls — a dead node no longer costs a connect timeout on every call, but is retried periodically rather than being dropped forever;
  - when all sentinels are healthy, the sentinel that just answered rotates to the end of the list, so repeated calls spread evenly across nodes (perfect round-robin for N healthy sentinels);
  - when a call had to look past the head of the list, the answering sentinel is promoted to the head — the pre-existing failover fast path (`test_discover_master_sentinel_down` / `test_discover_master_sentinel_timeout` still pass unchanged).
- discovery now iterates a snapshot of the sentinel list and installs the new ordering with a single list rebind, so concurrent `discover_master()`/`discover_slaves()` calls can neither skip a sentinel nor observe duplicated entries (regression-tested; reproduces on the previous in-place mutation).

No public API change. #654 (shuffle at init) didn't fix this because the swap re-pins a single node after the first successful discovery; #770 required a weighted-tuple API change. This PR keeps the list-manipulation internal.

## Testing

- new `test_discover_master_rotates_over_healthy_sentinels` (sync + async): repeated discovery is served by more than one sentinel;
- new `test_discover_master_keeps_failed_sentinel_last` (sync + async): while one sentinel is down the other serves every call without the dead node being retried at the head; after the quarantine window and recovery, both nodes serve traffic again;
- new `test_discover_master_order_permutes_never_dups` (sync + async): the sentinel order stays a duplicate-free permutation of the configured nodes across calls;
- new `test_discover_master_rotates_concurrently` (sync + async): interleaved discovery sequences keep the rotation consistent — fails with `duplicate sentinels: [('bar', 26379), ('bar', 26379)]` on the previous in-place mutation;
- RED→GREEN verified by sabotaging the rotation (tests fail on the old always-swap semantics);
- full local run (no live sentinel available, integration tests excluded): sync 29 passed / 9 failed (integration, identical failure set on pristine `master`), async 28 passed / 9 failed (identical failure set on pristine `master`); `ruff check` clean on all touched files.

Fixes #4330

---

> [!NOTE]
> **Medium Risk**
> Changes failover-sensitive discovery ordering and shared mutable `self.sentinels` state used on every master/replica lookup, though behavior is regression-tested and preserves the promote-on-failover fast path.
> 
> **Overview**
> **Sentinel master/replica discovery** no longer pins all traffic to the first healthy node. Sync and asyncio `Sentinel` now walk sentinels via `_discovery_order()` and update ordering through `_promote_answering_sentinel()` instead of swapping the answering client to index 0 only when a non-head node responded.
> 
> When every sentinel is healthy, the node that just answered moves to the **end** of the list so repeated `discover_master()` / `discover_slaves()` calls spread load evenly. If discovery had to skip the head (failover path), the answering sentinel is still **promoted to the head** for the next call.
> 
> Sentinels that fail with connection/timeout (and related errors on slave discovery) are **quarantined** for `FAILED_SENTINEL_QUARANTINE` (8) subsequent discovery attempts—they are tried last, then retried so recovery is not permanent blacklisting. Discovery iterates a **snapshot** of `self.sentinels` and replaces the list in one rebind so concurrent discovery cannot duplicate or drop entries.
> 
> New sync and async tests cover rotation across healthy nodes, quarantine behavior, permutation invariants, and concurrent discovery interleaving.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e03d5a50587b8adaa91479070d7972cf0d59b235. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redis/asyncio/sentinel.py
```diff
@@ -179,6 +179,10 @@ def rotate_replicas(self) -> AsyncIterator:
 
 
 class Sentinel(AsyncSentinelCommands):
+    # How many subsequent discovery calls keep a sentinel that just failed
+    # at the end of the try order before it gets another chance.
+    FAILED_SENTINEL_QUARANTINE = 8
```
vladvildanov: Please move the attribute after the class docstring, otherwise it breaks autodoc

vladvildanov:
@kokhlo Ensure resolve the comments from Cursor bot

kokhlo:
@vladvildanov Done — pushed b5678c7:

`_discovery_order()` now iterates a snapshot (`list(self.sentinels)`), and `_promote_answering_sentinel()` builds the next ordering into a new list and installs it with a single `self.sentinels = next_order` rebind instead of the in-place `pop(0)`/`append`/`insert` mutation. A rebind only swaps the name to a new list holding the same sentinel objects, so a concurrent discovery call that already iterates an older snapshot can neither observe duplicated entries nor lose a sentinel — it just finishes against the ordering it started with.

Regression tests added to both suites: `test_discover_master_order_permutes_never_dups` pins the duplicate-free permutation invariant across calls, and `test_discover_master_rotates_concurrently` runs interleaved discovery sequences — it fails on the previous implementation with `duplicate sentinels: [('bar', 26379), ('bar', 26379)]` and passes with this fix.

vladvildanov:
@kokhlo The concurrency issue raised by bot does makes sense

PR #4338, opened 2026-09-21
Avoid Python deprecation warnings for Redis commands

### LINKED ISSUE #4337: don't raise DeprecationWarning for "deprecated commands"
using `r.setex(...)` (among others) causes a DeprecationWarning to be raised:

    DeprecationWarning: Call to deprecated setex. (Use 'set' instead.) -- Deprecated since version 2.6.12.

[the docs](https://redis.io/docs/latest/commands/setex/) say this about that deprecation (emphasis mine):

> As of Redis version 2.6.12, this command is regarded as deprecated.

> It **can** be replaced by SET with the EX argument **when migrating or writing new code**.

my understanding is that redis is not going to remove them, and only suggests that the replacements be used in future code.

a python DeprecationWarning should only be used if the called API is going to be removed - which is not the case here.

GruffElixir:
### Description of change

Redis deprecates SETEX and HMSET at the server level while redis-py continues to expose them for compatibility. Emitting Python `DeprecationWarning` from these methods can break callers that treat warnings as errors. This removes those warning decorators, clarifies the Redis-side guidance in the docstrings, and adds sync/async regression tests.

Fixes #4337.

### Pull Request check-list

- [x] Do tests and lints pass with this change? Focused sync/async regression tests and repository-wide linters pass.
- [ ] Do the CI tests pass with this change? The full Redis integration suite could not be run locally because Docker is unavailable.
- [x] Is the new or changed code fully tested? The regression tests cover both commands on sync and async clients.
- [x] Is a documentation update included? Updated the command docstrings.
- [x] Is there an example added to the examples folder (if applicable)? Not applicable.

### Validation

- Focused pytest regression tests: 2 passed.
- `invoke linters`: passed.
- Full Redis integration suite: not run locally; Docker is unavailable.

---

> [!NOTE]
> **Low Risk**
> Behavioral change is limited to warning emission and documentation; command execution paths are unchanged.
> 
> **Overview**
> Stops emitting **Python `DeprecationWarning`** when calling `setex` and `hmset`, while keeping both methods and their Redis wire commands unchanged for compatibility.
> 
> The `@deprecated_function` decorators are removed from those two methods in `core.py`, and their docstrings now describe **Redis server** deprecation (since 2.6.12 / 4.0.0) and point callers to `set(..., ex=...)` and `hset` instead. Sync and async tests assert that `setex`/`hmset` no longer raise under `warnings.simplefilter("error", DeprecationWarning)`, and existing `hmset` tests no longer expect a deprecation warning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18e3e783ed317a221f4f23220ce84afa85a3acce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vladvildanov:
@GruffElixir I left a comment regarding this on #4337

PR #4340, opened 2026-09-22
fix(multidb): pass transaction options as keyword arguments

simpleqt:
### Description of change

`MultiDBClient.transaction()` (sync) forwards its `**options` dict **star-unpacked as positional arguments**:

```python
return self.command_executor.execute_transaction(func, *watches, *options)
```

`*options` on a dict yields its **keys**, so every transaction option is silently dropped and the option *names* are appended to `*watches`. `DefaultCommandExecutor.execute_transaction` then passes them on to `Redis.transaction(...)`, which ends up `WATCH`-ing strings like `"value_from_callable"` and `"watch_delay"` while ignoring the actual option values.

Minimal repro (no server needed — observe what the active-db client receives):

```python
from redis.multidb.client import MultiDBClient

class CaptureExecutor:
    def execute_transaction(self, func, *watches, **options):
        print("watches:", watches, "options:", options)

client = MultiDBClient.__new__(MultiDBClient)
client.initialized = True
client.command_executor = CaptureExecutor()
client.transaction(lambda pipe: None, "watched-key", value_from_callable=True, watch_delay=2.5)
# actual:   watches: ('watched-key', 'value_from_callable', 'watch_delay') options: {}
# expected: watches: ('watched-key',) options: {'value_from_callable': True, 'watch_delay': 2.5}
```

The async sibling (`redis/asyncio/multidb/client.py`) already forwards these as keyword arguments, and the sync client previously had the exact same star-unpacking bug class in its pubsub methods, fixed in #4276 — `transaction` was the missed sibling. This one-line fix makes the sync client forward options the same way:

```python
return self.command_executor.execute_transaction(func, *watches, **options)
```

A regression test is added in `tests/test_multidb/test_pipeline.py` asserting watches stay positional and options reach the active client as kwargs (fails without the fix, passes with it).

### Pull Request check-list

- [x] Do tests and lints pass with this change? (`pytest tests/test_multidb/test_pipeline.py` — 11 passed; `ruff check` + `ruff format --check` clean on both changed files)
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? — no API change; existing docstring already describes the correct behavior
- [ ] Is there an example added to the examples folder (if applicable)? — not applicable

---

> [!NOTE]
> **Medium Risk**
> Correctness fix for Redis transactions and WATCH semantics; behavior changes for callers that passed transaction kwargs, but prior behavior was wrong and could cause subtle data races.
> 
> **Overview**
> Fixes sync **`MultiDBClient.transaction()`** so transaction options are forwarded correctly to the active Redis client.
> 
> Previously `**options` was passed as `*options`, which star-unpacked dict **keys** into the watch list and dropped the option values (e.g. `value_from_callable`, `watch_delay` were treated as keys to `WATCH` instead of kwargs). The call now uses `**options`, matching the async multidb client and `DefaultCommandExecutor.execute_transaction`.
> 
> Adds **`test_transaction_forwards_watches_and_options`** to assert watched keys stay positional and options reach the underlying `client.transaction` as keyword arguments.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 292821ec9fe6809c19acd3dcc53d85ac28220c6b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vladvildanov:
@simpleqt Thanks for your contribution! Could you please also verify that async client has the same issue and apply the same changes?

simpleqt:
@vladvildanov Checked — the async client does **not** share the bug. `AsyncMultiDBClient.transaction` declares `shard_hint`/`value_from_callable`/`watch_delay` as named parameters and forwards them as keyword arguments (redis/asyncio/multidb/client.py:326), and `DefaultCommandExecutor.execute_transaction` does the same down to the active database client (redis/asyncio/multidb/command_executor.py:251). The star-unpacking regression from #4276 never reached these paths.

To make that guarantee durable, I added a regression test for the async client mirroring the sync one (85b52fd): it asserts watches stay positional and options arrive at the active client as keyword arguments (including the explicitly-forwarded `shard_hint=None`).

Locally: both sync and async multidb pipeline/transaction tests pass (8 passed), ruff 0.14.5 clean.

